PR #8, opened 2016-01-12
Fix a few typos

maxrothman:


stolksdorf:
You da man. I'll deploy this with my next set of changes. 

I'm so sorry you had to navigate that gross `welcomeMessage.js` file. I'll be replacing that with something more elegant.

Issue #61, opened 2016-03-09
Nested Lists Not Working


hephalumph:
**Browser Type/Version**: [Google Ultron v90.01]

**Operating System**: [GLaDOS v34.5.8]

**Issue Description**: [The thing won't thing]

**Markdown code to reproduce**:

*  list item one
  * nested list item 1
  * nested list item 2
* list item 2

**Related Images** :

SecretShop:
i would have to agree that i was going to use nested lists but realized that i couldn't find the syntax for it.

kkragenbrink:
it's just a CSS issue

http://www.naturalcrit.com/homebrew/edit/rylhOoaT

Fix here.

I'll submit a pull request to fix this later tonight.

kkragenbrink:
Might be able to close this now.

PR #82, opened 2016-04-19
Fix for Nested Lists Not Working

kkragenbrink:
Fix for #61


stolksdorf:
Looks good. I'm in the middle of a few changes, so I'll roll this out with the next release. Thanks!

Issue #71, opened 2016-03-30
Wide tables?


PariahSilver:
Heya, I'm not reporting an issue but I do have a question. You have wide monster blocks, which are fantastic, but do you have plans for adding wide tables?

stolksdorf:
Pretty easy to add, I'm just running out of syntax to use. Any suggestions?

tonymccallie:
Just add another line? Like they just add more hashes for headers.. 
```
##### The Linguist
___
___
___
| Level |
```

or maybe longer dashes?
```
##### The Linguist
______
| Level |
```

stolksdorf:
#62

kkragenbrink:
As a suggestion, stolksdorf, I think you should actually reuse the syntax you've got.

```
___
___
> Wide Monster Statblock

___
___
| Wide Table |

stolksdorf:
This has been added in v2.1.0 using the "Wide Block" snippet. Enjoy!

Issue #62, opened 2016-03-10
[Suggestion] Page-wide tables


Kayrajh:
In homebrewery, could it be possible to make page-wide tables? I'm currently using the "class table" code but the style is clashing with what I want.

Awesome work by the way, I'm totally in love with your work.

PR #84, opened 2016-04-19
Double HR leads to wider tables, but not class tables.

kkragenbrink:
#71 #62


stolksdorf:
This looks really interesting. I'll test it out tonight!

PR #200, opened 2016-09-16
Fix Dockerfile and migrate to Alpine as base

jonathonf:
This fixes the issues with the Docker container as well as switching to Alpine Linux as the base. This in turn reduces image size from ~932MB to 264MB.


jonathonf:
I'll wait until #203 is merged before updating this pull request.

jonathonf:
Updated PR: https://github.com/stolksdorf/homebrewery/pull/206

PR #203, opened 2016-09-18
Minor documentation changes

lasseborly:
Made some minor changes to the package.json and the readme.

* Corrected the license so that it was MIT everywhere.
* There was some minor spelling mistakes in the README.md
* We were already installing gulp locally so there is no need to have it globally with the small amount of commands we use it for. Added some local commands instead and changed the readme accordingly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,10 @@
   "description": "Create authentic looking D&D homebrews using only markdown",
   "version": "2.3.0",
   "scripts": {
-    "build": "node_modules/.bin/gulp fresh",
+    "build": "node_modules/.bin/gulp prod",
     "watch": "node_modules/.bin/gulp",
-    "postinstall": "gulp prod",
+    "libs": "node_modules/.bin/gulp fresh",
+    "postinstall": "npm run build",
     "start": "node server.js"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,11 +3,13 @@
   "description": "Create authentic looking D&D homebrews using only markdown",
   "version": "2.3.0",
   "scripts": {
+    "build": "node_modules/.bin/gulp fresh",
+    "watch": "node_modules/.bin/gulp",
     "postinstall": "gulp prod",
```
jonathonf: Does this also need to change to `node_modules/.bin/gulp prod` ?

stolksdorf:
Great PR, except I would change the `package.json` to this:

```
  "scripts": {
    "build": "node_modules/.bin/gulp prod",
    "watch": "node_modules/.bin/gulp",
    "libs": "node_modules/.bin/gulp fresh",
    "postinstall": "npm run build",
    "start": "node server.js"
  },
```

The commands are more clear and it still works with my deploys.

lasseborly:
I have changed according to previous comment and corrected the readme.

PR #206, opened 2016-09-19
Tweak Dockerfile and migrate to alpine as base

jonathonf:
This fixes issues I have with the Docker container as well as switching to Alpine Linux as the base. This in turn reduces image size from ~950MB to ~250MB.


stolksdorf:
I've decided to not explicitly support docker for this project anymore. I have never used it, nor do I test and make sure it works with continued updates. I encourage you to make a fork of this project that is docker'd and I'll link to it in the readme :)

PR #287, opened 2017-01-13
Fix Class feature snippet insertion

MildMastermind:
I'm not 100% sure this works, but it matches how the other snippets seem to be written.

Sorry, I'm new to GitHub


jamesbirkett:
I'm not @stolksdorf, but this isn't necessary - the `var` is only needed if the variable has not yet been defined. Since it's a parameter of the function, it is already defined :smile:

MildMastermind:
Regardless, this is still broken, it seems that classname contains the entirety of the current document when this runs. Looking at it now I think part of this issue is that classname shouldn't actually be a parameter of the function.

calculuschild:
We have an updated fix for this. See #687

Issue #348, opened 2017-02-26
Idea: More flexible styling of blockquotes and lists (e.g. for card templates)


Rae2che5:
This is a suggestion inspired by #212 -- @ryantheleach's request for printable card templates in the Homebrewery. @wrdclark did some great work mocking up some styles for cards based on `<div class='card'>`, but unfortunately, as @vlos observed, deeply nested `<div>` tags tend to switch the markdown parser into HTML mode, and thus lose the ability to use nice markdown formatting.

Looking into this, I discovered @stolksdorf's cool trick of styling lists differently if they immediately follow a `<hr>`, and I couldn't help thinking that this would also be useful for cards, allowing us to nest their content in a `<blockquote>` and not a `<div>` and thus preserve markdown formatting.

Unfortunately, there's only one kind of `<hr/>` tag, and so we need something else to attach styling to. This is where we can exploit two facts: Firstly, `<h6>` are currently unstyled and hence almost definitely unused; secondly, `marked` (our Markdown parser) generates IDs for the `<h6>` tags based on the text of the heading. To make a long story short, here's a bit of markup:

```markdown
###### card
> ### Goblin
>
> <img src="http://i.imgur.com/ET6Hp6D.png" style="height: 5cm;"/>
>___
>|STR|DEX|CON|INT|WIS|CHA|
>|:---:|:---:|:---:|:---:|:---:|:---:|
>|8 (-1)|14 (+2)|10 (+0)|10 (+0)|8 (-1)|8 (-1)|
>___
> <div class='descriptive'>
> - All markdown formatting works.
> - Even nested lists.
>   - And sublists.
> </div>
>
```

And here's how it could render: 
![image](https://cloud.githubusercontent.com/assets/26040349/23339820/30fedc9c-fc22-11e6-8fb6-e7211427c73e.png).

The key trick is to write selectors that begin with `h6[id|='card'] + blockquote`, meaning "the `blockquote` immediately following the `<h6>` tag whose ID starts with `card`". In particular, this also allows different styles of card -- `###### card back` could, for example, rotate the card by 180 degrees, so that you can simply fold to get a standee/tent.

[...]

PR #370, opened 2017-03-26
Parse markdown only once.

Rae2che5:
By the time `marked` invokes the `renderer.paragraph` callback, it has already processed the content and translated anything it could parse to HTML. After splitting on block markers, we should therefore avoid re-parsing the content as (sanitized) Markdown), because that will escape the HTML markup produced by the initial parse.

To see the problem, try a brew like this:

```
{{test
![Goblin](http://i.imgur.com/ET6Hp6D.png)
}}
```

Without this change, the `` tag produced by the first parse is escaped into `&lt;img...`, and therefore the image is not displayed. After the patch, the image renders as expected.

-------
I ran into this while investigating a port of #348 onto the v3 branch.

### DIFF
--- a/shared/homebrewery/markdown.js
+++ b/shared/homebrewery/markdown.js
@@ -10,7 +10,7 @@ renderer.paragraph = function(text){
 	if(!matches) return `\n<p>${text}</p>\n`;
 	let matchIndex = 0;
 	const res =  _.reduce(text.split(blockReg), (r, text) => {
-		if(text) r.push(Markdown(text, {renderer : renderer, sanitize: true}));
+		if(text) r.push(text);
 		const block = matches[matchIndex];
 		if(block && block[0] == '{'){
 			r.push(`\n\n<div class="block ${block.substring(2).split(',').join(' ')}">`);


stolksdorf:
Unfortunately, the problem isn't resolved that easily. 

The second call to the markdown renderer ensures that nested `<p>` tags get rendered within blocks. eg. this fails to render properly

```
{{note,alt
##### Gunslinger Corruptor
The alternate protagonist is revealed to be a different race/species than thought, which suddenly makes what's going on much clearer. Thanks to alien forces, the secondary protagonist ends up in a world after an apocalypse. There is a predicted hurricane - but it's not what was expected, and this complicates the plans of the protagonist. There is a sudden hurricane.
}}
```

I'll look into how to deal with the specific situation you've outlined.

stolksdorf:
Yeah, it'll take a deep dive into the actual Markdown lexer/parser to fix this issue. 

this breaks

```
{{test
**yo**
![Goblin](http://i.imgur.com/ET6Hp6D.png)
}}
```

But this works

```
{{test

**yo**

![Goblin](http://i.imgur.com/ET6Hp6D.png)

}}
```

The inline lexer is being executed prematurely for the block rendering to work properly. In more recent versions of the feature branches off of `v3` I've started using a custom version of `marked` that I can tweak to resolve issues like this. I'll try and push up a new version of `v3` when things are a bit more stable.

PR #432, opened 2017-07-02
Fix aside being rejected as unclosed link

Ry7n:
The current RegExp for validating HTML gets derived into includes this
value:

```
/(\<div|\<\/div>|\<span|\<\/span>|\<a|\<\/a>)/g
```

However, this leads to the following match on an `<aside />` element:

```
> ('<aside>foo</aside>').match(tagRegex)
[ '<a' ]
```

...which looks to homebrewery like an unclosed `<a />` element. This new
RegExp ensures that open tags are always matched with closing tags of
the same type.


Rae2che5:
Thanks for the diagnosis and patch. This now conflicts with master, so I've made the equivalent change in #678.

PR #539, opened 2017-10-31
Add QR-Code as snippet under Editor

bakgaard:
Adding a small snippet option under ``Editor`` in the client, which will insert a small a QR-code ``img`` in the code.

It will use the [QR code](http://fontawesome.io/icon/qrcode/) as icon.

It will look as follow:

```
<img 
    src='https://api.qrserver.com/v1/create-qr-code/?data=
    http://homebrewery.naturalcrit.com
    /&amp;size=100x100'
    style='width:100px;mix-blend-mode:darken '/>
```

The line ``http://homebrewery.naturalcrit.com`` should be replaced by the the Shared URL, and the image will update instantaneously.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -18,10 +18,9 @@ const SNIPPETBAR_HEIGHT = 25;
 const Editor = createClass({
 	getDefaultProps : function() {
 		return {
-			value    : '',
+			brew    : {},
 			onChange : ()=>{},
 
-			metadata         : {},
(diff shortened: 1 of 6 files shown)


calculuschild:
@bakgaard Hey there! After all these years, we are finally adding in some of these changes. It's still hidden behind a config variable so it won't show up on the live site yet, but we are working on it! Thanks for your contribution!

bakgaard:
@calculuschild Oh, I thought this project had been put to the grave.
Glad to hear it is of use :)

Issue #332, opened 2017-02-08
Wrap Source Text - option


tylerpaul:
This is to **improve user experience** and is **not an "Issue"**.

It would be nice to be able to wrap source text when viewed.
Some users put time into formatting their source, but, since the editor wraps text, many do not consider it. This leads to painful use of the horizontal scrollbar or Home and End keys to navigate source text.
There are workarounds, including copying the text into a simple text editor, but I thought it might improve your already excellent product (perhaps by only a small margin).

calculuschild:
This is good and will not be hard to implement. Thanks for the feedback!

Rae2che5:
Fixed by #665.

PR #665, opened 2018-04-18
Wrap the text in the "source" page.

calculuschild:
See #332.


Rae2che5:
LGTM, and I agree it's an improvement -- the source viewer of the brew already wraps on whitespace, so it makes sense to make the raw source viewer do it too.

PR #669, opened 2018-04-20
Fix Partial Page Rendering hiding the current page on longer documents

mlenser:
Problem: On large documents `this.state.viewablePageNumber` becomes out of sync with `viewIndex` as seen:

ppr

This PR temporarily increases the limit of pages shown to 10 so that large projects can see the current page. Otherwise it's just a spinner. The performance is marginally impacted, but that has less consequences than users not being able to use homebrewery imo

This can be changed at a later point when PPR is fixed, but currently homebrewery cannot handle a large amount of pages.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -78,13 +78,8 @@ const BrewRenderer = createClass({
 		if(!this.state.isMounted) return false;
 
 		const viewIndex = this.state.viewablePageNumber;
-		if(index == viewIndex - 3) return true;
-		if(index == viewIndex - 2) return true;
-		if(index == viewIndex - 1) return true;
-		if(index == viewIndex)     return true;
-		if(index == viewIndex + 1) return true;
-		if(index == viewIndex + 2) return true;
-		if(index == viewIndex + 3) return true;
+
+		if(Math.abs(viewIndex - index) < 10) return true;
 
 		//Check for style tages
 		if(pageText.indexOf('<style>') !== -1) return true;


Rae2che5:
I'm not sure I quite understand the problem this fixes. I agree that taking a performance hit would be better than "users not being able to use homebrewery", but we should fix that issue too. Could you please link to an example or give some more details?

mlenser:
This is experienced when creating a homebrewery of hundreds of pages. Around 100+ you can't see the page that you're editing. It shows a spinner instead. The cause is outlined in my description

I can create a demo of necessary.

mlenser:
@Rae2che5 demo: http://homebrewery.naturalcrit.com/share/HyrzF_RpTf
Scroll somewhere half way down and you'll see the issue.

I've stripped out all custom styles so it shouldn't be related to that.

Rae2che5:
Ah, I see -- the counter goes out of sync, and therefore the page that's actually displayed is suppressed by partial page rendering.

Could you please try whether #679 fixes the problem for you? It does so by attempting to fix the currently-visible-page computation, which, I believe, is the right approach.

Rae2che5:
(I should say -- thanks for providing the sample code, it was much easier to see the issue with it than with the synthetic multi-page brew I had created. The patch at #679 does make it work properly for me.)

mlenser:
It looks to fix the issue, but it also makes the scrolling quite laggy. Perhaps because it's actually showing the content or perhaps it is indeed worse.

calculuschild:
I notice the lag as well but it seems like it was already there with large documents. Do you not see this lag in the current live website on your large document?

Rae2che5:
I did not notice the lag in the (stripped-style) doc @mlenser provided; I wonder whether there's some hardware/software/configuration difference. Could you perhaps share the fully styled doc, in case that is what triggers the slowness?

calculuschild:
@Rae2che5 I am using his stripped-style document. It's not so much lag as a FPS issue. When I am at a relatively low number of pages, the scrolling transition is very smooth. Running a profiler in Chrome shows an FPS of 57-60 during the scrolling event. However with a larger document, around 2,500 lines for example, the FPS during scrolling drops to 18-20 FPS. Of course hardware might be a contributor, although my machine is by no means a laggard.

Cutting down the page rendering range to just one page on either side of the current page does help, jumping back up to ~40 FPS.

calculuschild:
I'm going to close this since it has been superseded by a new PR, #679.

mlenser:
Great, thanks for looking in to the issue!! I'll verify some things when I'm back from vacation next week.

PR #676, opened 2018-05-07
Config: Read MongoDB URL from config files too.

Rae2che5:
This turned out to be useful in some local work. By reading the MongoDB URLs via `nconfig`, we gain the ability to specify them on the command line or in config files -- while still respecting the previous environment variables, so that the patch hopefully does not break anybody's setup.


Rae2che5:
@calculuschild, if this looks safe to you perhaps we can merge it?

PR #677, opened 2018-05-07
Fix property name break-inside

ericscheid:
https://github.com/stolksdorf/homebrewery/issues/595


ericscheid:
Might as well..

>-webkit-column-break-inside: avoid; /* Chrome, Safari, Opera */
>          page-break-inside: avoid; /* Firefox */
>               break-inside: avoid; /* IE 10+ */

Rae2che5:
Thanks, lgtm now. I've tested to some extent, but mainly by validating that things don't break. @stolksdorf, @calculuschild -- do you agree this is safe to merge?

calculuschild:
This looks good to me. I don't see anything breaking on my end.

Issue #230, opened 2016-11-16
HTML Validator flagging custom tags starting with "a" as unmatched opening tag


calculuschild:
In my brew, I have made use of custom HTML tags to generate small icons inline with the text.  One of these icons is named < airEl >, and the HTML validator is flagging every use of this tag as "Unmatched opening tag, 'a' tag".  There doesn't seem to be an issue on the other custom tags (such as < fireEl >)

### Additional Details

**Share Link** : http://homebrewery.naturalcrit.com/share/SyEA7MxL (this might be temporarily changed to <bair> to avoid the issue so I can work on my brew.  Problem code is pasted below, but my HTML tags are not appearing in the code preview for some reason, so I replaced the leading < with a ( to keep them from hiding.)

or

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

<style>
airEl {
	background-image: url(https://d30y9cdsu7xlg0.cloudfront.net/png/281535-200.png);
    background-size: 14px 14px;
    background-repeat: no-repeat;
    display:inline-block;
    width:14px;
    height:14px;
    vertical-align: text-top;
    color: transparent;
}

fireEl {
	background-image: url(https://d30y9cdsu7xlg0.cloudfront.net/png/96564-200.png);
    background-size: 14px 14px;
    background-repeat: no-repeat;
    display:inline-block;
    width:14px;
    height:14px;
    vertical-align: text-top;
    color: transparent;
}
<style>

There are two primary essences that can be extracted from potent herbs, fungus, or animal parts for use as alchemical components: (airEl>(/airEl>**air**, and (fireEl>(/fireEl>**fire**.

</pre></code></details>

ericscheid:
This looks to be fixed in PR https://github.com/naturalcrit/homebrewery/pull/678

PR #678, opened 2018-05-07
Tag validation: Fix handling of <aside>/<a> and similar tags

Rae2che5:
Port of #432 to current master, including some tests.

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -20,6 +20,11 @@ const tagRegex = new RegExp(`(${
 		return `\\<${type}\\b|\\</${type}>`;
 	}).join('|')})`, 'g');
 
+// Special "void" tags that can be self-closed but don't need to be.
+const voidTags = new Set([
+	'area', 'base', 'br', 'col', 'command', 'hr', 'img',
+	'input', 'keygen', 'link', 'meta', 'param', 'source'
+]);
 
 module.exports = {
 	marked : Markdown,
@@ -29,6 +34,7 @@ module.exports = {
 
 	validate : (rawBrewText)=>{
 		const errors = [];
+		// We validate line by line.
 		const leftovers = _.reduce(rawBrewText.split('\n'), (acc, line, _lineNumber)=>{
 			const lineNumber = _lineNumber + 1;
 			const matches = line.match(tagRegex);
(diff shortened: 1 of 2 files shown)


calculuschild:
This looks fine, but I wonder if it would be helpful to handle more than just `span`, `div`, and `a`. I know that at the very least the `style` tag gets used a lot, and I've seen a few brews that use `<i>` or `<b>` instead of the markdown syntax. I myself have a few brews that use some custom tags as a shorthand for special icons.

Rae2che5:
OK, I've generalised the validation to handle arbitrary tags, not just the three listed tags. [Fortunately I had written some tests for the first, smaller change -- getting them to pass again was very helpful in terms of getting this right!]

calculuschild:
Good good!

Now, I just thought of one last thing.... What about something like `<br>` which is often written without the self-closing tag? As it is, `<br>` would be tagged as missing a closing tag, no? Same with `` I think... it's technically correct without the self-closing tag.

I don't know a graceful way to handle those, other than to special-case them, or maybe change the error message to say that they need to self-close those tags.  What do you think?

Here's a list I found of "void" HTML tags that can be self-closed but don't technically need to be:

> The void elements in HTML 4.01/XHTML 1.0 Strict are area, base, br, col, hr, img, input, link, meta, and param. HTML5 currently adds command, keygen, and source to that list. (http://www.456bereastreet.com/archive/201005/void_empty_elements_and_self-closing_start_tags_in_html/)

calculuschild:
Hmmm... That's tricky. I think overall we should be aiming for a system that functions correctly and is prone to fewer errors; a large portion of the userbase is not fluent with HTML syntax and frequently run into these issues that end up crashing their documents, or give them unexpected behavior that they get around by applying even worse HTML or CSS stylings.  So, yes, I personally think it is for the best to guide users toward valid syntax at the cost of some brews getting error popups now.

Perhaps we can make that error bar a bit less obnoxious by adding a button to close it. That way, if the document *seems* fine, the users can still choose to ignore it at their own risk; the brew should still render the same as it always has otherwise, right?

Rae2che5:
I can certainly see how hard it is to make the error dismissable.

Rae2che5:
@calculuschild I've now added support for void tags (see new test).

I continue to be slightly uneasy about detecting many more of these errors. On the one hand, it's definitely better to be up-front about the fact that the user made a mistake, but on the other regex-based validation of HTML will never be perfect (see e.g. the test in f7df192), and invalidating existing docs is a rather breaking change.

On the whole, though, I think I'd support just making the validation stricter. It'd be great if you could try this on as many brews as you have locally. @stolksdorf, is there a way of automatically testing how many existing brews it would report new errors on? [Do we consider brews on homebrewery.com "public" enough for this to be OK to do?]

Rae2che5:
Right. I meant to look into how hard that would be (or indeed whether the errors are blocking for things like PDF generation, beyond, of course, the risk of bad HTML they expose); that kind of slipped off my radar. I'll dig into it a bit.

ericscheid:
This will also fix issue https://github.com/naturalcrit/homebrewery/issues/230, right?

calculuschild:
Yes I think so. This probably needs to be checked to make sure it still solves the problem but it would be nice to merge this.

calculuschild:
Closing this in favor of #1583

PR #679, opened 2018-05-07
Fix calculation of currently-viewed page.

Rae2che5:
Previously, this relied on a hard-coded constant determining the
height of a page. However, that's unnecessary -- we know the scroll
height of the window and the number of pages, so we can compute it.

This addresses the problem reported in #669 in perhaps a more principled way.

@stolksdorf, I'd appreciate your thoughts on this one. I'm not familiar enough with react stuff to judge whether what I did is always safe. The semantic change is in the `handleScroll` function; the rest of the diff merely removes the now-unused `pageHeight` attribute.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -36,7 +36,6 @@ const BrewRenderer = createClass({
 		};
 	},
 	height     : 0,
-	pageHeight : PAGE_HEIGHT,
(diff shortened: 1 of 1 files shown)


calculuschild:
This looks good to me and makes a lot more sense. It was mentioned in #669 that the scrolling has also become laggy, but I notice the same lag on the live site after a document reaches about 1,200 lines, and gets progressively worse with more pages. I think that is a separate optimization issue.

PR #685, opened 2018-05-10
Throttle handleScroll function

calculuschild:
Hopefully reduce number of style recalculations due to accessing the DOM and speed up FPS when scrolling.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -76,7 +76,7 @@ const BrewRenderer = createClass({
 		this.setState({
 			viewablePageNumber : Math.floor(e.target.scrollTop / this.pageHeight)
 		});
-	},100),
+	}, 100),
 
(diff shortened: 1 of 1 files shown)


Rae2che5:
I like this more than the ad-hoc throttling in #684 (in particular, it will make sure we do update after a big scroll happens within the timeout). However, if #686 makes it unnecessary, perhaps we don't need to worry?

I'm still trying to reproduce the lag on some of my devices, with only occasional moderate success, so it's hard to tell...

Rae2che5:
Do we still think this is worth doing? On the whole I think I'm coming around to it -- there's some [discussions about throttling `onScroll` events](https://css-tricks.com/debouncing-throttling-explained-examples/) that are fairly convincing.

Stylistically, I'd slightly prefer to put the throttling in the `onScroll` handler registration, as `onScroll={_.throttle(this.handleScroll, 100)}`, rather than inline in `handleScroll` -- what do you think?

PR #686, opened 2018-05-10
Reduce repaints on scroll

calculuschild:
The `will-change` property allows the browser to optimize for smoother animations. This completely eliminates the scrolling stutter. Give it a whirl.

Had pull the page count and partial-page notification out of the BrewRenderer div because `will-change` messes up absolute positioning of child elements.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -25,6 +25,7 @@ const BrewRenderer = createClass({
 		return {
 			viewablePageNumber : 0,
 			height             : 0,
+			willChange		   : 'auto',
 			isMounted          : false,
 
 			usePPR : true,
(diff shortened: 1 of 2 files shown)


calculuschild:
Leaving `will-change: transform` active supposedly requires unnecessary "resources" which I am not exactly clear on. Everything I can find suggests that it is best practice to turn it on only when the animation will likely occur in the next few moments. I don't know if this means memory or CPU or what.

It doesn't look any different leaving it on all the time, and I honestly didn't see any "resource" issues, so maybe keeping it is fine. In that case we would just add it directly to the style sheet.

Now that you mention the keyboard scrolling, that's a really easy way to see the difference. Mouse over and scroll with the mouse wheel, then mouse away and scroll with the keyboard. On my machine it is pretty obvious that one is smoother than the other.

The source pane also has a similar scrolling lag, so we can probably work on adding it there as well.

calculuschild:
@mlenser Does this fix the scrolling lag you mentioned earlier?

mlenser:
I'm on vacation so I won't be able to verify, sorry!

Rae2che5:
@calculuschild You're right, the contrast in scrolling with the keyboard when the pointer is in the render pane vs. when it is not is quite visible.

I think we should just add `will-change:transform` to the stylesheet unconditionally, for both the source pane and the render pane, until we discover a reason not to.

As an alternative, we could take a slightly pragmatic approach of "we'll set it to `transform` when a scroll event happens, and then return to `auto` after a second (or five) of no scroll events"; this should at least not permanently use whatever nebulous extra resources are consumed.

The patch might be something like this (on top of your branch):

```diff
diff --git a/client/homebrew/brewRenderer/brewRenderer.jsx b/client/homebrew/brewRenderer/brewRenderer.jsx
index 5b1a6d5..818d75f 100644
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -62,10 +62,16 @@ const BrewRenderer = createClass({
 		});
 	},
 
+	resetWillChange : _.debounce(function() {
+		this.setState({ willChange: 'auto' });
+	}, 1000),
+
 	handleScroll : function(e){
+		this.setState({ 'willChange': 'transform' });
 		this.setState({
 			viewablePageNumber : Math.floor(e.target.scrollTop / e.target.scrollHeight * this.state.pages.length)
 		});
+		this.resetWillChange();
 	},
 
 	shouldRender : function(pageText, index){
@@ -127,27 +133,11 @@ const BrewRenderer = createClass({
 		return this.lastRender;
 	},
 
-	/**
-	* Optimize for smooth scrolling when mouse enters the rendering panel
-	**/
-	prepareScroll : function(){
-		this.setState({ willChange: 'transform' });
-	},
-
-	/**
-	* Unload smooth scrolling optimizations when mouse leaves rendering panel
-	**/
-	unprepareScroll : function(){
-		this.setState({ willChange: 'auto' });
-	},
-
 	render : function(){
 		return (
 			<React.Fragment>
 				<div className='brewRenderer'
 					onScroll={this.handleScroll}
-					onMouseOver={this.prepareScroll}
-					onMouseOut={this.unprepareScroll}
 					ref='main'
 					style={{ height     : this.state.height,
 							 willChange : this.state.willChange }}>
```

What do you think?

calculuschild:
@Rae2che5 Does this method you propose still have the smooth scrolling then? I don't have time to test it at the moment, but the documentation on `will-change` suggests that if you want the effect to work properly it needs to be enabled a few (~200) milliseconds before the animation occurs to allow for the optimizations to get loaded, otherwise you will get a little stutter at the beginning.

In any case, I think I'm in agreement that we should just add it to the stylesheet for now.

Rae2che5:
It's indeed laggy for the first ~200 milliseconds of scrolling, but becomes smooth thereafter. I don't think we can do better than that, given that keyboard scroll events are entirely unpredictable, so I'd like to think of it as a fallback if it turns out we don't want `will-change: transform` permanently.

calculuschild:
Ok, I have a new PR on #693 which just places `will-change` directly in the BrewRenderer style sheet.

I haven't tested it for the editor pane because we might have a similar issue with child elements having their absolute position broken and I haven't tested that yet.

PR #690, opened 2018-05-11
Revert "Fix duplicated text from Class Feature snippet"

Rae2che5:
Reverts stolksdorf/homebrewery#687

I hadn't rebuilt the bundle when I tested this. :-( It actually breaks all snippets, because `snippet.name` causes an exception -- `snippet` is undefined at that point.

@calculuschild, I'm also confused, because I don't see the duplication you described. Indeed `classfeature.gen.js` takes an argument, but [it is immediately overwritten](https://github.com/stolksdorf/homebrewery/blob/9519a0b4e4b6b4ecb06bb1fa45017112a8b95ec3/client/homebrew/editor/snippetbar/snippets/classfeature.gen.js#L5), so I don't think it can be responsible (on master) for the problem.

I'll merge this PR immediately, because it's a simple revert of something that breaks master.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -7,9 +7,7 @@ const cx    = require('classnames');
 const Snippets = require('./snippets/snippets.js');
 
 const execute = function(val, brew){
-	if(snippet.name == 'Table of Contents')
-		return val(brew);
-	else if(_.isFunction(val)) return val();
+	if(_.isFunction(val)) return val(brew);
 	return val;
 };
 


calculuschild:
Ok. Thanks for fixing that. I will double-check when I get back to my machine on the snippet.name. I wasn't seeing anything broken on my side so I might have forgotten to push up something.

To see the duplication issue, try typing something like "chicken nugget" into an empty brew document. Then, add in a "Class Feature" from the snippet bar. You should see a couple places in the generated text where you would expect "actuary" or whatever, but instead it will say "chicken nugget". Now add another Class Feature. You should have a giant mess with multiple copies of the text embedded within the new Class Feature. This happens for me both on the live site and on master. Could you double-check that this doesn't happen on your end?

You are right, the argument is immediately overwritten. *However*, the "brew" argument is a React prop, which is immutable, so the overwrite doesn't actually occur.

PR #691, opened 2018-05-12
Fix dupliate text from class feature snippet

calculuschild:
Reimplementation of stolksdorf/homebrewery#687 which had a small error.

Nearly identical fix; just moved the special case check so `snippet.name` is in scope.

@Rae2che5 Could you verify that the bug does exist? I am seeing it both on the live site and on master.

> To see the duplication issue, try typing something like "chicken nugget" into an empty brew document. Then, add in a "Class Feature" from the snippet bar. You should see a couple places in the generated text where you would expect "actuary" or whatever (*"As an **[actuary]** you gain the following class features..."*), but instead it will say "chicken nugget". Now add another Class Feature. You should have a giant mess with multiple copies of the previous text, which now includes the first "Class Feature" you made, embedded within the new Class Feature.

image

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -7,7 +7,9 @@ const cx    = require('classnames');
 const Snippets = require('./snippets/snippets.js');
 
 const execute = function(val, brew){
-	if(_.isFunction(val)) return val(brew);
+	if(snippet.name == 'Table of Contents')
+		return val(brew);
+	else if(_.isFunction(val)) return val();
 	return val;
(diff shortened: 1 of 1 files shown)


Rae2che5:
Thanks for the reproduction instructions. I can indeed see the problem on the live site, but not on `master` -- I believe 3247cab2148436fbf5be370381278ef484a76450 fixed it. Are you saying you can reproduce with a `master` bundle before this PR?

calculuschild:
Well I'll be... I re-pulled master and tried again. You are right: the problem isn't there on master. Thank you for the double-check! I'm going to just close this out.

PR #693, opened 2018-05-13
Embed brew renderer optimizations in style sheet

calculuschild:
Following the discussion on #686. This makes `will-change: transform` permanent in the brewRenderer stylesheet rather than trying to load and unload it programatically.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -25,6 +25,7 @@ const BrewRenderer = createClass({
 		return {
 			viewablePageNumber : 0,
 			height             : 0,
+			willChange		   : 'auto',
 			isMounted          : false,
 
 			usePPR : true,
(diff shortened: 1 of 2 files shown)


client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -134,20 +134,26 @@ const BrewRenderer = createClass({
...
-				{this.renderPages()}
-			</div>
-			{this.renderPageInfo()}
-			{this.renderPPRmsg()}
-		</div>;
+		return (
+			<React.Fragment>
+				<div className='brewRenderer'
+					onScroll={this.handleScroll}
+					onMouseOver={this.prepareScroll}
```
Rae2che5: We don't need this (and `onMouseOut`) any more, right? In particular, the functions no longer exist...

calculuschild:
You are right. It is now corrected.

Rae2che5:
Thanks, LGTM now. Let's see how this goes.

Issue #683, opened 2018-05-09
Table error when saving as pdf


ksogard:
*Delete the above section and the instructions in the sections below before submitting*

## Description

"Class tables" and "Descriptive Text Boxes" borders are glitching out when saving as a pdf. See attached images and code below:

![image](https://user-images.githubusercontent.com/39137116/39837192-576eab4c-539b-11e8-80ca-0d3937e260e0.png)

![image](https://user-images.githubusercontent.com/39137116/39837216-6ea9e506-539b-11e8-8fee-2799cbf2068e.png)

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

PASTE BREW CODE HERE
<div class='classTable wide'>
##### The Concierge
| Level | Proficiency Bonus | Features | Cantrips Known | Spells Known | 1st | 2nd | 3rd | 4th | 5th | 6th | 7th | 8th | 9th |
|:---:|:---:|:---|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| 1st | +2 | Genetic Banishing | 3 | 1 | 2 | — | — | — | — | — | — | — | — |
| 2nd | +2 | Spiritual Biochemistry | 3 | 1 | 3 | — | — | — | — | — | — | — | — |
| 3rd | +2 | Civil Alchemy | 4 | 2 | 4 | — | — | — | — | — | — | — | — |
| 4th | +2 | Ability Score Improvement, Consecrated Biochemistry | 4 | 3 | 4 | 1 | — | — | — | — | — | — | — |
| 5th | +3 | Seismological Divination | 4 | 3 | 4 | 2 | — | — | — | — | — | — | — |
| 6th | +3 | Ability Score Improvement | 5 | 4 | 4 | 3 | 1 | — | — | — | — | — | — |
| 7th | +3 | Civil Alchemy | 5 | 4 | 4 | 3 | 2 | — | — | — | — | — | — |
| 8th | +3 | Ability Score Improvement, Biochemical Sorcery | 6 | 4 | 4 | 3 | 2 | — | — | — | — | — | — |
| 9th | +4 | Genetic Banishing | 6 | 5 | 4 | 3 | 2 | — | — | — | — | — | — |
| 10th | +4 | ─ | 7 | 5 | 4 | 3 | 2 | — | — | — | — | — | — |
[...]

PR #700, opened 2018-05-22
Fix border-image bug on PDFs

calculuschild:
I think this might be caused by a new bug in Chrome or something since it seems to have popped up in GMBinder as well in the last couple weeks. See #683

Anyway, the error only occurs on the Class Table and Descriptive Text Box, which are also the only ones to use `border-image-repeat: round`. I changed it to `border-image-repeat: stretch` which solves the graphical issue and doesn't affect the appearance otherwise in this particular case.

### DIFF
--- a/client/homebrew/phbStyle/phb.style.less
+++ b/client/homebrew/phbStyle/phb.style.less
@@ -402,7 +402,7 @@ body {
 	border              : initial;
 	border-style        : solid;
 	border-image-outset : 25px 17px;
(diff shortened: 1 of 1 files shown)


Rae2che5:
The images in question are these:

image

image

Even though `stretch` might deform them a bit, `round` seems like definitely the wrong tiling method -- we never want to repeat those images. As such, merging -- thanks for the sleuthing!

Issue #695, opened 2018-05-17
Share page is broken on master


calculuschild:
## Description

The */share* page is broken on master. It seems to be generating a `<div class="homePage page">` rather than `<div class="sharePage page">` with components of an editor pane and everything. The result is a garbled display of the document scrunched over to one side of the screen and the snippet bar along the top where there should be nothing. I'm guessing this is some confusion with the routing in [Homebrew.jsx](https://github.com/stolksdorf/homebrewery/blob/a634b76117592158e348ee7843045222093c96c3/client/homebrew/homebrew.jsx) but I can't pinpoint why this is happening.

The issues also occurs with the */changelog* page.

**Edit:** After further searching, it looks like this is actually occurring on *every* page. The expected div is being replaced by `homePage` in each instance; it just isn't readily apparent in the `/edit/` page for example, because both `homePage` and `edit` are similar enough in structure that they both contain the `splitpane`, `editor` and `brewRenderer`.  It looks like somehow the [default URL](https://github.com/stolksdorf/homebrewery/blob/a634b76117592158e348ee7843045222093c96c3/client/homebrew/homebrew.jsx#L77) is overriding everything else.  Is this an issue with the latest Pico-Router maybe?

I'd like to figure out what happened here so we can work on pushing out a new version to live at some point.

@stolksdorf @Rae2che5 Are either of you aware of a particular commit that caused this regression?

**Share Link**:
[http://staging.homebrewery.naturalcrit.com/share/S1U_UU5Af](http://staging.homebrewery.naturalcrit.com/share/S1U_UU5Af)

and

[http://staging.homebrewery.naturalcrit.com/changelog](http://staging.homebrewery.naturalcrit.com/changelog)

LeviRosol:
Do you also see the weird `;` at the bottom of the document and the `Warning: Expected server HTML to contain a matching <div> in <div>.` thrown by libs.js when running locally? I'm not sure if these are related, but it seems like a place to start - as if there's an unclosed div or something somewhere.

FWIW, if I revert back to using `React.` as shown in this commit ( https://github.com/stolksdorf/homebrewery/commit/f18a181e2ec5e372121149b3fe9bafbccbe5884d )

along with reverting back to the old version of React, I can get the page to render the splitter correctly. There isn't any brew data showing, but I assume there are related changes to get it to render correctly.

Additionally, the version of modules will continue to be an issue if they are not specifically set in the package.json. It might be a good idea to remove the `^` in each one

[...]

PR #701, opened 2018-05-23
Fix regression with Pico-Router update

calculuschild:
This variable name changed with the v2.0.0 of Pico-Router and it wasn't changed here (commit [55c5294](https://github.com/stolksdorf/homebrewery/commit/55c529473a8b85f6f27eb8769ef7bcfb854827b8#diff-b9cfc7f2cdf78a7f4b91a753d10865a2)). This was causing every page to display a `<div class="homePage page">` rather than the appropriate `editPage` or `sharePage`. It just wasn't immediately noticeable since `homePage` and `editPage` are almost identical. The problem is clearly visible on `sharePage` however. See #695 .

`homebrew.jsx` was set to always render a defaultUrl of ` ` (empty string) which creates a `<div class="homePage page">` by default, overwriting the expected `<div>` on for the current url.

This is the only regression I have seen, and we have added several helpful fixes. What do we think about moving forward with a new release?


Rae2che5:
Ah, nice catch. This fix makes sense to me.

> This is the only regression I have seen, and we have added several helpful fixes. What do we think about moving forward with a new release?

That also makes sense. I'm not sure what the process is -- I guess @stolksdorf is the only one with access to staging/prod environments?

PR #734, opened 2018-08-20
Basic docker configuration

Rae2che5:
As suggested by @schemen, this adds a basic `Dockerfile` as well as a `docker-compose.yml`. I'm sure they can be refined further, but even so, I'm able to do `docker-compose up` to get `http://localhost:8000` as a rendered version of the site.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:8
+
+# Create app directory
+WORKDIR /usr/src/app
+
+# Bundle app source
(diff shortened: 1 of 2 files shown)


schemen:
I'll have a look and see how to we could improve it :)

calculuschild:
I'm not up to date on Docker, but it looks like it runs fine on my end.

Rae2che5:
Closing in favour of #737.

PR #735, opened 2018-08-20
Fix pooled react event property access

### LINKED ISSUE #733: Scrolling exception: Cannot read property 'scrollTop' of null
## Description

For context, I have a lot of spare hosting infrastructure and I am trying to dockerize & rehost several applications so we have some back up in case or offer the services for other dnd tool developers.

I am trying to host it, because it seems that the original author is not updating the website anymore.  I have dockerized the application, currently running at https://homebrewery.hoengg.tech with version 2.8.0. The main issue is that I do not understand react/node as much as basic work with it, I package applications into containers and run them.

I am currently running into the problem when I scroll, I receive the error:

`Uncaught TypeError: Cannot read property 'scrollTop' of null`

**How to reproduce** :
The introduction letter is enough already, if visit https://homebrewery.hoengg.tech you can try to scroll on the preview window and it will immediately crash, you have some lines of console output.

Rae2che5:
This resolves #733 -- thanks for reporting it, @schemen!

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -58,8 +58,9 @@ const BrewRenderer = createClass({
 	},
 
 	handleScroll : function(e){
+		let target = e.target;
 		this.setState((prevState)=>({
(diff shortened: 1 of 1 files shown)


calculuschild:
This looks fine. Good catch.

PR #737, opened 2018-08-21
Add dockerization and example compose file

schemen:
This is my take on #734. I have compressed the docker file to reduce the amount of layers and fixed the compose volume definition. The .dockerignore file is used so that when copying the files into the container it ignores npm related things (not sure this is related here as we use yarn?).

Also, the docker.json file is here to define the mongodb connection, I saw in #734 that you can use environment variables, if this holds true we can safely ignore this file and I will remove it.

This is the config/dockerfile that is currently running at https://homebrewery.hoengg.tech/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,3 +13,5 @@ services:
       MONGODB_URI: mongodb://mongodb/homebrewery
     ports:
       - "8000:8000"
+volumes:
+  mongodata:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,7 @@ services:
   mongodb:
     image: mongo:latest
     volumes:
-      - /your/data/location:/data/db
+      - mongodata:/data/db
   homebrewery:
     build:
       context: .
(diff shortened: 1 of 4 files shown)


schemen:
@calculuschild @Rae2che5 You guys might be interested in this PR :)

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: '2'
+services:
+  mongodb:
+    image: mongo:latest
+    volumes:
+      - /your/data/location:/data/db
```
Rae2che5: `/your/data/location` is just a placeholder string here? The [reference](https://docs.docker.com/compose/compose-file/compose-file-v2/#volume-configuration-reference) suggests that it may be best practice to have a top-level named volume (`database`?) and mount it here as `database:/data/db`.

schemen:
I'll take care of the requested changes in a bit! :)

schemen:
@Rae2che5 I have updated the PR

Rae2che5:
Thanks, that all looks good, with the exception of a missing volume declaration that I added in 30404ef. Merging.

schemen:
Ah thank you! I used so many bind mounts I forgot about that :)

PR #792, opened 2018-11-07
Updated packages

calculuschild:
Just updating PRODUCTION with dependencies/packages/CircleCI settings/etc. that never got pushed.


calculuschild:
Ok, gonna commit this to live and see if it actually deploys now that CircleCi tests are passing.

PR #797, opened 2018-11-08
Fix live meta data not working

calculuschild:
Fix metadata not working. Some earlier linting caused the handleFieldChange and handleSystem functions to not have access to event data. Lint is still happy.


calculuschild:
Gonna push this small change live now to fix more errors. Looks like our Master branch has a few of these small errors that are hidden in menu behavior, etc.

Issue #509, opened 2017-10-05
Statblock Format Improvement


SergentBullock:
In the perfection process, you might want to look at rectifying the Statblock background texture. Statblocks actually have a ragged edge to them on the "parchment" section for many monsters. 

This is best seen on the Beholder statblock inside the Monster Manual. Not all monsters have it or in great quantities, (the more badass they are the more likely it is that they seem to get it. Compare the Beholder to the Cyclops for example.) but it's a very minor detail worth looking into. Possibly the wide statblocks could get this treatment, since they're usually imposing, strong monsters?

### Additional Details

**Share Link** :

or

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

PASTE BREW CODE HERE

</pre></code></details>

calculuschild:
Added to the V3 changelog finally. Closing.

PR #805, opened 2018-11-12
Updating phb-style css

RKuerten:
Images are not that great, but things are visible.

- Changed first letter and line after h1 changes:
first-changes
After update - current state - book for comparison

W and Y as first letters are specially troublesome, because a decent upper part of them is cropped by the padding of the h1. Did my best to fix it while not dragging the other letters down. 

Made first line use `font-variant : small-caps` to keep it in line with the PHB. It messed up a bit with the first-letter, it needs clean-up.

- Monster stat block h3 and background changes
second-changes
After update - current state - book for comparison

Changed the h3 header inside the monster stats to use `font-variant : small-caps` and some other changes as well to keep it in line with the monster manual stat blocks.

Applied the phb background image to the block, following the background texture of the monster manual stat blocks. It is hard to see with original PHB background, but it is pretty clear with a different one.
Related issue: #509


AlexeySachkov:
The result looks nice!

Looks good to me, but I don't have contributor permissions, so you need to wait for @calculuschild or @Rae2che5 to get this PR merged.

As I understand, you tested this using default brew code from the main page of the site - is it true? If so, may be it is better to also check some user-created brews just to be sure.

RKuerten:
@AlexeySachkov 
I regularly use the Monster Manual changes (using style, etc). It does not change the size of a block inside a page as it is, but the background makes it heavier to render into pdf, etc.

The h1+p changes are hard to code within the editor, as they are made using less and not pure css. Depending of the content of the paragraph after h1, it will change the brew.

stolksdorf:
Looks good!

There is a font present that should be ScalySans but in small caps; `font-family   : ScalySansSmallCaps;` Does using that get it closer to the source material or does the `font-variant`?

RKuerten:
They are quite close. Take a look:
image

Top uses the same css of the pull request.

Bottom uses the following css:
```
.phb hr+blockquote h3{
       font-family    : ScalySansSmallCaps;
       font-weight    : bold;
       font-size      : 19px;
}
```

In my opinion, the top one looks better (and closer to the books), but I'm open to opinions and suggestions.

ericscheid:
I'll cast a dissenting opinion on the lighter small caps version - it looks like the capital "A" is in one font size, with the subsequent "ctions" typed in all-caps at a smaller font size and thus a lighter weight (i.e. smallcaps are faked).

From wikipedia: 
> Well-designed small capitals are **not simply scaled-down versions** of normal capitals; they normally **retain the same stroke weight** as other letters and have a wider aspect ratio for readability.

It might well look closer to what is in the book but that might be poor typography on their part.

calculuschild:
I think the changes make sense (small caps, background texture, etc.), but while we are at it, should we look at some other differences?

For instance, if we compare the current (and updated) styles against the actual monster manual (I scanned the orc page so it's at the proper angle here). When we scale things up to the same size (the "Orc" title is the same height in pixels, and the other fonts seem to be just about right as well), there are some clear differences, namely the width of the block and the margins around the text.

Updated - Current - Book
image

In the Monster Manual, the text in the stat blocks lines up very closely with the text in the main body columns so that is one thing we might look into if we are going for accuracy.

Book - Current
image

Also, if anyone cares to look, I also had a ["fancy" statblock](https://homebrewery.naturalcrit.com/share/SyMgMS0Lg) I made about a year ago. I still missed the small caps, but we can definitely make the blocks look nice on the borders especially:

image

RKuerten:
I favor your proposed changes, @calculuschild. I thought of including them in the initial pull request, but since those changes will effectively alter the formatting of many brews out there, I left them out.

In my opinion, the font-size is close, but a little bigger then the books. I tried to reproduce Strahd's stat block (which takes a complete page size) using the site, and it extrapolates the page by quite a lot. Didn't make more tests besides this one, but I got the general feeling that the font can be reduced.

To make it fit, I used `font-size: 102%`. It inherits its size from its parent element, which I suppose is the .phb sheet? 

Here it is my version (without the font-size changes) of the "closer" statblock using your "fancy" statblocks as the base, @calculuschild: 
image

Code for it: 
```
.phb hr+blockquote {
    background-color: transparent;
    background-image: url(http://i.imgur.com/NibirYB.png);
    background-repeat: no-repeat;
    background-position: -400px -10px;
    border-image: url(http://i.imgur.com/xNXvT71.png) 18 round;
    border-image-outset: 0px 2px; 
    padding: 0px;
    padding-top: 15px;
    border-width: 8px 6px;
}
```

PR #811, opened 2018-11-27
Added in a function to sanitize script tags specifically

stolksdorf:


shared/naturalcrit/markdown.js
```diff
@@ -13,6 +13,11 @@ renderer.html = function (html) {
 	return html;
 };
 
+const sanatizeScriptTags = (content)=>{
+	return content
+		.replace(/<script/g, '&lt;script')
+		.replace(/<\/script>/g, '&lt;/script&gt;');
```
Rae2che5: This sanitization is not sufficient. At the very least, we should add the `i` regex flag (as `<ScrIPt` is a perfectly valid way to start a script, and will not be caught at the moment).

What are we trying to guard against? Is the concern that a user may accidentally lock themselves out of a brew, or that a malicious author may share a link to a brew that runs scripts with their victims? In the latter case, there are many other things we need to worry about (for example, `onerror` handlers for `<img`...). If this is our concern, we'd be better off with a known-safe whitelist: the HTML parsing should only accept known tags and attributes, and strip/escape everything else.

stolksdorf: Good catch with the case-insensitivity. 

Two cases we can worry about in regards to script injection. One being a user wanted to add some functionality to their brew in some fashion. The other being someone trying to maliciously run code on other people's computers. The latter case being incredibly difficult to protect against 100% especially when the site uses user-generated content. 

So this change was trying to help out the first case and limit the footguns having access to `<script> tags can cause. 

We can explore the latter case, but that will require a different approach (as you mentioned), with more in-depth processes.

PR #815, opened 2018-11-30
Added middleware to force all routes to be https

stolksdorf:
Should have done this a long time ago.


stolksdorf:
As a heads up, I've already configured the server to handle https requests, so this is just forcing all requests to be secure.

AlexeySachkov:
Could you please also update the README.md and Github project description? I guess that patreon page should be also updated to link to https version

stolksdorf:
I can, but it's not needed, all http links will resolve to https.

PR #817, opened 2018-12-04
Updated admin page

stolksdorf:
- Updated the information and capability of the admin panel
- Removed the insecure `ADMIN_KEY` query param on the admin routes, instead using basic auth.
- Added in the UI and back-end for removing brews that are under a tweet in length and older than 30days. This is a temporary measure because the database is becoming quite full. We need a bit more time to implement a new storage solution.


calculuschild:
@stolksdorf , is there an easy way to view this page without being able to log in? I know the Admin page is for your own use but I'd like to see how it is set up.

stolksdorf:
You can clone down that branch and run it local, the default user/pass is in the codebase (it's obviously changed in production environment variables. 

But here's a quick screen shot of it 
image

It's pretty basic, but should be helpful

PR #827, opened 2019-01-18
Fix Recent Brews Tab

calculuschild:
Refactored the "Recent Items" navbar component. Greatly simplified the code to about 70% length and made it much easier to read. Results in the same thing.

Oh. And it works again. Recent brews should be showing up fine now.


calculuschild:
test

PR #839, opened 2019-02-07
Fix delete buttons

calculuschild:
Delete should now first remove the current user as an editor, and only delete the file if he was the last editor. Currently, anyone who views your brew is added as an editor and if they delete it, the whole brew is deleted for everyone.


calculuschild:
Had some time to test on my machine. Seems to be working fine. Merging!

PR #842, opened 2019-03-12
Implementing magic item snippet from Issue 671.

### LINKED ISSUE #671: A Magic Item Template
## Description

This would be a useful template for any dungeon master that would want to create custom magic items. When I first started, I had to figure out how to format it on my own. I personally believe it would be a great help to everyone.

**Magic Item Template** : <details><summary>Click to expand</summary>

<pre><code>

#### Doorknob of Niceness
*Wondrous condiment, quite rare (requires attunement)*

This knob is pretty nice. It opens the door you put it on.

#### Longsword of Leering
*Weapon (toy sword), really common (requires attunement by a child)*

This sword makes you scowl at anyone you see. Completely harmless.

</code></pre></details>

chrisbrickhouse:
I am trying to resolve #671 and think this should work, but haven't gotten around to testing it because of some issues getting nodejs installed. Will update once tested.


chrisbrickhouse:
I've managed to get nodejs running, however the current commit results in a server error and needs to be fixed. I believe the problem is in building the source as the server gives the error `Cannot find module './build/homebrew/bundle.js'` which should be built by `npm install`. I'll look into this more to see what is going wrong.

calculuschild:
You normally have to re-build every time you make code changes. You can run `npm build` or if that doesn't work, `npm run-script build`. That should generate the bundle.js file for you.

chrisbrickhouse:
I managed to get it working. The bundle.js file wasn't created because commit 9335608 had a syntax error that caused `npm build` to fail. This was fixed in commit f71a072.

PR #843, opened 2019-03-14
Readme update.

chrisbrickhouse:
I've updated the installation instructions in the Readme based on the steps I needed to get a local version running on my linux machine. I also cleaned up the formatting a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ Third, you will need to add the environment variable `NODE_ENV = local` to allow
 
 You can set this temporarily in your shell of choice:
 * Windows Powershell: `$env:NODE_ENV="local"`
-* Windows CMD: `set NODE_ENV=production`
-* Linux / OSX: `export NODE_ENV=production`
+* Windows CMD: `set NODE_ENV=local`
+* Linux / OSX: `export NODE_ENV=local`
 
 Fourth, you will need to install the program and run it using the two commands:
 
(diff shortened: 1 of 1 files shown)


jlwhelan28:
Hello, new face here. Incoming JS noob question... And likely an XY problem y'all can ignore. 

@chrisbrickhouse Commenting out `forcessl.mw.js` was also the only way I could get a localhost up and running without more knowledge of JS. This function seems to simply redirect `http://` to `https://`, but interferes when running locally somehow... 

I thought the `https` issue was not having a certificate for localhost, but after generating a certificate with openSSL, and creating the Express server as below, I was able to get past the secure connection error, but still can't connect to localhost properly. Then after omitting `forcessl.mw.js`, I can connect to localhost through both `http` and `https`, with either the below server configuration or the original use of `Express`

Basically what I'm trying to understand is why does commenting out the above require work?

```JavaScript
const PORT = process.env.PORT || 8000;

https.createServer({
	key:fs.readFileSync('server.key'),
	cert: fs.readFileSync('server.cert'),
}, app)
.listen(PORT, function () {
	console.log(`Homebrewery listening on port:${PORT}`)
})

// app.listen(PORT);
// console.log(`server on port:${PORT}`);
```

datdamnzotz:
Might want to just update the readme to say something like

Add an environment variable like this:
`NODE_ENV=local` 
bypasses the force SSL.

As seen here in the code https://github.com/naturalcrit/homebrewery/blob/master/server/forcessl.mw.js#L2

tcyrus:
@calculuschild Sorry for bothering you, but I'm curious on the status of this PR

calculuschild:
I have accepted this PR. Changed the text to encourage using environment variables rather than commenting out lines of code.

PR #845, opened 2019-03-21
Prevent offset split tables

### LINKED ISSUE #844: h1 header makes Split Table columns unaligned if said table is near the end of the page
This is a strange one, see [THIS](https://old.reddit.com/r/homebrewery/comments/b0a1vp/cant_figure_this_out_please_help/) reddit thread for a detailed explanation of what happened.

chrisbrickhouse:
Fixes #844 by removing padding from paragraphs that follow tables which caused split tables to be offset.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippets/snippets.js
+++ b/client/homebrew/editor/snippetbar/snippets/snippets.js
@@ -205,7 +205,7 @@ module.exports = [
 				icon : 'fa-th-large',
 				gen  : function(){
 					return [
-						'<div style=\'column-count:2\'>',
+						'<div class=\'split-table\' style=\'column-count:2\'>',
 						'| d10 | Damage Type |',
 						'|:---:|:------------|',
 						'|  1  | Acid        |',
--- a/client/homebrew/phbStyle/phb.style.less
+++ b/client/homebrew/phbStyle/phb.style.less
@@ -349,6 +349,7 @@ body {
 	//Text indent right after table
 	table+p{
 		text-indent : 1em;
+        padding     : 0em;
 	}
 	// Nested lists
 	ul ul,ol ol,ul ol,ol ul{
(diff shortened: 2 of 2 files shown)


ericscheid:
Yeah nah, this is bad.

Consider this case:

    |table|
    |-----|
    |data|

    paragraph-1 Lorem ipsum aliquam lectus orci, adipiscing et, sodales ac. 
    paragraph-2 Maecenas volutpat, nisl et dignissim pharetra.

That paragraph-1 won't have any padding-bottom ... and since all paragraphs have a negative margin-top the paragraph-2 will encroach upon paragraph-1, overlapping it.

Some of the .phb css styles are a tad weird, and so just patching here and there with case-by-case css tweaks is gonna just turn the big ball of mud into something .. worse.

chrisbrickhouse:
A good catch, I thought the `+` selector would only select the immediately adjacent sibling but I guess not. I made some more changes to account for it that I think winds up making the css slightly better.

I edited the split table snippet so that it adds in a div of the (new) class `split-table`. Then I changed the css so that `<p/>` tags of class `split-table` have 0em padding. Paragraphs will only be a member of this class inside this snippet (or if someone adds the class to something else for some strange reason) so it shouldn't cause problems anywhere else.

A side benefit of this is it makes split tables slightly more maintainable and modifiable. If we (or a user) wants to fiddle with things only in split tables, now there's a class for that.

Gazook89:
Can this PR be closed?  It seems that in #844 a fix is given, and also made irrelevant by the new parser.  No triple-tick marks are needed to force a column break, as the two tables should balance across the two columns without it.

calculuschild:
This has been fixed through improvements elsewhere (though I'm not 100% sure when). Even with backticks this seems to work on Legacy just fine now.

PR #859, opened 2019-12-18
fix fail if first #heading not end with \n

ericscheid:


calculuschild:
What was the issue on this one? I can't figure out how to reproduce the bug.

ericscheid:
The first heading in a fresh unsaved brew is usually used as the saved brew title.

Usually.

Do this though .. 

1. go to https://homebrewery.naturalcrit.com/new
2. type `# my new brew` (without a terminating CR)
3. hit save
4. observe that the brew title is simply `# `

A more broken example:

1. go to https://homebrewery.naturalcrit.com/new
2. select "A4 Page Size" from the Print menu to insert the CSS
3. hit CR to get a new line and type `# my new brew` (without a terminating CR)
4. hit save
5. observe that the brew title is now all the css and none of the heading  
`<style> .phb{ width : 210mm; height : 296.8mm; } </style> #`

ericscheid:
This is because `text.indexOf('\n', titlePos)` will return -1 if there is no `\n` in the input.

Then, the `-1` is input into `text.substring(titlePos + 2, ending)`, which JS will helpfully swap the parameters (because start is now greater than ending), and thus effectively does `text.substring(-1, titlePos + 2)`.

calculuschild:
I see. In that case, we may want to further check for "style" blocks because the "more broken example" will still give weird results even without a Header:

1. go to https://homebrewery.naturalcrit.com/new
2. select "A4 Page Size" from the Print menu to insert the CSS
3. hit CR to get a new line and type `hello!` (**with** a terminating CR)
4. hit save
5. observe that the brew title is now `<style>`

Do you want to look into that @ericscheid ?

ericscheid:
Hmm .. that's an unfortunate case of the fallback condition of not finding a heading, i.e. just use the first line of the brew for the title.

    const titlePos = text.indexOf('# '); // -1
	if(titlePos !== -1){ // false
        // ...
        } else { // no `#<space>` found
		return _.find(text.split('\n'), (line)=>{
			return line;
		});
	}

It did get me thinking of what happens if there's an #id css selector within that `<style>` block .. answer: nothing, because it would be `#<some-id>` and not `# <some-id>` (i.e. no space following the octothorpe)

But then again ... a brew with text like `this is an octothorpe: # isn't that a fun word` would also not do the correct thing because it would set the title to either `isn't that a fun word` (wrong because the heading marker needs to start on a new line) or to `this is an octothorpe: # ` (the original -1 bug due to no `\n` following the `#`).

ericscheid:
The first-line-fallback weirdness would be hard to work around. Would have to render the brew to html, and then use `innerText.split('\n')[0]` to extract the first line of actual output text. Seems overkill.

The other buggy case might be solved by finding by regex not literal .. maybe `/^#+ (.+)$/` ?

That'd also solve the original bug too.

calculuschild:
> Would have to render the brew to html, and then use innerText.split('\n')[0] to extract the first line of actual output text. Seems overkill.

Maybe not. That HTML is already rendered at that point. It might not be a bad idea to just use the existing rendered HTML rather than trying to decipher the editor box...?

ericscheid:
Certainly the html is rendered on the client side, but is the html is rendered on the server side?

ericscheid:
Given the brew is already rendered in the editing UI, the js that builds the save request can default the title to
`document.getElementsByClassName('pages')[0].innerText.split('\n')[0]`
which grabs the first line of rendered visible text.

We probably should put an upper limit of the length of that title string too (IIRC the model doesn't impose a limit, but stupidly long titles will break the UI elsewhere).

We dont need to guard here against malicious titles so long as we're always careful to defang the user-provided string to be html-escaped elsewhere (e.g. in the published directory of brews). A malicious title might be `&lt;script>alert(1)&lt;/script>`

PR #896, opened 2020-02-08
Correctly test for columnSpan support

### LINKED ISSUE #850: Only show RenderWarnings if features are really missing
Users using a standards-abiding browser (such as Firefox) shouldn’t have to see the warning. Instead of testing if the user uses Chrome, the actual features should be tested. Instead of

https://github.com/naturalcrit/homebrewery/blob/f28ed3d52ebcf048244d137252b1efd9f951a9a2/shared/homebrewery/renderWarnings/renderWarnings.jsx#L24

we could [test](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection#CSS) if the browser supports all necessary features and only display the warning if it doesn’t.

flying-sheep:
Fixes #850

### DIFF
--- a/shared/homebrewery/renderWarnings/renderWarnings.jsx
+++ b/shared/homebrewery/renderWarnings/renderWarnings.jsx
@@ -20,17 +20,14 @@ const RenderWarnings = createClass({
 		window.removeEventListener('resize', this.checkWarnings);
 	},
 	warnings : {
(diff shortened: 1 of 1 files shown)


calculuschild:
Superseded by #937

PR #912, opened 2020-03-15
Fix: Do not force SSL when running in a Docker container

brandon-fryslie:
This PR includes several minor changes related to Docker.  First off, thank you for keeping the Dockerfile and docker-compose.yml file around.  It makes running the application much easier for people who use docker.

Changes:
- Add a short README with basic information on how to run the application using docker
- Minor refactoring of the Dockerfile to improve image layer caching.  By rearranging the commands, we download the dependencies
- Make a small change to the application code to prevent forcing SSL redirect when running in Docker.  I considered using `process.env.NODE_ENV !== 'production'`, I can make that change if that is preferable.  This was preventing the application from being accessible when running in a docker container, and this fix is required to use the application in docker.  It's currently broken.

Thank you!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,19 @@
 FROM node:8
 
+ENV NODE_ENV=docker
+
 # Create app directory
 WORKDIR /usr/src/app
 
-# Bundle app source
+# Copy package.json into the image, then run yarn install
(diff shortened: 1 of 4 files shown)


calculuschild:
This looks good to me. Just going to check it over but I think I can merge this without too much issue. 👍

Issue #1085, opened 2020-11-02
Inconsistent list-item sizing


ericscheid:
Ordered lists get rendered as either rendered text directly within a `<li></li>`, or rendered text wrapped inside a `<p></p>` inside `<li></li>`, depending on whether a blank line exists between list items. This is a  Marked behaviour and not the issue.

When lists are embedded within a `<blockquote>`, the .phb CSS adjusts the font-family and font-size to match the `<p>` elements within a block quote. That is, they become ScalySans 11.98px.

_However_, the .phb css rules are incomplete, in that an `blockquote ol li` has a different font-size than `blockquote ol li p`, `blockquote p`, `blockquote ul li`, and `blockquote ul li p`.

Examples of the various cases here:
https://homebrewery.naturalcrit.com/share/h-ppujWlgKDG

ericscheid:
In bundle.css in the browser, there is this css rule:

```
.phb blockquote p, .phb blockquote ul {
    font-size: .352cm;
    line-height: 1.1em
}
```

There is no similar rule for `ol`. There should either be a rule for `ol`, or we change that "ul" to "li".

ericscheid:
Here are sample brew code do demonstrate the different behavior (because that share link will go poof eventually)

```
Observe the different line lengths due to different font-size:

> 1. THIS IS TEXT OF SAME LENGTH
> 1. THIS IS TEXT OF SAME LENGTH
> 
> 1. THIS IS TEXT OF SAME LENGTH
> 1. THIS IS TEXT OF SAME LENGTH
> 1. THIS IS TEXT OF SAME LENGTH
> 1. THIS IS TEXT OF SAME LENGTH


This doesn't happen with a non-blockquoted ordered list:

1. THIS IS TEXT OF SAME LENGTH
1. THIS IS TEXT OF SAME LENGTH

1. THIS IS TEXT OF SAME LENGTH
1. THIS IS TEXT OF SAME LENGTH
1. THIS IS TEXT OF SAME LENGTH
1. THIS IS TEXT OF SAME LENGTH

This doesn't happen with unordered lists:

> - THIS IS TEXT OF SAME LENGTH
> - THIS IS TEXT OF SAME LENGTH
> 
> - THIS IS TEXT OF SAME LENGTH
> - THIS IS TEXT OF SAME LENGTH
> - THIS IS TEXT OF SAME LENGTH
> - THIS IS TEXT OF SAME LENGTH
```

[...]

PR #1108, opened 2020-11-22
line-heights consistent in bq li (#1085)

ericscheid:
applies same line-height for both ol and ul inside a blockquote quote. See #1085.


calculuschild:
Thanks for this fix!

Issue #653, opened 2018-04-10
Not clear how to rename/edit metadata of brews


calculuschild:
Several users have asked how to change the title of their brews. This feature is hidden away behind the (☰) button in the toolbar. Perhaps we can find a more obvious way to do this. I suggest:

- Change the "☰" icon to a "ⓘ" info icon.
- Allow the user to edit the title of the brew by clicking directly on the title in the page heading

#622, #328

ericscheid:
Also add to the FAQ #810

ericscheid:
The current icon is from font-awesome using `<i class="fa fa-bars"></i>`, which is line 49 of /client/homebrew/editor/snippetbar/snippetbar.jsx

Alternatives are [fa-info-circle light](https://fontawesome.com/icons/info-circle?style=light), [fa-info-circle regular](https://fontawesome.com/icons/info-circle?style=regular), and [fa-info-circle solid](https://fontawesome.com/icons/info-circle?style=solid). The first two require FA Pro.

ericscheid:
homebrewery is using FA Classic, not pro, so this is what the solid info-circle looks like:

![fa-info-circle](https://user-images.githubusercontent.com/2605467/99919866-24b17d00-2d74-11eb-9851-64d695aa9b8d.png)

calculuschild:
I feel like maybe we need to straight up add some text there as well. We have space. Just make it a complete "brew info" button.

ericscheid:
> I feel like maybe we need to straight up add some text there as well. We have space. Just make it a complete "brew info" button.

Seems doable, lemme tweak my tweak

ericscheid:
The brew-meta button is now an "Properties" button with an "i" in circle icon:
![image](https://user-images.githubusercontent.com/2605467/111172179-b67e1800-85f9-11eb-985c-4983b775f53f.png)

This should go a good way towards fixing the original issue.

PR #1109, opened 2020-11-22
use fa-info-circle instead of fa-bars

ericscheid:
See #653

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -42,9 +42,10 @@ const Snippetbar = createClass({
 
 	renderMetadataButton : function(){
 		if(!this.props.showMetaButton) return;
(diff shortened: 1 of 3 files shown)


calculuschild:
See if you can make this a full button with text "Brew Info" in the same style as the snippet buttons, and we can merge this.

ericscheid:
OK. 

I can't see any built in styles directly suitable (the other items in the editor menu bar are all dropdown groups).

I can hand-craft the necessary css rules, will first take a closer look around for existing styles (is hb built upon a style library, or are they all hand-crafted?).

calculuschild:
Everything is hand-crafted. Even though the snippets are drop downs you can still borrow the css from them to get the look right.

calculuschild:
If you want to just apply the same class as the other snippet buttons, the CSS should be handled fine. We can merge this in with v2.10.7

Issue #1096, opened 2020-11-09
Fill out missing style snippets


calculuschild:
Once the new rendering pipeline is in place, we can start inserting more requested code snippets and styling features that users have created or requested for a long time. Part of the reason these have not been added until now, is the over-dependence on random markdown combinations to create textboxes, which leads to lots of conflicting styling issues in the CSS. Further, the use of invalid markdown due to an outdated markdown parser makes these conflicts even more common.

Blocked by #1093. Once that is complete, more of these tasks can be finished out.

This issue will collect the sub-tasks and related issues.

### Clean up existing styling and snippets
- [x] Inconsistent list font size in block quotes #1085
- [x] Background decoration on Class Table #302 

### Include snippets for common missing PHB items
*(collect from issues, a lot of these are already figured out)*
- [x] Table of Contents
- [x] Title page #640, #1334
  - [x] Add title fonts #2061
- [ ] Adventure Flowchart #936 
- [x] Monster Stat Block without Box #914 
- [ ] Notes on paper scraps (probably best handled by user images) #826 
- [x] Third-caster class table #1079, #191 
- [x] Magic Items #842, #671 
- [ ] ...etc.

### Include snippets for user convenience
*(collect from issues)*
- [x] Code style (often used for tutorial documents here)
- [x] QR Code #539 
- [ ] Credits to Homebrewery #283 

### Correct other existing style bugs
- [x] Line spacing in bulleted lists is inconsistent #1120
- [x] Long words are not wrapped #1136

Gazook89:
The "Page Number" snippet contains both a page number component and a 'footer' component, and I think they should be separated....or, the footer component be added to the "Auto Incrementing Page Number" snippet as well...it's weird that the footer is in one, but not the other.

calculuschild:
I don't mind separating them. That makes sense to me.

calculuschild:
User u/Kaiburr_Kath-Hound on Reddit (with help from others) has completed several of these and notified me that he would like to contribute them as official snippets. May need slight tweaks but 95% of the work is done:

https://www.reddit.com/r/homebrewery/comments/qubn28/phbstyle_including_title_page_credits_page/

ericscheid:
No longer blocked by #1093 as [on 12 Sep 2021](https://github.com/naturalcrit/homebrewery/issues/1093#event-5288558482).

5e-Cleric:
Cover page done in #2649, approval pending.
[...]

PR #1120, opened 2020-11-27
make all li inside bq smaller font for backwards compatibility

ericscheid:
Some brews were relying on the smaller font-size for column wrap
so make all `<li>` within `<bq>` smaller (including if inside `<p>`).

Some brews might now have extra space (vs some brews now overflowing the space they had before)
https://github.com/naturalcrit/homebrewery/issues/1085#issuecomment-734322530


calculuschild:
Is this behavior any different than just using the `ul` rule or whatever it was before? I'm leaning towards just fully reverting that change in case this extra `p` has any other unforseen side effects.

calculuschild:
Alright, I have just reverted the original change back to `ul` in favor of this to avoid any other side effects. Let's keep the related issue open though and make it part of the #1096 Epic.

PR #1122, opened 2020-11-28
Legacy Rendering Pipeline

calculuschild:


client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -110,13 +114,16 @@ const BrewRenderer = createClass({
 	},
 
 	renderPage : function(pageText, index){
-		return <div className='phb' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: Markdown.render(pageText) }} key={index} />;
+		if(this.state.renderer == 'legacy')
+			return <div className='phb' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(pageText) }} key={index} />;
+		else
+			return <div className='phb' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: Markdown.render(pageText) }} key={index} />;
```
AlexeySachkov, line 120:
```suggestion
                let HTML = "";
		if(this.state.renderer == 'legacy')
			HTML = MarkdownLegacy.render(pageText);
			
		else
			HTML = Markdown.render(pageText);
		return <div className='phb' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: HTML }} key={index} />;
```

client/homebrew/editor/editor.jsx
```diff
@@ -82,11 +83,49 @@ const Editor = createClass({
...
+			codeMirror.removeLineClass(lineNumber, 'text');
+
 			if(line.indexOf('\\page') !== -1){
 				codeMirror.addLineClass(lineNumber, 'background', 'pageLine');
 				r.push(lineNumber);
 			}
+
+			// New Codemirror styling for V3 renderer
+			if(this.props.renderer == 'V3') {
+				if(line.indexOf('\\column') === 0){
```
AlexeySachkov, line 103: I would assume that [`startsWith`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) should be faster and probably easier to read in the code

client/homebrew/editor/editor.jsx
```diff
@@ -82,11 +83,49 @@ const Editor = createClass({
 		if(!this.refs.codeEditor) return;
 		const codeMirror = this.refs.codeEditor.codeMirror;
 
+		//Clear all Text styles
+		const ranges=codeMirror.getAllMarks();
+		for (let i=0;i<ranges.length;i++) ranges[i].clear();
```
AlexeySachkov, line 88: Why this piece of code is needed? `range` is not used anywhere else

client/homebrew/editor/editor.jsx
```diff
@@ -82,11 +83,49 @@ const Editor = createClass({
 		if(!this.refs.codeEditor) return;
```
AlexeySachkov, line 83: Method should probably be renamed, because it now does more than just highlighting of `\page` lines

shared/naturalcrit/markdown.js
```diff
@@ -13,6 +13,63 @@ renderer.html = function (html) {
...
+			} else if(block == '}}' && blockCount !== 0){
+				r.push('</span>');
+				blockCount--;
+			}
+
+			matchIndex++;
+
+			return r;
+		}, []).join('');
+		return `\n<p>${res}</p>\n`;
```
AlexeySachkov, line 67: Why do we wrap inline span into a paragraph? Paragraph turns it into a block element instead of leaving it as inline one

client/homebrew/editor/editor.jsx
calculuschild: This resets the styling in the codemirror Editor. If you type some code that gets highlighted (\page, \column, {{etc...) and then move those blocks around (press the enter key in a weird spot, delete stuff) such that the code is no longer valid, the style stays applied, so you end up with highlighted bits of code all over that don't clear up until you refresh. This just grabs the array of custom code styling and clears them out before re-applying the styles in the proper places. `ranges` was just taken from the example on the codeMirror blog, but we can rename it for clarity. Maybe `customHilighting` .

shared/naturalcrit/markdown.js
calculuschild: This isn't wrapping the span in a paragraph. What the above code does, is investigate an existing `<p>` tag, search inside for any `{{` or `}}`, and replace with `<span>` or `</span>` as appropriate. Then, with the replacements done, it returns the modified paragraph contents, which we have to re-wrap in a new `<p>` tag to preserve the original paragraph. Does that make sense? Otherwise any `<p>contents contents {{tag tag}}</p>` with {{ inside will lose its `<p>` tag.

Although... I have just realized this means it does not work inside of certain contexts, such as within a list. The bullet point text is not counted as a paragraph.... Maybe I need to make it search *any* text instead of just paragraphs. Hrm...
```
* bullet
* bullet
* bullet with {{this span}}
```

client/homebrew/editor/editor.jsx
calculuschild: Yes I agree. I will change this.

shared/naturalcrit/markdown.js
calculuschild: This whole section has been totally rewritten and works much better now.

client/homebrew/editor/editor.jsx
calculuschild: Renamed this variable and rewrote comment for clarity.

client/homebrew/editor/editor.jsx
calculuschild: Done.

client/homebrew/editor/editor.jsx
calculuschild: Done.

client/homebrew/brewRenderer/brewRenderer.jsx
calculuschild: Still tweaking this area a bit. This might change, but if not I will apply this.

calculuschild:
Gah somehow I have two PRs on the same branch. Going to close this and keep the newest one.

PR #1165, opened 2020-12-24
FreeBSD installation

G-Ambatte:
# Installation in a FreeBSD/FreeNAS Jail

## Purpose

The purpose of this branch is to allow simplified installation of the Homebrewery project in a FreeBSD or FreeNAS jail. This is currently not possible as Docker is currently broken on FreeBSD and has been for some 15+ months, so the provided Dockerfiles do not allow the project to run.

### Work Completed So Far

- Get Homebrewery installed and working on a FreeBSD jail - **COMPLETE**  
- Determine minimum viable package (MVP) installation required for previous step - **COMPLETE**
- Script installation of MVP - **COMPLETE**
- Test with multiple available versions of MongoDB - **COMPLETE**
- Create rc.d files to run Homebrewery as a service, supporting service commands "start", "stop", and "restart" - **COMPLETE**
- Script installation of Homebrewery as a service - **COMPLETE**
- Parameterize configuration options NODE_ENV and PORT - **COMPLETE**
- Eliminate requirement for the current working directory to contain the **server.js** script - **COMPLETE**

### Future Improvements/In Development

- Add custom command "devmode" to service options to start in development auto-rebuild mode
- Write FreeBSD install documentation and add to repository - **IN PROGRESS** (see [instructions here](https://www.reddit.com/r/homebrewery/comments/k7rcfs/homebrewery_in_a_freenasbsd_jail_part_2))

### FreeBSD/Install.sh

This script currently clones from my fork (https://github.com/G-Ambatte/homebrewery.git) rather than the master project. This will need to be corrected during the integration process.
The script also currently clones the project into a new "homebrewery" folder in the root directory; whereas another location may be deemed to be more appropriate (i.e. "/usr/local/etc/homebrewery"). This will have flow on effects to the **freebsd/rc.d/homebrewery** file.

Finally, this script runs the command `npm audit fix -force`, which updates and corrects any security issues with the NPM modules. However the inclusion of the `-force` parameter allows the installation of potentially breaking changes to the required packages, and while this has not caused any issues to date, the `-force` parameter should be removed from the default installation script before wider promulgation, in order to reduce the potential for future issues.

### FreeBSD/RC.D/HomeBrewery

If the install directory for Homebrewery is changed, the "exec_cmd" will need to be changed to the new location of the "server.js" file.

### Server.js

- As the service starts the command from an as yet undetermined working directory, Homebrewery would immediately crash due to files being unable to be found. The addition of `process.chdir(__dirname);` towards the start of the "server.js" file changes the current working directory of the calling thread to the directory containing the script, eliminating this as a source of issues.
- Attempting to set the NODE_ENV environment variable prior to running the command "node server.js" was possible but not particularly clean, especially when writing the command into the script. Instead, I pushed this value into the **config/default.json** file as the key pair "environment", with a hard-coded default value of "local" if it is not set via environment variable or in the default configuration file.
- Similar arguments can be made of the PORT environment variable, so I pushed this to **config/default.json** as the key pair "web_port". Again, if not set via environment variable or default configuration, it reverts to the hard-coded value of "8000".

### Config/Default.json

New key pairs "environment" and "web_port" were added to support the changes as previously discussed in **server.js**.

## Conclusion

I hope that I have managed to lay out a clear and concise description of the changes made and the reasoning behind them in order to simplify the process of installing Homebrewery in a FreeBSD jail.

I look forward to receiving any feedback that you might have.
-- G

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -2,5 +2,5 @@
 	"host" : "homebrewery.local.naturalcrit.com:8000",
 	"naturalcrit_url" : "local.naturalcrit.com:8010",
 	"secret" : "secret",
-	"web_port" : 8001,
+	"web_port" : 8000
 }
--- a/config/default.json
+++ b/config/default.json
@@ -1,5 +1,7 @@
 {
 	"host" : "homebrewery.local.naturalcrit.com:8000",
 	"naturalcrit_url" : "local.naturalcrit.com:8010",
-	"secret" : "secret"
+	"secret" : "secret",
+	"web_port" : 8001,
+	"environment" : "local"
 }
--- a/config/default.json
+++ b/config/default.json
@@ -3,5 +3,4 @@
 	"naturalcrit_url" : "local.naturalcrit.com:8010",
 	"secret" : "secret",
 	"web_port" : 8001,
-	"environment" : "local"
 }
--- a/server.js
+++ b/server.js
@@ -27,6 +27,11 @@ const config = require('nconf')
 	.file('environment', { file: `config/${process.env.NODE_ENV}.json` })
 	.file('defaults', { file: 'config/default.json' });
 
+if ( !process.env.NODE_ENV && config.get('environment') ) {
+	process.env.NODE_ENV = config.get('environment');
+	console.log('NODE_ENV set from config')
+}
+
 //DB
 const mongoose = require('mongoose');
 mongoose.connect(config.get('mongodb_uri') || config.get('mongolab_uri') || 'mongodb://localhost/naturalcrit',
@@ -232,6 +237,6 @@ app.use((req, res)=>{
 });
 
 
-const PORT = process.env.PORT || 8000;
+const PORT = process.env.PORT || config.get('web_port') || 8000;
 app.listen(PORT);
 console.log(`server on port:${PORT}`);
--- a/server.js
+++ b/server.js
@@ -27,15 +27,7 @@ const config = require('nconf')
 	.file('environment', { file: `config/${process.env.NODE_ENV}.json` })
 	.file('defaults', { file: 'config/default.json' });
 
-if ( !process.env.NODE_ENV) {
-	if ( config.get('environment') ) {
-		process.env.NODE_ENV = config.get('environment');
-		console.log('NODE_ENV set from config');
-	} else {
-		process.env.NODE_ENV = 'local';
-		console.log('NODE_ENV set to "local" as no other values were set');
-	}
-}
+if(!process.env.NODE_ENV) { process.env.NODE_ENV = ( config.get('environment') || 'local' ); }
 
 //DB
 const mongoose = require('mongoose');
--- a/server.js
+++ b/server.js
@@ -27,9 +27,14 @@ const config = require('nconf')
 	.file('environment', { file: `config/${process.env.NODE_ENV}.json` })
 	.file('defaults', { file: 'config/default.json' });
 
-if ( !process.env.NODE_ENV && config.get('environment') ) {
-	process.env.NODE_ENV = config.get('environment');
-	console.log('NODE_ENV set from config')
+if ( !process.env.NODE_ENV) {
+	if ( config.get('environment') ) {
+		process.env.NODE_ENV = config.get('environment');
+		console.log('NODE_ENV set from config');
+	} else {
+		process.env.NODE_ENV = 'local';
+		console.log('NODE_ENV set to "local" as no other values were set');
+	}
 }
 
 //DB
--- a/server.js
+++ b/server.js
@@ -27,8 +27,6 @@ const config = require('nconf')
 	.file('environment', { file: `config/${process.env.NODE_ENV}.json` })
 	.file('defaults', { file: 'config/default.json' });
 
-if(!process.env.NODE_ENV) {process.env.NODE_ENV = (config.get('environment') || 'local');}
-
 //DB
 const mongoose = require('mongoose');
 mongoose.connect(config.get('mongodb_uri') || config.get('mongolab_uri') || 'mongodb://localhost/naturalcrit',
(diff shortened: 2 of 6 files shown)


calculuschild:
This looks pretty good so far. This will certainly need some documentation for users who want to try this; we can link to it from the main repository readme.

Also, since I am not a FreeBSD user myself, you would likely be the main point of contact for any questions about this process.

G-Ambatte:
> you would likely be the main point of contact for any questions about this process

I don't have an issue with that, although I will add that I personally rate my FreeNAS/BSD experience as "enthusiast" rather than "professional", so all I can promise is my best effort.

I've pushed a few more changes to the PR, as per the following:

## Task List

### Improvements

- [x] Write FreeBSD install documentation
- [ ] Add `devmode` to service commands

### Integration Considerations

- [x] Remove `-force` from `npm audit fix` in `freebsd/install.sh`
- [x] Point `install.sh` and install documentation at `naturalcrit/homebrewery` rather than the fork
- [x] Change install location from `/homebrewery` to a more appropriate location, `/usr/local/homebrewery`


-- G

G-Ambatte:
With the changes I've just pushed through, that should complete the last of the outstanding items from Improvements and Integration Considerations.

---

## Task List
### Improvements
 
- [x]  Write FreeBSD install documentation
- [x]  Add `dev_mode` to service commands
 
### Integration Considerations
 
- [x]  Remove `-force` from `npm audit fix` in `freebsd/install.sh`
- [x]  Point `install.sh` and install documentation at `naturalcrit/homebrewery` rather than the fork
- [x]  Change install location from `/homebrewery` to a more appropriate location, `/usr/local/homebrewery`

---

If you think there are any further changes that need to be made, or have any thoughts on improvements or changes, then please let me know.

-- G

calculuschild:
Awesome. I'll take another look over this and let you know if I notice anything.

G-Ambatte:
I just pushed a couple of commits through now; `npm run lint` and CircleCI returns cleanly now - the major one was eliminating a few lines of debugging code, which dropped the line count below the 200 limit.

server.js
```diff
@@ -25,6 +27,8 @@ const config = require('nconf')
 	.file('environment', { file: `config/${process.env.NODE_ENV}.json` })
 	.file('defaults', { file: 'config/default.json' });
 
+if(!process.env.NODE_ENV) {process.env.NODE_ENV = (config.get('environment') || 'local');}
```
calculuschild, line 30: If you want to set the default environment to 'local' (which is fine), it makes more sense to me to move this up before the configs are loaded up, and not load the environment from the config files. i.e., just:

```
if(!process.env.NODE_ENV)
    process.env.NODE_ENV = 'local';
```

Otherwise, we are loading the environment from the config files themselves, which could lead to some weird circular logic (Load up config file depending on environment -> use config file to set environment -> load up config file depending on environment).

If the environment needs to be set automatically, I'm thinking it should be done through the startup script, rather than based on a config file.

G-Ambatte: I've removed the NODE_ENV line from `server.js` and the `environment` variable from the `config/default.json`. I've updated the service file to include them in the "Env Setup" section, which appears to be working correctly.

config/default.json
```diff
@@ -1,5 +1,6 @@
 {
 	"host" : "homebrewery.local.naturalcrit.com:8000",
 	"naturalcrit_url" : "local.naturalcrit.com:8010",
-	"secret" : "secret"
-}
\ No newline at end of file
+	"secret" : "secret",
+	"web_port" : 8001,
```
calculuschild, line 5: Last things, and then I'll merge this:
1) Remove the comma after 8001, it crashes the build step
2) Is there a reason you need the web_port different than the default 8000 we already set? I don't see 8001 used anywhere else in this PR. I'm ok leaving the `web_port` key in here for people who want to easily customize the port, but I would prefer the default value in this file stay 8000, because otherwise it messes with my current development setup.

calculuschild (changes requested):

config/default.json
G-Ambatte: 8001 was only used to ensure that the value was actually being used; it can be reverted to 8000. I'll push a commit in a few minutes with that change and the fix for the hanging comma in the config file.

G-Ambatte: I've also changed the port number in the FreeBSD service file so that it will set the environment variable to run the service on port 8000.

calculuschild:
Ok, it all looks good to me now. Do you want to put a link to your readme on the main readme file, similar to the "how to install with Docker" section?

G-Ambatte:
I've done a basic copy and paste of the Docker section, and modified the link to point to the FreeBSD install instruction file.

G-Ambatte:
...and I've just seen that the MongoDB package has been updated, which breaks the installation script. I'll push a commit that fixes it through later today.

G-Ambatte:
I've pushed a change that fixes the install script - I've run this through my standard test process (creating a blank jail, running the script, and seeing if I have a working Homebrewery at the end of the process) and it worked.
I've also modified the service script so the user can easily change the NODE_ENV and port variables - they default to `local` and `8000` but the user can run `sysrc homebrewery_NODE_END="local"` or `sysrc homebrewery_port=8000` to set the values for the service to use.

calculuschild:
Looks good to me! I'll push this live!

server.js
```diff
@@ -14,6 +14,8 @@ app.use('/', expressStaticGzip(`${__dirname}/build`, {
 	index           : false
 }));
 
+process.chdir(__dirname);
```
AlexeySachkov, line 17: @G-Ambatte, could you please comment about why this line is needed?

G-Ambatte: This has been answered elsewhere, but for historical completeness, I'll include the answer here:

> node.js by default assumes any directories/files you name are relative to the calling directory. __dirname is the directory of the current running file (so the root folder, since that's where server.js resides). Normally, we execute server.js from the root, but in the case of running it from a docker container or BSD jail you might actually start up the app from some external source, which means the app would not be able to find any of the relative files. process.chdir(__dirname); ensures that the "working directory" of the app is always the same root folder no matter from where it is started up.

Issue #212, opened 2016-09-30
Card Mode


ryantheleach:
It would be awesome to have a "card mode" which could create Magic The Gathering sized cards that can fit into sleeves, and print to letter/A4

Homebrewery is so good, if it could just easily create cards from a template it would beat the absolute pants off of anything in existence.

stolksdorf:
Cool idea! Design the template and I can add it.

ryantheleach:
I'm not that great of a designer, when you say design the templates do you mean the CSS? 
some mockups?
Or do you think creating some form of actual templating substitution, then loading from a datasource would be relevant to the project?

wrdclark:
Here's some ideas I've just played with. Lots of CSS work to be done (standard monster stat blocks don't work, as it looks like the blockquote isn't being parsed), but it's a start!

ryantheleach:
@wrdclark That's a great start, but for cards I think it would be more flexible that instead of hard coding font's in cm that it used relative font sizes.

This would make it a lot easier for people to edit the default card template in case they wanted different sized cards (playing card vs MTG size sleeves) or printing onto index cards.

vlos:
Looks intereseting - side note, you may want to switch to landscape. Think you could get 2 rows of 4?

Gazook89:
I know this thread was opened 4 years ago, but wanted to thank @wrdclark for the idea to put cards inside the .phb div.  I had played with cards previously but was only thinking about turning the .phb div into individual cards, which resulted in bad pdfs.  I still like this method for just creating stuff online, but obviously if you are designing cards then you likely want to be able to print them too.  [These cards are set to "Tarot" dimensions.   
](https://homebrewery.naturalcrit.com/share/RjaYiyOc9RJQ) 

Creating the cards within the normal pages means no issue with printing.  Also, it means a table of contents could be created on the first page using the normal method (however...not sure if the ToC generator snippet even existed 3 years ago).

[My new cards are styled to match Magic the Gathering](https://homebrewery.naturalcrit.com/share/1UzpGWTa0WpKYlUCpuP6DnzIN0A2nHAVp7dsPYDGtTDEq), and likely can't be easily modified to do DnD cards (I don't know what that would even look like as it seems to me there is too much possible info/text needed on a card).  It could also be said that my method relies too much on divs to be considered "general use".  But posting here in case it helps.

AlexeySachkov:
[...]

PR #1167, opened 2020-12-24
Card & Deck Template Generator

G-Ambatte:
# Add Card & Deck Template Generator

## Purpose

The purpose of this pull request is to add a Card or Deck Template generator to Homebrewery's nav bar.

### Work Completed So Far

- Create a suitable template - **COMPLETE**  
- Write a generator script (**cardtemplate.gen.js**) - **COMPLETE**
- Add the template to the nav bar - **COMPLETE**

### Future Improvements/In Development

- Printing to PDF returns unexpected results - **UNDER INVESTIGATION**
- Improve in-template comments to simplify editing for non-power users

### Snippets.js

The menu location on the nav bar requires consideration; I added it under "Editor" as it makes extensive changes to the template.

### CardTemplate.gen.js

Returns a static block of text.

### Snippets.js

Changed to add the Deck Template item to the Editor menu, as previously discussed.

## Conclusion

I hope that I have managed to lay out a clear and concise description of the changes made and the reasoning behind them in order to add a new and hopefully useful template to the Homebrewery project.

I look forward to receiving any feedback that you might have.
-- G


client/homebrew/editor/snippetbar/snippets/cardtemplate.gen.js
```diff
@@ -0,0 +1,158 @@
+const _ = require('lodash');
+
+module.exports = function(){
+
+	return [
+		'<style>',
```
AlexeySachkov, line 6: This is a quite a significant snippet to insert and I'm not the biggest fan of putting a lot of CSS into what is supposed to be "only markdown" (but I do understand that this is sometimes impossible to avoid).

My question here: is it possible to apply the same styles not to `.page`, but to another `div`? I mean what if we ask users to create a `<div class="deck-container">...</div>` (like for example it is done with wide blocks) and put all of these styles into built-in homebrewery stylesheets instead of adding them directly to the page source?

G-Ambatte: On reflection, this is effectively a full-blown template, so it may not be appropriate to include it as such. It would likely benefit from being a template that can be applied in the separate CSS tab in the new editor, so that there is a clearer delineation between the CSS and the Markdown - the Markdown actually becomes quite simple once these CSS changes have been applied.

[Example with zoom and page layout](https://homebrewery.naturalcrit.com/share/1EbOq21BlSlvS-8bSvehk4jfzKINUYfhYVhimiKzT1Q2N)  
[Example without zoom/layout](https://homebrewery.naturalcrit.com/share/1XSOHsSodcSnv_UnVDxQOjMHYiVUCNYCuTylhQjnqbpTm)

So at this point, it may be best to put it on hold until the new Renderer is complete and the editor has a separate CSS tab.

G-Ambatte: I thought about this some more and came around to @AlexeySachkov's way of thinking. While the original is great for online viewing or creating an entire deck of magic items, it doesn't print to PDF well and it is not possible to embed a card in a larger document.  
After some manipulation, [this Homebrew](https://homebrewery.naturalcrit.com/share/140NrG9yBcfdlph10Bt2vs_R5Q-OQJY-rKL3TnfK7eqbR) is the result.

Unfortunately the Markdown parser doesn't parse inside `div`s, so I had to replicate the styles in their own classes, which makes it a little more opaque to the standard user, although this can probably be simplified by duplicating the class in the CSS build files.

The best way forward for this version of the card generator is probably to embed the `.card` style in the built CSS, and add a snippet that provides an example of creating a card with the various new div classes. I'll push a commit through for further review in the near future.

G-Ambatte:
The changes I've pushed this afternoon result in a functional card being added to the current page, with the option of adding the "back" class to create a double wide div with the card back included - the back is currently loaded from Imgur rather than being base64 encoded into the CSS.  
The snippet creates two examples, one with the back and one without.

Let me know what you think.

client/homebrew/phbStyle/phb.style.less
```diff
@@ -468,3 +468,99 @@ body {
...
+	column-rule: 2px solid black;
+	column-gap: 25px;
+	padding-left: 0px;
+}
+
+.phb .card.back::before {
+	content: url(https://i.imgur.com/iNJH8j9.png) !important;
+}
+
+.phb .card h1 {
```
AlexeySachkov, line 500: Note: you could use [`parent selectors`](http://lesscss.org/features/#parent-selectors-feature) to structure this code in a better (I suppose) way

client/homebrew/editor/snippetbar/snippets/cardtemplate.gen.js
```diff
@@ -0,0 +1,37 @@
...
+		'  <img src=\'https://i.imgur.com/iNJH8j9.png\'></img>',
+		'',
+		'  <span class="caption">Would you believe that a wizard did it?</span>',
+		'  <div class="statblock">',
+		'    **Effect:** The targeted player takes 1D10 *annoying-the-DM-in-a-dramatic-moment* damage. This damage cannot be mitigated in any way.',
+		'  </div><br />',
+		'  <div class="description centered">',
+		'      *"Don\'t say it, don\'t... You said it."*',
+		'  </div>',
+		'</div>',
```
AlexeySachkov, line 18: It seems like a lot of HTML-markup for a markdown-based tool. It probably will require some trickier CSS or some changes in markdown parser to use less HTML here, so I think it is okay to leave this as-is for the initial version of this feature. But still, we should probably submit an issue to discuss and track further improvements.

At the moment, the only thing I could suggest is to use header markup (like `####` or `#####`) for caption to simplify the markup, i.e. remove the need to type `<span class="caption"></span>".

We might also apply style using [`nth-of-type`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-of-type) selectors, i.e. the first paragraph within a card is styled like a caption, the next one is styled like a statblock and so on, but I guess it won't be a reliable solution as it is really easy to create an extra paragraph in Markdown.

client/homebrew/phbStyle/phb.style.less
```diff
@@ -468,3 +468,99 @@ body {
 		margin-bottom : 10px;
 	}
 }
+//*****************************
+// *      MAGIC ITEM CARD
+// *****************************/
+.phb .card {
+	width: var(--card-width);
+	height: var(--card-height);
```
AlexeySachkov, line 476: I suggest we set some default value here so users won't have to put `style="--card-width: 67mm; --card-height: 89mm;"` into each card. Docs for specifying the default value: [here](https://developer.mozilla.org/en-US/docs/Web/CSS/var()#Values)

client/homebrew/phbStyle/phb.style.less
```diff
@@ -468,3 +468,99 @@ body {
...
+	-webkit-print-color-adjust:exact;
+	border: 2px solid black;
+	border-radius: 20px;
+	column-count: 1;
+	padding-left: 10px;
+	padding-right: 10px;
+	column-span: all;
```
AlexeySachkov, line 483: What about allowing cards to be put into separate columns? Here is an example of how it could look like: [share link](https://homebrewery.naturalcrit.com/share/b6Qj3zcgxg-Y)

Here I removed `column-span: all` by default, but it is being added if card is also marked with `.centered` class.

Also note change in margin: cards are too close if they are not separated manually by `<br>` and I think it would be a nice addition, because it removes the need in an explicit HTML-markup in the brew source.

One more note about columns: we should probably use something like `break-inside: avoid-column;` to avoid cards being broken when they exceed column limits: I've also have this change in the shared brew and a page with an example

AlexeySachkov (commented):
Overall, this approach (with special markup just for a card) looks better to me. Just a few suggestion here and there to improve the feature

client/homebrew/editor/snippetbar/snippets/cardtemplate.gen.js
AlexeySachkov: > Unfortunately the Markdown parser doesn't parse inside divs, so I had to replicate the styles in their own classes, which makes it a little more opaque to the standard user, although this can probably be simplified by duplicating the class in the CSS build files.

This seems a bit strange to me, could you please outline this into a separate issue with some example so we could try to analyze and fix it later?

calculuschild: We have a custom rule set up so that markdown does indeed parse inside of divs. You can see it working on the live site such as in the "descriptive text box" snippet:
```
<div class='descriptive'>
##### Time to Drop Knowledge
Use notes to point out some interesting information. 

**Tables and lists** both work within a note.
</div>
```

client/homebrew/phbStyle/phb.style.less
G-Ambatte: This would also wrap the entire `card` style in a single set of braces, which should hopefully make it easier to understand what is and is not part of the style. I've rewritten the `phb.style.less` entries and am testing them now; if it works they way it should (in that it should be exactly the same) I'll push it into a commit.

client/homebrew/editor/snippetbar/snippets/cardtemplate.gen.js
G-Ambatte: I believe `h6` isn't widely used - it might be suitable for a `caption` style; I'll add that to the stylesheet. It may be worthwhile playing with the `h4` and `h5` styles as well - I'll make the change and see what it's like to use.

client/homebrew/phbStyle/phb.style.less
G-Ambatte: I like it - I was mostly approaching from the mindset of having both sides of the card; this is why I started with the "each page is a card face" approach. However now that it's a single face is confined to a `div` and can be included, this is a better approach.

G-Ambatte: Also `break-inside: avoid` seems to work better than `avoid-column`, but this may just be an artifact of cross-browser compatibility (Firefox vs Chrome).

AlexeySachkov (approved):

Gazook89:
Hey folks! Got pulled into this PR from Alexey's comment on #212, and man oh man it's crazy to see that a) we are working on the same thing at the same time and b) our solutions are very similar.  I just posted today on the subreddit my idea for cards (and in the #212 thread).  Let me start by saying that beyond html and css, I really have on idea what I'm doing; I never figured out how to get an offline homebrewery going on my macbook and I only recently got onto Github...well, I've had an account for awhile, but haven't done much on it.  So just want to put that out there before I go much further.

After reading through all the comments here, I have some Quick thoughts:
1. As you seem to have discovered in the same way I have, cards should be placed inside the pages, rather than making the pages themselves cards for easier printing. 

2. A standard trading card size will allow 9 cards on a page.  Use Flexbox or CSS Grid to set that up (I chose flex since grid seemed to be needlessly complicated for this task).

3. The cards should be flush with the edges of the pages, and have no gap between them.  This makes it MUCH easier to cut the cards apart after being printed.  My option utilizes an outline around the cards to show where those cut lines are....but honestly, It really shouldn't have those cut lines because it will mar the edges of the cards.  Possible solution is to replace those outlines with a custom .phb background that has those lines, except only displayed at the edges of the page and in the 'gaps' at the corners of the cards. 

4. My preference is to not use headers for anything besides actual headers.  I know using headers is a convenience thing in Markdown, but it can lead to issues down the road, similar to how we all know `blockquote + hr` and `hr + hr + blockquote` are bad.  In the examples I saw above, h2 is being used as a sort of "by-line" right after the h1....I think this should be avoided.  If a ToC was generated it would list all the card names, and this second line.  It may be very helpful to be able to create a simple ToC if people make many cards.

5. Gambatte- in the first examples you are using my code for the cards with the zooming....In my recent brews, I fixed some minor issues with that, and also it seems as if something else is "off", because the scroll bar shouldn't increase in size.  Since that format has been dropped in favor of "cards on a page", I won't look into it further unless you want me to.

6. The toughest part about cards is knowing what people want on the cards.  Unlike MtG, the game elements of dnd require so much text, and tables.  I don't even play D&D, which is another problem, so I'm not sure what people are using cards for. 

Anyway, sorry for the wall of text.

AlexeySachkov:
> Hey folks! Got pulled into this PR from Alexey's comment on #212, and man oh man it's crazy to see that a) we are working on the same thing at the same time and b) our solutions are very similar. I just posted today on the subreddit my idea for cards (and in the #212 thread). Let me start by saying that beyond html and css, I really have on idea what I'm doing; I never figured out how to get an offline homebrewery going on my macbook and I only recently got onto Github...well, I've had an account for awhile, but haven't done much on it. So just want to put that out there before I go much further.

Hi @Gazook89, I suggest that you and @G-Ambatte join our [gitter chat](https://gitter.im/naturalcrit/Lobby#), because it might be easier/quicker to discuss some matters in IM (and sometimes not overflow the review)

> The cards should be flush with the edges of the pages, and have no gap between them. This makes it MUCH easier to cut the cards apart after being printed.

Oh, I haven't tried that by myself and therefore haven't though about it. On contrary I suggested to insert some margins between cards because it looked better. This makes me thing that we might have two use-cases for cards: I was imaging that one would be a card within a regular homebrewery document, i.e. as some cheat sheet or quick reference or something like this; the other would be: we don't need the document itself, we just need cards to play.
So, for the former this PR is good, I guess; for the latter it seems to me that we might need a separate kind of page/layout for that to maximize our use of page (i.e. no paddings (or minimal) of the page itself - who wants to waste paper?). Do we need to have a further discussion about how this can be achieved and how it would look like from Homebrewery interface point of view? (just because we need to redefine some core `.phb` styles to fit more cards)

> My preference is to not use headers for anything besides actual headers. I know using headers is a convenience thing in Markdown, but it can lead to issues down the road, similar to how we all know `blockquote + hr` and `hr + hr + blockquote` are bad. 

On the one hand I agree with you that styling `blockquote + hr` and `hr + hr + blockquote` is not the way we should go, but I also don't like to have a lot of raw HTML in the document source and it would be nice (I think) to have it more user-friendly, i.e. closer to regular Markdown. We might want to find some Markdown extension that fits our purposes and it probably is a topic for a whole separate discussion (quite an interesting one!)

> The toughest part about cards is knowing what people want on the cards. Unlike MtG, the game elements of dnd require so much text, and tables. I don't even play D&D, which is another problem, so I'm not sure what people are using cards for.

Yeah, lack of knowledge of most wanted or potential use-cases it hunting me as well: shall we create a reddit post to gather general feedback about the feature and directions it should move to?

Gazook89:
I was just randomly going through PR's and this one hasn't been touched in a while.  [Here is a link to a newer card style](https://homebrewery.naturalcrit.com/share/1Wk4IvdSu1D_FDVyGmiUKZiDyK0YnIgrNr38rZJhxpDtR).   It is based on a simple  card {front, back}  div setup, which is laid out in a more [basic form here](https://homebrewery.naturalcrit.com/share/10SpSRX0cHBCjBhuBVIaJQwmBkEzRnuMbGvBYe8PuNow9).

The first share link is basically just made for monster stat blocks and is really just a showcase of "what is possible"....but the second link I think provides a solid base for people to place their own styling on, however they want it.  All the standard snippets work inside of it, but of course the cards are too small to really fit them....those adjustments would have to be up to the editor to figure out.   Unless standard card layout for all types of homebrew are decided on.

calculuschild:
This is pretty out-of-date and will work better as a separate "theme" once swappable themes are implemented.

Closing for now while we build out more of the framework needed for this to work smoothly.

PR #1175, opened 2020-12-31
Tweak spelling of username on user page

### LINKED ISSUE #575: User Page Minor Detail
Nothing major, no big bug or error to be fixed. I just noticed that if a user's name ends with an "S", the homebrew page adds an *'s* to the end of it, where the grammatically correct version would be just using an apostrophe.

It's a small detail, but a check for usernames ending in "S" and the exclusion of a single character based on that would make a big difference aesthetically for the site.

### Additional Details

My username for one of my accounts is *blackmagikarper-commissions*, and in the homebrews page it is printed as **blackmagikarper-commissions's brews**.

AlexeySachkov:
Fixes #575

### DIFF
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -26,6 +26,11 @@ const UserPage = createClass({
 			brews    : []
 		};
 	},
+	getUsernameWithS : function() {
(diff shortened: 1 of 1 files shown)


calculuschild:
This looks good. Thanks!

Issue #620, opened 2018-02-16
Table decoration split between columns


NoelMock:
I am trying to align my class table on the second column, but no matter what I do, it will not shift at all. The top of the table prints on the bottom of the left column, while the rest prints on the right column.

### Additional Details

**http://homebrewery.naturalcrit.com/share/H1Wazfkcef** :

or

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

PASTE BREW CODE HERE

</pre></code></details>

DoctorDan1986:
Use a column break to force it into the next column (under editor), and it's best not to post the edit link. Use the share link next time.

NoelMock:
Thanks. I tried using the column break, but then it created a whole empty column and pushed the table and text off the page.  (Because the text in column 1 goes all the way to the bottom.

granack:
yeah, table decoration doesn't go to the next column. it's annoying, and I'm not sure if it's a homebrewery issue or css in general. To fix it, you need something to render, like a non-breaking space, at the top of the page, so between the column break and the table `<div>`. Adjust the `<div>` margin to bring it up a bit since the extra bit will push it down.
Basically, do this:

    ```
    ```
    &nbsp;
    <div class='classTable' style="margin-top: 0px">

for reference, the default margin is somewhere around 25px.
I use lots of `&nbsp;` and margin adjustments to get my stuff working just right.

NoelMock:
Thanks! It at least prints okay now... :)

ericscheid:
Recreating from the source of the shared brew I see a different but related problem (the table is put into column-2 by virtue of column-1 being full).

![image](https://user-images.githubusercontent.com/2605467/99152893-cc85d580-26f8-11eb-8689-4653ad3ef6e0.png)

> I am trying to align my class table on the second column, but no matter what I do, it will not shift at all. The top of the table prints on the bottom of the left column, while the rest prints on the right column.

Was there a recent change that attempted to fix this .. because it looks like it did, but created a new problem.

----

I have also seen in the past where a blockquote note is at the bottom of column 1, but the overhanging bottom border is at the top of column 2. Attempting to recreate (using `<div style="margin-top:1px"></div>` to nudge it downwards .. and I see the overhang border disappear, but no longer re-appears at top of column 2.

Gazook89:
This was fixed with #1179 at the beginning of the year.  closing.

calculuschild:
[...]

Issue #610, opened 2018-01-25
Shadow not wrapping correctly


Smashman:
It appears that when column wrapping, the top of a wrapped stat block's shadow can be seen at the bottom of the first column.

Thanks.

### Additional Details

**Share Link** : http://homebrewery.naturalcrit.com/share/rko55LzHz

Smashman:
Might I suggest using flexbox and flex-wrap, rather than the coumn?

AlexeySachkov:
@naturalcrit/issue-managers, I suggest that we close this one, because we have applied a workaround in #1179

PR #1179, opened 2021-01-04
Fix for box-shadow/border-image issues on blockquotes in new columns

G-Ambatte:
As per issue reported on Reddit [here](https://redd.it/kpcbw6), with the fix found by /u/alexey152.

I was able to replicate the issue in Chrome and Edge, but not Firefox.  
I have confirmed that this change fixes the issue in both Chrome and Edge.

### DIFF
--- a/client/homebrew/phbStyle/phb.style.less
+++ b/client/homebrew/phbStyle/phb.style.less
@@ -189,6 +189,7 @@ body {
 		border-image        : @noteBorderImage 11;
 		border-image-outset : 9px 0px;
 		box-shadow          : 1px 4px 14px #888;
+		transform           : translateZ(0);
 		p, ul{
 			font-size   : 0.352cm;
 			line-height : 1.1em;
--- a/client/homebrew/phbStyle/phb.style.less
+++ b/client/homebrew/phbStyle/phb.style.less
@@ -189,7 +189,7 @@ body {
 		border-image        : @noteBorderImage 11;
 		border-image-outset : 9px 0px;
 		box-shadow          : 1px 4px 14px #888;
-		transform           : translateZ(0);
+		-webkit-transform           : translateZ(0);
 		p, ul{
 			font-size   : 0.352cm;
 			line-height : 1.1em;
--- a/client/homebrew/phbStyle/phb.style.less
+++ b/client/homebrew/phbStyle/phb.style.less
@@ -189,7 +189,7 @@ body {
 		border-image        : @noteBorderImage 11;
 		border-image-outset : 9px 0px;
 		box-shadow          : 1px 4px 14px #888;
-		-webkit-transform           : translateZ(0);
+		-webkit-transform   : translateZ(0); //Prevents shadows from breaking across columns
 		p, ul{
 			font-size   : 0.352cm;
 			line-height : 1.1em;


calculuschild:
Wow, weird solution... Can you verify if this shifts the elements around on the page at all? If not, we can apply this fix, but if it shifts boxes around it might ruin some old brews, and in that case we can save this for the new renderer.

G-Ambatte:
Doesn't look like it shifts the elements at all - here's some screengrabs:

[Before](https://i.imgur.com/unmHP47.png) || [After](https://imgur.com/eM8rq1R)  
[Before](https://imgur.com/LK8FyCz) || [After](https://imgur.com/eSc6utm)

My testing so far has shown that `transform: rotateZ(360);` doesn't resolve the issue. [This](https://homebrewery.naturalcrit.com/share/1jumEkGO1ICHEu0rLSaZqwNvVhTe5495NlkoXck9lLS4P) is the share link to my reproduction of the issue.

calculuschild:
Huh. Well alright then. Doesn't look like it will be a breaking change based on your screenshots. @AlexeySachkov, @ericscheid , any concerns merging this?

ericscheid:
Looks good to me - the modded result only shifts the shadow to where it should be, it doesn't bump the position of the box being shadowed.

ericscheid (approved):

AlexeySachkov:
I found this solution in [this](https://stackoverflow.com/questions/17881923/box-shadow-trimmed-in-css-columns-in-chrome) stackoverflow question and I believe that this is a bug in Chrome and we should probably put a `TODO`/`FIXME` comment in the code to remove it once Chrome is fixed. [This](https://bugs.chromium.org/p/chromium/issues/detail?id=269061) seems to be a bug report in Chromium project

Gazook89:
I'm on phone so can't be sure, but this is also related to issue #610? In which another solution is offered (use flexbox, not columns) but imagine that would create many more issues. But if this PR goes through, that issue can be closed.

client/homebrew/phbStyle/phb.style.less
```diff
@@ -189,6 +189,7 @@ body {
 		border-image        : @noteBorderImage 11;
 		border-image-outset : 9px 0px;
 		box-shadow          : 1px 4px 14px #888;
+		transform           : translateZ(0);
```
Gazook89, line 192: [Per this comment on the stackoverflow solution,](https://stackoverflow.com/questions/17881923/box-shadow-trimmed-in-css-columns-in-chrome#comment103838351_55110789) perhaps better to prefix with -webkit- so as to only kick in for Chrome users?

AlexeySachkov: We have a "Render Warning", that says that Homebrewery is built & tested for Chrome mostly, so I would expect the majority of users to use it and therefore, it doesn't seem like a critical fix to have; but yes, we can apply this change, I have no objections

AlexeySachkov:
> I'm on phone so can't be sure, but this is also related to issue #610? 
> Edit : and #620 probably

Yeah, those seem to be duplicates. I suggest that we close them and refer to this PR

> In which another solution is offered (use flexbox, not columns) but imagine that would create many more issues. But if this PR goes through, that issue can be closed.

Using "flexbox" might be a solution for this issue, but I wouldn't consider it as a quick fix, because most likely it will cause some rendering changes that will affect existing brews. We can for sure explore this way to see if it is easier to manage for users, maintain for developers and whether it is more reliable or flexible, but it should probably be done as a separate activity and not in scope of a bugfix.

Gazook89:
Yeah, the first thing I think about with flexbox is how to replace column spanning/wide blocks.  

> Using "flexbox" might be a solution for this issue, but I wouldn't consider it as a quick fix, because most likely it will cause some rendering changes that will affect existing brews. We can for sure explore this way to see if it is easier to manage for users, maintain for developers and whether it is more reliable or flexible, but it should probably be done as a separate activity and not in scope of a bugfix.

client/homebrew/phbStyle/phb.style.less
calculuschild: @G-Ambatte, do you want to make this change and then we'll merge it?

AlexeySachkov (approved):

client/homebrew/phbStyle/phb.style.less
G-Ambatte: I've pushed the change to `-webkit-transform` now.

PR #1183, opened 2021-01-04
[WIP] Setup some tests

AlexeySachkov:
I'm totally unfamiliar with JS ecosystem, but from what I read Jest seem to be a reliable and popular framework so I decided to go with it. Also it is [listed as recommended tool](https://reactjs.org/docs/testing.html#tools) for testing React applications.

My plan here is to create some skeleton which can be later filled with other tests for the rest of components and various Homebrewery details. I'm not aiming to create any good coverage in the first PR about testing.

This is a work in progress (and I will probably do force-pushes while this is still a draft PR), some TODO list:
- [ ] `"test:dev"` should probably be updated
- [ ] try to add a few tests for other components like Editor
- [ ] try to add a few tests for React components


server.js
```diff
@@ -1,236 +1,10 @@
...
-const GoogleActions = require('./server/googleActions.js');
-
-// Serve brotli-compressed static files if available
-app.use('/', expressStaticGzip(`${__dirname}/build`, {
-	enableBrotli    : true,
-	orderPreference : ['br'],
-	index           : false
-}));
-
-process.chdir(__dirname);
```
AlexeySachkov, line 17: I tried to preserve this line during the rebase, but it still seems to be lost. @G-Ambatte, could you please advise on where should it be in context of refactored `server.js` file?

G-Ambatte: From memory, it should proceed any calls to working directory specific resources, so the earlier the better. 
The simple test is to run `node server.js` from an alternate directory, i.e. with the current working directory as "C:/", run `node D:/Homebrewery/server.js`. If the line is in the correct place, Homebrewery will still run as normal.

server.js
```diff
@@ -1,236 +1,10 @@
...
-//app.use(express.static(`${__dirname}/build`));
-app.use(require('body-parser').json({ limit: '25mb' }));
-app.use(require('cookie-parser')());
-app.use(require('./server/forcessl.mw.js'));
-
-const config = require('nconf')
-	.argv()
-	.env({ lowerCase: true })
-	.file('environment', { file: `config/${process.env.NODE_ENV}.json` })
-	.file('defaults', { file: 'config/default.json' });
```
AlexeySachkov, line 28: `config` object is actually being used in `server.js`, so this code doesn't work as-is

server.js
AlexeySachkov: Thanks, restored this call in bfdde9104369dd0fc652a79ac6815a8ca60c5450. Tested locally: I'm able to run the server from an alternate directory like `node homebrewery/server.js`

server.js
AlexeySachkov: Refactored code a little bit in bfdde9104369dd0fc652a79ac6815a8ca60c5450 to fix this issue

package.json
```diff
@@ -30,19 +29,27 @@
...
+      "<rootDir>"
+    ],
+    "preset": "@shelf/jest-mongodb"
   },
   "babel": {
     "presets": [
-      "env",
-      "react"
+      "@babel/preset-env",
+      "@babel/preset-react"
```
calculuschild, line 42: I'm assuming this is functionally identical, right? Just using the more "correct" naming scheme here?

server.js
```diff
@@ -1,233 +1,12 @@
...
-const jwt = require('jwt-simple');
-const express = require('express');
-const app = express();
-
-const homebrewApi = require('./server/homebrew.api.js');
-const GoogleActions = require('./server/googleActions.js');
-const serveCompressedStaticAssets = require('./server/static-assets.mv.js');
-
-app.use('/', serveCompressedStaticAssets(`${__dirname}/build`));
+require('app-module-path').addPath(__dirname);
```
calculuschild, line 1: Can you clarify what this package does for us? Do we need to introduce another dependency for this PR to work?

tests/markdown.test.js
```diff
@@ -0,0 +1,18 @@
+const Markdown = require('shared/naturalcrit/markdown.js');
+
+// FIXME: tests below are disabled, becaues marked.js uses replaceAll function,
+// which is only supported starting with node.js 15.0.0:
+// https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V15.md#v8-86---35415
+// Need to enable them back when we uplifit node.js requirements of the project
```
calculuschild, line 6: Feel free to bump the Node version.

tests/markdown.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+test.skip('Escapes <script> tag', function() {
+	const source = '<script></script>';
+	const rendered = Markdown.render(source);
+	expect(rendered).toMatch('&lt;script&gt;&lt;/script&gt;');
+});
```
calculuschild, line 12: Is there a way to run tests on text files? It would be handy to have a folder of inputs / outputs that we can compare like Marked.js does (https://github.com/markedjs/marked/tree/master/test/specs/new). This would keep the actual test script small and easy to navigate while allowing lots and lots of tests of all different sizes.

tests/markdown.test.js
```diff
@@ -0,0 +1,18 @@
...
+
+test.skip('Processes the markdown within an HTML block if its just a class wrapper', function() {
```
calculuschild, line 14: It would also be really handy to do visual snapshot comparisons. Even if the HTML output is the same between updates to Homebrewery, we may change the CSS in such a way that messes up people's brews. Do you want to take a crack at image-pixel snapshot testing? Once this is set up? Looks like there's a good npm package to do this using Jest:

https://www.npmjs.com/package/jest-image-snapshot

package.json
AlexeySachkov: I assume the same, but I'm not so familiar with all that stuff to say for sure

server.js
AlexeySachkov: The problem I encountered is that we have two entry points to our app - a regular one through the existing main file, which is being launched under `node` and the new one which is being introduced with tests.

When you launch test, you basically just execute a `.js` file which contains test using `node`, but under some  different context/environment, i.e. all include paths are different comparing to a regular launch.

That `app-module-path` is my attempt to align things. If you have any other suggestions/ideas on how it can be fixed - I'm all for it

tests/markdown.test.js
AlexeySachkov: I guess we can organize something like this for sure, but I haven't explored it

tests/markdown.test.js
AlexeySachkov: > Do you want to take a crack at image-pixel snapshot testing after this PR is merged?

If time permits, I'm ok to continue working on those tests and expand the coverage we have (but apparently I appear here something like for a 1-2 months within a year in average?)

calculuschild:
This is pretty out of date now and never quite got off the ground. @AlexeySachkov if you want to take another crack at this you are welcome, but I'm going to close it for now since I don't think we have the resources to dedicate getting this up to speed with the current codebase.

AlexeySachkov:
> This is pretty out of date now and never quite got off the ground. @AlexeySachkov if you want to take another crack at this you are welcome, but I'm going to close it for now since I don't think we have the resources to dedicate getting this up to speed with the current codebase.

Yeah, sorry about that. No objections against closing this one. Hopefully, one day I will find enough time to finish that

Issue #1093, opened 2020-11-06
New Parser/Renderer


calculuschild:
Future updates and fixes for current bugs is becoming more difficult due to the current structure of the Markdown->HTML->Render pipeline. This is partly due to unforseen consequences of early design choices, layers of patches and band-aids, and users with custom styling and HTML. We propose creating a new "V2" pipeline which will become the new default, leaving behind the current implementation as a "legacy" option so as to preserve older brews, but limiting further updates on that path. We have learned a lot as the Homebrewery has grown and become more popular, and it is time to put it to use in a new, better version.

This issue will collect the sub-tasks and related issues.

Note: @stolksdorf did ~a lot of~ almost all of this work already, and even totally cleaned up and organized everything. The [NewRenderer](https://github.com/naturalcrit/homebrewery/tree/newRenderer/shared/homebrewery) branch is a goldmine. Some of it might not be compatible with the current codebase, but I really like what he's done.

### Create new render pipeline, "V3"

- [x] Collect legacy code into separate "legacy renderer" pipeline
- [x] Add "Homebrewery Version" tag to brews ("legacy" or "V3")
- [x] Switch between render pipelines based on Version tag
- [x] Update Markdown version for new pipeline (#852) 

### Restructure PHB Styling
- [x] Implement CSS Div classes in Markdown (#403)
  - [x] Define formatting (mustache-like vs Pandoc fenced divs, etc.)
  - [x] Add CodeMirror highlighting to the mustache syntax
- [ ] Update styling and rebuild Snippets to use mustache syntax instead of janky Markdown combinations
  - [x] Standard paragraph text
  - [x] Headers
    - [x] H1
    - [x] H2
    - [x] H3
    - [x] H4
    - [x] H5
  - [x] Drop caps (fancy first letter)  
  - [x] Monster stat blocks #509
    - [x] Standard
    - [x] Wide
    - [x] Border-less   
  - [x] Tables
    - [x] Standard table
    - [x] Wide table
    - [x] Split table
    - [x] Class table
    - [x] Half class table
  - [x] Note box
  - [x] Descriptive box
  - [x] Spell
  - [x] Spell list
  - [x] Class feature
  - [ ] Cover page
  - [x] Vertical spacing
  - [x] Column break
  - [x] Page break
  - [x] Wide block
  - [x] Image
  - [x] Background image
  - [x] Page number
  - [x] Auto-incrementing page number
  - [x] Link to page
  - [x] Table of Contents #304
[...]

PR #1184, opened 2021-01-04
Legacy renderer

calculuschild:
New Renderer pipeline, as described in #1093


Gazook89:
I'm not sure if I should open Issues for each of these thoughts, or leave them here, so let me know if I need to adjust.  This was all discussed in the gitter chat, but thought we should start a more permanent list.

- [ ] Code blocks should only display literal characters.
- [ ] But code blocks should also do syntax highlighting to match the Editor pane if possible.
- [ ] if the `.V3` class name is going to hang round, change it to `.v3` for faster typing.

For my 'code blocks' this is what I currently have...a lot of it would be rendered useless if the first two items above were implemented: https://homebrewery-stage.herokuapp.com/share/L3HE0wM5KoZF

For Monster Stat Blocks: https://homebrewery-stage.herokuapp.com/share/vklcUjeccCu9

And Notes: https://homebrewery-stage.herokuapp.com/share/S1qgpD4flDtQ

calculuschild:
Yup leave general comments in here. That's fine. If we can do code highlighting automatically with a plugin I think I would prefer that rather than disrupting the "only literal characters" in code. We can discuss the highlighting here: #1187 to keep it organized

client/homebrew/editor/editor.jsx
```diff
@@ -78,15 +79,53 @@ const Editor = createClass({
...
 	},
 
-	highlightPageLines : function(){
+	highlightCustomMarkdown : function(){
 		if(!this.refs.codeEditor) return;
 		const codeMirror = this.refs.codeEditor.codeMirror;
 
+		//reset custom text styles
+		const customHighlights = codeMirror.getAllMarks();
+		for (let i=0;i<customHighlights.length;i++) customHighlights[i].clear();
```
AlexeySachkov, line 88: I'm wondering why it is needed in context that we didn't have this reset before - does it mean that we have introduced something new that requires us to make a reset or we had a bug all this time and now it is being fixed?

calculuschild: There are several parts to this:
1) codeMirror has two style settings. "line classes" (the whole line is styled, which we currently do to highlight the `\page` section), and "text marks" (which styles a span of text, and we use on the `{{` characters).
2) The current live site only uses "line classes", but you can see the issue hat currently exists. If you type `\page` the code highlights as we expect, but then if you type extra characters `\pageblahblah` the line is still highlighted even though the code is no longer valid. Same thing if you delete or scramble characters `\pablahblahge` because once the line style is applied, it just remains until the line itself is deleted.
3) The above code handles the refresh for the `text marks` (otherwise once a `{{` block is colored, those colors will just hang around forever), but there is also an added block that now refreshes the "line classes" for `\page` and `\column`.

AlexeySachkov: Thanks for clarification!

Issue #1148, opened 2020-12-10
Add cmd+i and cmd+b for Mac users


Gazook89:
If possible to update the editor to allow the use of common keyboard hotkeys to make formatting text faster, that would be awesome.  Examples would be `ctrl/cmd + i `and `ctrl/cmd + b` to italicize and bold, respectively.  And/Or something like the FancyPants editor on Reddit.

Just throwing feature requests out there to see what sticks!

calculuschild:
Currently I believe ctrl I and ctrl b already insert asterisks in the position. Is this not working for you?

Keep the suggestions coming though! It's all good!

ericscheid:
The editor pane does already have keyboard shortcuts for bold and italics .. are these not working for you?

Gazook89:
Ah, ctrl does indeed work.  However, on a Mac, that isn't "right".  Control works, but for most using a Mac the expectation would be that Command would do the trick instead.  Perhaps if both were functional?

Also, this seems to be a one-way thing.  It's not possible to un-asterisk bold or italics using the same hotkeys.  It just doubles up.

calculuschild:
Ah I see. In that case, we can try to make both key combos work. That shouldn't be too hard.

There also might be a way to undo italics, but I can see that getting complicated.

For instance if you have this text and your cursor is at the X:

`This sentence *has somXe italics and **bold**** inside!`

What happens when you hit ctrl-i? Does the word "some" get asterisks around it?   Does a new set of asterisks appear in the middle of "some"? Do the outermost asterisks disappear? Not only do you have to parse all the nearby text for italics/not italics, but you have to count numbers of asterisks to make sure you don't mess up the bold on "bold". And that problem is a nightmare. Maybe possible, but probably rough. :/

Gazook89:
Why not just add the asterisks where the highlight/selection starts and ends? If nothing is selected, including when the cursor is just inside a word, nothing happens. 

It should be relatively intuitive that the set of asterisks you need (depending on whether you choose italic or bold) appear at the start or end of a selection, even if it sort goofs up the format around other asterisks. 

In your example, nothing would happen, since the cursor hasn't highlighted anything. If you highlighted the word Some, and hit ctrl I, a set of single asterisks get added to either side of Some and it no longer becomes italicized, but leaves the rest italicized.

calculuschild:
So here's where we are at now:

> Why not just add the asterisks where the highlight/selection starts and ends? 

[...]

PR #1191, opened 2021-01-10
Updated extraKeys (bold and italic) and added new shortcut (for span tags)

RKuerten:
So, this pull request is to address #1148. I also added a shortcut to wrap a selection in span tags (I chose Ctrl+F1 because it don't think it messes with any browser/system shortcuts afaik, but I'm open to suggestions). With the new renderer, it may be interesting to add a shortcut for the new divs (`{{ }}`) as well.

I can't test if the Cmd+B/+I/+F1 shortcuts work (I'm on Windows), but there's no reason for them not to work.

I left ~~strikethrough~~ key maps out, but I can add them too.


ericscheid:
I would caution away from mapping `Cmd-[x]`, as those are the keys the browser uses for commands. Right now, Chrome doesn't have any key bindings for `cmd-B` or `cmd-I`, but it does have a mapping for `cmd-S`.

So, while we could be cautious and judicious with selecting specific keys for homebrewery (to avoid clashes), we will end up having to choose some odd-looking alternatives (e.g. we can't use `cmd-S` for ~~strikethru~~ so we to choose something else which is less obvious). [Actually, we've already hijacked `cmd-S` from the browser, invoking the "save edits of brew" function.]

Although there are no _characters_ for `ctrl-{` (like there are for `ctrl-B`), JS doesn't detect characters, it detects keys and modifiers. So we could go ahead with `ctrl-{` for span (or for `{{..}}` blocks/spans).

calculuschild:
This seems fine to me. Other websites (like Github) also have keybindings for Ctrl-I, Ctrl-B for bold, italicize. It's not unusual.  CodeMirror even explicitly gives examples of doing this to override existing browser key-combinations. And we do already use CTRL-S for saving; it's a hotkey for most other text editors, so it makes sense to change the default key-combo behavior in that case. The Homebrewery is a web-based App, not a static page in the browser, and many other "text-editor" webapps do the same thing. This is very common practice.

I do like the idea of changing the "span" key combo to use the `{` button though. @RKuerten Would you make that change and then I can merge this?

Issue #1188, opened 2021-01-07
/source method does not escape &


ericscheid:
From the share page view of any brew (https://homebrewery.naturalcrit.com/share/:id), the `Source </>` button fetches a page view of the brew revealing the source for copy/paste (https://homebrewery.naturalcrit.com/source/:id).

This "source" view is actually the brew text presented within a `<code><pre>` block (probably because some browsers do content sniffing and might treat the response payload as html despite the response header saying it is text/plain).

The HB code thus converts any `<` or `>` to `&lt;` and `&gt;` so they'll _display_ as `<` and `>` in the browser.

The same escaping should occur for `&`, being html-escaped as `&amp;`, and consequently _displayed_ as `&`. 

It doesn't, and while generally this wouldn't matter much (brew source of `&times;` will get displayed as `×`, which when copy/pasted works just fine).

There are however edge cases that will cause problems. For example, brew source `&#42;hello;&#42;` will display as `*hello*` .. but when _that_ is copy/pasted it will get interpreted and displayed as *`hello`* (i.e. in italics).

A simple fix on two lines. PR incoming.

ericscheid:
Lines 76 and 86 of server.js, btw.

G-Ambatte:
Funnily enough, I was just looking at the same section of code for a completely different reason. Due to the duplication of code (in that exact same lines appear in duplicate places in `server.js`, it may be beneficial to push this to a separate function/module, so that corrections like this only need to be made once.

ericscheid:
This is now fixed via both Source and Download methods.

Issue #1129, opened 2020-12-02
[Feature Request] Add ability to download all brews as .zip folder from account


Gazook89:
I would like the ability to quickly download my brews to my computer.

1. Download all source code as text files
2. Download all rendered brews as PDFs  (less important, to me personally)

Allowing this will help with backing things up easily.  This came to mind as I consider linking my Google Drive account*, and the thought of backing up all of my brews by going into each one and copy/pasting all the text into separate files fills me with dread.  I have done this previously, but since I'm not sure which ones have since been revised I have to redo them all.

---
*side note, does linking to Google Drive create a new folder or something in my Drive, filled with text files?  I honestly thought I had linked already with Drive, but don't see anything to indicate as such in my Drive folders, so i'm not sure.  When I go to the 'log-in' screen it shows I'm logged in as my Homebrewery username...if I'm logged in with Google, woudl it should my google email instead?  This is way off topic here, sorry.

calculuschild:
As for your side note:

Linking to Google Drive will create a "Homebrewery" folder on your drive with text files. It doesn't automatically copy over existing brews though, so the folder will be empty to start. I *think* moving it around will still work fine if you don't want it in the root of your Google Drive. You would need to manually open each brew, then click the google Drive icon, at which point the next save will copy your brew to your google Drive and then remove the existing copy on the Hombrewery servers. If you linked your google account to an existing homebrewery account, the username is your old homebrewery name. Similarly, if you created a new account directly with the "log in with google" button, it would have asked you to sign in with an existing homebrewery account or create a new one so it could be linked.

You can tell if you are logged in with google by trying to create a new brew. If you have a linked Google account, it will default to creating the brew on your google drive when you save it for the first time.

Gazook89:
Ok, awesome.  I think what confused me most was that I clicked the G Drive icon in the editor, which "turned it on", but did nothing else...I didn't know I had to make an edit (type a little), which causes the auto-saver to trigger, and then it would create a copy.  Not until I made an edit did anything happen...and since I didn't know this, I sat staring at it for a few moments before thinking "this isn't working".  And then my understanding is that if I turn on this G Drive link with all my existing "how to" brews, that I'll have to go back to my reddit posts and update my share links, correct?  It's not really a problem as most of my Share links are in comments rather than being "link posts".

calculuschild:
We can probably make it save right away when you click the g drive button. That might be more clear to users.

And yes, you will get a new link if you move it to the g drive.

[...]

PR #1198, opened 2021-01-20
Download Brew Source as Text File

G-Ambatte:
# Download Brew Source

## Purpose

The purpose of this branch is to allow users to download the source of a brew as a plain text file.

### Work Completed So Far

- [x] Create code module to output Brew source as a text file
- [x] Add a new page (`download`) to get the Brew source using its ID
- [x] Incorporate the "Download" option into nav bar on Share page
- [x] Incorporate download option in brew items on User page

### Considerations

- [x] Lint/CircleCi - The 200 line maximum requires that the code for the new page is moved to a new module, which I have called `sourceFunctions.jsx` in the `sharePage` directory. It seemed logical to also move the code for the existing "source" page here as well, as large portions of the two functions are identical.
- [x] "Source" vs "Download Source" in Share page nav bar - I have changed the button text from "Source" to "View Source", to better differentiate the function of the two buttons.

## Conclusion

This functionality has been requested previously (e.g. #1129, or [this workaround in JS from Reddit](https://redd.it/aerj7c)). This also goes some way towards a user being able to download all of their brews at once, either as several text files or a single ZIP archive.

It's probably not the cleanest code I've ever written and it's well past midnight so I'm doing this with only one eye opened as I write this PR, so I'd appreciate any feedback that you might have.
-- G

---

EDIT: These changes should also fix the issue reported with the unmodified ampersand in the HTML source as reported in issue #1188. I added the check and replacement strings as a key/value array, so it should simplify adding any further items to be replaced in the future, if any more should be discovered.


G-Ambatte:
Shifted the functions back into `server.js` and removed both the unnecessary changes to `Homebrew.Model.js` and the new module. Only four files are now affected by this change, at least two of which are trivial changes.
At this point, I haven't got a working NaturalCrit login system yet, so I've been unable to test with a logged in user yet, and by extension, any of the GoogleActions.

- [ ] Test functionality with Google brews

G-Ambatte:
- [x] Test functionality with Google brews

Login system now working, including Google linking. Download confirmed working identically for Google-stored brews as it does for locally stored brews.

Discussion points:
- [ ] **Download filename prefix:** Prefixing the filename with "Homebrewery-" makes for quite a long filename. Would this feature benefit from a shorter filename prefix, e.g. "HB-"?
- [ ] **Other issues:** What else have I forgotten to consider?

G-Ambatte:
`/download` appears to be working correctly in all respects, for all brews - local and on Google drive.

calculuschild:
@G-Ambatte Just need the config value for `const prefix = config.get('name_prefix');` in `server.js`, since the config file is missing from this PR. Can you remind me what we wanted to use for that?

Gonna try to trim down the Server.js since a lot of stuff is duplicated between HB and Google brews (replacing `<` with `&lt`, prepping the file for download, etc). This does reveal a lot of other duplicated stuff in the other URLs as well so I might refactor those as well.

G-Ambatte:
> @G-Ambatte Just need the config value for `const prefix = config.get('name_prefix');` in `server.js`, since the config file is missing from this PR. Can you remind me what we wanted to use for that?

I was using "Homebrewery' but Gazook89 suggested "HB", which I have no objections to. From memory, I moved it to a config item so it could easily be changed later but also so I didn't actually have to make that decision at the time.

calculuschild:
Ok. I'm just cleaning up the duplicated functions a bit and then this can merge and go out with v2.11.0

PR #1242, opened 2021-02-15
Import from Share ID

G-Ambatte:
New functionality that pushes the text from an existing brew's ShareID into a New brew page.

It is expected that this will be useful for:
- assisting users with issues without interfering with their work, as they often accidentally share their edit links
- creating documentation demonstrating specific CSS/Markdown techniques
- other stuff that I haven't thought of yet

As always, please cast your eyes over this and let me know if/how/where the code can be improved.

Regards
-G


calculuschild:
I appreciate the idea here. However, I am not really a fan of making this into an entire separate page that is essentially a duplicate of the `new` page, which is itself a duplicate of `edit`, as is the `homepage`. I've been running into lots of small bugs with the v3 update because I make a change on one but forget to update all 3. In fact, unifying all of the "edit-type" pages into one has been rolling around as a needed maintenance step for a while now.

Given that, do you think you could modify this to instead send the imported text to the "new" page? You can have the `/import` path in the `server.js` generate the brew text, but then reroute to a `new` page after that point rather than a nearly-identical `/import` page.

G-Ambatte:
I've commented out the existing `/new` and changed the new `/import` to be `/new` instead. Initial testing shows this is working exactly as per the previous (`/new` results in a blank new, `/new/{shareID}` results in a new page loaded with the source from the brew with that shareID.

G-Ambatte:
> In fact, unifying all of the "edit-type" pages into one has been rolling around in my head as a needed maintenance step for a while now.

This has been percolating in the background of my mind for a while, ever since I realized just how similar big chunks of the pages are. However thus far I haven't committed to actually re-writing the page functions.

calculuschild:
> I've commented out the existing `/new` and changed the new `/import` to be `/new` instead. Initial testing shows this is working exactly as per the previous (`/new` results in a blank new, `/new/{shareID}` results in a new page loaded with the source from the brew with that shareID.

Alright, this is a good start. Now, ideally, I would like get rid of `importpage.jsx` and `importpage.less` so all of this logic is contained within `newpage.jsx` (if there is a brew imported from the share id, pass it in to the page to be displayed, but if not, pass in nothing so we get the same "empty" new page). Do you think you could make that work?

G-Ambatte:
> 
> 
> > I've commented out the existing `/new` and changed the new `/import` to be `/new` instead. Initial testing shows this is working exactly as per the previous (`/new` results in a blank new, `/new/{shareID}` results in a new page loaded with the source from the brew with that shareID.
> 
> Alright, this is a good start. Now, ideally, I would like get rid of `importpage.jsx` and `importpage.less` altogether so all of this logic is contained within `newpage.jsx` (if there is a brew imported from the share id, pass it in to the page to be displayed, but if not, pass in nothing so we get the same "empty" new page). Do you think you could make that work?

Actually, that's exactly how this branch is currently working; under the current iteration of the change, `newpage.jsx` and `newpage.less` never get called - everything runs through the `importpage` code.  
I've got a bit going on right now but I should be able to push a change to this branch in the next 48 hours where the entire `importpage` directory has been removed and all of the code references pushed back to `newpage`.

G-Ambatte:
> how this branch is currently working

Or rather, how I think it *SHOULD* be working. I'm not one to ignore the possibility that I've made a grievous error somewhere, or completely misunderstood how the code actually works.

calculuschild:
Good, good. Sorry for my confusion. Yeah, if you can make that change then I'll go ahead and merge it.

G-Ambatte:
Well, it WAS working fine... Something has gone awry after the conflict resolution with `master`. Chasing it down now...

G-Ambatte:
Got it; it's working now - `/new/{shareID}` loads the brew source into the text, but `/new` still gives a blank page as per the current functionality.

PR #1251, opened 2021-02-21
Fix title issue

G-Ambatte:
This change sets the maximum title length to 100 characters. This value is set in the constant `MAX_TITLE_LENGTH` in the `homebrew.api.js` file, although it likely could be moved to a config file if desired.

Merging this PR should close Issue #1250, or at least minimize the issue so a deeper investigation can take place at a later date.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
 	root          : true,
 	parserOptions : {
-		ecmaVersion  : 9,
+		ecmaVersion  : 2021,
 		sourceType   : 'module',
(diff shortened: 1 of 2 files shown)


G-Ambatte:
> Would be better having as two lines to keep the intent clear
> 
> line 18: selecting of text from elements: `let title = (tokens[...]).text;`
> line 19: limiting length to MAX_TITLE_LENGTH: `title = title.substr(0, MAX_TITLE_LENGTH);`

Sorry for the delay, for some reason I thought I'd already reverted that commit. Clearly working too hard - damned day jobs!

PR #1277, opened 2021-03-09
Parametrize fillMode for .useColumns

jsocol:
A @fillMode parameter for .useColumns9) allows setting 
`column-fill` property to `balance` for wide statblocks (or other
wide column types) instead of relying on Chrome's fallback
behavior that ignores `auto` when there's no set container height.
Set the default for the for `@fillMode` to `auto` to maintain the
current behavior.

### DIFF
--- a/themes/5ePhb.style.less
+++ b/themes/5ePhb.style.less
@@ -34,9 +34,9 @@ body {
 		letter-spacing : -0.02em;
 	}
 }
-.useColumns(@multiplier : 1){
+.useColumns(@multiplier : 1, @fillMode: balance){
 	column-count         : 2;
-	column-fill          : auto;
+	column-fill          : @fillMode;
 	column-gap           : 0.9cm;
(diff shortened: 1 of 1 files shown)


calculuschild:
Interesting. I didn't realize `balance` was a keyword we had available to us. I will look into this PR and make sure it doesn't break any existing brews, and if it does, we can always merge it into the v3 renderer.

Thanks!

jsocol:
I got to learn some things when I ran into this. `balance` is actually the default/initial value, and `auto` only makes sense in contexts with fixed heights. Unfortunately the spec isn’t particularly clear what to do when `auto` is combined with a flexible height. Chrome handles this by ignoring `auto` and falling back to the default. Firefox handles it by letting the first column just keep getting longer. I imagine Safari also has the fallback behavior but I haven’t tested it.

RKuerten (approved):

Issue #1285, opened 2021-03-18
Simplify the opting out of the fancy first letter


G-Ambatte:
The question is often asked in the subreddit: "How do I remove the fancy first letter after a title?" Stepping the user through the process of using CSS to select `.phb h1 + p::first-letter` and then remove the required formatting is both time consuming and error-prone.

Possible solutions:
1) add a snippet to insert the CSS to remove the fancy first letter; or
2) remove the fancy first letter from `phb.less` so it is no longer a default option and users must opt into using it.

Issues arising:
1) none apparent;
2) existing brews will need to be modified to maintain their existing styling.

Given the potential issues of option 2, I suspect option 1 is the better path to follow.

calculuschild:
Option 1 sounds fine to me. :thumbsup:

G-Ambatte:
I'll try to draft a PR tonight if I can get some time; the actual required code to be inserted has already been determined by /u/TheLurkerOne [here](https://www.reddit.com/r/homebrewery/comments/m82ybe/removing_fancy_first_letter_after_a_xxx_title/grf3x2s), so the PR is really just wrapping it in the necessary JS for the snippetbar change and finding an appropriate FA icon.

calculuschild:
https://fontawesome.com/icons/remove-format?style=solid

G-Ambatte:
> https://fontawesome.com/icons/remove-format?style=solid

Perfect!

G-Ambatte:
It really is... but you do raise a good point. Perhaps the snippet should be two snippets - one for disabling, which is `all: unset`; and one for tweaking, which is the explicit and verbose version.

G-Ambatte:
#1286 should close this issue.

PR #1286, opened 2021-03-19
Add snippet for removing drop cap (aka fancy first letter).

G-Ambatte:
This PR addresses issue #1285, adding a snippet (to both `snippets` and `snippetsLegacy`) that removes the drop cap from the first paragraph immediately after a big header (h1).


G-Ambatte:
Commit 7491f46 adds the "Remove Drop Cap" snippet for both Legacy and v3. Commit 712824d also adds the "Tweak Drop Cap" snippet.

PR #1354, opened 2021-05-17
Incorporate HTML "lang" attribute

G-Ambatte:
First pass at incorporating the HTML language attribute (`html lang='brew.metadata.language'`).

- [ ] Add functionality to set the brew language in the Metadata Editor.

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -18,7 +18,8 @@ const MetadataEditor = createClass({
 				published   : false,
 				authors     : [],
 				systems     : [],
-				renderer    : 'legacy'
+				renderer    : 'legacy',
+				language	: 'en'
 			},
 			onChange : ()=>{}
 		};
--- a/client/template.js
+++ b/client/template.js
@@ -1,7 +1,9 @@
 module.exports = async(name, title = '', props = {})=>{
+	const lang = this.props.metadata.language ? this.props.metadata.language : 'en';
+	
 	return `
 <!DOCTYPE html>
-<html>
+<html lang="${lang}">
 	<head>
 		<link href="//use.fontawesome.com/releases/v5.15.1/css/all.css" rel="stylesheet" />
 		<link href="//fonts.googleapis.com/css?family=Open+Sans:400,300,600,700" rel="stylesheet" type="text/css" />


G-Ambatte:
The template now includes `<html lang='en'>`, with `'en'` which updates to follow the value of `brew.language`. The `.less` probably needs some work to improve the UI/UX, so I'm marking this as Ready for Review.

server.js
```diff
@@ -188,7 +188,7 @@ app.use((req, res)=>{
 		account     : req.account,
 		enable_v3   : config.get('enable_v3')
 	};
-	templateFn('homebrew', title = req.brew ? req.brew.title : '', props)
+	templateFn('homebrew', title = req.brew ? req.brew.title : '', props, lang = req.brew ? req.brew.language : 'en')
```
calculuschild, line 191: `lang` will already exist inside of `props.brew.language`. We don't need to pass in another parameter here.

client/template.js
```diff
@@ -1,7 +1,7 @@
-module.exports = async(name, title = '', props = {})=>{
+module.exports = async(name, title = '', props = {}, lang='en')=>{
 	return `
 <!DOCTYPE html>
-<html>
+<html lang="${lang}">
```
calculuschild, line 4: I don't think this is where we want to apply the `lang` attribute. Since users will be setting this from the `metadata` panel, it will cause the brew document to update its HTML, but not trickle up to the full web page here without a page refresh. This probably needs to be applied at the `brewRenderer.jsx` page, possibly in the `initialContent` variable there...

server/homebrew.model.js
```diff
@@ -16,6 +16,7 @@ const HomebrewSchema = mongoose.Schema({
 	renderer    : { type: String, default: '' },
 	authors     : [String],
 	published   : { type: Boolean, default: false },
+	language   	: { type: String, default: 'en' },
```
calculuschild, line 19: Make similar changes for the Google Drive process. Look in `googleActions.js` at the `updateGoogleBrew`, `newGoogleBrew`, and `readFileMetadata` functons.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -18,7 +18,8 @@ const MetadataEditor = createClass({
 				published   : false,
 				authors     : [],
 				systems     : [],
-				renderer    : 'legacy'
+				renderer    : 'legacy',
+				language   	: 'en'
```
calculuschild, line 22: Run the lint script if you haven't and align the colons if it doesn't automatically. I think everything here is using spaces to align colons.

PR #1366, opened 2021-05-27
Fix #773 Wide Class Table snippet missing header

### LINKED ISSUE #773: Wide Class Table: missing header
## Description

The "Class Table" functionality is incomplete. There should be a header saying "—Spell Slots per Spell Level—" positioned directly above the headers labeled 1st, 2nd, 3rd, 4th, etc. There does not appear to be a way to manually put this header in place.

SRugina:
- There should be a "—Spell Slots per Spell Level—" header slightly above
the 1st-9th spell slot headers.

- This fix attempts to add this in as simple a manner as possible; it
involves a tiny bit of CSS, 1 new line to the snippet, and 1 moved line
within the snippet. The new snippet uses the heading row for the new
header and re-styles the first "normal" row to look like a header, and
subsequent rows to have the correct alternating background colours.
- The standalone stylesheet has been updated to reflect this new CSS as
well.

- The new snippet includes the `twoHeadingRows` class to be fully
backwards-compatible with documents using the old snippet.

**Edit:** Here's what the result should look like
image


calculuschild:
Thanks for contributing to this and good catch on the missing header part. I'll look this over.

If there is a way to do this without needing that extra twoHeadingRows class that would be preferred.... Once users have to start putting together lots of niggly options it can quickly turn into feature overload, and I would prefer if we can find a clean way that "just works", and also preferably without sneaking in bits of raw HTML in there, since most users are not HTML experts. It might make more sense if we can modify the Markdown parser so it recognizes multiple headers. Then the markdown also *looks* like it has two rows of headers instead of half above and half below the header line.

Tables are tricky.

[Here's](https://stackoverflow.com/questions/23571724/github-markdown-colspan) an example that still uses a bit of HTML but is a bit more "minimal" and gets you the column-spanning, but I still don't really like that approach.

calculuschild:
MultiMarkdown supports the syntax I think we would want. We can try to adopt that on the Homebrewery. https://fletcher.github.io/MultiMarkdown-5/tables.html

ericscheid:
For reference, since we use Marked.js, and that hews to CommonMark: https://talk.commonmark.org/t/tables-in-pure-markdown/81/126

SRugina:
If the class `twoHeadingRows` is to be removed, that would require changes to the markdown parser - namely, to allow multiple header rows. As mentioned, the amount of raw HTML can be shorted a tad by replacing `<th colspan=4></th>` with `| | | |` but at the moment the colspan=9 part is necessary - removing this would also require a parser change to support colspan.

Again, as mentioned, Multi-Markdown seems to be the only one that supports both of these. Personally, the colspan=9 part is acceptable as it communicates its purpose, so any parser/extension that supports multiple header rows would be a good choice.

However, that does seem like a lot of work and so I'm assuming this issue falls into the larger "V3" pinned issue for a new parser/renderer?

calculuschild:
Yes, it would require a change to the markdown parser, which would go into the V3 branch. Fortunately, I am one of the developers of the markdown parser we use so adding that in isn't too much of a stretch.

I think I'm going to hold off on merging this for now though. Having the snippet with an unusual mixture of HTML and Markdown I think might end up confusing users more than help, especially if they don't actually need that second header, for instance. This is a nice bonus feature we can add later on, but I'd like to do it "properly" via the parser.

I'll leave this open for now though because I think it's a good starting point, and we can add the parser fixes on top of this.

ericscheid:
Maybe we should have a discussion thread where we collect the various requirements for extending tables?

For the content space that HB targets, we already know it falls short with `colspan` for spell slots tables (as above). This use case also includes the twist of there being two rows of `<th>`.

Another `colspan` case would be the indenting typicallly visible on various weapon tables (where there are sub headings for "Simple melee weapons" etc). Currently I see hacks such as `&nbsp;` or even using code spans which our css renders as non-visible (but not hidden).

What other _content-driven_ use cases are there to consider?

Is it typical for some tables to have their `<th>` cells arranged vertically?

calculuschild:
Yep, we can open up a discussion thread for tables.

Just off the top of my head, I think we need:

* Colspan
* Rowspan
* Multiple headers

And ideally we still follow the *prime directive* of Markdown, which is that the code itself remains human-readable as much as possible.

The Commonmark thread you linked to @ericscheid  is a pretty good resource for possible approaches, but also keep in mind that even the Commonmark guys never actually settled on any single approach so at some point we are probably forging our own path here.

SRugina:
Considering that this PR is a fix to achieve the desired effect within the constraints of the current markdown parser, I think this can be safely closed in favour of a discussion thread / larger issue for tables in the new parser.

PR #1373, opened 2021-06-01
Update Dockerfile

earlng:
If a user were to execute the current README.DOCKER.md file as written, they would encounter an error along the lines of "node.js installed is incompatibe".

This is because the current Dockerfile is pulling node:8. But from package.json it seems that homebrewery is expecting something more like node:14.15.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:14.15
 
 ENV NODE_ENV=docker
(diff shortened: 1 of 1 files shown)


calculuschild:
Thanks for this. I personally don't actively maintain the Docker portion on this repo so I appreciate you providing a fix. I'll look it over and get this merged in the next version!

PR #1393, opened 2021-06-17
Incorporate Brew.Style into Source + Download pages

G-Ambatte:
This code is an early commit of an attempt to incorporate `brew.style` into the /source and /download pages.

### DIFF
--- a/server.js
+++ b/server.js
@@ -11,9 +11,11 @@ const sanitizeFilename = require('sanitize-filename');
 const asyncHandler = require('express-async-handler');
 const dedent = require('dedent-tabs').default;
 
+const brewAccessTypes = ['edit', 'share', 'raw'];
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -133,7 +139,8 @@ app.get('/download/:id', asyncHandler(async (req, res)=>{
 		'Content-Type'        : 'text/plain',
 		'Content-Disposition' : `attachment; filename="${fileName}.txt"`
 	});
-	res.status(200).send(brew.text);
+	const text = brew.text;
+	res.status(200).send(text);
```
calculuschild, line 143:
```suggestion
	res.status(200).send(brew.text);
```

We don't need to declare a new variable here since we don't do any processing on it before sending it out.

PR #1394, opened 2021-06-19
Fix merge of undefined Style data

G-Ambatte:
This only merge `brew.style` with `brew.text` if `brew.style` exists or is an empty string.


calculuschild:
Ah. This doesn't actually solve the issue the way I though it would. It's a deeper problem. I need to re-think how to approach this.

G-Ambatte:
With that as the goal, it may be better to close this PR and expand on #1397 instead.

calculuschild:
Superseded by #1397

PR #1397, opened 2021-06-21
Split codefenced CSS from in New Brews from Content to Style tab

G-Ambatte:
Initial pass at code to split any CSS data stored in a triple backtick CSS codefence at the start of the Brew text (`brew.text`) to the Style editor (`brew.style`) before the first save.

### DIFF
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -135,6 +135,13 @@ const NewPage = createClass({
 
 		console.log('saving new brew');
 
+		// Split out CSS to Style if CSS codefence exists
+		if(brew.text.startsWith('```css') && brew.text.indexOf('```\n\n') > 0) {
+			const index = brew.text.indexOf('```\n\n');
+			brew.style = `${brew.style}${brew.text.slice(7, index - 1)}`;
+			brew.text = brew.text.slice(index + 5);
+		}
+
 		if(this.state.saveGoogle) {
 			const res = await request
 			.post('/api/newGoogle/')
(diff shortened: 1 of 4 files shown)


client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -135,6 +135,13 @@ const NewPage = createClass({
 
 		console.log('saving new brew');
 
+		// Split out CSS to Style if CSS codefence exists
+		if(brew.text.startsWith('```css') && brew.text.indexOf('```\n\n') > 0) {
+			const index = brew.text.indexOf('```\n\n');
+			brew.style = `${brew.style}${brew.text.slice(7, index - 1)}`;
```
calculuschild, line 141: Maybe insert a newline `\n` between the existing brew.style and the newly inserted part, so we don't end up with

```
styleFromTab {
 color: red;
}stylePastedIn {
 color: blue;
}
```

G-Ambatte: Added in new commit.

calculuschild:
Ok, so lets modify a bit then, to only append to the existing Tab data if it is not the example text. But we don't want to simply test if `brew.style == exampleText", in case the user has fiddled with the brew, and then saved, and then did an undo to restore the example text, we would now want that example text saved as part of their brew.

I think what we might need to do is have an `exampleText` variable in `editor.js`. 

https://github.com/naturalcrit/homebrewery/blob/d61fda9cff13cf755994da5b6d8813a74ca15291/client/homebrew/editor/editor.jsx#L167-L181

Maybe in the `Render()` function of the editor, we tell it to display `value = exampleText` by default until the user modifies the tab and `brew.style` becomes filled. And then we make sure `brew.style` is not defined anywhere until the user types something or if it was loaded from the brew.

PR #1401, opened 2021-06-21
Add local storage key for Brew Style data

G-Ambatte:
Add a new local storage key to the `/new` page for storing Brew Style data.

### DIFF
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -158,6 +158,7 @@ const NewPage = createClass({
 
 			const brew = res.body;
 			localStorage.removeItem(BREWKEY);
+			localStorage.removeItem(STYLEKEY);
 			window.location = `/edit/${brew.googleId}${brew.editId}`;
 		} else {
 			request.post('/api')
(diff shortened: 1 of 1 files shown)


client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -70,12 +73,18 @@ const NewPage = createClass({
...
+				brew : { text: brewStorage }
 			});
-		}
+		};
+		const styleStorage = localStorage.getItem(STYLEKEY);
+		if(!this.props.brew.style && styleStorage){
+			this.setState({
+				brew : { style: styleStorage }
+			});
+		};
```
calculuschild, line 87: Unfortunately this doesn't quite work. If both localStorage keys exist, then you are first setting the state to be a brew object with only `text = brewStorage` and then immediately overwriting it with a brew object that only has `style = styleStorage`. One option is to use something like below, which you can see in other places in this page. The other option would be to just rewrite this so both parts get applied in the same `setState` call. I'm going to go ahead and do that to merge, but just FYI for the future because it comes up a lot in the Homebrewery.

```
this.setState({
  this.set_.merge(brew, {text: brewStorage});
});
```

PR #1408, opened 2021-06-23
Add dropdown to `/share` for `source` to unify options:

G-Ambatte:
* "view" - View Source
* "download" - Download Source
* "clone to new" - Create New Brew from this Brew

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -3,17 +3,26 @@ const React = require('react');
 const createClass = require('create-react-class');
 const _ = require('lodash');
 const cx = require('classnames');
+const dedent = require('dedent-tabs').default;
 
 const CodeEditor = require('naturalcrit/codeEditor/codeEditor.jsx');
 const SnippetBar = require('./snippetbar/snippetbar.jsx');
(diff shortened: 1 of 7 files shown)


calculuschild (changes requested):
This is all very good. Since this has expanded the button to be more complicated than a one-liner (including several functions), could you break it out into its own component, similar to the `recent.navitem.jsx` component? (even better would be to create a generic "dropdown" component that both this and `recent.navitem` can share to reduce code duplication, but that's probably a different PR)

Second, it looks like the dropdown sub-items are not quite the same width as the top `source` button. Can you try to align those?

image

Third, could we vertically-center the text? Something looks just a bit off. Might also be the "Clone to New" wrapping into a second line that looks a bit funny, but I'm no sure a better way to display that.

G-Ambatte:
On the second issue, I'm not seeing it on my local install: image

G-Ambatte:
> (even better would be to create a generic "dropdown" component that both this and recent.navitem can share to reduce code duplication, but that's probably a different PR)

I think that this is the best solution, but probably exceeds the scope of this PR. I'll raise a separate issue to split that functionality to it's own module.

PR #1415, opened 2021-06-28
Display state of Google credentials on New page

G-Ambatte:
This is an early coding pass at an attempt to display the current state of the Google credentials associated with the account on the `/new` page.

Checklist:
- [x] Add Google Drive icon to `/new`
- [x] Add error message window with link to login page
- [ ] test for the current state of credentials (poss. `GoogleActions.authCheck()` ???)

### DIFF
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -20,6 +20,9 @@ const BrewRenderer = require('../../brewRenderer/brewRenderer.jsx');
 const BREWKEY = 'homebrewery-new';
 const STYLEKEY = 'homebrewery-new-style';
 
+const googleDriveActive = require('../../googleDrive.png');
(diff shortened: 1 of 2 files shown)


calculuschild:
After looking at this, it seems like we're ending up with almost exact duplication of the `edit` page. Why don't we work on just merging the two pages at this point?

G-Ambatte:
> After looking at this, it seems like we're ending up with almost exact duplication of the `edit` page. Why don't we work on just merging the two pages at this point?

Thinking about it, isn't `/new` really just `/edit` that isn't saved? So `/edit` (as opposed to `/edit/:id`) could/should work just the same as `/new` does now. Although this then raises the issue of how to maintain the "Clone to New" functionality...
I'll think on it some more before pushing ahead in either direction.

G-Ambatte:
Closing in favour of continuing with #1423.

PR #1423, opened 2021-07-03
Unify new and edit pages

G-Ambatte:
Draft PR for the unification of the `newPage` and `editPage` files. At this point, the code appears to be completely functional and all `newPage` files have been removed.

**CHECKLIST:**
- [x] `/new` saves to and loads from local storage automatically
- [x] clicking save on `/new` redirects to `/edit` and wipes local storage 
- [x] importing from a Share ID via `/new/:id` works
- [x] saving on `/edit` works locally
- [ ] transfer to/from Google Drive works
- [ ] other functionality that I've currently forgotten about

Notes:
**Transfer to/from Google Drive:** should work, but testing is required.


client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -52,35 +57,81 @@ const EditPage = createClass({
...
+				authors     : this.props.brew.authors || [],
+				systems     : this.props.brew.systems || [],
+				renderer    : this.props.brew.renderer || 'legacy'
+			},
 			isSaving               : false,
 			isPending              : false,
 			alertTrashedGoogleBrew : this.props.brew.trashed,
 			alertLoginToTransfer   : false,
-			saveGoogle             : this.props.brew.googleId ? true : false,
+			saveGoogle             : (global.account && global.account.googleId ? true : false),
```
calculuschild, line 78:
```suggestion
			saveGoogle             : this.props.brew.googleId || (global.account?.googleId && this.props.editType == 'new') ? true : false,
```
We want to load the page with `saveGoogle` true in two cases:
1) The file is already saved to Google (this.props.brew.googleId exists)
2) It is a new page (editType = 'new') and the user is signed in with Google

This currently only sets `saveGoogle` true if the user is signed in with Google. It will ignore whether the brew was previously saved to Google, and if a user visits the edit page accidentally without being logged in, any save will transfer the brew to HB.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -6,26 +6,31 @@ const _ = require('lodash');
...
-
 const googleDriveActive = require('../../googleDrive.png');
 const googleDriveInactive = require('../../googleDriveMono.png');
 
 const SAVE_TIMEOUT = 3000;
 
+const BREWKEY = 'homebrewery-new';
+const STYLEKEY = 'homebrewery-new-style';
+
+const editTypes = ['edit', 'new'];
```
calculuschild, line 32: I don't think this is necessary. We shouldn't ever have a case when an invalid editType comes through since the only place it can come from is us, at `homebrewery.jsx`

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -52,35 +57,81 @@ const EditPage = createClass({
...
+			if(this.props.editType == type) {
+				this.editType = this.props.editType;
+			};
 		});
+		if(!this.editType){
+			// ERROR: Unknown editType
+			console.log('Unknown EditType parameter!');
+			// Redirect to Home page; need to revisit with better Error messaging in the future
+			window.location.href = `/`;
+		};
```
calculuschild, line 100: See above comment. All of these checks aren't necessary. We also don't need to worry about redirecting at this stage. That is already handled in the React router (if a user visits anything that is an invalid path, it takes them to the home page).
```suggestion
```

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -52,35 +57,81 @@ const EditPage = createClass({
...
+
+		if(this.editTypeIsEdit()) {
+			this.saveGoogle = this.props.brew.googleId ? true : false;
+		}
```
calculuschild, line 104:
```suggestion
```
Not needed. This will already be set in `getInitialState()` (with my changes). Also, keep in mind `this.saveGoogle` is not the same as `this.state.saveGoogle`. I don't think this does anything as it is anyway since `this.saveGoogle` isn't referenced again.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -52,35 +57,81 @@ const EditPage = createClass({
...
+				editId      : this.props.brew.editId || null,
+				gDrive      : this.props.brew.gDrive || false,
+				title       : this.props.brew.title || '',
+				description : this.props.brew.description || '',
+				tags        : this.props.brew.tags || '',
+				published   : this.props.brew.published || false,
+				authors     : this.props.brew.authors || [],
+				systems     : this.props.brew.systems || [],
+				renderer    : this.props.brew.renderer || 'legacy'
+			},
```
calculuschild, line 73:
```suggestion
			brew : this.props.brew,			
```
Leave this as it was. Both `/edit/:id` and `new/:id` should come with a `this.props.brew` object that already has all of these filled. `new/` doesn't get any brew object from the server, so `this.props.brew` instead uses the `getDefaultProps()` value above. In either case, by the time we get to this step, we have a fully valid `this.props.brew` and we don't need to copy each individual brew property component line by line anymore.

I know several of our pages have something like this where the props are copied line by line back into `this.state`, but I think that was a misunderstanding of Scott when he first built this years ago.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -52,35 +57,81 @@ const EditPage = createClass({
...
-			if(this.state.isSaving || this.state.isPending){
+			if(!editTypeIsNew && (this.state.isSaving || this.state.isPending)){
 				return 'You have unsaved changes!';
 			}
 		};
 
+		console.log(brew);
+
 		this.setState((prevState)=>({
+			brew       : _.merge({}, prevState.brew, brew),
```
calculuschild, line 133: Just FYI: You can have separate `this.setState` calls; you don't have to bunch all of the variables together in one call like this if it makes the code easier to decipher. The only issue is if you have two or more `setState` that modify the same variable, then you might have a conflict.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -91,6 +142,14 @@ const EditPage = createClass({
 		document.removeEventListener('keydown', this.handleControlKeys);
 	},
 
+	editTypeIsNew : function() {
```
calculuschild, line 145: Can you come up with a shorter name for these? The purpose of a function like this is essentially shorthand for `this.props.editType == 'edit'`, but with a longer name like this it's not that much shorter. I might actually prefer just checking `this.props.editType == 'edit'` directly because it's a bit more straightforward to me.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -91,6 +142,14 @@ const EditPage = createClass({
...
+		return (this.editType == 'new');
```
calculuschild, line 146: There is no real reason to create an identical variable like above that I can see:
```
if(this.props.editType == type) {
	this.editType = this.props.editType;
};
```

`editType` should never change while the user is viewing the page. If they save a `new` brew, the page redirects to `/edit` which will give us a new editType, but we aren't ever changing it within the page itself. For constant settings like this, it is preferred to just check `this.props` directly.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -139,11 +198,18 @@ const EditPage = createClass({
 	},
 
 	trySave : function(){
-		if(!this.debounceSave) this.debounceSave = _.debounce(this.save, SAVE_TIMEOUT);
-		if(this.hasChanges()){
-			this.debounceSave();
-		} else {
-			this.debounceSave.cancel();
+		if(this.editTypeIsNew()) {
```
calculuschild, line 201: `trySave()` is meant just to be a check so we don't save too frequently (we don't want to save with *every* character the user types or we overload the servers.) but instead limits to every 3 seconds. All the actual "save" logic needs to go in the `save()` function, but I think the "debounce" stuff can apply to both New and Edit pages.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -194,86 +260,134 @@ const EditPage = createClass({
 			htmlErrors : Markdown.validate(prevState.brew.text)
 		}));
 
-		const transfer = this.state.saveGoogle == _.isNil(this.state.brew.googleId);
+		if(this.editTypeIsNew()) {
```
calculuschild, line 263: I feel like there should be some way we can better condense the saving logic between the New and Edit pages, since they both have to go through the hoops of whether to save to Google or not, etc. It would require some deeper digging though.

calculuschild (commented):
I know this is a lot, so please ask if you disagree with any of it or have questions; this is super awesome work so far and I think it's getting close.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: As I recall, I was getting errors where `this.state.brew.text` was still undefined after `setState` had been called to set it. I put it down to the asynchronous nature of `setState` and reduced the number of calls by creating a `const brew = this.state.brew`, applying the changes to that object, then merging it in a single `setState` at the end of the function. I stopped getting the error, so I moved on.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: `this.editType` was only created as a separate entity to hold the result of the `this.props.editType` validation. Now that it's been/being removed, we can shift to using `this.props.editType` for exactly the reasons you've outlined.

client/homebrew/pages/editPage/editPage.jsx
calculuschild: I see. Yes, that is another thing to think about; `setState` doesn't immediately update, so you can't start querying it's values like, in the same function. If this is working for you now then you can keep it as is.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: Agreed - I've been pushing towards "functional" and less worried about "clean" or "elegant". But there's certainly space for it once the merging is complete.

client/homebrew/pages/editPage/editPage.jsx
calculuschild: Perfect.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: Yes, it can absolutely be renamed to something shorter. `typeNew()`, `typeEdit()` or something to that effect.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: I'm just experimenting with a change now, instead of pushing the `/new` data to localStorage immediately on any change, it instead checks for New/Edit when `debounceSave` is triggered; if new it pushes the data to localStorage and cancels the debounce. If the type is Edit, it pushes the data through the normal `this.save()` function.

client/homebrew/pages/editPage/editPage.jsx
calculuschild: I do like the `is` style for straight boolean checkers like this. How about `isNew()` and `isEdit()`?

client/homebrew/pages/editPage/editPage.jsx
calculuschild: That sounds reasonable to me.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: I've just pushed through a new commit where the functions are renamed to `isNew()` and `isEdit()`.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: Hopefully the current commit is both cleaner and more elegant. I'll resolve this conversation and we can open new ones for anything that leaps out from the current commit.

calculuschild (changes requested):

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -141,7 +166,15 @@ const EditPage = createClass({
...
-			this.debounceSave();
+			if(this.isEdit()) {
+				this.debounceSave();
+			}
+			if(this.isNew()) {
+				//console.log(this.state.brew.text);
+				localStorage.setItem(BREWKEY, this.state.brew.text);
+				localStorage.setItem(STYLEKEY, this.state.brew.style);
+				this.debounceSave.cancel();
+			}
```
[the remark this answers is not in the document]
G-Ambatte: It can be moved; `save()` is called by the Save button click handler but also this automatic save function, so those two states will need to be disambiguated.

G-Ambatte:
Reverted all commits after the initial merge was functional.

client/homebrew/pages/editPage/editPage.jsx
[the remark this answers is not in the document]
calculuschild: `trySave()` should trigger automatically when a user types (or edits metadata, or changes google Drive status) and we detect that the content has changed. This just sets a timer for `SAVE_TIMEOUT` miliseconds and ignores any other save requests until that time is up, at which point it calls `save()`. If the user undoes their change before that happens, the save will not trigger if I remember right.

We can keep the save button calling `save()` directly since I don't think users want to click and nothing happens for 3 seconds.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -170,11 +207,20 @@ const EditPage = createClass({
...
+				errors     : null
+			}));
+		}
+		if(this.isEdit()) {
+			this.setState((prevState)=>({
+				saveGoogle : !prevState.saveGoogle,
+				isSaving   : false,
+				errors     : null
+			}), ()=>this.save());
+		}
```
calculuschild, line 223:
```suggestion
		this.setState((prevState)=>({
			saveGoogle : !prevState.saveGoogle,
			isSaving   : false,
			errors     : null
		}), ()=>{
			if(this.isEdit())
				this.trySave();
		});
```

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -285,10 +379,14 @@ const EditPage = createClass({
 
 			{this.state.confirmGoogleTransfer &&
 				<div className='errorContainer' onClick={this.closeAlerts}>
-					{ this.state.saveGoogle
+					{ this.isNew() && (this.state.saveGoogle
+						?	`Use Homebrewery storage when you save this brew?`
+						: `Use Google Drive storage when you save this brew?`
```
calculuschild, line 384:
```suggestion
						? `Use Homebrewery storage when you save this brew?`
						: `Use Google Drive storage when you save this brew?`
```

client/homebrew/pages/editPage/editPage.jsx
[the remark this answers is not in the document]
G-Ambatte: Yes, absolutely; what I meant is that because `save()` is called by clicking the SAVE button, it causes the Brew to be saved to the appropriate HB/Google storage and redirects to the `/edit` page, whereas this automatic saving from `/new` only saves to localStorage. So if we put it in the `save()`, we'll need a way to identify that this should be saved to localStorage, not MongoDB or Google Drive.

We could also create a new function, e.g. `saveLocal()`, and drop the three lines of code into that.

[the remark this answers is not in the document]
G-Ambatte: I'm now wondering about a `this.saveLocation` variable with possible values of `Homebrewery`, `GoogleDrive`, `BrowserCache`, which direct the `save()` logic. It would be more extensible than just using the single `this.saveGoogle` boolean.

Probably outside the scope of this PR, though. Maybe in a refactor PR.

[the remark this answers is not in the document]
calculuschild: Ah I understand now. Yes that sounds like a good approach. I don't mind if you want to add that in so you can move the local storage saving to the save function.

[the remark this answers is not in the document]
calculuschild: Although you are right it might not be that straightforward.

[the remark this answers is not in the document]
calculuschild: Maybe if saving comes from TrySave, pass a "autosave" parameter to the Save function and then we can handle it in there depending if it's true or false.

calculuschild:
@G-Ambatte What stage is this at now? I know this is a pretty huge task, but unfortunately that makes it easy to fall out of date with the current master branch. Would it be better to break this up into a series of smaller PRs focused on refactoring just one section or function at a time? I think it might be easier to manage (and for me to review).

For instance, take the saving function (or something smaller) that we need to kind of meld two versions together. Update that function on both /new and /edit with the only difference being a flag at the top (`if page == "edit"`) that determines how the function runs. Start copying over bits from /new that need to be in the combined version over to /edit, and vice-versa, even if they aren't going to be used on those pages right now. And slowly make the two pages closer and closer to identical until we can just delete one.

G-Ambatte:
Superceded by #1799.

calculuschild:
Superseded by #1801

PR #1453, opened 2021-07-25
Prepend 'CLONE:- ' to title of cloned brew

G-Ambatte:
Prepend 'CLONE:-' to the title of the brew when cloning it via `/new/:id` so as to disambiguate it from the original.

### DIFF
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -26,6 +26,7 @@ const NewPage = createClass({
 		return {
 			brew : {
 				text      : '',
+				style     : undefined,
 				shareId   : null,
 				editId    : null,
 				createdAt : null,
@@ -46,6 +47,7 @@ const NewPage = createClass({
 		return {
 			brew : {
 				text        : this.props.brew.text || '',
+				style       : this.props.brew.style || undefined,
 				gDrive      : false,
 				title       : this.props.brew.title || '',
 				description : this.props.brew.description || '',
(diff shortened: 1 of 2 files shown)


calculuschild:
I don't think the live version even copies the title at all. Just testing the live site, the title of a cloned brew is always blank from what I can see, so this PR doesn't have any effect yet (the title is still blank.)

G-Ambatte:
image

I'm seeing the metadata (title/description/systems/etc) being cloned across on the live site?  
https://homebrewery.naturalcrit.com/new/2RfZXoJSJRXE

calculuschild:
How strange. It doesn't work in this case here: https://homebrewery.naturalcrit.com/new/WTYbPj4C_vIO

G-Ambatte:
Really weird - because the window title was set to "Why title no copy?", it flashed up in the normal title location, THEN a fraction of a second later it disappeared - the metadata is being removed for some reason???

calculuschild:
I think there might be some conflict with the localStorage wanting to overwrite it? I'm not in a good position to test it at the moment but that's the only thing that comes to mind for now.

G-Ambatte:
Oh! I wonder if it's related to the Style data issue? The one I linked has Style data that doesn't get cloned, whereas your one doesn't have Style data to clone... I'll experiment with that when I have more time.

G-Ambatte:
That does seem to be related somehow - I cloned an old brew with no Style data and the Metadata (including the title) was removed in the clone; I then copied the data between the `<style>` tags to the Style tab, cloned it again, and now the Metadata remains while the Style data is absent.

How curious. I'll dig further when I have a little more time.

G-Ambatte:
I think I got it... The when Style wasn't set, it triggered the code to use `setState` to set the state of `brew.text` and `brew.style`; this call was almost immediately before the `setState` to set `htmlErrors`. Combining the two `setState`s into a single call seems to have fixed the issue on my local install; I'll push a commit with just that change shortly.

PR #1454, opened 2021-07-25
Sanitize Style tab data by parsing it with Markdown

G-Ambatte:
Initial code pass at Style data sanitization via Markdown parsing.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -130,6 +130,7 @@ const BrewRenderer = createClass({
 	},
 
 	renderStyle : function(){
+		if(!this.props.brew.style) return;
 		if(this.props.renderer == 'legacy')
 			return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(`<style> ${this.props.style} </style>`) }} />;
 		else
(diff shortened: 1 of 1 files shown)


client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -129,6 +129,14 @@ const BrewRenderer = createClass({
 			return <div className='phb3 page' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: Markdown.render(pageText) }} key={index} />;
 	},
 
+	renderStyle : function(){
+		if(!this.props.brew.style) return;
+		if(this.props.renderer == 'legacy')
+			return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(`<style> ${this.props.style} </style>`) }} />;
+		else
+			return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: Markdown.render(`<style> ${this.props.style} </style>`) }} />;
```
calculuschild, line 137: We can just render all the style panels using `Markdown.render` since being in Legacy or V3 mode shouldn't affect the way the Style panel is processed.

calculuschild:
I'm not sure this solves the root problem as we've been discussing on the Gitter. Instead we probably need to use a separate sanitizer script. From Marked.js : "Instead use a sanitize library, like DOMPurify (recommended), sanitize-html or insane on the output HTML!"

G-Ambatte:
I believe you are correct - none of these changes have been changing the result of the test cases, as we've seen in the Gitter chat.

calculuschild:
Probably the first thing we need to solve is getting the title to be cloned at all.

G-Ambatte:
Closing this PR as Markdown is not the correct way to sanitize the CSS.

PR #1455, opened 2021-07-25
Make WelcomeText and ChangeLogText conditional

G-Ambatte:
Initial pass at changes to `server.js` so as to only include `welcomeText` or `changelogText` on the required pages.

### DIFF
--- a/server.js
+++ b/server.js
@@ -108,8 +108,7 @@ app.get('/robots.txt', (req, res)=>{
 //Home page
 app.get('/', async (req, res, next)=>{
 	const brew = {
-		text        : welcomeText,
-		welcomeText : welcomeText
+		text : welcomeText
 	};
 	req.brew = brew;
 	return next();
--- a/server.js
+++ b/server.js
@@ -105,6 +105,18 @@ app.get('/robots.txt', (req, res)=>{
 	return res.sendFile(`${__dirname}/robots.txt`);
 });
 
+//Home page
+app.get('/', async (req, res, next)=>{
+	req.welcomeText = welcomeText;
+	return next();
+});
+
+//Changelog page
+app.get('/changelog', async (req, res, next)=>{
+	req.changelogText = changelogText;
+	return next();
+});
+
 //Source page
 app.get('/source/:id', asyncHandler(async (req, res)=>{
 	const brew = await getBrewFromId(req.params.id, 'raw');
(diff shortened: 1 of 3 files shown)


server.js
```diff
@@ -105,6 +105,18 @@ app.get('/robots.txt', (req, res)=>{
...
+	req.welcomeText = welcomeText;
+	return next();
+});
+
+//Changelog page
+app.get('/changelog', async (req, res, next)=>{
+	req.changelogText = changelogText;
+	return next();
+});
+
```
calculuschild, line 119: Let's see... In `Homebrew.jsx` we just use changelogText to create a barebones `brew` object that gets passed into a share page. How about we just create that brew object here instead, and clean up the number of "special-case" props below?

calculuschild: > Sounds remarkably sensible, something like:
>
> ```
> app.get('/changelog', async (req, res, next)=>{
>	req.text = changelogText;
>	return next();
> });
> ```

I would just use the minimal brew you already made. No need for a separate `req.text`.  By making a `brew` we no longer use the `changeLogText` prop at all and we can just remove it.

calculuschild: And yes, we can just pass in `this.props.brew.text` to the Homepage route in Homebrew.jsx since Homepage is weird and doesn't accept a full brew (to be fixed later as you say...).

server.js
```diff
@@ -105,6 +105,22 @@ app.get('/robots.txt', (req, res)=>{
 	return res.sendFile(`${__dirname}/robots.txt`);
 });
 
+//Home page
+app.get('/', async (req, res, next)=>{
+	req.welcomeText = welcomeText;
+	return next();
```
calculuschild, line 111: Can we make a minimal brew here too?

```suggestion
	const brew = {
		title : 'WelcomeText',
		text  : welcomeText
	};
	req.brew = brew;
	return next();
```

And then in Homebrew.jsx have `welcomeText={this.props.brew.text}`

server.js
```diff
@@ -204,8 +220,8 @@ app.use((req, res)=>{
 	const props = {
 		version     : require('./package.json').version,
 		url         : req.originalUrl,
-		welcomeText : welcomeText,
-		changelog   : changelogText,
+		welcomeText : req.welcomeText,
+		changelog   : req.changelogText,
```
calculuschild, line 224: Then if everything is contained in the brews both of these lines can just be deleted.

calculuschild: And in Homebrew.jsx we can delete them from the initialProps.

server.js
G-Ambatte: This is slightly different because HomePage compares welcomeText to brew.text to detect if any changes have been made and thus Save should be available. But we should be able to do that via a brew.

G-Ambatte: I have now pushed `welcomeText` into the `brew` object on the HomePage, and changed the check from `this.props.welcomeText` to `this.props.brew.welcomeText`.

server.js
G-Ambatte: Deleted in the latest commit; confirmed that both the home page and change log page still work correctly.

PR #1456, opened 2021-07-25
Clean up template.js

G-Ambatte:
As per comments from Eric Scheid in Gitter chat: 

- [x] `<link href=/homebrew/bundle.css rel='stylesheet'></link>` — don't need that `</link>`
- [x] `</body>↩<script src=/homebrew/bundle.js></script>↩<script>start_app(...` — both of those two script elements should be before the `</body>`

This PR corrects these issues in the `template.js` file.


calculuschild:
Looks like it works. Thanks!

PR #1457, opened 2021-07-25
Fix issue with undefined Style tab data

G-Ambatte:
Modify the BrewRenderer to only insert the Style data between `<style>` tags if such data exists. Uses the pre-existing `renderPages` functionality as a template.


client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -189,7 +194,9 @@ const BrewRenderer = createClass({
 
 						<div className='pages' ref='pages'>
 							{/* Apply CSS from Style tab */}
-							<div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: `<style> ${this.props.style} </style>` }} />
+							{this.state.isMounted
```
calculuschild, line 197: Can we combine this check with the identical one below? Something like this? (my syntax might be slightly off)

```
{this.state.isMounted &&
  <>
    {this.renderStyle()} 
    {this.renderPages()}
  </>
}

G-Ambatte: Checks are now combined; IMO your logic here is better than what was in the code, so I've changed it.

PR #1459, opened 2021-07-26
Add options for ordering brews on User page

G-Ambatte:
This PR is an initial pass at adding ordering controls to the `/user` page.

### DIFF
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -53,7 +53,7 @@ const UserPage = createClass({
 
 	sortBrews : function(brews, sortType){
 		if(sortType == 'alpha') {
-			return _.orderBy(brews, (brew)=>{ return brew.title; }, this.state.sortDir);
+			return _.orderBy(brews, (brew)=>{ return brew.title.toLowerCase(); }, this.state.sortDir);
 		}
 		if(sortType == 'created'){
 			return _.orderBy(brews, (brew)=>{ return brew.createdAt; }, this.state.sortDir);
@@ -67,7 +67,7 @@ const UserPage = createClass({
 		if(sortType == 'latest'){
 			return _.orderBy(brews, (brew)=>{ return brew.lastViewed; }, this.state.sortDir);
 		}
-		return _.orderBy(brews, (brew)=>{ return brew.title; }, this.state.sortDir);
+		return _.orderBy(brews, (brew)=>{ return brew.title.toLowerCase(); }, this.state.sortDir);
 	},
 
 	handleSortOptionChange : function(event){
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -162,7 +162,7 @@ const UserPage = createClass({
 						<button
 							onClick={this.handleSortDirChange}
 						>
-							{`${(this.state.sortDir == 'asc' ? '⮝ ASC' : '⮟ DESC')}`}
+							{`${(this.state.sortDir == 'asc' ? '\u2B9D ASC' : '\u2B9F DESC')}`}
 						</button>
 					</td>
 				</tr>
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -53,7 +53,7 @@ const UserPage = createClass({
 
 	sortBrews : function(brews, sortType){
 		if(sortType == 'alpha') {
-			return _.orderBy(brews, (brew)=>{ return brew.title.toLowerCase(); }, this.state.sortDir);
+			return _.orderBy(brews, (brew)=>{ return _.deburr(brew.title.toLowerCase()); }, this.state.sortDir);
 		}
 		if(sortType == 'created'){
 			return _.orderBy(brews, (brew)=>{ return brew.createdAt; }, this.state.sortDir);
@@ -67,7 +67,7 @@ const UserPage = createClass({
 		if(sortType == 'latest'){
 			return _.orderBy(brews, (brew)=>{ return brew.lastViewed; }, this.state.sortDir);
 		}
-		return _.orderBy(brews, (brew)=>{ return brew.title.toLowerCase(); }, this.state.sortDir);
+		return _.orderBy(brews, (brew)=>{ return _.deburr(brew.title.toLowerCase()); }, this.state.sortDir);
 	},
 
 	handleSortOptionChange : function(event){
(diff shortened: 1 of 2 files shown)


client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -36,13 +44,132 @@ const UserPage = createClass({
...
+		  		</label>
+				  </td>
+				  <td>
+						<h6>Direction :</h6>
+					</td>
+					<td>
+						<button
+							onClick={this.handleSortDirChange}
+						>
+							{`${(this.state.sortDir == 'asc' ? '⮝ ASC' : '⮟ DESC')}`}
```
ericscheid, line 165: Are those unicode characters there? Can we write them with `\uXXXX` format?
image

G-Ambatte: Should be done now.

G-Ambatte:
We should be able to generate a page count at `brew.save` by a `brew.text.split('\\page').length`... Unfortunately the data that is currently passed to the User page has all of the `brew.text` data removed, so we can't just do it in the page.

I've had a *very* quick look at expanding the Sort to include a Filter option, which might also act as a limited search (e.g. type "adv" into the title filter; the User page lists brews titled "Advanced D&D Homebrews", "Adventure Module 1", "Template: Adventure Module").
Although this might constitute a large enough deviation from simply re-ordering the listed brews that it should have it's own PR.

calculuschild:
We can get this basic functionality out first and the add more bonus filters later on as we need.

G-Ambatte:
I've marked this as Ready for Review. Filters and what not can be a later PR to expand this functionality.

With that in mind, should the Sort options be moved to the NavBar, or left in their own little window like they currently are?

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -24,7 +24,15 @@ const UserPage = createClass({
 	getDefaultProps : function() {
 		return {
 			username : '',
-			brews    : []
+			brews    : [],
+			sortType : 'alpha',
+			sortDir  : 'desc'
```
calculuschild, line 29: Props represent any values that have been (or are expected to have been) "passed in" from a container object higher up (or from `Server.js`). At the moment, this page seems like it doesn't get passed those values in any way so we can just remove them here. Just tracking it in State is good enough.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -24,7 +24,15 @@ const UserPage = createClass({
...
+		};
+	},
+	getInitialState : function() {
+		return {
+			sortType : 'alpha',
+			sortDir  : 'desc'
```
calculuschild, line 35: The only thing here is I was surprised the default sorting was reverse-alphabetical (brews starting with Z were at the top). Can we make it "alpha", "asc" by default?

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -36,13 +44,88 @@ const UserPage = createClass({
...
+			return _.orderBy(brews, (brew)=>{ return brew.updatedAt; }, this.state.sortDir);
+		}
+		if(sortType == 'views'){
+			return _.orderBy(brews, (brew)=>{ return brew.views; }, this.state.sortDir);
+		}
+		if(sortType == 'latest'){
+			return _.orderBy(brews, (brew)=>{ return brew.lastViewed; }, this.state.sortDir);
+		}
+		return _.orderBy(brews, (brew)=>{ return _.deburr(brew.title.toLowerCase()); }, this.state.sortDir);
+	},
```
calculuschild, line 71: Can we make this like a switch statement or something?

Also, if you want the "fallback" sorting to just be alphabetical, we don't need a separate "if-then" check for alphabetical mode. Just let it fall through to the default sorting mode if `sortType == "alpha"`

client/homebrew/pages/userPage/userPage.jsx
G-Ambatte: Actually I made that the default to see if the `_.orderBy` function worked as I expected before I had created any sort of interface. So yes, it can absolutely be changed to default to 'asc'.  
Although I was thinking about whether it would be beneficial to store and load these values from localStorage, so that the UserPage remembers your last sort parameters, rather than resetting to Alphabetical every time.

G-Ambatte: That functionality is probably better as a separate PR, though.

calculuschild: I agree.

client/homebrew/pages/userPage/userPage.jsx
G-Ambatte: I'll pull those out in the next commit.

client/homebrew/pages/userPage/userPage.jsx
G-Ambatte: There's a lot of DRY violation going on here. I feel like there should be a better way; I'll look into it further tonight.

Something sort of like this:

```
brewSortingData : function(brew){
    switch(sortType) {
        case 'alpha':
            return _.deburr(brew.title.toLowerCase());
        case [....]
    };
},

sortBrew : function(brews, sortType){
    return _.orderBy(brews, (brew)=>{ brewSortingData(brew) }, this.state.sortDir);
}
```

Something like that, so all the stuff that changes from line to line in the current `sortBrews` function takes place in a separate sub-function `brewSortingData` or some such name, and thus the function is reduced to effectively a single line containing the `_.orderBy`.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -36,13 +44,88 @@ const UserPage = createClass({
...
+		this.setState({
+			sortType : event.target.value
+		});
+	},
+
+	handleSortDirChange : function(event){
+		const newDir = (this.state.sortDir == 'asc' ? 'desc' : 'asc');
+		this.setState({
+			sortDir : `${newDir}`
+		});
```
calculuschild, line 83:
```suggestion
		this.setState((prevState)=>{
			sortDir : prevState.sortDir == 'asc' ? 'desc' : 'asc'
		});
```

client/homebrew/pages/userPage/userPage.less
```diff
@@ -30,4 +30,31 @@
 
 		}
 	}
+	.sort-container{
```
calculuschild, line 33: Just minor comments on the styling:

* Vertical-align the text with the radio buttons if possible.
* "Open Sans" doesn't seem to be applying to the label items. See if you can get that to stick.
* Remove the border on the "dir" button and try to get the padding even all around its text.
* Change the arrow to something like U+1F83B or U+25BC to fit the blockier style of our other menu icons.
* "SORT TYPE:" to "SORT BY"

PR #1468, opened 2021-07-28
Fix issue 1467 - Source on Change Log doesn't work

G-Ambatte:
This draft PR is an initial pass at code to hide the Source and Get PDF buttons from the NavBar on the `/changelog` page, as they do not function there.

### DIFF
--- a/client/homebrew/pages/sharePage/sharePage.jsx
+++ b/client/homebrew/pages/sharePage/sharePage.jsx
@@ -79,21 +79,6 @@ const SharePage = createClass({
 		</div>;
 	},
 
-	renderPrint : function(){
-		if(!this.props.brew.shareId) return;
-		return <PrintLink shareId={this.processShareId()} />;
-	},
-
-	renderSourceDropdown : function(){
-		if(!this.props.brew.shareId) return;
-		return <Nav.item icon='fas fa-code' color='red' className='source'
-						onMouseEnter={()=>this.handleDropdown(true)}
-						onMouseLeave={()=>this.handleDropdown(false)}>
-						source
-						{this.renderDropdown()}
-					</Nav.item>;
-	},
-
 	render : function(){
 		return <div className='sharePage sitePage'>
 			<Meta name='robots' content='noindex, nofollow' />
@@ -103,8 +88,13 @@ const SharePage = createClass({
 				</Nav.section>
 
 				<Nav.section>
-					{this.renderPrint()}
-					{this.renderSourceDropdown()}
+					{this.props.shareId && <PrintLink shareId={this.processShareId()} />
+					<Nav.item icon='fas fa-code' color='red' className='source'
+						onMouseEnter={()=>this.handleDropdown(true)}
+						onMouseLeave={()=>this.handleDropdown(false)}>
+						source
+						{this.renderDropdown()}
+					</Nav.item>}
 					<RecentNavItem brew={this.props.brew} storageKey='view' />
 					<Account />
 				</Nav.section>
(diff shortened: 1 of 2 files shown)


calculuschild:
I think I prefer @ericscheid 's approach of just having a check for shareId == undefined.

G-Ambatte:
Can certainly do that; my initial thought process was that it might be useful to be able to enable/disable these things in the SharePage route. However that might also be using a sledgehammer to crack a single walnut.

ericscheid:
> I think I prefer @ericscheid 's approach of just having a check for shareId == undefined.

I don't know react, but this looks promising:

https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -79,6 +79,21 @@ const SharePage = createClass({
...
+
+	renderSourceDropdown : function(){
+		if(!this.props.brew.shareId) return;
+		return <Nav.item icon='fas fa-code' color='red' className='source'
+						onMouseEnter={()=>this.handleDropdown(true)}
+						onMouseLeave={()=>this.handleDropdown(false)}>
+						source
+						{this.renderDropdown()}
+					</Nav.item>;
+	},
```
calculuschild, line 95: Probably doesn't need to even be in separate functions up here anymore. Since they both use the same check, just put the check directly in the render function.

```
      {this.props.brew.shareId  &&
        <PrintLink shareId={this.processShareId()} />
	<Nav.item icon='fas fa-code' color='red' className='source'
		onMouseEnter={()=>this.handleDropdown(true)}
		onMouseLeave={()=>this.handleDropdown(false)}>
		source
		{this.renderDropdown()}
	</Nav.item>
      }
```

client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -103,8 +88,13 @@ const SharePage = createClass({
 				</Nav.section>
 
 				<Nav.section>
-					{this.renderPrint()}
-					{this.renderSourceDropdown()}
+					{this.props.shareId && <PrintLink shareId={this.processShareId()} />
```
calculuschild, line 91: You can put this on two lines for legibility:

```
{this.props.shareId &&
  <PrintLink shareId={this.processShareId()} />
```

client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -103,8 +88,13 @@ const SharePage = createClass({
...
+					<Nav.item icon='fas fa-code' color='red' className='source'
+						onMouseEnter={()=>this.handleDropdown(true)}
+						onMouseLeave={()=>this.handleDropdown(false)}>
+						source
+						{this.renderDropdown()}
+					</Nav.item>}
```
calculuschild, line 97: Similarly, we can take the closing `}` and move it to a new line so it's more legible which components are enclosed by this check.

```
  </Nav.item>
}
```

I think Lint might actually force this anyway.

G-Ambatte: I should be able to run it through the Linter shortly in the next hour or so.

client/homebrew/pages/sharePage/sharePage.jsx
calculuschild: Ah, and since React expects only one "element" at a time when we do this, we have to wrap it in an empty "fragment" `<>` so it can process this correctly:

```
<>
  <PrintLink shareId={this.processShareId()} />
  <Nav.item icon='fas fa-code' color='red' className='source'
	onMouseEnter={()=>this.handleDropdown(true)}
	onMouseLeave={()=>this.handleDropdown(false)}>
	source
	{this.renderDropdown()}
  </Nav.item>
</>
```

G-Ambatte:
This commit is clean, elegant, simple - and as an added bonus, it even works, too.

PR #1482, opened 2021-07-30
v3 Snippets: Class Tables

Gazook89:
[Relevant Share Link](https://homebrewery-stage.herokuapp.com/share/jn7Gokj_w8Hh)

- Converts class tables to v3 moustache syntax.
- adds a `.full` class which spans 2 columns (like `.wide`) but also adds "-- Spell Slots per Spell Level --" header using `:after'
- adds `--row-color="rgba(250,0,0,.50)"` to the snippets and css so that odd row colors are easily changeable.

I only just thought that maybe I'm forgetting to do something so that these snippets continue to work for legacy?  Not sure where it determines which renderer is being used and which snippet to generate?


G-Ambatte:
As I recall, if the renderer is set to `legacy` the snippets are loaded from the `snippetsLegacy` folder; if set to `v3`, they are loaded from the `snippets` folder. So what you're doing is absolutely fine.

calculuschild:
I'm going to put this one on pause for a bit and see if I can't get some cool table markdown working quickly. If so we can modify this PR to work with that instead of relying on `:after` elements.

PR #1483, opened 2021-07-30
v3 Snippets: Page Numbers

Gazook89:
This is a small PR for Page Numbers (Manual & Auto):
- Included the `footnote` element with the Auto Page Number snippet to match the Manual Page Number snippet.  Since many people may just always choose auto, they may not ever see the footnote bit as the snippet is currently setup.
- changed both snippets to be a single line, or `<spans>`.  Reduces line count and is easier to read, IMHO.

(Is this PR too small?  I assume I shouldn't lump a bunch of unrelated snippet changes together but let me know if I should be dreaming bigger).


calculuschild:
Small PRs are fine. Fixing one problem at a time means makes it much easier to review.

Issue #849, opened 2019-06-11
Feature Request: Smart quotes option


sjml:
Would it be possible to use the marked library's smartypants option so that quotes and dashes get rendered into a more pleasing typography?

ericscheid:
https://marked.js.org/using_advanced#options

This sounds simple. I can't imagine what could possibly go wrong. #tweak

Worst case, we uncover a bug in _markedjs_ that they would care about to fix and we can disable in the interim.

calculuschild:
This is now functioning in the v3 renderer on commit https://github.com/naturalcrit/homebrewery/commit/8bd09e58cb41be7f2e3468ac67b9fea9981395f0

I'm hesitant to apply it to Legacy since any slight change in spacing as a result is bound to ruin somebody's brews.

calculuschild:
Added to the V3 changelog finally. Closing.

PR #1490, opened 2021-07-31
Enable Marked SmartyPants option

G-Ambatte:
This PR creates an independent `MarkedOpts.json` file to contain the options for Marked, and sets both `Markdown` and `MarkdownLegacy` to use those options.  
Currently, the only option set in `MarkedOpts.json` is `"smartypants" : true`, but this file can be extended later to set other options for Marked.

This PR addresses issue #849.

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -1,6 +1,6 @@
 /* eslint-disable max-lines */
 const _ = require('lodash');
-const Markdown = require('marked');
+const Markdown = require('marked').setOptions(MarkedOpts);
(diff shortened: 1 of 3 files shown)


calculuschild:
Note we have already applied the smartypants option to the v3 renderer in a previous commit:

https://github.com/naturalcrit/homebrewery/commit/8bd09e58cb41be7f2e3468ac67b9fea9981395f0

I'm hesitant to apply it to Legacy since everything we touch on there makes somebody mad and ruins somebody's spacings and alignments.

calculuschild:
Going to close this since we already apply smartypants directly in v3 without needing a separate config file.

PR #1491, opened 2021-07-31
Display Brew page count on User Page

G-Ambatte:
This PR adds a page count to the Brew Item on the User Page. The page count defaults to '1' but updates when the Brew is saved.

### DIFF
--- a/client/homebrew/pages/userPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/userPage/brewItem/brewItem.jsx
@@ -107,8 +107,6 @@ const BrewItem = createClass({
 			<h2>{brew.title}</h2>
 			<p className='description'>{brew.description}</p>
 			<hr />
-			<div className='spacer'></div>
-
 			<div className='info'>
 				<span title={`Authors:\n${brew.authors.join('\n')}`}>
 					<i className='fas fa-user'/> {brew.authors.join(', ')}
--- a/client/homebrew/pages/userPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/userPage/brewItem/brewItem.jsx
@@ -114,13 +114,16 @@ const BrewItem = createClass({
 				<span title={`Last viewed: ${moment(brew.lastViewed).local().format(dateFormatString)}`}>
 					<i className='fas fa-eye'/> {brew.views}
 				</span>
-				<span
-					title={
-						`Created: ${brew.createdAt ? moment(brew.createdAt).local().format(dateFormatString) : 'UNKNOWN'}\n` +
-						`Last updated: ${brew.updatedAt ? moment(brew.updatedAt).local().format(dateFormatString) : 'UNKNOWN'}`
-					}>
-					<i className='fas fa-sync-alt'	/>
-					{moment(brew.updatedAt).fromNow()}
+				{brew.pageCount &&
+					<span>
+						<i className='far fa-file' /> {brew.pageCount}
+					</span>
+				}
+				<span>
+					<i className='far fa-file' /> {brew.pageCount || 1}
+				</span>
+				<span>
+					<i className='fas fa-sync-alt' /> {moment(brew.updatedAt).fromNow()}
 				</span>
 				{this.renderGoogleDriveIcon()}
 			</div>
--- a/client/homebrew/pages/userPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/userPage/brewItem/brewItem.jsx
@@ -115,13 +115,10 @@ const BrewItem = createClass({
 					<i className='fas fa-eye'/> {brew.views}
 				</span>
 				{brew.pageCount &&
-					<span>
+					<span title={`Page count: ${brew.pageCount}`}>
 						<i className='far fa-file' /> {brew.pageCount}
 					</span>
 				}
-				<span>
-					<i className='far fa-file' /> {brew.pageCount || 1}
-				</span>
 				<span>
 					<i className='fas fa-sync-alt' /> {moment(brew.updatedAt).fromNow()}
 				</span>
--- a/client/homebrew/pages/userPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/userPage/brewItem/brewItem.jsx
@@ -104,13 +104,12 @@ const BrewItem = createClass({
 		const dateFormatString = 'YYYY-MM-DD HH:mm:ss';
 
 		return <div className='brewItem'>
-			<h2>{brew.title}</h2>
-			<p className='description'>{brew.description}</p>
+			<div className='text'>
+				<h2>{brew.title}</h2>
+				<p className='description'>{brew.description}</p>
+			</div>
 			<hr />
 			<div className='info'>
-				<span title={`Authors:\n${brew.authors.join('\n')}`}>
-					<i className='fas fa-user'/> {brew.authors.join(', ')}
-				</span>
 				<span title={`Last viewed: ${moment(brew.lastViewed).local().format(dateFormatString)}`}>
 					<i className='fas fa-eye'/> {brew.views}
 				</span>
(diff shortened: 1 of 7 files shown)


G-Ambatte:
Currently only working for locally saved brews; Google brews still to be implemented.

client/homebrew/pages/userPage/brewItem/brewItem.jsx
```diff
@@ -113,6 +114,9 @@ const BrewItem = createClass({
 				<span>
 					<i className='fas fa-eye' /> {brew.views}
 				</span>
+				<span>
```
ericscheid, line 117: could this be made to not display at all if `!brew.pageCount` ?

client/homebrew/pages/userPage/brewItem/brewItem.jsx
```diff
@@ -113,6 +114,9 @@ const BrewItem = createClass({
...
+					<i className='far fa-file' /> {brew.pageCount || 1}
+				</span>
```
ericscheid, line 119: could this be made to not display at all if `!brew.pageCount` ?

calculuschild:
What's the status of this one @G-Ambatte ? Are you still planning to add to this or does it need to simmer for a while? I'd like to see if we can cut down the number of "open but not finished" PRs.

G-Ambatte:
> What's the status of this one @G-Ambatte ? Are you still planning to add to this or does it need to simmer for a while?

I need to add support for Google Brews, which I expect will be by pushing a pageCount properties into the file.properties fields.
I'll have a look at it tonight, as Google Brews are currently working again on my local install.

G-Ambatte:
I got Google Brews fully working, and then rebased the branch so it should be up to date and not conflict with the current `master` branch.

client/homebrew/pages/userPage/brewItem/brewItem.jsx
G-Ambatte: I've modified it so that it will only display if `pageCount>1`.

server/googleActions.js
```diff
@@ -154,13 +155,15 @@ GoogleActions = {
...
-																	   tags       : brew.tags,
-																	   systems    : brew.systems.join() }
+										 							    published  : brew.published,
+																	    lastViewed : brew.lastViewed,
+																	    views      : brew.views,
+																	    version    : brew.version,
+																	    renderer   : brew.renderer,
+																	    tags       : brew.tags,
+																	    systems    : brew.systems.join(),
+																	    pageCount  : (brew.text.match(/\\page/g) || []).length + 1
```
calculuschild, line 165:
```suggestion
																	    pageCount  : (brew.text.match(/^\\page/gm) || []).length + 1
```

We've actually got two slightly different ways of counting pages, depending on the renderer. Legacy splits pages *anywhere* there is a `\page`. V3 only splits if the line starts with `\page`.

We should use the V3 version preferably (faster Regex since it knows to only start at the beginning of a line instead of scanning the *entire text*) *or* make sure to handle both cases specifically.

*or*, pass the page count as a `brew` property into the save function instead of trying to calculate it here. Might be cleaner and be a bit more DRY.

**Edit:** yeah, after reading the rest of this, let's do that: pass in pageCount as a brew property.

server/googleActions.js
```diff
@@ -191,11 +194,12 @@ GoogleActions = {
...
-				'shareId'  : nanoid(12),
-				'editId'   : nanoid(12),
-				'title'    : brew.title,
-				'views'    : '0',
-				'renderer' : brew.renderer || 'legacy'
+				'shareId'   : nanoid(12),
+				'editId'    : nanoid(12),
+				'title'     : brew.title,
+				'views'     : '0',
+				'pageCount' : (brew.text.match(/\\page/g) || []).length + 1,
```
calculuschild, line 201: Same comments as above. Seeing this appear in so may places makes me think that we should indeed have it calculated ahead of time so we don't have to do that here.

server/googleActions.js
```diff
@@ -301,6 +306,7 @@ GoogleActions = {
 				createdAt  : obj.data.createdTime,
 				updatedAt  : obj.data.modifiedTime,
 				lastViewed : obj.data.properties.lastViewed,
+				pageCount  : obj.data.properties?.pageCount || 1,
```
calculuschild, line 309:
```suggestion
				pageCount : obj.data.properties.pageCount,
```
Same comments as above.

server/googleActions.js
```diff
@@ -230,6 +234,7 @@ GoogleActions = {
 			updatedAt : new Date(),
 			gDrive    : true,
 			googleId  : obj.data.id,
+			pageCount : fileMetadata.properties.pageCount || 1,
```
calculuschild, line 237:
```suggestion
			pageCount : fileMetadata.properties.pageCount,
```
Same comments as above

server/homebrew.api.js
```diff
@@ -38,6 +38,7 @@ const newBrew = (req, res)=>{
 
 	brew.authors = (req.account) ? [req.account.username] : [];
 	brew.text = mergeBrewText(brew.text, brew.style);
+	brew.pageCount = (brew.text.match(/\\page/g) || []).length + 1;
```
calculuschild, line 41: Same comments as above. This should be passed in as a brew property.

server/homebrew.api.js
```diff
@@ -66,6 +67,7 @@ const updateBrew = (req, res)=>{
 		.then((brew)=>{
 			brew = _.merge(brew, req.body);
 			brew.text = mergeBrewText(brew.text, brew.style);
+			brew.pageCount = (brew.text.match(/\\page/g) || []).length + 1;
```
calculuschild, line 70: Same comments as above. This should be passed in as a brew property.

calculuschild (changes requested):
Mostly ready. I have some suggestions for cleanup.

ericscheid (approved):
LGTM with calculuschild's amends

server/googleActions.js
G-Ambatte: Committed.

server/googleActions.js
G-Ambatte: Committed.

client/homebrew/pages/userPage/brewItem/brewItem.jsx
```diff
@@ -120,6 +121,11 @@ const BrewItem = createClass({
 				<span>
 					<i className='fas fa-eye' /> {brew.views}
 				</span>
+				{(brew?.pageCount>1) &&
```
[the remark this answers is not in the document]
G-Ambatte: > I think it's ok if we display this for brews with just 1 page

That was actually my initial implementation; only displaying the page count when it was higher than 1 was a suggestion from Gitter - @ericscheid, I think.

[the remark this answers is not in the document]
calculuschild: His suggestion was just to hide it if it doesn't exist. I think we are all on the same page.  https://github.com/naturalcrit/homebrewery/pull/1491#discussion_r680360876

client/homebrew/pages/userPage/brewItem/brewItem.jsx
G-Ambatte: As per request from @calculuschild, this is now set to display if `brew.pageCount` exists, even if it is only a single page.

client/homebrew/pages/userPage/brewItem/brewItem.less
```diff
@@ -69,4 +69,7 @@
 		padding : 0px;
 		margin  : -5px;
 	}
+	.spacer {
+		height : 25px;
+	}
```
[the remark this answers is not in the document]
G-Ambatte: Eliminated the spacer `div`.

G-Ambatte:
I've applied the requested changes, rebased to the current master, then switched PR base branch to `PRODUCTION` and back to `master`, so it *should* be ready for review again.

Issue #1303, opened 2021-04-01
Add a Comment Selection editor action


ericscheid:
Suggestion: add an option to the Editor menu to wrap any selection with `<!--` and `-->` (with line breaks if selection includes one or more line breaks).

This is useful for temporarily removing brew content from the render, and for adding editorial comments for work-in-progress.

Possibly not worth the trouple of getting the code to detect if the selection is inside a `<style />` block (and thus using `/*` and `*/` instead)

Gazook89:
Was going to close this issue as PR #1492 was merged, adding an html comment snippet and hotkeys for converting existing selected text....but now that the Styles are separated into their own pane, i don't think it should be too hard to detect which editor frame is open and adjust the comment style accordingly?  For both the hotkeys and the snippet.

### Hotkey code:
[codeEditor.jsx](https://github.com/naturalcrit/homebrewery/blob/53de59940f11251e7f7093dc69ee4d5ceef16022/shared/naturalcrit/codeEditor/codeEditor.jsx#L93-L100)
```
	makeComment : function() {
		const selection = this.codeMirror.getSelection(), t = selection.slice(0, 4) === '<!--' && selection.slice(-3) === '-->';
		this.codeMirror.replaceSelection(t ? selection.slice(4, -3) : `<!-- ${selection} -->`, 'around');
		if(selection.length === 0){
			const cursor = this.codeMirror.getCursor();
			this.codeMirror.setCursor({ line: cursor.line, ch: cursor.ch - 4 });
		}
	},
```

### Snippet Code
[snippets.js](https://github.com/naturalcrit/homebrewery/blob/53de59940f11251e7f7093dc69ee4d5ceef16022/client/homebrew/editor/snippetbar/snippets/snippets.js#L116-L122)  (v3)
```
			{
				name : 'Add Comment',
				icon : 'fas fa-code',  /* might need to be fa-solid fa-comment-code --not sure, Gazook */
				gen  : dedent`\n
					<!-- This is a comment that will not be rendered into your brew. Hotkey (Ctrl/Cmd + /). -->
					`
			},
```

Now I say "should" be easy...but I created a branch to do just this (only the hotkeys) and couldn't figure it out, but I'm hoping someone who isn't like a week into JS would have an easier time.  I may give the Snippet side of things a shot tonight.

calculuschild:
If you want, open your branch for the split snippets, and then make a new branch based on that one. Then you don't have to wait for me to merge your splitting pr first

[...]

PR #1492, opened 2021-07-31
Add Hotkeys for HTML Commenting Selection

Gazook89:
This addresses #1303 for both macOS/not macOS.  Surrounds selection in `<!-- -->`, and does not check if it needs to be `/*  */` for css.....however, if someone knows how to do that, please do so.  I suspect it is easier to determine if the selection is in the Style Editor vs Brew Editor, and that might be "good enough"?

Please double check my code, though...it was mostly a copy/paste of other hotkeys.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippets/snippets.js
+++ b/client/homebrew/editor/snippetbar/snippets/snippets.js
@@ -113,6 +113,13 @@ module.exports = [
 						'  }\n' +
 						'</style>'
 			},
+			{
+				name : 'Add Comment',
+				icon : 'fas fa-comment-code',  /* might need to be fa-solid fa-comment-code --not sure, Gazook */
+				gen  : dedent`\n
+					<!-- Add a comment to your code here without adding it to your live brew. Hotkey Ctrl/Cmd K -->
+					\n`
+			},
 		]
 	},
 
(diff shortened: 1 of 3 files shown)


calculuschild:
If we add a hotkey for this we should also make it a snippet. 

What about changing it to CTRL + backslash? I think I see that as a shortcut more often than CTRL+ K.

calculuschild:
I'd prefer this hotkey only creat HTML comments when in the brew tab,, since we should be encouraging users to put all CSS into the Style tab. We can have it make CSS comments when in the style tab.

Gazook89:
I did a quick Google search of commonly used hotkeys for commenting, and saw ctrl k more often (but did also see the slash as well). I think it depends on language and IDE, so nothing universal....and it makes no difference to me except maybe K is easier to remember..../ and \ can be mixed up easily. 

Detecting which editor is in use is beyond my skillz, so will leave that to someone else.

calculuschild:
Do you think you could make a snippet for this? It can be in both v3 and legacy.

Gazook89:
I'm working on snippets for this, but I'm going to reverse course and agree with you on the backslash `\` rather than `k` for the hotkey.   And reserve `K` for a hyperlink shortcut, as `ctrl`/`cmd` `K` is much more commonly for adding a link.  Don't know why that is a standard, but not going to fight it.

Gazook89:
Added snippets for both legacy and v3, including some explanatory text and provided hotkey info as well.  Can be shortened to just "comment" if that is easier and good enough.

Also, changed comment hotkey to backslash.

Took a few tries to figure out what CircleCI is, but I got it now and found where the rules are in the repo.  Learning!

calculuschild:
If you have it running on your local machine you can run the script `npm run-script lint:dry`, and it will automatically scan your files and tell you what needs to be fixed to match our style. Or just `npm run-script lint` and it will *usually* be able to auto-fix them, and then tell you what it couldn't

calculuschild:
Superseded by #1529

PR #1509, opened 2021-08-03
Add Artist Credit Snippet

Gazook89:
Idea here is to encourage crediting artists, including with a link to the artists portfolio, by making a simple snippet to do so.  The snippet includes an example with an Image Name, and Artist Name.   I wanted to somewhat match the text occasionally found in the PHB (example here is from pg 13): 

image

Here is the result:

image

The snippet itself is created with a `top` and `left` property included and the stylesheet has it absolutely positioned, so hopefully it's easy to position it.  

I created snippets for both legacy and v3, and added a font for each...I'm not 100% married to the font, and the job can be done with just ScalySansSmallCapsRemake if you want to reduce font usage.  Font is Walter Turncoat found on Google Fonts with a permissive license.

~~One issue I am having on my local instance is the stylesheet doesn't seem to pull through...?  I have access to the new snippet, but the new style doesn't come through (but the original version of the stylesheet comes through).  So I'm hoping that's just an issue on my end.~~  Edit:  It seems to have gone through now.  But it is now evident that i borked the addition of the new font.


Gazook89:
Found my missing `@` in the font-faces, so font comes through now and all is well.  Here is the snippet:

```
<div class='artist' style='top:20px;left:20px;'>
##### Starry Night
[Van Gogh](https://www.vangoghmuseum.nl/en)
</div>
```

calculuschild:
This looks just about ready. What do you think about adding the backgroundSnippet text onto this so it appears with an example image tied to it? Or vice-versa? Just so people can clearly see that this goes with an image and how it might be positioned.

PR #1513, opened 2021-08-04
Confirm `Brew.Title` exists before ordering by it

### LINKED ISSUE #1512: Reports of "Internal Server Error" on User Page
Reports on Reddit [here](https://redd.it/ox6px7)

```
TypeError: Cannot read property 'toLowerCase' of undefined
2021-08-03T23:04:12.946324+00:00 app[web.1]:     at Object.sortBrewOrder (/app/build/homebrew/ssr.js:3657:36)
```

Appears to be an issue in the new Brew sorting logic.

G-Ambatte:
This PR should resolve #1512.

The ordering logic checks that `brews` exists, then derives `brew` from it. However it does not confirm that the derived `brew` object contains a `title` element before attempting to order the `brews` set by that element.
This PR sets a default value of 'No Title' to any derived `brew` that fails the `(!brew.title)` test.

### DIFF
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -42,7 +42,7 @@ const UserPage = createClass({
 	renderBrews : function(brews){
 		if(!brews || !brews.length) return <div className='noBrews'>No Brews.</div>;
 
-		const sortedBrews = this.sortBrews(brews, this.state.sortType);
+		const sortedBrews = this.sortBrews(brews);
 
 		return _.map(sortedBrews, (brew, idx)=>{
 			return <BrewItem brew={brew} key={idx}/>;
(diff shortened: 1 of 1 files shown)


client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -42,14 +42,15 @@ const UserPage = createClass({
...
 
 		return _.map(sortedBrews, (brew, idx)=>{
 			return <BrewItem brew={brew} key={idx}/>;
 		});
 	},
 
 	sortBrewOrder : function(brew){
+		if(!brew.title){brew.title = 'No Title';};
 		const mapping = {
 			'alpha'   : _.deburr(brew.title.toLowerCase()),
```
calculuschild, line 55:
```suggestion
		const mapping = {
			'alpha'   : _.deburr(brew.title?.toLowerCase() || "No Title"),
```

How about this?

G-Ambatte: Unfortunately that change results in the following Brew entry in the User Page:

image

Intention is to remain with the original code at this time.

PR #1523, opened 2021-08-07
Add tooltip and classNames to UserPage and BrewItem

### LINKED ISSUE #1511: microdata on user page?
Could we get the actual value of `brew.updatedAt` expressed into the html of the page as an html attribute?

Something like this:
```
<span class="updatedAt" title="8:14pm on July 12th, 2018" datetime="2018-07-12T20:14:13Z">
  ...
</span>
```

This line here, I think:
https://github.com/naturalcrit/homebrewery/blob/6f6c4acf7efc02b32a77291081851c1b0bec0c68/client/homebrew/pages/userPage/brewItem/brewItem.jsx#L116-L118

----

And while we're at it, express the `brew.createdAt` value into the page as another span, but use CSS to hide that by default.

```html
<span class="createdAt" title="8:14pm on July 12th, 2018" datetime="2018-07-12T20:14:13Z">
  ...
</span>
```

```css
.userPage .brewItem .info .createdAt { display: none; }
```

----

No immediate need, but it will enable some creative hacking, and possibly some accessibility improvedments.

The `title=<human-readable-moment>` will enable a tooltip on the html element.

G-Ambatte:
Draft PR to resolve #1511.

### DIFF
--- a/client/homebrew/pages/userPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/userPage/brewItem/brewItem.jsx
@@ -99,11 +99,18 @@ const BrewItem = createClass({
 		</span>;
 	},
 
+	getTooltipData : function(){
+		const dateFormatString = 'YYYY-MM-DD HH:mm:ss';
+		let outputString = `Created: ${this.props.brew.createdAt ? moment(this.props.brew.createdAt).local().format(dateFormatString) : 'UNKNOWN'}\n`;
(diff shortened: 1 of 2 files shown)


calculuschild:
At some point it might be cool to have this same functionality but embedded into the brew item itself. If you hover over the box for a couple seconds (or click a little "expand" button), the box animates and expands out to reveal more info.

But this is good for now. 🎉

PR #1524, opened 2021-08-07
Fix incorrect ordering on UserPage

### LINKED ISSUE #1506: Created and Updated Dates not correct
I mentioned it briefly in gitter, but figured an issue should be opened:  Last Updated and Created Date fields on the user page are not correct.  At least on Google linked brews.  If needed I can find out when some of these were *actually* created by finding my old reddit posts, but some are at least 6 months old, and then others were made this week....with many inbetween which are found elsehwere in the list.  And some of these I haven't opened in a long time.

Edit: and I don't think this is a "new" issue, but likely ties into the Google integration.  I've long noticed it, but didn't care because it was irrelevant information to me...now with sorting, it's actually a "problem".

image

G-Ambatte:
Draft PR to resolve #1506.

### DIFF
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -4,6 +4,8 @@ const createClass = require('create-react-class');
 const _     = require('lodash');
 const cx    = require('classnames');
 
+const moment = require('moment');
+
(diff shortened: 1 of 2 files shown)


calculuschild:
#1506 has another issue still going on so I'm going to keep that open for now.

Issue #1499, opened 2021-08-01
Reloading a cloning will overwrite any changes done


ericscheid:
When cloning a brew the browser is sent to `/new/:id` where changes can be made.

Hitting reload or similar will cause the cloning to re-occur, overwriting any changes made in that `/new` editor.

What should happen is that after the cloning the `/:id` should be stripped out of the browser location.

ericscheid:
This is tricky. Usually, in web architecture, a POST to a generic url is often responded with a HTTP 302 response with a specific url pointing to a newly minted resource. In our case a request for `/new/:id` needs to eventuate with a browser location of `/new` only, but we can't do that by 302 redirecting `/new/:id` to `/new` because the editor localStorage hasn't been set up.

We'll need do finagle the browser location via `history.replaceState(stateObj, title, [url])` on load somewhere.

ericscheid:
If the user truly wants to reclone the original brew, they can go to that original brew and clone it.

G-Ambatte:
One option would be for `/new/:id` to load the contents of the brew to `localStorage`, and then immediately redirect to `/new`. Of course, if the user had anything that was only saved in localStorage, it would be lost...
An alternative is that we were discussing a programmatic option the other day to create a Brew that a user could then save to their account, e.g. `/template?title=MY%20NEW%20BREW&text=data_goes_here`. Cloning could use such a template brew by loading the content of the source brew's title/text/style/whatever to the URL without touching localStorage.

ericscheid:
Forcing a 2nd request to the server seems excessive. But also possibly necessary, as the page itself contains the cloned brew content and reloading that might have the same result as banging `/new/:id`. Will have to test.

G-Ambatte:
> We'll need to finagle the browser location via `history.replaceState(stateObj, title, [url])` onload somewhere.

I randomly thought of this issue this morning, and this is the simplest solution to the problem. We now set the content of local storage on page load, so the issue of whether we should wipe the user's local content is now resolved, as it already happens.

So in  `./client/homebrew/pages/newPage/newPage.jsx` at the end of the `componentDidMount` function, add a line or three that checks the current URL and if NOT `/new`, perform a `history.replaceState()`.

dbolack-ab:
@ericscheid Does this PR solve this issue to your liking?

ericscheid:
> @ericscheid Does this PR solve this issue to your liking?

[...]

PR #1532, opened 2021-08-09
Add Route for external tool access

G-Ambatte:
This PR adds a `/template` route, which will allow external apps to create a Homebrewery document via URL. This is achieved by allowing `text`, `style`, `title`, and `description` to be set via parameters in the URL query.

image

image

### DIFF
--- a/client/homebrew/homebrew.jsx
+++ b/client/homebrew/homebrew.jsx
@@ -45,6 +45,7 @@ const Homebrew = createClass({
 						<Route path='/share/:id' component={(routeProps)=><SharePage id={routeProps.match.params.id} brew={this.props.brew} />}/>
 						<Route path='/new/:id' component={(routeProps)=><NewPage id={routeProps.match.params.id} brew={this.props.brew} />}/>
 						<Route path='/new' exact component={(routeProps)=><NewPage />}/>
+						<Route path='/template' exact component={(routeProps)=><NewPage query={queryString.parse(routeProps.location.search)}/>}/>
 						<Route path='/user/:username' component={(routeProps)=><UserPage username={routeProps.match.params.username} brews={this.props.brews} />}/>
 						<Route path='/print/:id' component={(routeProps)=><PrintPage brew={this.props.brew} query={queryString.parse(routeProps.location.search)} />}/>
 						<Route path='/print' exact component={(routeProps)=><PrintPage query={queryString.parse(routeProps.location.search)} />}/>
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -38,7 +38,8 @@ const NewPage = createClass({
 				tags        : '',
 				published   : false,
 				authors     : [],
-				systems     : []
+				systems     : [],
+				query       : {}
 			}
 		};
 	},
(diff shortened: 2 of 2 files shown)


ericscheid:
What happens if .. 

1. /template route is used to create the unsaved brew
2. some edits are made in the editor (but not yet saved)
3. the page is reloaded?

G-Ambatte:
This re-uses the existing `NewPage` so changes are saved to `localStorage` and can be recovered at `/new` (although I note that only the parts that have been edited are saved; if `brew.style` is not changed but `brew.text` is, only `brew.text` data will appear at `/new`).
However, the explicit purpose of the `/template` path is specifically to provide a pathway for other tools to provide data to HB, so if an external tool produces a template link, a document with the same content will be created every time that the link is followed - so refreshing the page will eliminate any changes that have been made.
It is worth noting that with current state of the commit, if there is data in `localStorage` (for example, from entering data on `/new`) that is not explicitly overwritten by the query parameters in the `/template` path, then the `localStorage` values will be used.

ericscheid:
Yes. So we need to be clear on whether this route is to be used as an API (in which case it should instead be returning a json resource) or whether it should produce an user-editing environment (in which case it should alter the eventual document.location so a refresh doesn't destroy any user edits).

This is similar to the #1499 issue.

G-Ambatte:
> (in which case it should alter the eventual document.location so a refresh doesn't destroy any user edits)

I floated this idea briefly on Gitter, but what if we went in the opposite direction? Disable autosaves (with a big warning bar) so that nothing is stored in `localStorage`, meaning that if the user has been working on a Brew in `/new` then it isn't overwritten when they play with a Templated or Cloned Brew. This should also mean that if there are unsaved changes, a confirmation box will appear when they try to close the tab, warning them again that they have changes that they have not saved.
In this theoretical, there *should* be no user expectations that do not align with the actual results as to exactly what will happen if/when they close or refresh the tab.

Unfortunately the problem with that last sentence is the word that starts with "s".

---

An alternate idea along a similar line would be to lock the Editor on Cloned and Templated Brews, so the only way to unlock it for editing is to save a unique copy first (thus redirecting to the `/edit` page).

---

> So we need to be clear on whether this route is to be used as an API (in which case it should instead be returning a json resource) or whether it should produce an user-editing environment (in which case it should alter the eventual document.location so a refresh doesn't destroy any user edits)

Based on previous discussions with @calculuschild, I don't believe the intention is to allow external tools to write brews to the HB database directly; but rather to allow the production of a brew which the user may then save/print/whatever at their own discretion.

ericscheid:
I'm uncomfortable with disabling localStorage saving. It is there as a safety net for those brave folk that work for hours on their masterpiece before hitting the save button.

If anything, the `/template` and `/clone` (sic) routes should prompt for replacing the current `/new` scratchspace at initial load. 

> "You already have brew text in your editor, do you want to replace it?"

ericscheid:
> Based on previous discussions with @calculuschild, I don't believe the intention is to allow external tools to write brews to the HB database directly; but rather to allow the production of a brew which the user may then save/print/whatever at their own discretion.

Agree, rollout of an API to facilitate 3rd-party tools to integrate deserves deeper thought.

G-Ambatte:
> If anything, the /template and /clone (sic) routes should prompt for replacing the current /new scratchspace at initial load.
>
> > "You already have brew text in your editor, do you want to replace it?"

I could see this working; on "Yes", update `localStorage` and redirect to `/new`. On "No", disable autosave to `localStorage`, so that content will not be affected and changes will only be saved if the user manually hits 'Save'.

Or, an alternate line of thought: Clones/Templates save to a different `localStorage` key(s), so they do not affect the contents of `/new`.

<!- When I was working on #1534 last night, I used `JSON.stringify` to push an object to `localStorage` and `JSON.parse` to retrieve the data, specifically so it could be easily expanded on later. However, this morning I was struck by a thought: why not just push the entire `this.state.brew` through `stringify`, then write a function in `getInitialState()` that parses all the properties in the object retrieved from `localStorage` into the page's Brew? Then `/new`, `/clone`, `/template` and any other similar future page could all have their own `localStorage` key and not cross-contaminate their data. -!> Of course, this is a different scope than what this PR originally sets out to achieve.

calculuschild:
Closing. This needs further discussion on the underlying issue to make sure how/whether this needs to be handled going forward.

PR #1534, opened 2021-08-09
Save Renderer data in LocalStorage and GoogleDrive

### LINKED ISSUE #1498: Choice of renderer on /new page is not preserved
On staging, if the `/new` page is reloaded then the choice of selecting "V3" as the renderer resets back to default. 

It needs to be preserved in localStorage like other stuff.

G-Ambatte:
This PR should resolve #1498.

While the original intention was to merely save and retrieve the `renderer` data to/from `localStorage`, during the process of testing it was discovered that while this setting worked on the NewPage, it was lost when saved to Google Drive.
Further investigation revealed that the `renderer` data was not being saved to `file.properties`. Correcting this resulted in correct operation on both NewPage and EditPage after saving to Google Drive.


calculuschild:
LGTM!

PR #1541, opened 2021-08-10
Update BrewItem tooltips

G-Ambatte:
PR based on feedback from Gitter:

> the tooltip for created/updated microdata should really only appear when hovering over the refresh item in the `<div class="info">`, not when hovering the `<div class="brewItem">`.  
Also, the items within the `<div class="links">` should have tooltips (e.g. "share", "edit", "download", "delete").

However, as I believe that the `<div class='info'>` is a small un-intuitive target when compared to the entire Brew Item, at this time I have added a line to the start of the tooltip containing "Title: {`brew.title`}" without changing the element that the tooltip is attached to.


G-Ambatte:
Tooltip images:
image
image
image
image
image
image
image

calculuschild:
LGTM!

Issue #1221, opened 2021-01-28
Systems are not removed from database


G-Ambatte:
## Issue

It appears that when the `systems` array is written to the database, items that are in positions greater than the current length of the `systems` array are not removed.

---

### Reproduction Steps:

1. Tag a brew with all four of the systems in the Metadata Editor tab.
2. Save the brew.
3. Return to the Metadata Editor and untick the Pathfinder systems box
4. Save the brew.
5. Refresh the page.
6. Check the Metadata Editor - Pathfinder has returned.

---

### Second Repro:

1. Tag a brew with one or more systems.
2. Save.
3. Untick all systems boxes.
4. Save.
5. Refresh.
6. All ticks have returned.

---

### Third Repro:

1. Tag a brew with any two non-5e systems.
2. Save.
3. Tick 5e.
4. Save.
3. Untick one non-5e system.
4. Save.
5. Untick the other non-5e system.
6. Save.
7. Refresh - 5e is the only remaining system, but cannot be removed. Checking this brew in the database will show that '5e' is in all positions 0-2 in the `systems` array.

![Example image](https://user-images.githubusercontent.com/19826659/106108698-7e3b8b00-61ad-11eb-974e-b9c9875e83bf.png)

ericscheid:
The bug is here:
https://github.com/naturalcrit/homebrewery/blob/d15bec08a3cb1c80fdb34a28f2bf50f50112a597/server/homebrew.api.js#L55

`_.merge({arr:[1,2,3]}, {arr:[5,6]})` produces `{arr: [5, 6, 3]}` .. i.e. _.merge recursively merges objects and elements, so `target.arr[0]` is replaced with `source.arr[0]` (etc).

Thus:
`_.merge({systems:[5e,4e,3.5e,PF]}, {systems:[4e,3.5e,PF]})` produces `{systems:[4e,3.5e,PF,PF]}`.


Additionally, this: https://github.com/naturalcrit/homebrewery/blob/d15bec08a3cb1c80fdb34a28f2bf50f50112a597/client/homebrew/editor/metadataEditor/metadataEditor.jsx#L32-L39 
wasn't written with the possibility that the system being pushed into the array is already in the array, and so by particular sequences of updates, saves, and reloads we end up with `systems:["5e","4e","3.5e","3.5e","4e","4e","4e"]`.

ericscheid:
[...]

PR #1543, opened 2021-08-10
Stop view count being reduced by an open Edit page

### LINKED ISSUE #406: View count on user's views reset
Not sure exactly what happened or what caused it. But the view count on my user page for one of my published brews dropped from over 500 to under 100. I suspect something triggered it to reset entirely, and it grew back up to 82 before the next time I saw it.

May or may not have something to do with the fact that minor alterations to the brew were made. However, I could not reproduce by altering other published brews.

http://homebrewery.naturalcrit.com/user/Zagorath

http://homebrewery.naturalcrit.com/share/4kCj3fbal

G-Ambatte:
Currently, when an EditPage is opened, the `brew.views` data is loaded into the page. However, this data can and does change while editing is taking place, so it should not overwrite the values in MongoDB/Google Drive when the brew is saved.
This PR simply removes the `brew.views` property from the EditPage at the time that the brew is loaded.

This should resolve #406.

### DIFF
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -51,18 +51,16 @@ const EditPage = createClass({
 	},
 
 	getInitialState : function() {
-		const newBrew = this.props.brew;
-		delete newBrew.views;
 		return {
-			brew                   : newBrew,
+			brew                   : this.props.brew,
 			isSaving               : false,
 			isPending              : false,
-			alertTrashedGoogleBrew : newBrew.trashed,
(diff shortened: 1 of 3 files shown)


ericscheid:
This feels like the wrong way to fix this — it's patching just one symptom of a larger problem
(https://github.com/naturalcrit/homebrewery/issues/1221#issuecomment-812151150)

calculuschild:
Alright. I'mma go ahead and merge this as-is. We can address the rest of #1221 in a separate PR since it looks like it still needs a bit more hacking to finish up.

PR #1562, opened 2021-08-12
Editor-Specific Snippet Groups, and Print Snippet Menu Changes

Gazook89:
The main thrust of this PR is to make changes to the current "Print" snippet menu.  The initial changes were a new "square" dimension page snippet similar to the A4 option, but scope grew...

### New/Updated Snippets
- Renaming "Print" to "Page":  IMO, "Print" menu should actually have the option to print a brew, so it's a bit misleading as-is.  Snippets in the menu affect the display of the page itself.
- Square Page:   Change page width & height to be square, for users who would like to share brews on sites like instagram.  Sets dimensions, reduces page padding, and reverts column count to 1.
  - <details><summary>Example Image</summary>image</details>
- Updated Ink-Friendly snippet:   updated for v3 so that is using selectors `.monster`, `.note` etc rather than the old selectors `hr + blockquote`.  Also removes background colors from monster blocks and notes to be even more printer friendly.

### Editor Specific Snippet Groups
When making the above changes to the Print menu, it occurred to me that I could try to display the Print/Page snippet menu only when viewing the Style editor, so they are all css specific.  As is, they utilize `<style>` tags, which is something we want to get away from...especially they are making changes across the entire brew.

### Help needed
~~So in snippets.js  (v3) I added a "editor" property in the snippet groups, with values either 'text' or 'style'.   Then each snippet group could be sorted to the right editor....except, for now, I've reached the limit of my nascent JS skills after staring at snippetbar.jsx for several hours. Just wanted to get a PR out there in case this is an easy thing.~~

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -40,19 +40,9 @@ const Snippetbar = createClass({
 		let snippets = [];
 
 		if(this.props.renderer === 'V3')
-			if(this.props.view === 'text') {
-				snippets = SnippetsV3.filter((SnippetsV3)=>SnippetsV3.view === 'text');
-			} else if(this.props.view === 'style') {
-				snippets = SnippetsV3.filter((SnippetsV3)=>SnippetsV3.view === 'style');
-			} else
-				snippets = null;
+			snippets = SnippetsV3.filter((snippetGroup)=>snippetGroup.view === this.props.view);
 		else
-		if(this.props.view === 'text') {
-			snippets = SnippetsLegacy.filter((SnippetsLegacy)=>SnippetsLegacy.view === 'text');
-		} else if(this.props.view === 'style') {
-			snippets = SnippetsLegacy.filter((SnippetsLegacy)=>SnippetsLegacy.view === 'style');
-		} else
-			snippets = null;
+			snippets = SnippetsLegacy.filter((snippetGroup)=>snippetGroup.view === this.props.view);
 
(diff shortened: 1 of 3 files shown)


calculuschild:
Take a look at the lines in `snippetBar.jsx` here:

https://github.com/naturalcrit/homebrewery/blob/9726fb5666689c8f879cc91a89e87c90adb0ac40/client/homebrew/editor/snippetbar/snippetbar.jsx#L42-L47

There you can see how we swap between legacy and v3 snippet files, as well as only display snippets when in the `text` view. You could finagle some extra logic in there to split out the snippets from the file that have the correct `editor` property you made (although I would rename it `view` so the variables match) depending on what `this.props.view` is equal to.

Gazook89:
With the hint from @calculuschild , got the editor-specific snippet menus working (Thanks again!).  So Print/Page menu shows only in the Style view.

- Matched Legacy with v3 changes in terms of snippets, with appropriate changes in syntax.
- Added css-commented snippet titles to promote well-organized stylesheets (for Print/Page menu only)
- Small tweaks to Ink Friendly snippet.

### Issues
- The "welcome page" sandbox *does not* have different views, and thus it contains all menus.   If a user builds a brew from this page the Style snippets that have been adapted to work without the `<style>` tags will not work out of the box.  
  - The Welcome Page should display a Style view toggle (but no Metadata toggle).  
  - Alternatively, separate snippets could be written for just the welcome page, or all style snippets could be passed into function to wrap `<style></style>` tags around the existing snippets.

**Tested in my docker build.**

calculuschild:
Thanks @Gazook89

PR #1568, opened 2021-08-13
Hotkey for hyperlinks (`Ctrl`/`Cmd` + `K`)

Gazook89:
Adds a simple and common hotkey for changing text to a markdown hyperlink:

- Windows/Linux:  `Ctrl` + `K`
- macOS:  `Cmd` + `K`

Running the shortcut will place selected text inside the `[selection]` and then leave the cursor in the parentheses `(url)`.    `[selected text](cursor)`

Selecting an entire link and hitting the hotkeys again will remove the link (however, unlike other hotkeys, the selection is lost after creating the link so you *will* have to reselect your text if you want to remove the link syntax).

### Fixes needed:
- [x] multi word selections do not work properly...sends final cursor position to wrong place.
- [x] Haven't tested removing Hyperlinks that actually contain URLs which will not work. If going to implement, not sure if it should remove the link syntax AND the url, or just the syntax and leave the url there as a raw link.


Gazook89:
Calling this "done" unless someone disagrees.  Now the generated syntax with the hotkeys is `[](url)` and `url` is highlighted.

Earlier I thought that I would need to find a way to remove the link syntax but now am not so worried about it.  The functionality in this PR matches the functionality of Github comments in terms of link hotkeys.  If it's good enough for github...

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -99,6 +101,15 @@ const CodeEditor = createClass({
 		}
 	},
 
+	makeLink : function() {
+		const selection = this.codeMirror.getSelection(), t = selection.slice(0, 1) === '[' && selection.slice(-6) === '](url)';
```
calculuschild, line 105: `t` (whether to "create" or "undo" the hotkey) is only true if the selection literally ends with `](url)`. This means the "undo" hotkey won't work if the user has changed the url to anything else. i.e., selecting `[](url2)` and pressing the hotkey just creates another link since `url2` != `url`.

Can you change this to accept undoing *any* value instead of requiring exactly `url`?

calculuschild (changes requested):
Just one small change.

shared/naturalcrit/codeEditor/codeEditor.jsx
Gazook89: So this was something I fiddled with a bit, but was left wondering if the end result of undoing should be:
```
a) [alt text](duckduckgo.com)  -->  alt text duckduckgo.com
b) [alt text](duckduckgo.com)  -->  alt text
c) [alt text](duckduckgo.com)  -->  duckduckgo.com
d) something else?
```
I was going to post about it and ask for input, but then I checked to see what Github does in the same situation...and it actually has the same behavior as this PR currently has.  
```
[alt text](duckduckgo.com) -->  [[alt text](duckduckgo.com)](url)
```
So I thought if it's good enough for GitHub, it's good enough for and left it at that.  But it can be updated to another option, I'm just interested to hear which is preferred.  In my mind, no information should be lost, so option (a) would be my choice (if any change happens at all).

calculuschild: I would prefer option A if you can do it.

The issue I was trying to address though, was that `[alt text](duckduckgo.com)` turns into `[[alt text](duckduckgo.com)](url)`, but `[alt text](url)` turns into `url`. The behavior should be the same whether the url is "url" or "duckduckgo".

Gazook89: Okay check it out now.  It uses regex to sort out anything in the URL field, and I went for option A above as the desired result.  It also finishes the "undoing" of a hyperlink with the URL selected, so the next step of deleting that URL is a little easier.

```
alt text  -->  [alt text](url)
[alt text](duckduckgo.com)  -->  alt text duckduckgo.com
*blank*  -->  [](url)
[](url)  -->  url
```

Some more changes I was fiddling with but couldn't figure out:

A) regex to determine if the selection is inbetween `](` and `)` so that link syntax could be removed by just having the URL highlighted and doing the hotkey.
B) just cleaning the code up a bit and using the capture groups to assign the `altText` and `url` constants.
C) should a blank zero-length selection end up creating `[alt text](url)`....adding "alt text" in as a placeholder rather than just empty brackets?

calculuschild: I added parts B and C that you suggested. Part A would be nice but is a little more involved because we need to grab text *around* the selection which I'd need to look into. If people complain we can change that.

calculuschild:
I'll finally merge this now. Lots of PRs to clean up. Maybe a good next PR after all these hotkeys are added would be to document them somewhere, maybe even in the snippet dropdowns.

Issue #1489, opened 2021-07-30
curly-injectors to target parent "page" element if no preceding sibling target


ericscheid:
If the `{curly-injector}` appears as the first chunk on a page (i.e. very beginning of brew, or following a `\page`) then have the values injected into the enclosing `<div class="page V3">` element.

That is, instead of defaulting to treating the `{curly-injector}` as plain text.

----
#### Example: adding a class to a page:

```
\page
{songsheet}
Non, rien de rien
Non, je ne regrette rien
Ni le bien, qu'on m'a fait
Ni le mal, tout ça m'est bien égal
```
Note that the usual classes of "page" and "V3" should be retained, even in face of `{class="overwrite"}`.

----

#### Example: adding a `lang` attribute to a page:

```
\page
{lang=fr}
Non, rien de rien
Non, je ne regrette rien
Ni le bien, qu'on m'a fait
Ni le mal, tout ça m'est bien égal
```

(note that adding a `lang` attribute to the whole brew is a separate issue #1343, and separately needed.)
(note 2: here we are injecting a html-attribute, not a style class or style rule. See #1489 for that.)

An attempt to override the `id` of a `<div class="page" id="p123">` should be ignored.

ericscheid:
See also #1488

jeddai:
If we set it up where it could be anywhere in the page like the pageNumber syntax, something like `{{page.class}}`, I think that would be better than specifically right after `\page`.

calculuschild:
Right after `\page` seems more logical syntax to me. Injector curlies already can be used to target the immediately-preceding element. This expands the syntax to apply to the immediate parent if there is nothing preceding. This could apply to other nesting structures, for instance table cells? Divs?

```
|  header |  header |
|:-------:|:-------:|
|{blue} hi| not blue|

{{A div
{blue}
content?
}}

```

Not that we want to implement any of those necessarily.

But my main concern is so far none of our other syntax uses "keywords" to trigger something except for `\page` and `\column`, and I'm hoping we can keep those special cases to a minimum. The pagenumber syntax uses `pageNumber` as a CSS class, but not a functional keyword, if that makes sense. It can be placed anywhere because it's a standard DIV with an absolute position, so it doesn't matter where in the markdown it goes.

That's my 2 cents.

Gazook89:
>This expands the syntax to apply to the immediate parent if there is nothing preceding.

[...]

Issue #846, opened 2019-04-05
Feature request: page indicator for the source pane


ericscheid:
When the author has a large document, they can scroll the preview pane and there is displayed a page indicator widget in the bottom left:

    <div class="pageInfo">7 / 13</div>

However, over on the left side the source pane has no equivalent indicator. Further, since that pane only displays an extract of the actual source, an author can't use in-browser-find to locate particular a section of content. They have to scroll and scan, scroll and scan.

It would be useful to see an indicator of which page the source will get rendered to. Possibly as a widget in the same style, or maybe even insert the information in a `<span style="float:right;"> 7 / 13</span>` just after the `<span role="presentation" style="padding-right: 0.1px;">\page</span>`

HeruAsatru:
Maybe it would be useful to be able to click on certain elements (like headers, or bold/italic text, or...) of the brew (in the render view) and jump directly to that position in the source code?

Gazook89:
PR #1586 would add page numbers to the source pane

calculuschild:
Handled in a slightly different way in #1586 , which adds a page number onto the end of each `\page`

PR #1586, opened 2021-08-18
Update editor.less for .pageLine

Gazook89:
Change the .pageLine (page break) border from bottom to top, since any text on the same line is part of the following page rather than the preceding page.  Also, `background-color` had a value of `fade( ... )` which I wasn't familiar with, and didn't seem to be doing anything.  Replaced with a slight linear gradient.  I'm not very attached to the gradient.

image

Some changes to the file to line up the colons.

I took this up with the thought of maybe someday allowing a toggle to add page number and footer snippets to every new page.


Gazook89:
In the latest commit I basically undid what I originally set out to do, because along the way it was determined that `\page` would only work when alone on a line.  And I wasn't married to the idea.

NOW....this PR is about adding a page number to the `.pageLine` in the editor.  It does this with only CSS using an `:after` pseudo element and css counters, in the same way auto page numbering snippet works.   This may help users navigate long brews in the editor.

This should not be interpreted as a match to a `.pageNumber` snippet, as users can start those however often they like.  However, it *should* match the page counter in the bottom right of the preview window.

image

client/homebrew/editor/editor.less
```diff
@@ -4,41 +4,47 @@
 	width    : 100%;
 
 	.codeEditor{
-		height : 100%;
+		height        : 100%;
+		counter-reset : page;
 		.pageLine{
-			background-color : fade(#333, 15%);
```
calculuschild, line 9: So I looked it up. `fade` is part of Less. Essentially takes the color and applies an opacity of x% to it.

calculuschild:
I do like the page numbering.

I'm not sold on the CSS styling of the line though. Removing all the background color makes it *much* harder to locate the start and end of a page simply because it looks like standard paragraph text with an easy-to-miss underline.

How about something like
```
background: linear-gradient(to bottom, white, #33333328 40%);
border-bottom: #339 solid 1px;
```

This way we keep the underline which I think it important, we keep a distinctive background, but it has just the slightest of gradients at the top to blend into the page above.

This does bring up another question though... I feel like the `\page` separator should actually belong to the page *below*. Like, the divider that created page 2 seems like it should be a part of page 2, not page 1? And scrolling to the *bottom* of the page content to find what page it is doesn't quite click for me. What do you think?

Gazook89:
So to your last question:  That is what I was originally concerned about when I first opened this PR, before I changed direction.  When `\page` was still possible even with other text on the same line (`\page more text here`), it was evident that the following text was being placed on the "second" or new page.   This meant it was more clear to have the solid black border at the *top* of the .newLine.   

However, the change was made that nothing else could appear after \page on the same line, if you wanted to actually create a new page, so this became less relevant.  

In my mind's eye I could see re-allowing text on the same line as \page (only after).   This could play well with the future possibility of #1489, allowing the `.page` element to be targeted by a curly injector right on the same line as `\page`.  For instance, adding a pre-made page layout using a class name:
```
\page {coverPage}
```
Or replacing the default background image:
```
\page {background-image:url(imgur.com)}
```
In both examples the style is being applied to the new page, below \page.  And more to the point for this PR, a page number could be included as done in the latest commit and that page number would be at the "top" of the page in the editor as you prefer.

Of course, the first page has to be considered, since it currently doesn't have a \page line.   So if we do the above, the first page wouldn't have the page number etc.  But what if *all* pages required `\page` at the start, including the first page?  A new document would already come preloaded with \page at the top, along with it's page number.  If the user deletes that first line, it would just show the brewRenderer background with a tip on how to get started (or a random rotating choice of 'tips'):

image

Overall, this makes the first page consistent with the rest of the pages, allows page numbering from "1" to be displayed at the top of the page, and a convenient place for parent-element injection.  Finally, it is also a good spot for *commented* "page names" for easy reference if we ever introduce code folding by page:

```
\page /* Table of Contents */
\page /* Class Descriptions */
\page /* Equipment */
```

calculuschild:
There's still possibility that we can allow content after `\page` on the same line at a later date. Don't let that be the deciding factor, but also keep in mind that is probably *far* down the road because it involves writing a whole separate Markdown extension just for that special case.

We can discuss requiring `\page` at the start (as it is now, the *last* page will be missing a page number) in a separate PR I think, since this one is aimed at just adding page numbers.

I say we move ahead with associating `\page` with the start of a new page. So lets move the border to the top to make it clear `\page` belongs to the page contents below, and let's just leave Page 1 unlabeled for now. It's obvious enough that the top of the document coincides with page 1 for now, and we can address it later after some discussion on the Gitter perhaps.

Gazook89:
Updated this PR to match the change requests.....Page Number starts at '2' now.   Removed the gradient entirely though as I now think that a single line height gradient looks a little silly, and doesn't fit the design of the overall site (gradients are used nowhere else).  

*If* ever we allow `\page` at the start of the document as discussed, the start of the page incrementing will need to be worked out....right now, the first incrementation from `0` is done by counting `.codeEditor`, and then subsequent `.page` elements.  

Does this address #846?  Wasn't sure if we should leave that open and plan for a more explicit page indicator to match the preview side.

calculuschild:
> Does this address #846? Wasn't sure if we should leave that open and plan for a more explicit page indicator to match the preview side.

I think it addresses it. I'll try to get this PR merged tonight. Good work!

PR #1592, opened 2021-08-20
Move Change Log link to Version Number, add Report Issue to User Page

### LINKED ISSUE #1166: [UI Suggestion] Have the version # in the top toolbar link to github repository
I noticed that the "link" from the toolbar where the version number is given doesn't actually go anywhere.  It might be helpful to have that lead to this github repository, or to the changelog.  Or, with the new version coming up, possibly a dialog that offers to migrate a brew to the new version, explaining that it might/likely will break some formatting.

I see now that the Welcome page has a link to the Changelog on it, labelled as such, but that link is not in the normal Editing page.

Gazook89:
This is an effort to resolve and close #1166.  

### Changes
- Adds a short sentence to changelog.md pointing to Github for development updates
- Removes Change Log link from navbar, adds link via version number.
- Adds Report Issue link to User Page

> The Change Log page content could also directly link to GitHub (eric)

> Compromise: the Change Log link goes to /changelog and then in the content of that page there is an explanatory preamble which links to GitHub for all the gory details. (eric)

This PR adds a short sentence about visiting the project repo on Github, including a link. (Commit b11d130)

----

> We don't have Report Issue on the Brew Share page - we have seen issues reported regarding the Share page

> Also I suggest that we put "Report Issue" button on every page, as it seems important - a bug can occur in any part of the app, apparently. (alexey)

Not sure this is a good idea ("on every page"), since I suspect there are many page visitors to a Share page that are not familiar with Homebrewery.  If they see some "issue", they should communicate it to the author first who can then see if it's something they can fix on their end or then mention it on the subreddit.  In particular, I am thinking of issues like "third column" or even "problems" with the homebrew content itself.   

---

> We don't have Report Issue on the User home page - we have seen issues reported regarding the User page

Added on this PR.  (Commit e80e482)

----

> We don't have Report Issue on the Change Log page - I doubt we've seen or will see issues regarding this

I also don't see this as being a problem.  But more to the point, I don't see in the code where the Change Log page is built, so not sure where to add Report Issue code.

### Further Work
- It drives me a little nuts that the various navItems are not not vertically aligned to the same baseline due to various fonts, sizes, and line heights (and some weird margin stuff).
- I think maybe the google drive icon could be grouped into the same box as the Save Now box.


changelog.md
```diff
@@ -31,6 +31,7 @@ pre {
 ```
 
 # changelog
```
calculuschild, line 33:
```suggestion
## Changelog

```

Otherwise the text you added below will trigger dropcaps on the first letter, and this will be cleaner than adding custom styling to the page.

calculuschild (changes requested):

changelog.md
Gazook89: added

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -135,6 +135,7 @@ const UserPage = createClass({
 			<Navbar>
 				<Nav.section>
 					<NewBrew />
+					<ReportIssue />
```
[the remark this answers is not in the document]
Gazook89: ah, haven't figured out the log-in side of things yet so can't ever check user page stuff.  thanks, added

calculuschild:
Looks good! 🎉

PR #1594, opened 2021-08-20
Handle `brew.style` being `undefined` when injecting snippets.

### LINKED ISSUE #1593: Snippets won't insert on CSS tab
Attempting to insert a snippet on the CSS tab fails when no changes have been made to the content of the editor on a new brew.

The resultant error is `Uncaught TypeError: text is undefined
    handleInject https://homebrewery.naturalcrit.com/homebrew/bundle.js:72863`

As the style data is left undefined until the user changes the data in anyway, it appears that `handleInject` is failing when attempting to process the `undefined` data. It should be a simple matter to detect this situation and insert the default content instead.

G-Ambatte:
This PR will resolve #1593.

When `this.props.brew.style` is still in it's default state of `undefined`, attempting to inject a snippet will fail. This PR changes the logic for setting the `text` variable so that if `this.props.brew.style` is `undefined` when it is passed to the `text` variable, it will instead send the value set as `DEFAULT_STYLE_TEXT` - which is what is displayed in the editor at this time anyway.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -68,7 +68,10 @@ const Editor = createClass({
 	},
 
 	handleInject : function(injectText){
-		const text = (this.isText() ? this.props.brew.text : this.props.brew.style);
+		const text = (
+			this.isText() && this.props.brew.text ||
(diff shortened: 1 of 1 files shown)


calculuschild (approved):
*chef's kiss* :ok_hand:

PR #1605, opened 2021-08-25
Add tooltip to the Time Since Last Update `info <span>`

G-Ambatte:
This PR adds a tooltip to the Time Since Last Update `info <span>` on the UserPage, like every other item in the BrewItem.

I believe this was originally part of the PR; it must have been lost in one of the merges/updates. This PR rectifies that mistake.


ericscheid (approved):

calculuschild:
LGTM

PR #1606, opened 2021-08-25
Rearrange BrewItem metadata items - put `authors` first on it's own line

G-Ambatte:
This PR rearranges the metadata in the BrewItem object on the UserPage to put `authors` before the other metadata, while retaining it's own line in the BrewItem.


calculuschild:
@G-Ambatte Did this have an associated issue with it that need to be closed out?

PR #1617, opened 2021-08-28
Moved server start log to `app.listen` callback

blitzher:
This is done in order to ensure that the app is properly initialized and listening, before logging in the console.


calculuschild:
Thanks for submitting this! Looks good to me.

Issue #1422, opened 2021-07-02
[FEATURE]: Built-in color picker/wheel


Gazook89:
Since the tool is primarily for spiffy-ing up brews, it seems it should include a color wheel/picker as part of the interface.  Personally I already have a color wheel browser extension, but I imagine that many users do not and may not have a great understanding of CSS color values.

ericscheid:
As a custom menu item on the Style pane? Or maybe like [this](https://github.com/easylogic/codemirror-colorpicker), where `ctrl-k` invokes the picker?

G-Ambatte:
> [...] Or maybe like [this](https://github.com/easylogic/codemirror-colorpicker), where `ctrl-k` invokes the picker?

I like that one; if I'm reading the documentation correctly, we could set it up with a custom palette for colours that are already commonly used in HB.

calculuschild:
This may be folded into #1870

dbolack-ab:
Is this still a concern?

5e-Cleric:
## Following the introduction of CodeMirror 6

A color picker is easily introduced by getting `import { color } from '@uiw/codemirror-extensions-color';` and setting that in the extension list. It just works.

### But

That said, it works with an appropriate language, in this case, CSS.  If using the Markdown language, lezer doesn't provide the appropriate nodes and the extension doesn't do anything.

This can be fixed by modifying the markdown parser to catch these blocks, but that is easier said than done.

To make custom nodes, you need a .grammar file and lezer-generator , with a syntax as defined in https://lezer.codemirror.net/docs/guide/#building-a-grammar, then follow the instructions to create a parser. That should send the nodes with the rest and they can be caught like so:

```javascript
import { LRLanguage, LanguageSupport, parseMixed } from '@codemirror/language';
import { cssLanguage } from '@codemirror/lang-css';

const homebreweryMarkdownParser = MarkdownParser.configure({
  wrap: parseMixed((node) => {
    return node.name === "mdStyle" ? { parser: cssLanguage.parser } : null;
  })
});

const hbMarkdown = LRLanguage.define({parser: homebreweryMarkdownParser});
```

---

I have stopped developing this because i cannot really understand the lezer grammar syntax, keeps giving very obscure error logs and the documentation (previously linked) is not helpful at all.

Anyone feel free to tackle this, as for right now i wont.

PR #1620, opened 2021-08-30
(WIP) Color Picker

G-Ambatte:
This draft PR is an initial pass at adding a color picker tool to the Text and Style tabs. As of this commit, the color picker is functional and updates the text in the editor:
- If text is selected, any color values in the selection are replaced with the picked color (identified as "#" followed by 3, 4, 6, or 8 hexadecimal values).
- If no text is selected, the hex value of the picked color is injected at the current cursor position.

---

Once completed, this PR will resolve issue #1422.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -1,9 +1,12 @@
+/*eslint max-lines: ["warn", {"max": 300, "skipBlankLines": true, "skipComments": true}]*/
 require('./editor.less');
 const React = require('react');
 const createClass = require('create-react-class');
 const _ = require('lodash');
 const cx = require('classnames');
 const dedent = require('dedent-tabs').default;
+
+import { SketchPicker } from 'react-color';
 
 const CodeEditor = require('naturalcrit/codeEditor/codeEditor.jsx');
 const SnippetBar = require('./snippetbar/snippetbar.jsx');
@@ -41,7 +44,9 @@ const Editor = createClass({
 	},
 	getInitialState : function() {
 		return {
-			view : 'text' //'text', 'style', 'meta'
+			view            : 'text', //'text', 'style', 'meta'
+			hideColorPicker : false,
+			sketchColor     : '#000000'
 		};
 	},
 
@@ -67,7 +72,7 @@ const Editor = createClass({
 		}
 	},
 
-	handleInject : function(injectText){
+	handleInject : function(injectText, overwrite=false){
 		const text = (
 			this.isText() && this.props.brew.text ||
 			this.isStyle() && (this.props.brew.style ?? DEFAULT_STYLE_TEXT)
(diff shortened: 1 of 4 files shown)


Gazook89:
Out of curiosity, why not use [codemirror-colorpicker](https://github.com/easylogic/codemirror-colorpicker)?  It seems to be well integrated with codemirror.

I think it would be handy to have the inline color picker which opens when clicking on a :before element, such as the below (done with just browser inspector):

image

calculuschild:
Is this something we can integrate with our v3 markdown or is it obly compatible with a css tab?

Gazook89:
@calculuschild i am not sure, in regards to codemirror-colorpicker.

With react-color, I tried building on this PR to add an element in front of hexadecimal color codes to get started, but didn't get very far.  I was using the functions in markdown.js that convert the `{{ }}` into divs as a guide, but really didn't know what i was doing after a point.

G-Ambatte:
> Out of curiosity, why not use [codemirror-colorpicker](https://github.com/easylogic/codemirror-colorpicker)? It seems to be well integrated with codemirror.

I tried using `codemirror-colorpicker` but had issues trying to get it to play nicely with React - I found it easier to get `react-color` to play nicely with CodeMirror.

> I think it would be handy to have the inline color picker which opens when clicking on a :before element, such as the below (done with just browser inspector):

As do I; I don't expect the current iteration to be the final product. However if someone else wants to run with this feature, either using this work as a starting point or ignoring it completely, I would not be upset at all.

calculuschild:
I've finally had a chance to play around with this.  I do like the idea of it, but here are my thoughts for moving forward:

- Having icons inline such as using [codemirror-colorpicker](https://github.com/easylogic/codemirror-colorpicker) would be especially handy, as agreed above. I think it's worth another visit to try getting that to work.
- Perhaps additionally, we can keep the little button, but alongside the undo/redo buttons as another "tool" separate from the snippets.
- I don't like having to select text for the color picker to do anything. Instead I think it would be easier if selecting any color just injected that hex right at the cursor, and then subsequent selections if you haven't moved the cursor would just overwrite that hex again.

Gazook89:
Just a thought while I was doing the code editor page numbers: this might be a good use of `cm.setBookmark()` to add an inline widget/node. Or maybe addWidget(). 

I don't have anything to add beyond that. 🤷

calculuschild:
This might be overtaken by https://github.com/naturalcrit/homebrewery/issues/1870, although this is a good reference for that.

Any opposed if we close this down?

ericscheid:
I support the general principle of doing-the-work / keeping-notes in the one issue, so a maybe-yes to closing.

That said, would be sad if progress on this one feature was blocked by slower progress on a much larger effort. Could we re-point/revise this issue to be compatible with #1870 and marking it as Blocked? Are there things in #1870 that would be a major blocker to this?

calculuschild:
#1870 also proposes a whole different mechanism for summoning the color wheel, so it's built upon a different code structure that this PR isn't really compatible with. In other words, we could spend the effort to make this PR work, but it would just be torn town and rebuilt later once #1870 is underway, which is in turn waiting for "themes".

jeddai:
@G-Ambatte 
> I tried using codemirror-colorpicker but had issues trying to get it to play nicely with React - I found it easier to get react-color to play nicely with CodeMirror.

I have some tricks for handling CodeMirror+React issues from when I was setting up the snippet options proof of concept. Was it issues with elements being redrawn on changes?

calculuschild:
Closing. This needs further discussion on the underlying issue to make sure how/whether this needs to be handled going forward.

PR #1636, opened 2021-09-04
Add Textbox with Share Link to Metadata Editor

Gazook89:
This PR adds a read-only text field that displays the Share URL for a brew.  It also adds a 'copy to clipboard' button.

This removes the need to click through to the Share page to get the share url, which can be tedious.  

image

(my local build doesn't display the icons for some reason I haven't figured out yet).

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -7,6 +7,8 @@ const request = require('superagent');
 
 const SYSTEMS = ['5e', '4e', '3.5e', 'Pathfinder'];
 
+
+
 const MetadataEditor = createClass({
 	getDefaultProps : function() {
 		return {
@@ -69,6 +71,7 @@ const MetadataEditor = createClass({
 		const meta = this.props.metadata;
 
 		const shareLink = (meta.googleId || '') + meta.shareId;
+
 		const title = `${meta.title} [${meta.systems.join(' ')}]`;
 		const text = `Hey guys! I've been working on this homebrew. I'd love your feedback. Check it out.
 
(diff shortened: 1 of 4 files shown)


calculuschild:
First though in my head, is there a clean way we could make sure the *end* of the share link is always visible in the box even if it's a bit small? So the `/share/ID` part is always visible?

Gazook89:
I thought about moving the Reddit button to the next line, full width below the textbox.  Also, turning the copy-to-clipboard button into a onHover element that displayed over the textbox.   Those two adjustments would help.   Of course, it's limited by the width of the panel which the user sets.

In my opinion, the font-size shouldn't be different from the rest of the text.

calculuschild:
I was thinking more just some text-alignment or something so if there's not enough room it favors the right side.

Gazook89:
> I was thinking more just some text-alignment or something so if there's not enough room it favors the right side.

This is done with commit 9121c2e.

And per discussion in Gitter, I added the same options to a dropdown Share menu in commit 4c2a1df....but might just want to toss that commit and have someone else do it, as I'm sure it needs to be refactored.  But it does work on my local machine.

calculuschild:
After sitting on this for a while.... I think it's going to be all-around cleaner to just move this fully into the Share button as dropdown menu items (including the "share to reddit" button). I'd rather keep the Metadata menu less cluttered if we can, in anticipation of "theme" and other options that will eventually need to fit in there too.

@Gazook89 , would you be willing to build a new PR with *just* the items in the Share dropdown menu in favor of this? (and it would then count toward Hacktoberfest as well 😉 )

calculuschild (commented):
After sitting on this for a while.... I think it's going to be all-around cleaner to just move this fully into the Share button as dropdown menu items (including the "share to reddit" button). I'd rather keep the Metadata menu less cluttered if we can, in anticipation of "theme" and other options that will eventually need to fit in there too.

@Gazook89 , would you be willing to build a new PR with just the items in the Share dropdown menu in favor of this? (and it would then count toward Hacktoberfest as well 😉 )

Gazook89:
Superseded by #1732 , closing this PR.

Issue #181, opened 2016-08-29
Indent 2nd paragraph in monster stat block


Fralexion:
If the description of a monster's ability or action takes up multiple paragraphs to describe (such as a dryad's Fey Charm action), there should be no vertical separation between the paragraphs and each one after the first should have a slight indent, as normal text outside a stat block does. 

Currently, the only way to avoid separating the paragraphs is to end the upper one with two spaces before hitting Enter (which I discovered by accident) and there is no way to put an indent in the lower ones.

calculuschild:
We might be able to leverage our `:` blank line syntax here, since paragraphs afterward are unindented.

```
**First item** La dee dah dee dah
:
**Second Item** Zippity-doo-dah

Another paragraph of Second Item

And another. Both indented.
:
**Third Item** Back to unindented
```

ericscheid:
For reference - note the **Enslave** action:
![Aboleth in WotC Monster Manual](https://user-images.githubusercontent.com/2605467/132294826-a2289aae-443a-43ab-a92b-928fc2e030f9.png)

ericscheid:
Another possibiity is to syntax those actions as definition list items and then the 2nd paras would just be paras. Might need an extra curly-div for semantics and also css-targetting:
````
{{monster,frame
[...]

{{monster-actions
### Actions
***Action Thingo One*** :: description of the action
***Another Action*** :: description of the action
***Enslave (3/Day).*** :: The aboleth targets one creature it can see within 30 feet of it. The target must succeed on a DC 14 Wisdom saving throw or be magically charmed by the aboleth until the aboleth dies or until it is on a different plane of existence from the target. The charmed target is under the aboleth's control and can't take reactions, and the aboleth and the target can communicate telepathically with each other over any distance.
New paragraph .. Whenever the charmed target takes damage, the target can repeat the saving throw. On a success, the effect ends. No more than once every 24 hours, the target can also repeat the saving throw when it is at least 1 mile away from the aboleth.
}}

}}
````

But then again, on second thoughts ... this is stylised the same both within and without the `{{monster block}}`, which is encouraging and predictable to authors (and they don't have to learn about any weird css we special case).

````
***Bolded words.*** Morbi non erat non ipsum pharetra tempus.
:
***Aliquam ultrices.*** Scelerisque sem. In elit nulla, molestie vel, ornare sit.
:
[...]

PR #1644, opened 2021-09-07
fix: address issue #181

BertoLucca:
Addresses issue #181

Current homebrewery render 
image

Purposed render
image

### DIFF
--- a/client/homebrew/editor/snippetbar/snippets/monsterblock.gen.js
+++ b/client/homebrew/editor/snippetbar/snippets/monsterblock.gen.js
@@ -96,12 +96,12 @@ const getStats = function(){
 
 const genAbilities = function(){
 	return _.sample([
-		'***Pack Tactics.*** These guys work together like peanut butter and jelly.',
-		'***Fowl Appearance.*** While the creature remains motionless, it is indistinguishable from a normal chicken.',
-		'***Onion Stench.*** Any creatures within 5 feet of this thing develops an irrational craving for onion rings.',
-		'***Enormous Nose.*** This creature gains advantage on any check involving putting things in its nose.',
-		'***Sassiness.*** When questioned, this creature will talk back instead of answering.',
-		'***Big Jerk.*** Whenever this creature makes an attack, it starts telling you how much cooler it is than you.',
+		'\`Pack Tactics.\` These guys work together like peanut butter and jelly.',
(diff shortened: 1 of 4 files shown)


Gazook89:
I think this is clever but in my mind may cause unexpected issues, particularly if a user decides to bold some text inside an ability (that isn't at the start of the ability).  Obviously, I don't think they SHOULD do that, but...

I also recently spent a lot of time looking at #181 and just couldn't work out a good way doing it *without* adding unpredictable behavior.  Maybe it boils down to nesting div's inside a monster statblock for each section, so they can be styled independently?  Would be gross...

BertoLucca:
Oh, I really hadn't thought about these marginal cases. 🤔 Life would be so much easier if p tags tracked child position vs content position... Just being able to style the paraph only if it starts with bold would fix this issue.

Gazook89:
Yeah, that is a common demand that I see mentioned a lot in general css blogs etc. To be able to detect the contents of a container and style the container based on that (typically called Container Queries).  I think it's something that is being worked on by the people designing css but who knows if it'll ever happen. 

And yes, edge cases are annoying especially in a case like this where there is no reason for someone to bold inside one of these abilities. But, between either not having the appropriate spacing/indentation or suddenly getting unexpected line breaks when just trying to bold something, I think the latter is more jarring and hard to understand to a lay person. 

Hopefully someone else will weigh in here too as this is just one persons opinion.

BertoLucca:
@Gazook89 I gave some more thought to this and found that using code instead of strong would very much improve the situation, as the styling:

```css
.phb hr+blockquote p+p code::before {
	content: 'br';
	display: block;
	color: #0000;
	margin-left: -1em;
	line-height: 0.8em;
}
	
.phb hr+blockquote code {
	visibility: visible !important;
	font-weight: 800;
	font-style: italic;
}
```
Produces this layout (and, to be fair, people are less likely to mess with code tags than with bold ones), also, I like the hilight it gives to the pane:
image

calculuschild:
@BertoLucca Also take a look at the discussion on #181 . We might have a different way to go about this that doesn't rely on `bold` at all.

Unfortunately `code` is going to be used for something else already come V3, so that's not going to be an option.

calculuschild:
Superseded by #1680 . @BertoLucca Thanks for your work on this. Although we are going with a different solution, this helped us get on the right track.

PR #1645, opened 2021-09-07
Add Hotkeys for Underline and makeDiv

### LINKED ISSUE #732: Aligning Right / Horizontal Spaces
## Aligning Right

With a large pyramid-shaped image in the bottom-center of my homebrew, I need the text in the bottom of my right column to either align right, or to step down and indent right with each successive line. Any advice on how to achieve this?

A code to insert horizontal breaks would serve the same function.

Danke schoen!

### LINKED ISSUE #1303: Add a Comment Selection editor action

Gazook89:
Adds hotkeys:

## Underline:  `ctrl`/`cmd` + `u`
<details>
<summary>Click to expand</summary>

```
 <u>|</u>
```
*_pipe indicates final cursor position when no selection_

- I am happy to change this to `++ text ++` --> `<ins> text </ins>` to match the convention in [markdown-it](https://github.com/markdown-it/markdown-it-ins) but I cannot yet figure out how to add custom inline markdown in CodeMirror.  Since html tags will always still work in CodeMirror, it's my opinion it would be fine to add this hotkey as-is and make a change to it later to work with `++`.  
</details>

## makeDiv:   `shift` + `ctrl`/`cmd` + `m` 
<details>
<summary>Click to expand</summary>

```
{{
|
}}
```
*_pipe indicates final cursor position when no selection_

### Considerations
- makeDiv is for v3 only, using curly syntax. 
- I considered setting the final cursor position on makeDiv to be immediately after the opening brackets, to facilitate quickly adding classes etc....
  - but not sure how to find that position when creating from non-empty selection (when using replace)
  - if going ahead with this, likely best to eliminate the empty line between brackets, so that the first "empty" line is created by hitting enter after entering class names etc.
</details>

##  makeNbsp: `ctrl` / `cmd` + `.`
<details>
<summary>Click to expand</summary>

```
&nbsp;
```
Finishes with the cursor at the end.  May look into an additional hotkey (`ctrl` + `,`) to remove as well.
</details>

## makeSpace: `shift` + `ctrl` / `cmd` + `.`
<details>
<summary>Click to expand</summary>

```
{{width:10% }}
```
Finishes with span selected.  Repeated use increases the percentage by 10% increments to a max of 100%.

## removeSpace: `shift` + `ctrl` / `cmd` + `,`
Reduces the above span by 10% increments until 0%, at which point the span is removed entirely.

Overall, these hotkeys should make it easier to work with wrapping text around images, if not using shape-outside etc.  While shape-outside might be the ideal method of wrapping text, it is still difficult to use and doesn't work in every situation.

</details>

## makePage: `ctrl` / `cmd` + `Enter`
<details>
<summary>Click to expand</summary>

```

\page

|
```
*_pipe indicates final cursor position when no selection_

Creates a new page.

</details>

## makeColumn: `shift` + `ctrl` / `cmd` + `Enter`
<details>
<summary>Click to expand</summary>

```
\column

|
```
*_pipe indicates final cursor position when no selection_

Creates a new column

</details>

### Minor Changes
a. This PR does also change the current `makeItalic` hotkey to use `*` instead of `_`.   This is sort of an arbitrary change, but I feel it is more consistent with the syntax used in existing snippets.
b. makeComment() was added in a previous PR/release, but in this PR I just updated it to be responsive to either the GFM editor or CSS editor (commit 3bf5d7a).  It will now comment correctly in either editor.  It also fixes a minor issue where undoing a comment left a leading and trailing space.  This resolves #1303.

### Closes
a. possibly closes #732 
b. resolves #1303

### Related PR's
- #1719:  Lists
- #1568:  Links

### DIFF
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -155,12 +155,24 @@ const CodeEditor = createClass({
 	},
 
 	makeComment : function() {
-		const selection = this.codeMirror.getSelection(), t = selection.slice(0, 4) === '<!--' && selection.slice(-3) === '-->';
-		this.codeMirror.replaceSelection(t ? selection.slice(4, -3) : `<!-- ${selection} -->`, 'around');
+		let regex;
+		let cursorPos;
+		let newComment;
+		const selection = this.codeMirror.getSelection();
+		if(this.props.language === 'gfm'){
+			regex = /^\s*(<!--\s?)(.*?)(\s?-->)\s*$/gs;
+			cursorPos = 4;
+			newComment = `<!-- ${selection} -->`;
+		} else {
+			regex = /^\s*(\/\*\s?)(.*?)(\s?\*\/)\s*$/gs;
+			cursorPos = 3;
+			newComment = `/* ${selection} */`;
+		}
+		this.codeMirror.replaceSelection(regex.test(selection) == true ? selection.replace(regex, '$2') : newComment, 'around');
 		if(selection.length === 0){
 			const cursor = this.codeMirror.getCursor();
-			this.codeMirror.setCursor({ line: cursor.line, ch: cursor.ch - 4 });
-		}
+			this.codeMirror.setCursor({ line: cursor.line, ch: cursor.ch - cursorPos });
+		};
 	},
 
 	//=-- Externally used -==//
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -51,6 +51,8 @@ const CodeEditor = createClass({
 				'Cmd-B'  : this.makeBold,
 				'Ctrl-I' : this.makeItalic,
 				'Cmd-I'  : this.makeItalic,
+				'Ctrl-U' : this.makeUnderline,
+				'Cmd-U'  : this.makeUnderline,
 				'Ctrl-M' : this.makeSpan,
 				'Cmd-M'  : this.makeSpan,
 				'Ctrl-/' : this.makeComment,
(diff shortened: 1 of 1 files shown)


Gazook89:
[Here is a demo page](https://markdown-it.github.io/#md3=%7B%22source%22%3A%22%2B%2BUderlined%20text%2B%2B%22%2C%22defaults%22%3A%7B%22html%22%3Afalse%2C%22xhtmlOut%22%3Afalse%2C%22breaks%22%3Afalse%2C%22langPrefix%22%3A%22language-%22%2C%22linkify%22%3Atrue%2C%22typographer%22%3Atrue%2C%22_highlight%22%3Atrue%2C%22_strict%22%3Afalse%2C%22_view%22%3A%22html%22%7D%7D) for markdown-it showing use of the plug-in markdown-it-ins

Gazook89:
Added another set of hotkeys, inspired by #732.   

###  makeNbsp: `ctrl` / `cmd` + `.`
```
&nbsp;
```
Finishes with the cursor at the end.  May look into an additional hotkey (`ctrl` + `,`) to remove as well.

### makeSpace: `shift` + `ctrl` / `cmd` + `.`
```
{{width:10% }}
```
Finishes with span selected.  Repeated use increases the percentage by 10% increments to a max of 100%.

### removeSpace: `shift` + `ctrl` / `cmd` + `,`
Reduces the above span by 10% increments until 0%, at which point the span is removed entirely.

Overall, these hotkeys should make it easier to work with wrapping text around images, if not using shape-outside etc.  While shape-outside might be the ideal method of wrapping text, it is still difficult to use and doesn't work in every situation.

Gazook89:
makeComment() was added in a previous PR/release, but in this PR I just updated it to be responsive to either the GFM editor or CSS editor (commit 3bf5d7a).  It will now comment correctly in either editor.  It also fixes a minor issue where undoing a comment left a leading and trailing space.  This would close #1303.

calculuschild:
Ok, this all seems to be working and I don't have any strong feelings on how the hotkeys work, as I rarely use them myself. We can release this into the wild and if people complain we can make tweaks.

At some point we should move all of these out into a separate file, since Hotkey scripts are becoming their own beast. That will probably also make it easier to change behavior based on the current view / renderer (i.e., `MakeDiv` would be different depending on Legacy or V3, etc.)

PR #1650, opened 2021-09-07
GMB Compatibility

Gazook89:
Looking to make it as easy as possible to bring GMB documents to Homebrewery (and back again, if there is time).  One goal is to increase user count with easy migration process, but also to help troubleshoot issues occurring in GMB using Homebrewery.

### First Commit
This commit adds `\pagebreak`, `\pagebreakNum`, and `\columnbreak` to the watch words for page and column breaks.

It does nothing further for `\pagebreakNum` (does not add page numbers).

### Further Work
- [ ] apply page numbers using \pagebreakNum
- [ ] identify baked in class names such as "cover-page"
- [ ] more...

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -30,7 +30,7 @@ const BrewRenderer = createClass({
 		if(this.props.renderer == 'legacy') {
 			pages = this.props.text.split('\\page');
 		} else {
-			pages = this.props.text.split(/^\\page$/gm);
+			pages = this.props.text.split(/^\\pagebreakNum$|^\\pagebreak$|^\\page$/gm);
 		}
 
 		return {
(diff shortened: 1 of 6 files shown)


calculuschild:
I think this needs to be discussed before spending too much more time on this PR.

For one, tracking more than one "watch word" like this is only going to complicate things unnecessarily and introduce more confusion points between users ("wait, what's the difference between page and pageBreak?"), especially since it is trying to accommodate somebody *else's* project that we have no control over. I don't think we need to try to handle any syntax other than our own. We have diverged enough with V3 that even if we match \pageBreak, etc., 90% of the brew is still going to be broken.

Second, I am not comfortable with the idea of directly incorporating GMB's syntax into our codebase. Making a conversion script to translate a GMB doc into an HB doc is one thing. Starting to "borrow" their code and formatting however starts to look suspicious.

Gazook89:
Okay, let's close this PR.   I was having similar thoughts as well but went ahead with it as a "let's see what happens" sort of thing.  

[Search and Replace](https://codemirror.net/demo/search.html) from codemirror would effectively make this a moot point, too.   Easily replace the pieces that need to be changed.

PR #1661, opened 2021-09-10
change legacy box shadows to drop-shadow

### LINKED ISSUE #1569: Grey boxes / shadows around blockquotes (Mac Preview)
As described here: https://old.reddit.com/r/homebrewery/comments/ch3v0d/psa_grey_boxesshadows_around_notes_stat_blocks_etc/ 

This issue is still a thing. 
OS: macOS 11.5.1
Chrome: 92

Seems like certain renderers like the macOS Preview App still have problems with the exported box-shadows.

Stupidly, if I replace the box-shadows with a drop-shadow filter, everything renders fine in every PDF App I have:

```css
.phb blockquote {
	/* workaround for grey boxes around blockquotes */
	box-shadow: none; 
	filter: drop-shadow(2px 4px 6px #888);
}
```

Same holds for other elements using box-shadows like elements with the class descriptive:
```css
.phb .descriptive {
	box-shadow: none;
	filter: drop-shadow(0 0 6px #faf7ea);
}
```

Gazook89:
Closes #1569 

Changes box-shadows to drop-shadows in the Legacy styling.   

Ignore the note in the commit description--- this does change monster stat blocks as well.

Should be good to push through as a minor change.

### DIFF
--- a/themes/5ePhbLegacy.style.less
+++ b/themes/5ePhbLegacy.style.less
@@ -188,7 +188,7 @@ body {
 		border-width        : 11px;
 		border-image        : @noteBorderImage 11;
 		border-image-outset : 9px 0px;
-		box-shadow          : 1px 4px 14px #888;
+		filter              : drop-shadow(1px 4px 6px #888);
 		p, ul{
 			font-size   : 0.352cm;
 			line-height : 1.1em;
@@ -424,7 +424,7 @@ body {
 	border-width        : 7px;
 	border-image        : @descriptiveBoxImage 12 stretch;
 	border-image-outset : 4px;
-	box-shadow          : 0px 0px 6px #faf7ea;
+	filter              : drop-shadow(0 0 3px #faf7ea);
 	p{
 		display        : block;
 		padding-bottom : 0px;


calculuschild (commented):
After applying drop-shadow to V3, we are running into this issue from Reddit, if you remember, where using `filter` causes text boxes to be rendered as images in the resulting PDF, rather than selectable text:

https://www.reddit.com/r/homebrewery/comments/pr9k5l/issue_v3_pdf_block_render_problems/

I'm not sure what the best approach here is.

**Box-shadow** doesn't have that issue, but doesn't work in *one* PDF viewer, Apple Preview, due to a bug on their part. 

**Drop-shadow** *looks* fine, but breaks text selection in *all* PDF viewers. Not sure if this is a bug in Chrome or intentional. If intentional, I don't think it's the behavior we want.

Firefox is even worse; it seems to downsample the resulting image to the point it looks nearly unreadable:

image

calculuschild:
This actually seems to make things objectively worse. I'm also thinking we might want to go back and undo the change to Legacy as well. Selectable text and good image quality in almost every PDF viewer is preferable to just catering to the weird edge case in Mac Preview.

G-Ambatte:
I've been experimenting with a work around for this, because it doesn't seem to be going away - see the styling here:
https://homebrewery.naturalcrit.com/share/H8lmhcqi75KA

HugoLepage:
> I've been experimenting with a work around for this, because it doesn't seem to be going away - see the styling here: https://homebrewery.naturalcrit.com/share/H8lmhcqi75KA

Hey! I've tried this and it's great, but I'm still getting one problem. If more than one descriptive exists on the same page, the shading bleeds out of the borders. I've attached a screenshot of your example brew with another descriptive added so you can see what I mean.

Would you have any idea how to fix this?

Screenshot 2025-09-17 at 16 52 15

PR #1662, opened 2021-09-10
Add Page Count to the Google brew update function

### LINKED ISSUE #1660: Google Brews don't update pageCount
`pageCount` is just not included in the properties that gets sent to Google in the `updateGoogleBrew()` function.  Easy fix.

G-Ambatte:
Add Page Count to the Google brew update function to fix the User Page bug.

This will resolve #1660.


G-Ambatte:
Should resolve #1660.

PR #1693, opened 2021-09-16
Add `.pages` container div to printPage.jsx

### LINKED ISSUE #1612: even/odd footers flipped on /print vs /share pages
> https://homebrewery.naturalcrit.com/share/n_qEpe-cH__4
> 
> I have a bit of an issue. When I attempt to save my homebrewery document as a pdf, it seems to either insert or remove a phantom page, and completely messes up the page count. [The footer graphic that] had once been on the left side of the page is now on the right and vice versa. This doesn't appear while viewing the document on the Homebrewery, only when printing to PDF.

Gazook89:
Fixes #1612 where the page footer is getting reversed on alternate pages when sending to the /print/ page.  Wraps the `.page` divs in a `.pages` container so that the :nth-child selector isn't getting mixed up.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -188,7 +188,6 @@ const BrewRenderer = createClass({
 	        : null}
 
 				<Frame initialContent={this.state.initialContent}
-					head = <link href={`${this.props.renderer == 'legacy' ? '/themes/5ePhbLegacy.style.css' : '/themes/5ePhb.style.css'}`} rel='stylesheet'/>
 					style={{ width: '100%', height: '100%', visibility: this.state.visibility }}
 					contentDidMount={this.frameDidMount}>
 					<div className={'brewRenderer'}
@@ -200,17 +199,17 @@ const BrewRenderer = createClass({
 							<RenderWarnings />
 							<NotificationPopup />
 						</div>
-
-						<div className='pages' ref='pages'>
-							{/* Apply CSS from Style tab and render pages from Markdown tab */}
-							{this.state.isMounted
-								&&
-								<>
-									{this.renderStyle()}
+						<link href={`${this.props.renderer == 'legacy' ? '/themes/5ePhbLegacy.style.css' : '/themes/5ePhb.style.css'}`} rel='stylesheet'/>
+						{/* Apply CSS from Style tab and render pages from Markdown tab */}
+						{this.state.isMounted
+							&&
+							<>
+								{this.renderStyle()}
+								<div className='pages' ref='pages'>
 									{this.renderPages()}
-								</>
-							}
-						</div>
+								</div>
+							</>
+						}
 					</div>
 				</Frame>
 				{this.renderPageInfo()}
--- a/client/homebrew/pages/printPage/printPage.jsx
+++ b/client/homebrew/pages/printPage/printPage.jsx
@@ -35,6 +35,11 @@ const PrintPage = createClass({
 		if(this.props.query.dialog) window.print();
 	},
 
+	renderStyle : function() {
+		if(!this.props.brew.style) return;
+		return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: `<style> ${this.props.brew.style} </style>` }} />;
+	},
+
 	renderPages : function(){
 		if(this.props.brew.renderer == 'legacy') {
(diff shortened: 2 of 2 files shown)


ericscheid (approved):
LGTM

calculuschild:
Can we make the `brewrenderer.jsx` and `print.jsx` share the same structure?  That is, arrange the three blocks `style`, `theme` (the "link" to v3 or legacy CSS), and `pages` in the same order to be *sure* the print results will be equal?  This PR superficially fixes the issue but there are still differences that I think might bite us later.

Gazook89:
Can we push this one through though so the issue is "fixed" for users in the meantime? The difference between this fix and the one you are suggesting shouldn't make a difference to at least most users right now so I don't see a drawback in doing the quick fix now and changing it later. 

I'm mostly worried that if we are waiting on me it might be awhile, as I'm just learning the stuff as I go. Of course, if someone else jumps to it then there isn't anything to worry about.

ericscheid:
I'm also in favour of bug fixes having priority over refactoring. 
We don't want to is introduce a new bug while fixing an existing bug.

See, when I look at printPage.jsx, I can't even see where the `<head/>` stuff is happening — is the output of printPage.jsx being used by another source file that is actually doing the page creation? There are many other small differences too, like the splitting of brew text into pages happening inside `getInitialState` vs happening inside `renderPages` (and comingled with the rendering).

Yes, a refactor is needed. Separately.

calculuschild:
> We don't want to is introduce a new bug while fixing an existing bug.

This is precisely the reasoning for the refactor. If we are going to go rearranging things on the print page let's do it in a way that's consistent with brewrenderer.jsx to avoid accidentally breaking something else.

But also, "refactor" is a pretty heavy word for rearranging 3 lines.

Let's do it right the first time.

Gazook89:
Okay, so I have now matched the output of both printPage.jsx and brewRenderer.jsx.  

- [x] <link/theme> shows up just before the `.pages` div (meaning I had to take it out of the <head> on brewRenderer).
- [x] <style> shows up just after that, above `.pages` as well.
- [x] <style> only appears if there is actually a brew style (matches the behavior of brewRenderer):  shouldn't see 'undefined' anymore.

One concern is that I reused the renderStyle() function in both files, except in printPage.jsx there is a tiny difference: `this.props.BREW.style` rather than `this.props.style`.  I'm not sure if this is a big deal or not, but it seems like a waste of a good function. I just don't know how to adjust things so that one function can be reused in this case.

calculuschild:
I'm trying to test this now, but I can't reproduce the issue on the live site to compare. @Gazook89 Do you have an example where this is broken so I can verify that this PR is working properly?

Gazook89:
Here is a share Link:

https://homebrewery.naturalcrit.com/share/1gVB7rDeZIt1E5kqNYUk1KCZeaJCJQJB54_2v65FrNRYY

I have noted on which side of the page the "page number circle" is when viewing the Edit page.  When you go to Print page, it ends up on the other side.

Note, the behavior changes if you create a new brew and don't touch the Style editor at all (leaving the default text in there).  **In fact, just create a brand new brew.  In the text editor, add a line indicating which side of the page the page number circle is, and then in the Style editor delete the default text and just add anything**....Note that the footer flips.

calculuschild:
Ok, thank you.

So last question, with this PR, it looks like it reverses the page order again, putting the page number for P1 on the right side, where in Live the page number is on the left. (until you print).

Following the PHB, it looks like we want odd-numbered pages to have the number on the right, which matches "unstyled" brews behavior.

Gazook89:
Yeah, I think it is good to go at least until someone breaks it.

One small thing that I might consider changing but I doubt you will agree:  In the .less, replace the footer `background-image` property with a css variable like `--background-uri` and then declaring the image uri in the `:root` or elsewhere....and then doing the same anywhere we have an URI.  This is just because it is a pain in the butt to scroll past those URI's when working with the Inspector.  Having the URI's localized at the :root will pull them away from the middle of the rest of the declarations.

But that's just like my opinion man.

calculuschild:
Cool, I'll merge this then and send it out.

As for the other thing... yeah, I hear what you're saying, but I think it will end up scrambling our CSS around a lot for very little benefit. ProTip though: If you use Chrome, it automatically collapses those URIs into one line in the inspector.... just sayin'....

image

calculuschild (approved):

PR #1701, opened 2021-09-19
Initial pass at Editor Toolbar - Undo and Redo.

G-Ambatte:
Initial pass - Undo and Redo are functional. Still needs work on styling, position, and tooltips.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -1,3 +1,4 @@
+/*eslint max-lines: ["warn", {"max": 300, "skipBlankLines": true, "skipComments": true}]*/
 require('./editor.less');
 const React = require('react');
 const createClass = require('create-react-class');
@@ -198,6 +199,21 @@ const Editor = createClass({
 		}
 	},
 
+	renderEditorToolbar : function(){
+		return <div className='editorToolbar'>
+			<span className='undo'
+				onClick={this.refs.codeEditor?.undo}
+				title='Undo'>
+				<i className='fas fa-undo' />
+			</span>
+			<span className='redo'
+				onClick={this.refs.codeEditor?.redo}
+				title='Redo'>
+				<i className='fas fa-redo' />
+			</span>
+		</div>;
+	},
+
 	render : function(){
 		this.highlightCustomMarkdown();
 		return (
@@ -211,6 +227,7 @@ const Editor = createClass({
 					renderer={this.props.renderer} />
 
 				{this.renderEditor()}
+				{!this.isMeta() && this.renderEditorToolbar()}
 			</div>
 		);
 	}
--- a/client/homebrew/editor/editor.less
+++ b/client/homebrew/editor/editor.less
@@ -42,4 +42,16 @@
 		.tooltipLeft("Jump to brew page");
 	}
 
+	.editorToolbar{
+		position: absolute;
+		top: 5px;
+		left: 50%;
+		color: black;
+		font-size: 13px;
+		z-index: 9;
+		span {
+			padding: 2px 5px;
+		}
+	}
+
 }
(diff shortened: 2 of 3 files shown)


calculuschild:
Functionality seems good. Nice work.

Feedback:
- I think for now this can be placed in-line as part of the snippet bar, rather than absolutely-positioned on top. If/When we get more editor buttons we can expand it into a separate menu.
- It needs some kind of indication that an undo has become available. Perhaps "gray out" the buttons if there is no stored "redo" or "undo" state, respectively. I think `historySize()` might be a function we can use from CodeMirror to get those values.
- Buttons need some indication that they are interactable. At a minimum highlight the block behind them in darker gray like all the other buttons when hovered over. Better might be putting black borders to separate them and create visible "boxes" like we do with the snippets.
- BONUS: Would it be helpful to shrink the "history" delay? Codemirror defaults to waiting 1.25 seconds between keypresses before it will register a separate "history" event for undo/redo purposes. See `historyEventDelay` on the CodeMirror manual. I think even 0.25 or 0.5 seconds might be better behavior, since I often find myself typing large blocks of text and then undo kills a whole paragraph.

calculuschild (commented):
Functionality seems good. Nice work.

Feedback:
- I think for now this can be placed in-line as part of the snippet bar, rather than absolutely-positioned on top. If/When we get more editor buttons we can expand it into a separate menu.
- It needs some kind of indication that an undo has become available. Perhaps "gray out" the buttons if there is no stored "redo" or "undo" state, respectively. I think `historySize()` might be a function we can use from CodeMirror to get those values.
- Buttons need some indication that they are interactable. At a minimum highlight the block behind them in darker gray like all the other buttons when hovered over. Better might be putting borders on left and right to separate them and create visible "boxes" like we do with the snippets and navbar buttons.
- BONUS: Would it be helpful to shrink the "history" delay? Codemirror defaults to waiting 1.25 seconds between keypresses before it will register a separate "history" event for undo/redo purposes. See `historyEventDelay` on the CodeMirror manual. I think even 0.25 or 0.5 seconds might be better behavior, since I often find myself typing large blocks of text and then undo kills a whole paragraph.

G-Ambatte:
Turning this into a checklist:

- [ ] I think for now this can be placed in-line as part of the snippet bar, rather than absolutely-positioned on top. If/When we get more editor buttons we can expand it into a separate menu.
- [ ]  It needs some kind of indication that an undo has become available. Perhaps "gray out" the buttons if there is no stored "redo" or "undo" state, respectively. I think historySize() might be a function we can use from CodeMirror to get those values.
- [ ] Buttons need some indication that they are interactable. At a minimum highlight the block behind them in darker gray like all the other buttons when hovered over. Better might be putting borders on left and right to separate them and create visible "boxes" like we do with the snippets and navbar buttons.
- [x] BONUS: Would it be helpful to shrink the "history" delay? Codemirror defaults to waiting 1.25 seconds between keypresses before it will register a separate "history" event for undo/redo purposes. See historyEventDelay on the CodeMirror manual. I think even 0.25 or 0.5 seconds might be better behavior, since I often find myself typing large blocks of text and then undo kills a whole paragraph.

G-Ambatte:
A 250ms delay is a pretty sweet spot; I have a pretty decent typing WPM but even at top speed, I still end up with about three undo history events per line.

G-Ambatte:
Moving the Undo and Redo buttons to the snippet bar has automatically included the onHover highlighting.

G-Ambatte:
- [ ] Clear undo history after loading

G-Ambatte:
Current state of UI:
- Undo
image
- Redo
image

G-Ambatte:
Previous screenshot was from `/new`, this one is from `/edit`:

image

As you can see here, Undo is available but Redo is not.

PR #1707, opened 2021-09-21
Added heading shortcuts (H1 to H6)

RKuerten:
This can probably be improved upon but here's a suggestion for heading shortcuts, from H3 to H6 (I felt that shortcuts aren't needed for H1/H2 headers). I've been using a lot of H6 recently and I thought it would be nice to have shortcuts for it.


Gazook89:
I think it would be okay to include first and second headers...just for continuity sake. If ever we need that particular combo of hot keys for something else, we can take away h1 and h2, but I suspect that won't ever be the case.

Gazook89:
Oh also, you should include MacOS commands too (`Shift` `Cmd` `3` etc).

Although, my macbook pro does have a `control` key...but I don't think it's quite same as `ctrl`.

ericscheid:
The <kbd>control</kbd> key on MacOS is exactly the same as the <kbd>ctrl</kbd> key elsewhere.

But yes, add <kbd>Cmd</kbd> equivalents, as that is the convention we have with the other shortcuts.

ericscheid:
Side note: <kbd>Cmd</kbd>-<kbd>1</kbd>(no <kbd>shift</kbd>) is a shortcut built in to Chrome MacOS to select the 1st tab of the current window (and similar for 2,3,4,etc). There are a few other <kbd>Cmd</kbd> key shortcuts we should avoid stomping on (some are ok to override, like <kbd>Cmd</kbd>-<kbd>S</kbd> .. tis better to do our Save Brew routine, than for Chrome to prompt to save the html source).

Gazook89:
All of these have shift key as well.

RKuerten:
Added h1 and h2 commands, and Cmd version too.

> Side note: Cmd-1(no shift) is a shortcut built in to Chrome MacOS to select the 1st tab of the current window (and similar for 2,3,4,etc). There are a few other Cmd key shortcuts we should avoid stomping on (some are ok to override, like Cmd-S .. tis better to do our Save Brew routine, than for Chrome to prompt to save the html source).

This a feature of Chrome, but it works with Ctrl on Windows (probably Linux too, but can't test right now). That's why I used Ctrl+Shift. At least on Windows, using it with Shift works fine and doesn't trigger the original `Ctrl+1` command.

Issue #611, opened 2018-01-30
[Suggestion] Spell Checker


LukeUpThere:
Hey man, great work on homebrewery. Love it.

The only thing that bugs me is that I can't really type up stuff on the web platform due to my horrible spelling. A built-in spellcheck would be amazing.

ericscheid:
Duplicate #312

Gazook89:
An issue with a spellchecker is that this tool is specifically geared towards fantasy fiction writing--- documents will be riddled with red squiggly lines due to unfamiliar nouns.  A true word processor can have it's library added to, but I doubt codemirror can do this (or that HB could reliably store those saved words for a user).

5e-Cleric:
https://stackoverflow.com/questions/12343922/codemirror-with-spell-checker/66641365#66641365 might be of help, and i believe it uses chrome's spellcheck, which means you can add words to your dictionary.

dbolack-ab:
Possibly relevant.

https://gist.github.com/kofifus/4b2f79cadc871a29439d919692099406

dbolack-ab:
Marking this as blocked until Codemirror update

5e-Cleric:
#3511 linking blocking issue

5e-Cleric:
After a poll within the dev team, this feature has been decided to be not implemented.

Gazook89:
Just want to put my comments from gitter here for posterity:

My issue with right-click context menu for adding/ignoring words:

Suppose someone's context menu has additional options in it that they want. Something added by their OS that you haven't thought of, or by a browser extension. Once you get replace the default context menu with your own, you are basically saying "your cursor isn't yours anymore". Not that that is the intent, but I'm skeptical that there is such a need for something like adding a word to a library.

Alternatives:
- Just have a text area list like I mentioned before that makes it easy to add words
- selecting an unknown word could open a tooltip/toolbar above it to add the word (this could later be used to offer simple commands like _bold_, _italicize_, etc).
- selecting an uknown word could display a button in our existing toolbar/snippet bar, or flip such a button from `disabled` to `disabled=false`.

Issue #517, opened 2017-10-11
Feature Request: Spell check


Gadget2020:
Could a spell-check be added? 
I don't know how the likes of Facebook etc do it, but I currently copy and paste into (eg) Word, spell check and copy/paste back again.

ericscheid:
Duplicate #312

Issue #312, opened 2017-01-29
Add Spell Check?


dmmagic:
Is it possible to incorporate Chrome's native spell check feature into the code area on the left?

stolksdorf:
Unfortunately, [it's a hard problem](https://github.com/codemirror/CodeMirror/issues/1017).

dmmagic:
Darn. No problem. :)

PR #1672, opened 2021-09-12
WIP: Add UserInfo framework

G-Ambatte:
This PR adds the initial framework of a UserInfo MongoDB database collection.

The intention of the UserInfo collection is to store user information, such as preferences (renderer: legacy/v3) or default themes (once implemented). It is also possible that the UserInfo collection could be used to store notification or other user messaging data.

PR #1720, opened 2021-09-29
Add browser-based Spell Check

Gazook89:
Looking for options to cover spell checking, as mentioned in #611, #312, and #517. 

### Some options to address this:
- [x] Option used in this PR:  set the codeMirror options `inputStyle : contentEditable, spellcheck : true,`  [from this SO answer](https://stackoverflow.com/a/66641365).  Seems to work for both FF and Chrome.  From the CodeMirror 5 manual: 

> inputStyle: string
>    Selects the way CodeMirror handles input and focus. The core library defines the "textarea" and "contenteditable" input models. On mobile browsers, the default is "contenteditable". On desktop browsers, the default is "textarea". Support for IME and screen readers is better in the "contenteditable" model. The intention is to make it the default on modern desktop browsers in the future.

> spellcheck: boolean
>    Specifies whether or not spellcheck will be enabled on the input.

#### Minor complaint
- I think it is annoying that the words are highlighted as typos as they are being typed, but I'm not sure what can be done about that using a browser spell checker. 

----

- [ ] Use typo.js as suggested a few places, [here](https://gist.github.com/kofifus/4b2f79cadc871a29439d919692099406) but it seems this might have issues outside a few languages.

### DIFF
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -45,6 +45,8 @@ const CodeEditor = createClass({
 			lineWrapping   : this.props.wrap,
 			mode           : this.props.language, //TODO: CSS MODE DOESN'T SEEM TO LOAD PROPERLY
 			indentWithTabs : true,
+			inputStyle     : 'contenteditable',
+			spellcheck     : true,
 			tabSize        : 2,
 			extraKeys      : {
 				'Ctrl-B' : this.makeBold,


calculuschild:
It seems to function fine, but I have some thoughts after playing around:

- Spellcheck on a markdown / CSS document is going to be *filled* with detected "typos". This starts to get overwhelming pretty quickly. I couldn't see an obvious way to avoid spellchecking certain parts (div classes, etc.).
- Spellcheck probably shouldn't be on the Style Tab at all.
- Perhaps we need a toggle button to turn it on and off. Some people probably don't want annoying squiggles on everything.
- I also don't like the marking on partially-typed words but I'm not sure of any way around that.

**Edit:** We may be able to use this guy's extension as a template to make our own spellchecker. https://github.com/sparksuite/codemirror-spell-checker

We can see the logic where he determines which words to check in [here](https://github.com/sparksuite/codemirror-spell-checker/blob/master/src/js/spell-checker.js). It's not that complicated, but perhaps it's getting in over our heads.

Thoughts @jeddai , @G-Ambatte , @ericscheid ?

jeddai:
> Perhaps we need a toggle button to turn it on and off

I would be willing to argue that this should be a requirement before this goes out. I also think it should default to off, but could have its own icon in the editor menu bar.

On a related note, I'm actually about to open an issue brought to me by the DOMT team about having user editor settings that persist across documents, maybe that could be the solution for this concern.

> We may be able to use this guy's extension as a template to make our own spellchecker. https://github.com/sparksuite/codemirror-spell-checker 

That's certainly an option, but I think one of the biggest benefits of the browser spell check is the ability to add words to the dictionary. Replicating that functionality is possible, but would require some work.

G-Ambatte:
> On a related note, I'm actually about to open an issue brought to me by the DOMT team about having user editor settings that persist across documents

This was part of the UserInfo framework (#1672). The idea behind it was that some user specific information could be stored; in the immediate future, items like editor preferences, but in the longer term, it offers the potential for things such as storing Users "liked" brews, and intra-User messaging (which could potentially be useful for user notifications, for required messages like DMCA takedown processes).

However the framework is still very much WIP.

G-Ambatte:
>     * Spellcheck probably shouldn't be on the Style Tab at all.

Agreed.

>     * Perhaps we need a toggle button to turn it on and off. Some people probably don't want annoying squiggles on everything.

Absolutely, I almost never use spelchking.

>     * I also don't like the marking on partially-typed words but I'm not sure of any way around that.
> 
>     * The typo lines flash with each keystroke.

If we can debounce the checking functioning, it could potentially resolve both of these issues. However, I haven't had a chance to dig in to see if this is actually possible.

PR #1724, opened 2021-09-30
Fix QR code snippet when `brew.shareId` doesn't exist

### LINKED ISSUE #1710: Insert QR Code fails vs `/new` brew
On the `/new` brew editor page, the *Insert QR Code* snippet inserts this:

     {width:100px;mix-blend-mode:multiply}

Note the `undefined`. Entirely understandable how that happens .. but what do we do to fix?

Maybe detect the editor is on `/new` and a) use `SHARE-ID-HERE` instead of `undefined`, and maybe b) also insert `<!-- ATTENTION: You will need to insert your share id into the url above, once you do save this brew -->`.

G-Ambatte:
This PR modifies the snippet generator for the QR code to use the URL for the Homebrewery home page when a `shareId` does not currently exist.

This resolves #1710.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippets/snippets.js
+++ b/client/homebrew/editor/snippetbar/snippets/snippets.js
@@ -53,10 +53,10 @@ module.exports = [
 				name : 'QR Code',
 				icon : 'fas fa-qrcode',
 				gen  : (brew)=>{
-					return `![]` +
-							`(https://api.qrserver.com/v1/create-qr-code/?data=` +
-							`https://homebrewery.naturalcrit.com` + `${brew.shareId ? `/share/${brew.shareId}` : ''}` +
-							`&amp;size=100x100) {width:100px;mix-blend-mode:multiply}`;
+					return dedent`![]
+							(https://api.qrserver.com/v1/create-qr-code/?data=
+							https://homebrewery.naturalcrit.com${brew.shareId ? `/share/${brew.shareId}` : ''}
+							&amp;size=100x100) {width:100px;mix-blend-mode:multiply}`;
 				}
 			},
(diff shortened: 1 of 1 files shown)


calculuschild (changes requested):
Can you use `dedent` to clean up the formatting a little? (Instead of the plus signs concatenating everything)

calculuschild (changes requested):
Oh shoot. I guess this does need to stay on a single line in the output. I misread.

Can you go back to not using dedent and put the plus signs back at the end of lines as needed so the output is one single line? That's my fault. Otherwise looks fine.

G-Ambatte:
Reverted the `dedent` commit, so back to original now.

client/homebrew/editor/snippetbar/snippets/snippets.js
```diff
@@ -55,7 +55,7 @@ module.exports = [
 				gen  : (brew)=>{
 					return `![]` +
 							`(https://api.qrserver.com/v1/create-qr-code/?data=` +
-							`https://homebrewery.naturalcrit.com/share/${brew.shareId}` +
+							`https://homebrewery.naturalcrit.com` + `${brew.shareId ? `/share/${brew.shareId}` : ''}` +
```
calculuschild, line 58:
```suggestion
							`https://homebrewery.naturalcrit.com${brew.shareId ? `/share/${brew.shareId}` : ''}` +
```

calculuschild (approved):

calculuschild:
Thanks @G-Ambatte ! 🎉

G-Ambatte:
Looks like it's working fine on the staging site!

PR #1732, opened 2021-10-04
Add Share to Reddit and Copy Share URL to Share Menu

Gazook89:
This address the request on PR #1636 to simplify Share options:

a.  Removes Share to Reddit from metadata panel
b.  Adds Share to Reddit in Share Menu
c.  Adds Copy Share URL to Clipboard in Share Menu
d.  Fixes issue with sharing to reddit when no Systems selected creating a post title with empty `[ ]` at the end.  Checks to see if `systems.length > 0`.

One opportunity to improvement is to change the text of "Copy Share URL" to "copied!" when the button is clicked, to provide some visual cue to the user that something has happened successfully.  But I'm not sure yet how to do that and don't want to hold this up.

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -65,18 +65,6 @@ const MetadataEditor = createClass({
 			});
 	},
 
-	getRedditLink : function(){
-		const meta = this.props.metadata;
-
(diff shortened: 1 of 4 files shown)


Gazook89:
Further, I think it would be nice to add the brew description to the reddit post by default.  [I did that on my branch here](https://github.com/Gazook89/homebrewery/commit/ca301dc37108391b4ccb274bb2a780d1330e8180), and can submit a PR for it if desired.

PR #1759, opened 2021-10-18
Fix text wrapping in code blocks

### LINKED ISSUE #1736: CSS - `.page code` does not wrap long lines to next line.
As [evidenced in this reddit post](https://old.reddit.com/r/homebrewery/comments/q2528l/code_block_error_not_limiting_text_width/), long strings of text do not break and wrap to the next line inside `pre code` blocks.   Causes trouble when needing hyperlinks in the code block.

image

This would fix it:

```css
.page code {
    word-break : break-word;
}
```

G-Ambatte:
Add `word-break: break-word` to CSS in order to fix code wrapping.

This resolves #1736.

### DIFF
--- a/themes/5ePhb.style.less
+++ b/themes/5ePhb.style.less
@@ -512,7 +512,7 @@ body {
 		background-color : #faf7ea;
 		border-radius    : 4px;
 		white-space      : pre-wrap;
-		word-break       : break-word;
+		overflow-wrap    : break-word;
 	}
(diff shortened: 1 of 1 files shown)


calculuschild:
I think we can merge this pretty easily and get it out of the way. @G-Ambatte what do you think about my comments above?

calculuschild (approved):

PR #1760, opened 2021-10-19
Limit divider position

### LINKED ISSUE #1674: Disappearing Pane Divider (From old FAQ)
In the current reddit FAQ there is talk about what to do when the editor/preview window divider disappears entirely.  This seems like it should be an Issue that can be fixed?  Rather than using the convoluted "fix" provided in the answer. 

I haven't ever had this happen before, but just messing around it seems like the divider can be moved behind the vertical scroll bar on the preview window, which might be the cause?  

> Q: The preview window is suddenly gone, I can only see the editor side of the homebrewery (or the other way around)
A: Press CTRL+SHIFT+i (or right-click and select "Inspect") while in the homebrewery.
>
> Expand body -> main -> div class="homebrew" -> div class="editPage page" -> div class="content" -> div class="splitPane".
> 
> There you will find 3 divs: <div class="pane" [...]>, <div class="divider" [...]>, and <div class="pane" [...]>. The first (or second, depending on which side of the homebrewery went missing) <div class="pane"> has a style tag, that looks similar to this: <div class="pane" data-reactid="36" style="flex: 0 0 auto; width: 925px;">.
> 
> Change whatever stands behind width: to something smaller than your display width. Here's a screenshot of how it looks expanded: imgur.

G-Ambatte:
This PR limits the editor pane divider to the central 80% of the screen - that is, it can never be dragged beyond the final 10% of the screen width to either the left or the right of the screen.

As per the discussion on the Issue page, this resolves #1674.

### DIFF
--- a/shared/naturalcrit/splitPane/splitPane.less
+++ b/shared/naturalcrit/splitPane/splitPane.less
@@ -28,5 +28,8 @@
 				color     : #666;
 			}
 		}
+		&:hover{
+			background-color: #999;
+		}
 	}
 }


Gazook89:
Could it be set to a certain pixel amount rather than percentage, and that pixel amount is equal to the width of the divider?  That way the divider can be brought all the way to the edge of the screen, but can't be hidden.

calculuschild:
Good point. Yeah, if it's still easily clickable/draggable at the very edge of the screen that sounds preferable to me.

The only issue I foresee is someone not recognizing the draggable bar at the edge of the screen mistaking it as just part of the border. That's not likely... right?

Gazook89:
> That's not likely... right?

Right...Right...?  Who knows.  But maybe the divider can have a :hover affect that slightly darkens it to give some sort of interactive feedback that it is a moveable element?

I just think that if it can't be moved all the way over to the side, some people will fuss...particularly those who still use HB on mobile and need all the screen they can get.

G-Ambatte:
> I just think that if it can't be moved all the way over to the side, some people will fuss...particularly those who still use HB on mobile and need all the screen they can get.

I'd like to create a button that toggles the position of the divider - extreme left, extreme right, center - but that's a new feature, rather than this QoL fix, so it should be entered into it's own separate issue. I think that would really help those on small screens... My second right now is a little 4:3 monitor that was surplus at work so I could grab it for WFH due to the pandemic. Unfortunately Homebrewery only fits nicely when 2/3 of the viewable screen space is editor or preview, depending on which is the more important view space at the time.

Issue #629, opened 2018-03-04
Collapsible Sections


KippKipper:
Over the past month, I have been using Hombrewery a lot. I think it would be nice to collapse sections of my brew that I don't use so that it is easier to scroll to the 25 page or find the section I want to edit. 25 pages seem like a wall of text to scroll through to find anything. I have had to split up my brew into manageable chapters so that it was easier to edit it.

Thank you for the homebrewery, this is my feedback.

</pre></code></details>

Gazook89:
Old issue/feature request, but would like to +1 this.  I was thinking that in the editor pane it would be nice to set this up like:

\page [Page Title]

A user could collapse portions of the markup in the editor pane by maybe double clicking the \page line, and the \page line would also have a user-entered title so that it is easy to find which page they need.

ericscheid:
CodeMirror has collapsing built in by the looks of things, so this shouldn't be too complicated.

I would limit it to `\page` boundaries, i.e. not objects within pages.

https://codemirror.net/demo/folding.html

calculuschild:
Looks like we would need to edit `shared/Naturalcrit/codeEditor/codeEditor.jsx` and add:

```
foldGutter: {
                rangeFinder: function(startPos) //given a starting line of code, return null if this doesn't start a foldable block, or return line number where this foldable block ends. startPos.line is the starting line number.
            },
gutters: ["CodeMirror-linenumbers", "CodeMirror-foldgutter"]
```

ericscheid:
Oh .. integration with #1147 might prove weird - that is currently deferred to #1093 V3 Renderer because it might break brews where `\page` does not occur at line start.

Suggestion: implement #629 but only fold `\page` if at line-start .. rendering would unaffected (yay), folding is available (yay), and the fact a given `\page` doesn't fold because it isn't at line-start might encourage editorial clean-up.

jeddai:
I've got some pretty big brews as well and definitely agree it would be a helpful addition -- I'm going to start working on this!

calculuschild:
@jeddai Great! New PRs are always welcome!

jeddai:
> I would limit it to \page boundaries and maybe # Headings down to some depth, but not other objects.

@ericscheid could it make sense to include xml folding for things like `<style>` blocks?

calculuschild:
[...]

PR #1792, opened 2021-11-03
Add code folding feature for all content within a single page

jeddai:
Added the gutter definitions and css for code folding. Enabling code folding in the editor was
tricky due to how CodeMirror loads its files. At the moment, the CodeMirror code-folding code has
been copied into the fold-code.js file. Additionally, that file contains the helper registration
for the Homebrewery-specific code folding function.

Issue: #629


shared/naturalcrit/codeEditor/fold-code.js
```diff
@@ -0,0 +1,338 @@
+/* eslint-disable max-lines */
+module.exports = {
+	enableCodeFolding : function (CodeMirror) {
```
calculuschild, line 3: Ok, so all of this file except for `RegisterHomebreweryHelper()` is just copied out from the CodeMirror source, right? Just want to be sure we're looking at the right parts as we try to get the normal `require` to work.

jeddai: Yep! The contents of the `enableCodeFolding` function are from CodeMirror files directly, only the contents of the IIFEs in those files. I made adjustments to them to make them conform to the code style requirements, such as var -> let, const where applicable, and switching `function <name>() {` to `const <name> = function() {` and reordering of the declarations where necessary as a result. There are two comment lines there that explain what code came from what files.

The other function in there is one I made to handle the code folding on pages and only pages. It only matches on `\page` when it is at the beginning of the line, as discussed in the issue.

shared/naturalcrit/codeEditor/fold-code.js
```diff
@@ -0,0 +1,338 @@
...
+
+		const onFold = function(cm, from) {
+			const state = cm.state.foldGutter;
+			if(!state) return;
+			const line = from.line;
+			if(line >= state.from && line < state.to)
+				updateFoldInfo(cm, line, line + 1);
+		};
+	},
+	registerHomebreweryHelper : function(CodeMirror) {
```
jeddai, line 311: Just to add some context to this, I based this function off of the Markdown code folding helper that is packaged with CodeMirror. It will allow for folding the first line in the document to the first `\page`, then the line immediately after each following `\page` to the next `\page`. It will not allow for folding from the last `\page` to the end of the document as it is currently designed.

jeddai:
General thoughts:
* I think there are a number of improvements that can be made to this code folding functionality to make it more helpful for users in the long run
  * The code-folded widget (the small arrows that show up when code is folded) could contain a snippet of the lines being folded so users can more easily recognize what is folded
  * A button or shortcut to fold/unfold all the available code blocks
  * The folded-code state being stored in state somewhere so it will retain its state on a refresh

If any of the above seems like something that should be in the mvp for this, I'm happy to add it in this PR or if not I'll add it to the issue for later.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -74,8 +78,15 @@ const CodeEditor = createClass({
 				'Ctrl-M' : this.makeSpan,
 				'Cmd-M'  : this.makeSpan,
 				'Ctrl-/' : this.makeComment,
-				'Cmd-/'  : this.makeComment
-			}
+				'Cmd-/'  : this.makeComment,
+				'Ctrl-,' : this.toggleCodeFolded,
+				'Cmd-,'  : this.toggleCodeFolded
```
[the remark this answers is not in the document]
ericscheid: `⌘\` is available on MacOS and used by some editors for text-folding.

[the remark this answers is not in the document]
jeddai: Ah that's a good call! Thanks!

shared/naturalcrit/codeEditor/fold-code.js
calculuschild: Alright, so I think we definitely should take another look at getting this to work without having to copy and paste all those functions, otherwise this adds a whole layer of extra maintenance that we have to think about if CodeMirror has an update.

@jeddai Do you think you can change this PR to remove all of those functions and get it back as close as possible to how you *think* `require` should work?  Even if it's not functioning at that point it's at least in a state where the rest of us can try our hand debugging it.

calculuschild (changes requested):

shared/naturalcrit/codeEditor/fold-code.js
jeddai: Yep I'll make that change tonight -- I'm still working on getting it to keep the fold state on change, I have it working but it is a very sketchy solution to say the least. (timeout of 0ms re-folding all the code in the componentDidUpdate function)

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -57,6 +66,12 @@ const CodeEditor = createClass({
 		} else if(this.codeMirror?.getValue() != this.props.value) { //update editor contents if brew.text is changed from outside
 			this.codeMirror.setValue(this.props.value);
 		}
+
+		setTimeout(()=>{
+			snapshot.forEach((fold)=>{
+				this.codeMirror.foldCode(fold, { scanUp: false }, 'fold');
+			});
+		}, 0);
```
calculuschild, line 74: I agree with you, this seems like a hacky way to do this. Can you explain again your findings before you tried this method out? I know you mentioned it in Gitter but here we can have it in a note that we can refer to as we keep investigating.

jeddai: Absolutely! I did a lot of debugging around this, opening console logs in the various lifecycle hooks to catch where it ends up unfolding the code. The code unfolds on the dom after `componentDidUpdate` completes, which is seemingly why the timeout (even though it's set to 0ms) successfully refolds the code that didn't move.

An important note: removing the `this.codeMirror.on('change', (cm)=>{this.props.onChange(cm.getValue());});` event did prevent the code from being unfolded on a change during my testing, so the issue likely lies within how React interacts with CodeMirror and not in CodeMirror itself.

Whatever is unfolding the code does result in a CodeMirror `unfold` event being fired. You can use `this.codemirror.on('unfold', (cm)=>{console.log('unfold');});` to see that.

calculuschild: Are you able to run through the debugger in Chrome to locate exactly when the code unfolds? I think you are right, this sounds like an unintended consequence of something we did interacting with React.

jeddai: Yeah my next plan with it was to step over in the debugger until it finally unfolds (I haven't had much time to keep working on it in the last few days). I'm concerned this is going to be something that's not necessarily a result of how CodeMirror was implemented with React, but that it might be a limitation imposed onto CodeMirror by React itself and how it interacts with the dom. No evidence to support that, just a thought that keeps running through my head.

shared/naturalcrit/codeEditor/fold-code.js
jeddai: After adding the new .js files to the project libs code-folding now works with the `require` method!

shared/naturalcrit/codeEditor/fold-code.js
```diff
@@ -0,0 +1,28 @@
...
+						return {
+							from : CodeMirror.Pos(start.line, 0),
+							to   : CodeMirror.Pos(end, cm.getLine(end).length)
+						};
+					}
+					++end;
+					nextLine = cm.getLine(end + 1);
+				}
+
+				return null;
```
calculuschild, line 22: At this point, if nothing has been returned yet from the while loop, by definition we *must* be at the end of the document, right? So can we just return everything from the start point to the end?

```suggestion
				return {
					from : CodeMirror.Pos(start.line, 0),
					to   : CodeMirror.Pos(end, cm.getLine(end).length)
				};
```

jeddai: That would allow the section from the last `\page` to the end of the document to be folded, however while I was writing this method I wondered if that might be a pointless area to fold. I was thinking that section is probably where most people will be doing work on a document, and thus folding it would not provide any benefit. 

If we think it makes more sense for the functionality to be as consistent as possible then yes by all means we should add that.

calculuschild: Yes, I think we should make it consistent that every page is foldable.

1) We can't assume how and where users want to fold. Users might want to go back and edit somewhere in the middle of the document and close up everything else, etc.

2) If it works everywhere except one spot, users will assume it's a bug.

jeddai: Added that!

calculuschild:
So far I have been able to track the unfolding to here:

https://github.com/naturalcrit/homebrewery/blob/fb52618ce95ae25a33edeb7fea64c9d08b420288/client/homebrew/pages/editPage/editPage.jsx#L110-L119

Calling https://github.com/naturalcrit/homebrewery/blob/fb52618ce95ae25a33edeb7fea64c9d08b420288/shared/naturalcrit/codeEditor/codeEditor.jsx#L82 sends a signal up the chain which calls `handleTextChange()` in `editPage.jsx` and it's the `setState` there which is causing the unfolding. Even if `setState()` is empty and changes nothing, when React finally fires it, it undoes the folding.

calculuschild:
FOUND IT!

In `Editor.jsx` we update all of our custom syntax highlighting as the user types by clearing all the marks from the document and then re-rendering. I didn't realize until I looked at your `getsnapshotbeforeupdate` that code folds are also included in that list of "marks".

So..... we need to so something like this to make sure we aren't deleting code folds when we refresh the highlighting:

Change this line:
https://github.com/naturalcrit/homebrewery/blob/fb52618ce95ae25a33edeb7fea64c9d08b420288/client/homebrew/editor/editor.jsx#L111

to this:

```
const customHighlights = codeMirror.getAllMarks().filter((mark)=>!mark.__isFold);
```

jeddai:
> So..... we need to so something like this to make sure we aren't deleting code folds when we refresh the highlighting:

That did it! Good call, I didn't think to look for mark changes in the editor.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -67,15 +74,40 @@ const CodeEditor = createClass({
...
+					const maxLength = 50;
+					while (currentLine <= to.line && text.length <= maxLength) {
+						text += this.codeMirror.getLine(currentLine);
+						if(currentLine < to.line) {
+							text += ' ';
+						}
+						currentLine += 1;
+					}
+
+					return `\u21A4${text.substr(0, maxLength)}\u21A6`;
```
calculuschild, line 107: At this point I'm nitpicking, but:

1) Can we make the text inside another span that is *not* shadowed? It turns into a blurry mess. The little arrows on the side are fine shadowed.
2) Can we `.trim()` the text and add a single space between the arrows and the contents?
3) Not fully decided on this, but perhaps we can add elipses if the length is cut off?

image

shared/naturalcrit/codeEditor/fold-code.js
```diff
@@ -0,0 +1,31 @@
...
+						};
+					}
+					++end;
+					nextLine = cm.getLine(end + 1);
+				}
+
+				return {
+					from : CodeMirror.Pos(start.line, 0),
+					to   : CodeMirror.Pos(end, cm.getLine(end).length)
+				};
```
calculuschild, line 25: Just a simplification in the logic here (I *think* this still works)

```suggestion
				while (end < lastLineNo) {
					if(nextLine.match(matcher))
						break;

					++end;
					nextLine = cm.getLine(end + 1);
				}

				return {
					from : CodeMirror.Pos(start.line, 0),
					to   : CodeMirror.Pos(end, cm.getLine(end).length)
				};
```

calculuschild (changes requested):

shared/naturalcrit/codeEditor/fold-code.js
calculuschild: Last few little nitpicks. It does seem to work just fine, so this is just a matter of cleaning up a bit.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -67,15 +74,40 @@ const CodeEditor = createClass({
...
-			}
+				'Ctrl-B'  : this.makeBold,
+				'Cmd-B'   : this.makeBold,
+				'Ctrl-I'  : this.makeItalic,
+				'Cmd-I'   : this.makeItalic,
+				'Ctrl-M'  : this.makeSpan,
+				'Cmd-M'   : this.makeSpan,
+				'Ctrl-/'  : this.makeComment,
+				'Cmd-/'   : this.makeComment,
+				'Ctrl-\\' : this.toggleCodeFolded,
```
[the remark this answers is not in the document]
jeddai: It does work, but your cursor has to be on a line that can be folded. We can add `scanUp: true` to the fold options to have it scan up for the closest fold when users hit the shortcut if that would be preferred.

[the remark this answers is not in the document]
calculuschild: Ah... yes I was assuming it would work with the cursor anywhere inside the fold range. Let's add that change.

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: How does this look? Turns out it seemingly sets the innerhtml to the value of the widget function, so we probably can't add additional html elements inside it.

image

client/homebrew/editor/editor.jsx
```diff
@@ -108,7 +108,7 @@ const Editor = createClass({
 			const codeMirror = this.refs.codeEditor.codeMirror;
 
 			//reset custom text styles
-			const customHighlights = codeMirror.getAllMarks();
+			const customHighlights = codeMirror.getAllMarks().filter((mark)=>!mark.__isFold);
```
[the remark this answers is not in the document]
jeddai: Will be added with my next push

shared/naturalcrit/codeEditor/fold-code.js
jeddai: Will be updated with my next push

shared/naturalcrit/codeEditor/codeEditor.jsx
calculuschild: Huh. Ok, yeah this is fine I think.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -91,20 +91,24 @@ const CodeEditor = createClass({
 			},
 			foldGutter  : true,
 			foldOptions : {
+				scanUp      : true,
```
calculuschild, line 94: Everything looks good except this last bit. For some reason, Ctrl-\ still isn't quite right. It now folds any region just fine, but pressing the keys again it just folds the next block above instead of toggling back the current block. Do you have any ideas how to get the toggling to work?

jeddai: Toggling a folded section open seems to work as expected when the cursor is on the starting position of the folded segment. 

That was part of why I chose not to enable scanUp, I'll do some digging into the logic and see if it's something I can fix.

calculuschild: When press the hotkey to close, it places my cursor at the end of the section rather than at the front. Also I can't seem to click the starting position of the folded segment to put my cursor there even if i wanted, since clicking anywhere in the line just unfolds it anyway.

>That was part of why I chose not to enable scanUp 

Without scanUp though, you have to manually place the cursor into position before you can hit the hotkey, which kind of defeats the purpose of having a hotkey.

jeddai: Hmm maybe we consider if it even needs to be a hotkey? There might also be some additional variables I could set in the foldCode call that could fix these issues.

calculuschild: If you want, we can remove that particular hotkey for now and get this merged. Then if you want to take a closer look at it later it can be in a separate PR.

jeddai: Just removed the shortcut, I'll do some more research on it and add it at a later date.

calculuschild (approved):
Great work! Thanks for digging into this and getting it working smoothly!

Issue #1780, opened 2021-10-29
Clean up redundant Google Drive code


G-Ambatte:
Not an issue for this PR, but this reveals how redundant my Google Drive code was.... So many slightly different versions converting from the `brew` object and back to a Google document. So many places to forget to add a property. You had to change what... 15 lines here just to add the one property? Good opportunity for a new PR in the future.

_Originally posted by @calculuschild in https://github.com/naturalcrit/homebrewery/pull/1777#discussion_r737808293_

dbolack-ab:
Has this  been worked out?

calculuschild:
> Has this  been worked out?

Still not handled.

dbolack-ab:
Looking over api.getBrew() I'm not sure this hasn't been resolved over time?

PR #1794, opened 2021-11-03
Fix UserPage ordering of Google Brews

### LINKED ISSUE #1793: UserPage - Sorting by views is incorrectly ordered
On the User Page, sorting by view count results in a strange order that does not appear to be by view count.

As reported at https://redd.it/qlh2hn

image

G-Ambatte:
Update `listGoogleBrews` to ensure `pageCount` and `views` are integers. This will correct the incorrect ordering of views and page counts on brews stored on Google Drive.

This will resolve #1793.


calculuschild:
Looks good. Really makes me want to fix #1780 more than ever :P

Issue #1413, opened 2021-06-27
"New" page doesn't show the "not logged in" error when saving to Google


calculuschild:
This is one of the more common issues on the Subreddit. If you are signed in with Google, the Homebrewery will by default try to save the new brew to your Google Drive. However, if your credentials are expired, the save button will just get stuck with no warning to the user (and the user can't create any new brews). See for example: https://www.reddit.com/r/homebrewery/comments/o24njc/issue_anything_generated_from_new_wont_save/

Signing in and out via google fixes the issue, but we need to communicate this to the user. The Edit page already has an error bubble that pops up telling the user to sign in to Google, but this isn't present on the New page. Should be relatively simple to copy that behavior over.

calculuschild:
Would be fixed by #1423

dbolack-ab:
Is this now fixed by #1801 ?

calculuschild:
> Is this now fixed by #1801 ?

#1801 was never completed and so was not merged.

dbolack-ab:
@calculuschild is this covered by teh recent page merging work?

calculuschild:
It will be eventually, yes.

PR #1801, opened 2021-11-04
[WIP] Create Base Page structure - EditorPage

### LINKED ISSUE #1800: Unify base page structure - EditorPage
#1423 was an attempt to unify the underlying editor page for NewPage and EditPage. However the rest of the project moved on before the PR was completed.

This issue is to record the idea - unifying the underlying page structure so that functions do not need to be modified in multiple places - so that it is not lost with the closure of #1423.

My intention is to close #1423 and open a new PR with a smaller, sleeker set of changes, implementing a minimal combined EditorPage to be called by NewPage and EditPage, which can be merged further through later PRs.
I also intend to push UserPage through an underlying ListPage, which can be used for any future pages that need to list multiple brews.

---

This issue is to track to the creation and progress of the EditorPage base page.

G-Ambatte:
This PR is to implement underlying base pages, which are then called by the unique pages that require that format (i.e. `/user` draws from a base EditorPage prototype).

This will prevent duplication of effort - when the function is updated in the underlying page, all derived pages are updated.

- [x] EditorPage base page 
- [x] NewPage implements EditorPage
     - [x] Save works
     - [x] Google Drive selection works
     - [x] Print works
- [x] EditPage implements EditorPage
     - [x] Save works
     - [x] Google Drive selection works
     - [x] Print works
- [ ] Split functions from `if(isType()){...}` to callbacks from calling page
     - [x] Print
     - [x] NavElements
          - [x] Include Print in NavElements?
     - [ ] Save/Autosave
          - [x] NewPage - error handling NYI
          - [ ] EditPage
     - [ ] Transfer to/from Google Drive/HB MongoDB - EditPage only

This will resolve #1800.


G-Ambatte:
I believe this is now functional, so it's probably time to get an extra set of eyes on it to see if I've missed anything.

client/homebrew/pages/basePages/editorPage/editorPage.jsx
```diff
@@ -0,0 +1,593 @@
...
+				trashed   : false,
+
+				title       : '',
+				description : '',
+				tags        : '',
+				published   : false,
+				authors     : [],
+				systems     : [],
+				renderer    : 'legacy'
+			},
```
calculuschild, line 53: As I've become more familiar with React, I'm noticing more weird things with props and state that we have an opportunity to fix here. This section is one we can handle better, especially since I see that the related section in `getInitialState` is commented out.

So, props we know are passed in to components. `defaultProps` are what the component uses if nothing was passed in. A problem occurs however when the passed prop is an object with nested properties, like a brew with `title`, `text`, `renderer` and such. `getDefaultProps` only applies props that are fully missing, so if we pass in a brew here that is missing a few values (say an old brew where `renderer` or `style` isn't defined), it will only see that an object called `brew` exists and it won't supply any of the missing values.

This is why in `getInitialState` we had to break out each key in the brew and double-check that the value was there, and if not, set some default value. (continued in comment below)...

client/homebrew/pages/basePages/editorPage/editorPage.jsx
```diff
@@ -0,0 +1,593 @@
...
+			googleDriveOptions : [
+				'DRIVE > HB',
+				'HB > DRIVE'
+			]
+		};
+	},
+
+	getInitialState : function() {
+		return {
+			brew                   : this.props.brew,
```
calculuschild, line 64: ...(continued) Just setting the whole `props.brew` object here is skipping that vital step of checking for `brew.renderer` and applying a default value of `legacy`.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -3,28 +3,8 @@ require('./editPage.less');
...
-const Markdown = require('naturalcrit/markdown.js');
-
-const googleDriveActive = require('../../googleDrive.png');
-const googleDriveInactive = require('../../googleDriveMono.png');
-
-const SAVE_TIMEOUT = 3000;
+const EditorPage = require('../basePages/editorPage/editorPage.jsx');
 
 const EditPage = createClass({
 	getDefaultProps : function() {
```
calculuschild, line 10: ...(continued) And one thing we can do to streamline, is to make those checks for missing values just *once* somewhere. Even without this PR, we have *multiple* separate components that all have this `getDefaultProps` with a brew, all just in case no brew got passed down. None of that needs to be there, if the highest component in the chain that ever sees the brew does the check, then we *know* all the child components will have a good brew.

G-Ambatte: I was thinking on a similar line - there's not much point to process the brew on the `NewPage` or `EditPage` if we're just going to process it all again in the base `EditorPage`. But I think the other PR we discussed, to implement a base default Brew at a higher level - either in `server.js` or `homebrew.jsx` - that is passed to all Routes will prevent the need to process the brew properties on each Page.

client/homebrew/pages/basePages/editorPage/editorPage.jsx
```diff
@@ -0,0 +1,593 @@
...
+			</Nav.section>
+
+			<Nav.section>
+				{this.renderGoogleDriveIcon()}
+				{this.renderSaveButton()}
+				{this.isNew() && <>
+					<PrintLink url={this.getPrintLink()}/>
+				</>}
+				{this.isEdit() && <>
+					<NewBrew />
```
calculuschild, line 545: Let's try to remove entirely having to pass in what page it is so we don't have to do the `isNew()` / `isEdit()` check, when anything "unique" to the different pages should be handled in that page itself.

For the navbar, we can have `newPage.jsx` and `editPage.jsx` define the unique elements of the navbar on their own page and pass them in similar to below where we display a `<SplitPane>` component and insert `<Editor>` and `<BrewRenderer>` with all the special info they need, rather than doing that inside of `splitPane.jsx`.

For example, look below at my comment in `newPage.jsx`.

client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -82,238 +73,13 @@ const NewPage = createClass({
...
-					<BrewRenderer text={this.state.brew.text} style={this.state.brew.style} renderer={this.state.brew.renderer} errors={this.state.htmlErrors}/>
-				</SplitPane>
-			</div>
-		</div>;
+		const googleDriveOptions = [
+			'Set save location to the Homebrewery?',
+			'Set save location to your personal Google Drive storage?'
+		];
+
+		return <EditorPage pageType='new' googleDriveOptions={googleDriveOptions} brew={this.props.brew}></EditorPage>;
```
calculuschild, line 82: We could do something like this:

```suggestion
return (
	<EditorPage pageType='new' googleDriveOptions={googleDriveOptions} brew={this.props.brew}>
		<NavBar>
			<PrintLink url={this.getPrintLink()}/>
		</NavBar>

	</EditorPage>
);
```

Then inside of `editorPage.jsx` (or perhaps it needs to be inside of `NavBar.jsx`) we can use `{props.children}` wherever we want the insides of `NavBar` to be inserted like so:

```
render : function(){
return <Nav.base>
	<Nav.section>
		<Nav.logo />
		<Nav.item href='/' className='homebrewLogo'>
			<div>The Homebrewery</div>
		</Nav.item>
		<Nav.item>{`v${this.state.ver}`}</Nav.item>
		<PatreonNavItem />
		{this.props.children}   //<======================= HERE
		{/*this.renderChromeWarning()*/}
	</Nav.section>
	{this.props.children}
</Nav.base>;
}
```

See https://reactjs.org/docs/composition-vs-inheritance.html for the details. Might take some finagling but this would get rid of `editorPage` having to know which type the containing page is.

calculuschild: Perhaps what we really need is more like this:

return (
	<EditorPage pageType='new' googleDriveOptions={googleDriveOptions} brew={this.props.brew}
		navElements={
			<PrintLink url={this.getPrintLink()}/>
		}
	</EditorPage>
);

And then inside `editorPage` we can  choose where to render `this.props.navElements`.

client/homebrew/pages/basePages/editorPage/editorPage.jsx
G-Ambatte: Implemented passing the printLink value as a prop (`<PrintLink url={this.props.printLink}}/>`).
A future iteration may include this in the NavElements.

client/homebrew/pages/newPage/newPage.jsx
G-Ambatte: Implemented NavElements as a prop. Currently EditPage passes NavElements while NewPage does not.

G-Ambatte:
I have split the Print and NavElements from the `isNew()`/`isEdit()` blocks inside the EditorPage functions to instead use props passed to EditorPage. At this stage, this has caused some issues with the Share links after transferring to/from Google Drive/Mongo, but I expect that this will be resolved once the final steps are completed.

Functions to split:
- [x] Print
- [x] NavElements
- [ ] Save/Autosave
- [ ] Transfer to/from Google Drive/HB MongoDB

calculuschild:
@G-Ambatte Do you need help knocking this one out?

dbolack-ab:
Is this still a going concern and given its age, does it need redone from master?

calculuschild:
> Is this still a going concern and given its age, does it need redone from master?

Yeah it still needs to happen (we have a lot of redundancy that means multiple points need to be updated when fixing an issue with the editor, and its easy to miss one.)

But yeah needs to be done fresh from Master. A lot has changed since then.

dbolack-ab:
Does this solve for #1413 as it's predecessor was supposed to?

calculuschild:
> Does this solve for #1413 as it's predecessor was supposed to?

This PR was never completed and so was not merged.

dbolack-ab:
What remains for this?

calculuschild:
> What remains for this?

Basically needs to be redone from scratch at this point.

5e-Cleric:
I have wanted to do this for a while, but every time i start it becomes really overwhelming, really fast. Too many features, and they should all be done together, in my opinion. Also lots of baggage from older react versions and that.

dbolack-ab:
> > What remains for this?
> 
> Basically needs to be redone from scratch at this point. Too many changes since this was originally made.

Would it be best, then to close this and create a feature branch to target with smaller PRs until it completes?

calculuschild:
Yeah, I will close this one and attempt with smaller PRs.

Before attempting again, let's please discuss in #1800 some specific PR targets, so we can ensure the main branch remains functional as we gradually merge each chunk.

PR #1809, opened 2021-11-07
Page numbers via js

Gazook89:
This is a fix for an issue (that was never created on Github) where the current page numbers are not reliably updated or consistent.   The current page numbers are done via CSS pseudo-elements-- because CodeMirror doesn't render* the entirety of the editor at a time the css `counter` was not counting the pages correctly.

This PR eliminates the CSS method and adds a new function in Markdown.js similar to `addCustomHighlights()`.   In fact it could likely just be combined with the `addCustomHighlights()` function but I kept it separate for now for clarity.   The function adds an absolutely-positioned `div` for each `\page` line using the CodeMirror addWidget() method.  The div can be styled with CSS using the class `editor-page-count`.

*\*I'm just assuming this is the reason.*

-----

I have tested this with longer brews and seems to work fine, *except* when copy\pasting in some content....it doesn't register that as a "change" and the numbers can get misaligned with their row until typing resumes.  I'm not sure how to fix this, and am not bothered about it.

This is pretty much done with vanilla JS, and I highly suspect that React could get this done better, so I won't be offended if we go another route.....though I do think this would be a good stop-gap between now and then.

------

Further exploration can be done to change these to links which jump to the corresponding page in the preview pane.  I'm not sure how that works across the iframe, though, and seems like a separate future PR if someone wants to explore that.  For now, this fixes an existing bug that was annoyingly close to being a good navigation feature.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -140,7 +140,7 @@ const Editor = createClass({
 						editorPageCount = editorPageCount + 1;
 					}
 				}
-			}, []);
+			});
 		}
 	},
 
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -115,11 +115,6 @@ const Editor = createClass({
 
 			let editorPageCount = 2; // start page count from page 2
 
-			// get top and bottom line numbers currently in editor viewport
-			const viewportRect = codeMirror.getWrapperElement().getBoundingClientRect();
-			const topVisibleLine = codeMirror.lineAtHeight(viewportRect.top, 'window') - 50;
-			const bottomVisibleLine = codeMirror.lineAtHeight(viewportRect.bottom, 'window') + 50;
-
 			_.forEach(this.props.brew.text.split('\n'), (line, lineNumber)=>{
 
 				//reset custom line styles
@@ -129,17 +124,15 @@ const Editor = createClass({
 				// Legacy Codemirror styling
 				if(this.props.renderer == 'legacy') {
 					if(line.includes('\\page')){
-						// add a check here to see if in current viewport
-						if(lineNumber > topVisibleLine && lineNumber < bottomVisibleLine){
-							// add back the original class 'background' but also add the new class '.pageline'
-							codeMirror.addLineClass(lineNumber, 'background', 'pageLine');
-							const pageCountElement = Object.assign(document.createElement('span'), {
-								className   : 'editor-page-count',
-								textContent : editorPageCount
-							});
-							codeMirror.setBookmark({ line: lineNumber, ch: line.length }, pageCountElement);
-						}
-						editorPageCount = editorPageCount + 1;
+						// add back the original class 'background' but also add the new class '.pageline'
+						codeMirror.addLineClass(lineNumber, 'background', 'pageLine');
+						const pageCountElement = Object.assign(document.createElement('span'), {
+							className   : 'editor-page-count',
+							textContent : editorPageCount
+						});
+						codeMirror.setBookmark({ line: lineNumber, ch: line.length }, pageCountElement);
+		
+						editorPageCount += 1;
 					};
 
 				}
@@ -147,8 +140,6 @@ const Editor = createClass({
 				// New Codemirror styling for V3 renderer
 				if(this.props.renderer == 'V3') {
 					if(line.match(/^\\page$/)){
-						// add a check here to see if in current viewport,
-						if(lineNumber > topVisibleLine && lineNumber < bottomVisibleLine){
 							// add back the original class 'background' but also add the new class '.pageline'
 							codeMirror.addLineClass(lineNumber, 'background', 'pageLine');
 							const pageCountElement = Object.assign(document.createElement('span'), {
@@ -156,8 +147,8 @@ const Editor = createClass({
 								textContent : editorPageCount
 							});
 							codeMirror.setBookmark({ line: lineNumber, ch: line.length }, pageCountElement);
-						}
-						editorPageCount = editorPageCount + 1;
+						
+						editorPageCount += 1;
 					}
 
 					if(line.match(/^\\column$/)){
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -105,22 +105,20 @@ const Editor = createClass({
 	highlightCustomMarkdown : function(){
 		if(!this.refs.codeEditor) return;
 		if(this.state.view === 'text')  {
+			console.log('change');
 			const codeMirror = this.refs.codeEditor.codeMirror;
 
 			//reset custom text styles
 			const customHighlights = codeMirror.getAllMarks().filter((mark)=>!mark.__isFold); //Don't undo code folding
-			for (let i=0;i<customHighlights.length;i++) customHighlights[i].clear();
+			for (let i=customHighlights.length - 1;i>=0;i--) customHighlights[i].clear();
 
-			// remove all widgets (page numbers in Editor)
-			const pageCountWidgets = document.getElementsByClassName('editor-page-count');
-			for (let x=pageCountWidgets.length - 1;x>=0;x--) pageCountWidgets[x].remove();
 
 			let editorPageCount = 2; // start page count from page 2
 
 			// get top and bottom line numbers currently in editor viewport
 			const viewportRect = codeMirror.getWrapperElement().getBoundingClientRect();
-			const topVisibleLine = codeMirror.lineAtHeight(viewportRect.top, 'window');
-			const bottomVisibleLine = codeMirror.lineAtHeight(viewportRect.bottom, 'window');
+			const topVisibleLine = codeMirror.lineAtHeight(viewportRect.top, 'window') - 50;
+			const bottomVisibleLine = codeMirror.lineAtHeight(viewportRect.bottom, 'window') + 50;
 
 			_.forEach(this.props.brew.text.split('\n'), (line, lineNumber)=>{
 
@@ -131,20 +129,16 @@ const Editor = createClass({
 				// Legacy Codemirror styling
 				if(this.props.renderer == 'legacy') {
 					if(line.includes('\\page')){
-						// add back the original class 'background' but also add the new class '.pageline'
-						codeMirror.addLineClass(lineNumber, 'background', 'pageLine');
-						// add a check here to see if in current viewport, then create widget if true
+						// add a check here to see if in current viewport
 						if(lineNumber > topVisibleLine && lineNumber < bottomVisibleLine){
-							const testElement = Object.assign(document.createElement('div'), {
+							// add back the original class 'background' but also add the new class '.pageline'
+							codeMirror.addLineClass(lineNumber, 'background', 'pageLine');
+							const pageCountElement = Object.assign(document.createElement('span'), {
 								className   : 'editor-page-count',
 								textContent : editorPageCount
 							});
-							codeMirror.addWidget({ line: lineNumber, ch: 0 }, testElement);
-							// addWidget() sets an inline style for 'top' and 'left' by default; the below overrides or removes the default values
-							testElement.style.top = `${parseInt(testElement.style.top) - codeMirror.defaultTextHeight()}px`;
-							testElement.style.left = null;
+							codeMirror.setBookmark({ line: lineNumber, ch: line.length }, pageCountElement);
 						}
-						// end createWidget process
 						editorPageCount = editorPageCount + 1;
 					};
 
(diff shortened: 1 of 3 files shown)


calculuschild (changes requested):

client/homebrew/editor/editor.jsx
```diff
@@ -102,6 +104,50 @@ const Editor = createClass({
 		}, 1);
 	},
 
+	addEditorPageNumbers : function(){
+		if(!this.refs.codeEditor) return;
+		if(this.state.view === 'text')  {
+			const codeMirror = this.refs.codeEditor.codeMirror;
+
+			const pageCountWidgets = document.getElementsByClassName('editor-page-count');
+			for (let x=pageCountWidgets.length - 1;x>=0;x--) pageCountWidgets[x].remove();
```
calculuschild, line 113: I asked in the CodeMirror forums here https://discuss.codemirror.net/t/best-way-to-update-widgets-live-while-typing/3691 if there's some better way to do this than scan-and-delete with each keystroke the user types. This seems to work fine and it's basically what I did in the highlightCustomSyntax, but perhaps this is an opportunity to find a more efficient way to do this.

client/homebrew/editor/editor.jsx
```diff
@@ -102,6 +104,50 @@ const Editor = createClass({
...
+			const lineNumbers = _.reduce(this.props.brew.text.split('\n'), (r, line, lineNumber)=>{
+				if(this.props.renderer == 'legacy') {
+					if(line.includes('\\page')) {
+						const testElement = Object.assign(document.createElement('div'), {
+							className   : 'editor-page-count',
+							textContent : editorPageCount
+						});
+						codeMirror.addWidget({ line: lineNumber, ch: 0 }, testElement);
+						testElement.style.top = `${parseInt(testElement.style.top) - 12.5}px`;
+						testElement.style.left = 'unset';
```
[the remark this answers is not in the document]
Gazook89: The div that is created is absolutely positioned and the `top` property is set by codemirror's addWidget method.  So each widget gets an inline-style for `top` and `left`.....which are based on the line's {line: , ch: } location in the editor.  By default, codemirror places the widget in the "next line", just below where we want it.

Since the `top` and `left` properties are set as inline styles, I think using an external stylesheet won't have any effect. 

Same with the other comment.

client/homebrew/editor/editor.jsx
```diff
@@ -102,6 +104,50 @@ const Editor = createClass({
...
+		if(!this.refs.codeEditor) return;
+		if(this.state.view === 'text')  {
+			const codeMirror = this.refs.codeEditor.codeMirror;
+
+			const pageCountWidgets = document.getElementsByClassName('editor-page-count');
+			for (let x=pageCountWidgets.length - 1;x>=0;x--) pageCountWidgets[x].remove();
+
+			let editorPageCount = 2; // start counting from page 2
+
+			const lineNumbers = _.reduce(this.props.brew.text.split('\n'), (r, line, lineNumber)=>{
```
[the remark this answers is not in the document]
Gazook89: I will check this out tomorrow, likely.  As I was working on this I was trying to figure out exactly what was going on here and it didn't quite add up for me, but I figured "why mess with success?".   So I'm glad you've said something.

client/homebrew/editor/editor.jsx
Gazook89: I also thought this seemed a bit 'overkill'.  

When widgets are created, they are created not on the actual lines, but as a pile down at the bottom of a common container div (and then positioned with 'absolute' over the editor).  So possibly the 'search' can be focused on just that div rather than the whole document for a little extra efficiency?

client/homebrew/editor/editor.jsx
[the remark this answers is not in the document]
Gazook89: This is completed with c3e24ef.

client/homebrew/editor/editor.jsx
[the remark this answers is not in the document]
calculuschild: I think this is a case that we can justify using `!important`, if it helps keep the styling all in one place instead of scattered between multiple documents.

How about in the .less file: `left : unset !important, top: transform: translateY(-12.5px)`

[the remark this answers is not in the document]
calculuschild: Huh, that's obnoxious.

Well then, I think that this is a case where we can justify using `!important`, if it helps keep the styling all in one place instead of scattered between multiple documents.

How about in the .less file: `left : unset !important, top: transform: translateY(-12.5px)`

client/homebrew/editor/editor.jsx
calculuschild: Hm... Well I have a couple responses from the CodeMirror dev here: https://discuss.codemirror.net/t/best-way-to-update-widgets-live-while-typing/3691/2

Doesn't look like there's a built-in thing, but possibly we can use some additional checks to only look at a particular range of lines. But then we need to track which widget belongs to which line so we only update the ones that are affected. And I suppose we also need to update any widgets *below* that line too, since the page count needs to be cascade down if any pages are added/removed above.

**Edit:**  Ooh, what about this? If we only worry about adding widgets if they are actually in the current scroll view? https://discuss.codemirror.net/t/get-visible-lines-with-respect-to-current-scroll/1429

We can still delete "all widgets" with each pass (to clear any offscreen widgets), but the total number of widgets we will ever be creating or destroying at a time will be limited to how many `\page` s can fit into the visible text.

calculuschild:
> In fact it could likely just be combined with the addCustomHighlights() function but I kept it separate for now for clarity. 

Yes, once the above points are addressed lets combine these together so we only have to scan through the document once to apply all of our styles and widgets.

client/homebrew/editor/editor.jsx
```diff
@@ -102,6 +104,46 @@ const Editor = createClass({
...
+							className   : 'editor-page-count',
+							textContent : editorPageCount
+						});
+						codeMirror.addWidget({ line: lineNumber, ch: 0 }, testElement);
+						testElement.style.top = `${parseInt(testElement.style.top) - 12.5}px`;
+						testElement.style.left = 'unset';
+						editorPageCount = editorPageCount + 1;
+					}
+				}
+			}, []);
```
calculuschild, line 143:
```suggestion
			});
```

Since we aren't using `reduce` anymore I don't think we need this empty array that we were collecting line numbers into.

calculuschild (changes requested):

client/homebrew/editor/editor.jsx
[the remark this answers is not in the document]
Gazook89: What about for `left` just removing that property from the inline style attribute?  So it won't even show inline, and then we don't need to unset it in the stylesheet, and in the future it's easy to change to `left` for whatever reason.

And regarding `top`, my gut feeling is to leave it inline since that is the output of addWidget().  Rather than having the top position set in a single spot, it would be set in *two* spots.  And transforms skeeve me out for some reason.  However, my 'gut' doesn't have very many strong feelings so I'm amenable to changing per your recommendation, but wanted to give a chance for further reflection on it.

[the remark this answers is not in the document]
calculuschild: > What about for left just removing that property from the inline style attribute?

Hmmmm.... that might be ok. Alright, if you fully remove `left` we can keep `top` here for now. But if possible can we change the magic 12.5px number to something that respects the actual line size in case somebody wants to zoom in and out or do something funky with the appearance of the editor? I wonder if `cm.defaultTextHeight()` or something similar would work?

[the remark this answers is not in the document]
Gazook89: removed `left` entirely, and used `cm.defaulTextHeight()` to offset position of the widget.

client/homebrew/editor/editor.jsx
calculuschild: > do we want to include the "add pageLine class" into the "check if inside the current viewport" process?

Yes if it behaves properly. The one think I'm not sure of is scrolling up and down to see more of the document, if that will trigger redraws or if it only updates while typing.

Gazook89:
```js
const viewportRect = codeMirror.getWrapperElement().getBoundingClientRect();
const topVisibleLine = codeMirror.lineAtHeight(viewportRect.top, "window");
const bottomVisibleLine = codeMirror.lineAtHeight(viewportRect.bottom, "window");

_.forEach(this.props.brew.text.split('\n'), (line, lineNumber)=>{
if(this.props.renderer == 'legacy') {
	if(line.includes('\\page')) {

		// add a check here to see if in current viewport, then create widget if true
		if(lineNumber > topVisibleLine && lineNumber < bottomVisibleLine){
			const testElement = Object.assign(document.createElement('div'), {
				className   : 'editor-page-count',
				textContent : editorPageCount
			});
			codeMirror.addWidget({ line: lineNumber, ch: 0 }, testElement);
			testElement.style.top = `${parseInt(testElement.style.top) - codeMirror.defaultTextHeight()}px`;
			testElement.style.left = null;
		}
		
		// end createWidget process

		editorPageCount = editorPageCount + 1;
	}
}
```

So this is what I have.  It "works", except as you predicted it only updates on input, not on scroll.  I guess this is the same with highlightCustomMarkdown() as well.  

I haven't committed this yet to avoid a revert if possible.  Let me know if you have thoughts on how to make it update on scroll, as this is now beyond my current knowledge of JS/react.

calculuschild: I was going to say, there's a line of code in CodeEditor.jsx that says CodeMirror.on("change") or something that fires the "redraw highlights" function. We could also add a similar line for "scroll". But "viewportchange" is probably better because it would cover resizing the window as well.

Gazook89: @calculuschild I have experimented with CodeEditor.jsx and that onChange() but I'm stuck.  I can keep 'poking around' but it's kind of like a monkey at a typewriter at this point.

I can commit & push what I wrote above if you want, and let you (or someone else) figure out that last piece, or I can make some notes for myself and come back to it later when I think I have a firmer idea on it.

calculuschild: @Gazook89 

Did you try adding a new line next to the `on("change")` one but using `on("viewportChange")` ?

```
this.codeMirror.on('viewPortChange', (cm)=>{this.props.onChange(cm.getValue());});
```

Gazook89: Ah, didn't think it was that simple.  

So, that seems to work, but....scrolling takes a performance hit; it becomes kind of choppy.  I went ahead and utilized a new-to-me concept of "debouncing", and so it looks like this now

```js
this.codeMirror.on('viewportChange', _.debounce((cm)=>{this.props.onChange(cm.getValue());},200));
```

So there is a slight delay after scrolling, but improves the performance dramatically.

If this works for you, and there are no other changes needed, I'll go ahead and start to merge this function in with highlightCustomMarkdown() assuming we still want to go that route.

calculuschild: Cool! Go ahead and merge it over.

calculuschild:
Uh oh. Something strange is going on when I adjust the editor column by moving the divider. The text on the page rearranges to fit, but the widget stays at the same height so it falls off of the `\page` line.

Gazook89:
Simple fix: don't change the divider. 


Ugh... I'll look at it some more, but might not be until tonight

Gazook89:
> Uh oh. Something strange is going on when I adjust the editor column by moving the divider. The text on the page rearranges to fit, but the widget stays at the same height so it falls off of the `\page` line.

@calculuschild -- What about this, in editor.jsx:

```js
	updateEditorSize : function() {
		if(this.refs.codeEditor) {
			let paneHeight = this.refs.main.parentNode.clientHeight;
			paneHeight -= SNIPPETBAR_HEIGHT + 1;
			this.refs.codeEditor.codeMirror.setSize(null, paneHeight);
			this.highlightCustomMarkdown();   // Add this
		}
	},
```

It doesn't *smoothly* update the position of the page numbers while dragging the divider, but it does update their position once you release the mouse button.  If we need it to reposition the numbers while still holding onto the divider, in realtime, I think it might cause performance issues...?  Not sure, though, and not sure where I would look for fixing getting that set.

calculuschild:
Was the position updating correctly before we tried the whole "only redraw the current page" thing?

Gazook89:
No, I don't believe it was updating correctly when the divider was moved.  And thinking back, I don't think it would--- it's just calculating the position from the top of the editor element in pixels.  The process is still the same, but instead of calculating distance from top of the document, it's calculating from top of viewport, right?

calculuschild:
The more I look at the documentation, the more I get the feeling the `addWidget()` function is meant for temporary popups like a color wheel or dialog box that just float on top of the page.

I wonder if [setBookmark()](https://codemirror.net/doc/manual.html#setBookmark) might be more appropriate? It seems to inject an element in-line rather than absolutely positioned which would help it scroll better.

Gazook89:
hmmmm, you are probably right.  Use addBookmark() with the `widget:` option to set a DOM node inline.  It looks like this would tie into the existing loop in highlightCustomMarkdown() that deletes all marks since bookmarks are included in that, I think.

If this is the case, I imagine it is probably best to get rid of the "in viewport only" code and just apply across the entire document to match the behavior of the custom highlighting?   Or, alternatively, start only highlighting inside the viewport;  either way, unifying the redraw of the page numbers and the markdown highlighting.

Unrelated but just FYI, I merged the pageNumbers function with highlightCustomMarkdown and merged in upstream/master (with jeddai's changes) without issue.

calculuschild:
At some point I think the "only in view" will still be nice (and we could add it to *all* the custom highlighting), but yeah, let's remove that for now because it's complicating things a lot and I feel like we're still missing something that would do the same thing in a simpler way.

Gazook89:
Okay so I removed addWidget() in favor of setBookmark().  I also left in the "change on viewport change", slightly modified.  Now the "viewport" includes 50 lines beyond it in either direction.   On scroll there is occasionally a bit of jitter, though?  Not sure if you'll see the same thing.  And at some point in testing this the on("viewportChange") was not firing consistently.  I added a simple console.log('change') near the top of the highlightCustomMarkdown() function just to see when viewportChange was firing, and it was just not registering that change.  ....possibly due to something code folding is doing.... 🤷 ?   But this seemed to disappear after I added the +/- 50 lines to the viewport.

As suggested it may be best to remove the viewport check entirely but I thought I could try to include it, see what we think, then remove later.  The "highlight block backgrounds in editor" suggestion in issue #1845 will play into this, too, if that moves forward.

calculuschild:
@Gazook89 After playing some more I still think we should just remove the viewport check. I thought it would speed things up but I think in the end it's too much complexity for not really any gain. Lets keep this PR simple and just solve the Page Numbering.

Gazook89:
Okay, removed all that stuff.  Double checked it still works for me, so let me know if I missed anything.

Issue #1622, opened 2021-08-30
Make tabs visible in codemirror pane


ericscheid:
It can be helpful to see if white space is a tab or spaces, (sometimes this can also affect the rendering, which might be a bug but if tabs are invisible it's also maddening to debug)

So .. add this to styles..

````
.cm-tab { 
  background: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADAAAAAMCAQAAACOs/baAAAARUlEQVR4nGJgIAG8JkXxUAcCtDWemcGR1lY4MvgzCEKY7jSBjgxBDAG09UEQzAe0AMwMHrSOAwEGRtpaMIwAAAAA//8DAG4ID9EKs6YqAAAAAElFTkSuQmCC);
  background-position: right;
  background-repeat: no-repeat;
}
````

![image](https://user-images.githubusercontent.com/2605467/131410149-db967059-9073-42e5-bcf3-13c140d55260.png)

ericscheid:
This can go into both V3 and Legacy as a non-breaking tweak.

calculuschild:
Added via #1819

ericscheid:
This was the image we used to indicate a tabbed space (effected via a `data:` uri)

[![image.png](https://files.gitter.im/5b92953ad73408ce4fa74346/2Kd6/thumb/image.png)](https://files.gitter.im/5b92953ad73408ce4fa74346/2Kd6/image.png)

It was not well received by the community. "Too much clutter".

It might work better as a 1 character wide arrow, with the rest of the ~~width being transparent~~ being a floating dot for each character space (to make clear cases of `SPACE` `SPACE` `TAB`)

Would also be good to have some kind of toggle for "show invisibles" in the snippet bar, possibly as a toggle button adjacent to the undo/redo buttons. (We could later put other editor-ui-toggle widgets there too, e.g. something to invoke find/replace).

5e-Cleric:
the toggle seems to be the most important part of this, we should probably add a button with the same icon MS Word uses, ¶ ([PILCROW U+00B6](https://www.compart.com/en/unicode/U+00B6))

ericscheid:
Not just <kbd>tab</kbd>s though .. would also be helpful to show trailing spaces that trigger line breaks.

See #3743

Mikle103:
From a discord suggestion and feedback there to add here.
[...]

Issue #1202, opened 2021-01-21
Some CodeMirror tweaks


ericscheid:
Some options available in CodeMirror for consideration:

- https://codemirror.net/demo/activeline.html - highlights the line the cursor is on
- https://codemirror.net/demo/closetag.html - auto-inserts closing tags, handy when you type `<style>` or `<div>`
- https://codemirror.net/demo/trailingspace.html - highlights those pesky md-significant trailing spaces
- https://codemirror.net/demo/theme.html - would require UI for selecting theme, so .. meh

calculuschild:
Added via #1819

Issue #1201, opened 2021-01-21
Add Search and Search/Replace in the Brew Editor?


ericscheid:
CodeMirror has built in options for doing search and search/replace .. looks pretty simple to turn on (famous last words).

https://codemirror.net/demo/search.html

lucastucious:
Add a screenshot here for suggestion on UI overhaul
![image](https://user-images.githubusercontent.com/28046122/141188794-69cab2cc-501e-4766-9a3d-03db70e66a82.png)

jeddai:
Doing some more digging that appears to be the find/replace box in the code editor Ace, while this project uses CodeMirror. 

I'm pretty sure we can re-style it and create our own find/replace box, but for the initial setup of it I'm focusing on getting search/replace functioning. Thanks for posting the screenshot!

calculuschild:
Added via #1819  (use CTRL + F for just find, CTRL + SHIFT + F for replace)

davideghz:
what's the shortcut on Mac?

PR #1819, opened 2021-11-10
Codemirror tweaks

jeddai:
**This is branched off the code folding PR, and should not be merged until it is merged.**

Added the following CodeMirror features:
* Code search/replace
* Auto-closing HTML tags
* Auto-closing curly braces (`{{ }}`)
* Active-line highlight
* Trailing space highlight
* Making tabs visible in code editor

Issues: #1201 #1202 #1622


shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -142,8 +154,12 @@ const CodeEditor = createClass({
 					return `\u21A4 ${text} \u21A6`;
 				}
 			},
-			gutters : ['CodeMirror-linenumbers', 'CodeMirror-foldgutter']
+			gutters           : ['CodeMirror-linenumbers', 'CodeMirror-foldgutter'],
+			showTrailingSpace : true,
```
calculuschild, line 158: I like the idea of this, but I'm not a fan of the angry red squiggles that are usually on misspelled words. Is it possible to just draw the trailing whitespace as dots (or arrows for tabs)? 
image Like this but only on trailing?

jeddai: Ah that's a good call! I pulled the squiggles from the code mirror tutorial for something, don't remember what that was though at this point. 

Yes I like the dots idea, I'll make that change tonight!

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -18,6 +18,16 @@ if(typeof navigator !== 'undefined'){
 	//Addons
 	require('codemirror/addon/fold/foldcode.js');
 	require('codemirror/addon/fold/foldgutter.js');
+	require('codemirror/addon/fold/xml-fold.js');
```
calculuschild, line 21: Xml folding doesn't seem to be working for me.

In any case, I kind of want to keep code folding limited to \pages at the moment. Can we remove this for now?

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -115,6 +125,8 @@ const CodeEditor = createClass({
 				'Shift-Cmd-Enter'  : this.newColumn,
 				'Ctrl-Enter'       : this.newPage,
 				'Cmd-Enter'        : this.newPage,
+				'Ctrl-Alt-F'       : this.findPersistent,
```
calculuschild, line 128: Is it possible to make persistent Find dialog the default with just `ctrl + f` ?

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -18,6 +18,16 @@ if(typeof navigator !== 'undefined'){
 	//Addons
 	require('codemirror/addon/fold/foldcode.js');
 	require('codemirror/addon/fold/foldgutter.js');
+	require('codemirror/addon/fold/xml-fold.js');
+	require('codemirror/addon/search/search.js');
+	require('codemirror/addon/search/searchcursor.js');
+	require('codemirror/addon/search/jump-to-line.js');
+	require('codemirror/addon/search/match-highlighter.js');
+	require('codemirror/addon/search/matchesonscrollbar.js');
+	require('codemirror/addon/dialog/dialog.js');
```
calculuschild, line 27: Can you clarify to me what each of these does?

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: Code folding is set to only be our page folding, this is a requirement of the auto-close html feature and not enabled. https://codemirror.net/doc/manual.html#addon_closetag

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: Yep! I'll probs need to un-bind a CodeMirror listener but definitely should be doable!

shared/naturalcrit/codeEditor/codeEditor.jsx
calculuschild: Ah ok. In that case, can we move it next to the closetag require line and maybe  just add a small comment?

jeddai: Absolutely!

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai:
```javascript
    require('codemirror/addon/search/search.js');
	require('codemirror/addon/search/searchcursor.js');
	require('codemirror/addon/search/jump-to-line.js');
	require('codemirror/addon/search/match-highlighter.js');
	require('codemirror/addon/search/matchesonscrollbar.js');
	require('codemirror/addon/dialog/dialog.js');
```
All of the above are for find/replace. 
https://codemirror.net/demo/search.html
https://codemirror.net/doc/manual.html#addon_matchesonscrollbar

```javascript
	require('codemirror/addon/fold/xml-fold.js');
	require('codemirror/addon/edit/closetag.js');
```
These are for auto-closing html tags

```javascript
	require('codemirror/addon/edit/trailingspace.js');
```
This adds a css class to trailing spaces, which allows us to customize it to our liking

```javascript
	require('codemirror/addon/selection/active-line.js');
```
This highlights the active line

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: Updated!

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: Comments added and lines re-ordered

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: `findPersistent` is now on ctrl+f, and I also bound `findPersistentPrevious` to shift+enter because using ctrl+shift+g was awful.

shared/naturalcrit/codeEditor/codeEditor.less
```diff
@@ -1,10 +1,25 @@
...
+    background: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAADAAAAAMCAQAAACOs/baAAAARUlEQVR4nGJgIAG8JkXxUAcCtDWemcGR1lY4MvgzCEKY7jSBjgxBDAG09UEQzAe0AMwMHrSOAwEGRtpaMIwAAAAA//8DAG4ID9EKs6YqAAAAAElFTkSuQmCC) no-repeat right;
+  }
+
+  .cm-trailingspace {
+    background: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAgAAAAQAgMAAABW5NbuAAAACVBMVEVHcEwAAAAAAAAWawmTAAAAA3RSTlMAPBJ6PMxpAAAACXBIWXMAAC4jAAAuIwF4pT92AAAAFUlEQVQI12NgwACcCQysASAEZGAAACMuAX06aCQUAAAAAElFTkSuQmCC) repeat-x left;
+    background-size: 7.16px 15px;
+
+    .cm-tab {
+      background: none;
+    }
```
calculuschild, line 23: This looks good, although I would prefer if tabs still show as arrows.

Do you think we could use something like `specialChars` to add a special rule just for whitespaces? https://discuss.codemirror.net/t/zero-width-spaces-and-styling-of-white-spaces/2054

Perhaps better would be to use the `overlay` system, which is what some other sites uing

jeddai: Yeah I can play around with that and see if it works better

calculuschild: FYI I pasted the wrong link to the overlay example. Fixed now.

jeddai: So I've got this working, but as it currently exists it works on all spaces, not just ones that are trailing the line. Do you have a preference between those two?

calculuschild: I think we can just apply the CSS to spaces inside the trailing whitespace. That will keep documents looking cleaner.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -115,6 +131,9 @@ const CodeEditor = createClass({
 				'Shift-Cmd-Enter'  : this.newColumn,
 				'Ctrl-Enter'       : this.newPage,
 				'Cmd-Enter'        : this.newPage,
+				'Ctrl-F'           : 'findPersistent',
+				'Cmd-F'            : 'findPersistent',
+				'Shift-Enter'      : 'findPersistentPrevious',
```
calculuschild, line 136: Shift Enter doesn't seem to do anything for me. What is the `findPersistentPrevious` search? Do we even need it?

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -16,8 +16,24 @@ if(typeof navigator !== 'undefined'){
...
+	require('codemirror/addon/search/search.js');
+	require('codemirror/addon/search/searchcursor.js');
+	require('codemirror/addon/search/jump-to-line.js');
+	require('codemirror/addon/search/match-highlighter.js');
+	require('codemirror/addon/search/matchesonscrollbar.js');
+	require('codemirror/addon/dialog/dialog.js');
+	//Trailing space highlighting
+	require('codemirror/addon/edit/trailingspace.js');
+	//Active line highlighting
+	require('codemirror/addon/selection/active-line.js');
```
calculuschild, line 32: active line is getting cleared as soon as you type. Probably related to our `highlightcustommarkdown()` function clearing out line styles? Unless that's the intended behavior.

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: You'll want to open up the find dialog, then search for something. Enter will go to the next instance, shift enter should go to the previous instance.

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: Yes I discovered that last night and hadn't committed the fix. I'm not sure what happened but I think a merge from master undid the change I made to fix that, because the change was no longer in my codebase.

shared/naturalcrit/codeEditor/codeEditor.less
jeddai: So this does not appear to be viable. We effectively have two options from what I can tell:
1. We keep the CSS here, even though it is admittedly a bit odd.
2. We introduce the specialChars setup to add in the dots on trailing spaces, BUT we accept that there will be random dots littered throughout documents in places we wouldn't expect them. The part that finds/replaces things that match the specialChars regex apparently happens _after_ it has been partially rendered, thus causing certain spaces to have dots in them even though they aren't at the end of the line.

Example: 
With the regex `/ +$/`, all trailing spaces will have `•` in them. However, `This is a **test**` will show up as `This is a•**test**`, because `**test**` being bolded is causing it to be rendered in a different html element, thus meaning that the space after `a` is _technically_ at the end of the line, even though it's not when you look at the editor.

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: Fixed!

shared/naturalcrit/codeEditor/codeEditor.less
calculuschild: What I meant is we handle the display with the CSS rules. So, rather than `/ +$/`, we can just use `/ /` to capture *all* spaces. Then, in the CSS, we use

```
.cm-trailingspace {
    .cm-space {
      background: DOTS;
    }
}
```

So *every* space is a `specialChar`, but only ones inside `.cm-trailingspace` will have the display changed.

jeddai: 🤦

jeddai: Yeah that worked lol

calculuschild:
Nice. This looks good. Ready to merge I think!

jeddai:
Yay!

Issue #1838, opened 2021-11-18
Handle expanded brew metadata storage (tags, thumbnails, etc)


calculuschild:
Storing metadata for brews works just fine if we are strictly using MongoDB.

We start running into issues with alternative storage sources however (e.g. Google Drive) because each system has varying limitations on the types and quantity of metadata that can be stored for a given document. Google Drive, for instance, allows 100 custom properties per file, but [only 30 can come from a single app](https://developers.google.com/drive/api/v3/properties), *and* they are each limited to 124 bytes, including the property name. This means longer properties such as thumbnail image links, long strings of tags, etc. simply can't be stored on Google Drive using the native properties. We will likely run into similar issues if we ever support GitHub, etc. as further storage options. See https://github.com/naturalcrit/homebrewery/pull/1777#issuecomment-960377433

We have two possible solutions as I see it:

1) We've had success with for storing the "style" tab for each brew, by simply embedding additional properties directly into the text of the file in some sort of header structure. I.e., the entire brew contents and metadata will be concatenated into one large string. This does mean that those particular properties will not be "indexable" or "searchable" directly from the database, but if users are hosting files in Google Drive we can't search all of Google Drive by property either. What does this mean for future "sharing" capabilities? Do we ever want to be able to hop onto Homebrewery and search for other published brews by tag? If so, we probably need to be satisfied that Google Brews will never be searchable in this way and users will be forced to transfer brews to our HB database storage if they want to have their brews publicly searchable.

2) We create brew "stubs" in the HB database alongside our other brews (suggested by @ericscheid at some point). Google Brew contents are saved to Google exactly as always, but we also store a "stub" in our MongoDB database that contains only a link to the Google Drive file, plus that extended metadata. This means it *will* be possible to make "published" Google Drive documents publicly searchable via our database. This also means sharding the storage of each Google document into multiple locations and neither one has the complete document information. A user who downloads their brew from Google will not find any of the related metadata attached. Maybe this is fine. We also run the risk of orphaned "stubs" if the Google Drive file is deleted by the user via the Google Drive interface, since Hombrewery wouldn't be notified of the deletion to update the stub appropriately.

One solution is more "unified" (make everything self-contained within the document itself), one is more functional (documents should just be the brew contents themselves, any other metadata is bonus that the HB tracks).

There are several issues that can't really be solved until we implement a solution here, but we need to carefully choose how we do this. Once we implement a storage strategy it becomes very hard to undo it later without losing data.

[...]

PR #1841, opened 2021-11-19
Adding metadata block to brew text for Google brews

jeddai:
Update brew text merge and split functions to include metadata

#1838


jeddai:
In talking with @G-Ambatte we are thinking the stub document is better served to be implemented in their PR.

server.js
```diff
@@ -10,7 +12,7 @@ const serveCompressedStaticAssets = require('./server/static-assets.mv.js');
 const sanitizeFilename = require('sanitize-filename');
 const asyncHandler = require('express-async-handler');
 
-const brewAccessTypes = ['edit', 'share', 'raw'];
+const brewAccessTypes = ['edit', 'share', 'raw', 'view'];
```
calculuschild, line 15: What is the `view` type meant to do? I don't see `raw` being used anywhere now.

server/utils/brew.js
```diff
@@ -0,0 +1,41 @@
...
+		}
+		if(metadata) {
+			let metadata = {
+				title       : brew.title,
+				description : brew.description,
+				tags        : brew.tags,
+				systems     : brew.systems,
+				renderer    : brew.renderer,
+				authors     : brew.authors,
+				published   : brew.published
```
calculuschild, line 20: Let's also remove `authors` from this. I don't want to have some accidental backdoor for people to manually add other people as authors to brews by just typing into this field. Or accidentally deleting themselves as an author and "losing" their brew somehow.

jeddai: Ah good call -- authors can be added to the `fullMetadata` section below but I need to add a section to `splitTextStyleAndMetadata` that deletes the values that shouldn't be set on the brew from loaded documents. Something like 

```javascript
if(brew.text.startsWith('```metadata')) {
  const index = brew.text.indexOf('```\n\n');
  const metadata = brew.text.slice(12, index - 1);

  const deletionList = ['shareId', 'pageCount', 'authors', 'views', ...];
  for(let key of deletionList) {
    delete metadata[key];
  }

  Object.assign(brew, yaml.load(metadata));
  brew.text = brew.text.slice(index + 5);
}
```
That should prevent anything we don't want set on the brew from being set when the brew is loaded from an external source.

server.js
jeddai: I effectively replaced the use of `raw` with `view` since the metadata block would need to be removed and re-added to handle source/download. Raw doesn't really have a use anymore that I can think of so if it makes sense I'm happy to remove it.

calculuschild:
> In talking with @G-Ambatte we are thinking the stub document is better served to be implemented in their PR.

Yes let's just make this PR focus on the in-file metadata block.

server.js
calculuschild: Yeah if we aren't using it go ahead and remove it.

server/utils/brew.js
```diff
@@ -0,0 +1,41 @@
+const yaml = require('js-yaml');
+
+module.exports = {
+	mergeBrewText : (brew, { style, metadata, fullMetadata })=>{
```
calculuschild, line 4: Is there a particular reason this was split out into a separate file? I can see the benefit of Utils files but where this is a single function on its own that isn't used anywhere else, I don't think we need to split this out just yet.

Could we merge this back into `homebrew.api.js` for now to minimize the amount of movement in the code, and then in a future PR deal with any cleanup and refactoring?

jeddai: It's also used in `server.js` now in the download/source routes

server/homebrew.api.js
```diff
@@ -46,7 +37,7 @@ const newBrew = (req, res)=>{
 	}
 
 	brew.authors = (req.account) ? [req.account.username] : [];
-	brew.text = mergeBrewText(brew.text, brew.style);
+	brew.text = brewUtils.mergeBrewText(brew, { style: true });
```
calculuschild, line 40: Is there ever a case where `style` *isn't* included? I think we want to just default to always including the style text. Same with `metadata`. I can't think of an instance where we want to update a brew without saving its metadata into the text or download a file without the metadata (especially if we are not including sensitive info like ids).

server/homebrew.api.js
```diff
@@ -4,13 +4,36 @@ const router = require('express').Router();
...
+			`\`\`\`\n\n` +
+			`${text}`;
+	}
+	const metadata = {
+		title       : brew.title,
+		description : brew.description,
+		tags        : brew.tags,
+		systems     : brew.systems,
+		renderer    : brew.renderer
+	};
```
calculuschild, line 29: This works fine. I wonder if it looks any cleaner to use https://lodash.com/docs/#pick ?

jeddai: Since the values in the object share the name between instances yes it'd definitely be cleaner to use pick -- I'll update that.

server.js
```diff
@@ -45,8 +46,19 @@ const sanitizeBrew = (brew, full=false)=>{
...
+	if(brew.text.startsWith('```metadata')) {
+		const index = brew.text.indexOf('```\n\n');
+		const metadataSection = brew.text.slice(12, index - 1);
+		const metadata = yaml.load(metadataSection);
+		brew.title = metadata.title;
+		brew.description = metadata.description;
+		brew.tags = metadata.tags;
+		brew.systems = metadata.systems;
+		brew.renderer = metadata.renderer;
+		brew.text = brew.text.slice(index + 5);
```
calculuschild, line 60: `_.pick()` here too

jeddai: Both have been updated!

package-lock.json
```diff
@@ -1,9595 +1,8 @@
 {
   "name": "homebrewery",
   "version": "3.0.4",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
calculuschild, line 4: Do you have an old version of NPM? I think since v7 lockfileVersion defaults to 2 when it generates package-lock.json. Just a different format for the same thing I think but make sure you're updated in case it creates accidental conflicts somewhere.

calculuschild:
Just tested this a bit and it seems to work pretty well!

- [ ] Metadata saves to both Mongo Database and Google Drive
- [ ] Changing the metadata block by editing the document / Google Drive file externally takes priority over the old method of metadata storage, but both will be updated on the next save.
- [ ] Fully deleting the metadata block still loads fine (and instead uses the last values saved to file properties)
- [ ] Swapping a file between Google and Mongo preserves metadata
- [ ] No style tab / with style tab CSS blocks both work.

I think we can merge this once the conflicts are resolved, then start into some of the other features that needed more room for file metadata properties!

PR #1849, opened 2021-11-21
Change conditions for Google error

### LINKED ISSUE #1848: Google Drive Error Handling
I couldn't find another issue that addressed the frequent 'Save to Google Drive' errors and resulting Github Issues; if I missed it, let me know.

Right now it seems we only have logic to give a descriptive error message for '403' errors, and nothing for '503' or the less common '400'.  And we have still received github issues for '403's, so I think possibly the error handler isn't configured correctly?   Here is the relevant set of lines from the editPage:

https://github.com/naturalcrit/homebrewery/blob/0d685acfca27a68cd8384b350505e77eb89d7e32/client/homebrew/pages/editPage/editPage.jsx#L352-L359

Should `.reason == 'insufficientPermissions'` instead be `.reason == 'insufficientFilePermissions'`?  I am looking [at this error code description](https://developers.google.com/drive/api/v3/handle-errors#resolve_a_403_error_the_user_does_not_have_sufficient_permissions_for_file_fileid) on the Google Developers site.

And to capture the other error codes, possibly just remove the need for a matching `.reason` entirely, and then go based strictly off the code number (inserting '503' and '400' into the first check)?  That is, if we are okay with the same error message and call to action for all three errors:

```js
if(this.state.errors.status == ('403' || '503' || '400')){
```

Thoughts?  This is something I don't know how to test on my local install, just fyi.

Gazook89:
Resolves #1848 by including all google related errors:

```js
if(this.state.errors.url.match(/^\/api\/.*Google.*$/m)){
```

And makes a small change to the resolution option...rather than saying "Sign in", it offers "Go to Log In Page".  I feel this removes some ambiguity about what the button is doing.

I *haven't* been able to test this besides just checking the regex works.

### DIFF
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -349,7 +349,7 @@ const EditPage = createClass({
 				</Nav.item>;
 			}
 
-			if(this.state.errors.status == '403' && this.state.errors.response.body.errors[0].reason == 'insufficientPermissions'){
+			if(this.state.errors.url.match(/^\/api\/.*Google.*$/m)){
 				return <Nav.item className='save error' icon='fas fa-exclamation-triangle'>
 					Oops!
 					<div className='errorContainer' onClick={this.clearErrors}>
@@ -360,7 +360,7 @@ const EditPage = createClass({
 						<a target='_blank' rel='noopener noreferrer'
 							href={`https://www.naturalcrit.com/login?redirect=${this.state.url}`}>
 							<div className='confirm'>
-								Sign In
+								Go to Log In Page
 							</div>
 						</a>
 						<div className='deny'>
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -360,7 +360,7 @@ const EditPage = createClass({
 						<a target='_blank' rel='noopener noreferrer'
 							href={`https://www.naturalcrit.com/login?redirect=${this.state.url}`}>
 							<div className='confirm'>
-								Go to Log In Page
+								Sign In
 							</div>
 						</a>
 						<div className='deny'>
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -349,7 +349,7 @@ const EditPage = createClass({
 				</Nav.item>;
 			}
 
-			if(this.state.errors.url.match(/^\/api\/.*Google.*$/m)){
+			if(this.state.errors.response.req.url.match(/^\/api\/.*Google.*$/m)){
 				return <Nav.item className='save error' icon='fas fa-exclamation-triangle'>
 					Oops!
 					<div className='errorContainer' onClick={this.clearErrors}>
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -354,9 +354,9 @@ const EditPage = createClass({
 					Oops!
 					<div className='errorContainer' onClick={this.clearErrors}>
 					Looks like your Google credentials have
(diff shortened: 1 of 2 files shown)


client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -237,7 +237,7 @@ const NewPage = createClass({
 						<a target='_blank' rel='noopener noreferrer'
 							href={`https://www.naturalcrit.com/login?redirect=${this.state.url}`}>
 							<div className='confirm'>
-								Sign In
+							Go to Log In Page
```
calculuschild, line 240: Undo this for now. It doesn't fit in the button.

client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -226,7 +226,7 @@ const NewPage = createClass({
 				</Nav.item>;
 			}
 
-			if(this.state.errors.status == '403' && this.state.errors.response.body.errors[0].reason == 'insufficientPermissions'){
+			if(this.state.errors.url.match(/^\/api\/.*Google.*$/m)){
```
calculuschild, line 229:
```suggestion
			if(this.state.errors.response.req.url.match(/^\/api\/.*Google.*$/m)){
```

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -349,7 +349,7 @@ const EditPage = createClass({
 				</Nav.item>;
 			}
 
-			if(this.state.errors.status == '403' && this.state.errors.response.body.errors[0].reason == 'insufficientPermissions'){
+			if(this.state.errors.url.match(/^\/api\/.*Google.*$/m)){
```
calculuschild, line 352:
```suggestion
			if(this.state.errors.response.req.url.match(/^\/api\/.*Google.*$/m)){
```

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -360,7 +360,7 @@ const EditPage = createClass({
 						<a target='_blank' rel='noopener noreferrer'
 							href={`https://www.naturalcrit.com/login?redirect=${this.state.url}`}>
 							<div className='confirm'>
-								Sign In
+								Go to Log In Page
```
calculuschild, line 363: Undo this for now. It doesn't fit in the button.

client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -226,7 +226,7 @@ const NewPage = createClass({
...
 				return <Nav.item className='save error' icon='fas fa-exclamation-triangle'>
 					Oops!
 					<div className='errorContainer' onClick={this.clearErrors}>
```
calculuschild, line 232: I can't comment on the right line, but change the error text to:

```
Looks like your Google credentials have
expired! Visit our log in page to sign out,
then and sign back in with Google
and try saving again!
```

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -349,7 +349,7 @@ const EditPage = createClass({
 				</Nav.item>;
 			}
 
-			if(this.state.errors.status == '403' && this.state.errors.response.body.errors[0].reason == 'insufficientPermissions'){
+			if(this.state.errors.url.match(/^\/api\/.*Google.*$/m)){
 				return <Nav.item className='save error' icon='fas fa-exclamation-triangle'>
 					Oops!
 					<div className='errorContainer' onClick={this.clearErrors}>
```
calculuschild, line 355: I can't comment on the right line, but change the error text to:

```
Looks like your Google credentials have
expired! Visit our log in page to sign out,
then and sign back in with Google
and try saving again!
```

calculuschild (changes requested):

calculuschild (approved):

PR #1850, opened 2021-11-21
Fix `undefined` inside CSS code fence on /Source/ and /Download/

### LINKED ISSUE #1846: `undefined` css code blocks in /source
If a brew doesn't have any css added via the editor panel, the brew when viewed via `/source` will have a css code-block prepended that looks like this:

````
```css
undefined
```
````

That's a bug, easy to fix.

Also appears in the `/download` resource.

### PR DESCRIPTION
**Pull request by Gazook89**
Fixes #1846.  A single tiny change:

```js
		const brew = this.props.brew;

		if(typeof window !== 'undefined') { //Load from localStorage if in client browser
			const brewStorage  = localStorage.getItem(BREWKEY);
			const styleStorage = localStorage.getItem(STYLEKEY);
			const metaStorage = JSON.parse(localStorage.getItem(METAKEY));

			if(!brew.text || !brew.style){
				brew.text = brew.text  || (brewStorage  ?? '');
				brew.style = brew.style || (styleStorage ?? '');   //  <--- changed this from undefined to '' (empty string)
				// brew.title = metaStorage?.title || this.state.brew.title;
				// brew.description = metaStorage?.description || this.state.brew.description;
				brew.renderer = metaStorage?.renderer || brew.renderer;
			}
		}
```

This does leave the code fence there, but empty.  I thought leaving the empty code fence would be helpful in showing that there is no additional code, rather than making it uncertain if it exists, but is missing or hidden.  However, possibly in the future there would exist an option to 'hide' the contents of your Style editor from the source?  Though I imagine that "closed source" brews would hide the entire source, rather than just portions.  That's a future discussion in any case.

I am not sure why this only needed a change on the /new/ page, but it works for the tests I did.  I know some unification is going to happen so am not too bothered about it unless someone says something.

### DIFF
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -54,7 +54,7 @@ const NewPage = createClass({
 
 			if(!brew.text || !brew.style){
 				brew.text = brew.text  || (brewStorage  ?? '');
-				brew.style = brew.style || (styleStorage ?? undefined);
+				brew.style = brew.style || (styleStorage ?? '');
 				// brew.title = metaStorage?.title || this.state.brew.title;
 				// brew.description = metaStorage?.description || this.state.brew.description;
 				brew.renderer = metaStorage?.renderer || brew.renderer;


calculuschild:
I don't see that it actually changes anything as far as fixing #1846 

> So #1850 does this:
> 1. If style tab isn't touched from "new", the example styling doesn't appear, and no code fence block.
> 2. If it's changed, the change appears
> 3. If it's changed, but entirely deleted, it shows the code fence block but is empty:

This is already the behavior without this PR as far as I can tell. I think we might need to clarify with @ericscheid the steps to reproduce the issue because I can't reproduce it with or without this PR.

calculuschild:
Closing, since the issue in #1846 was due to a different cause. Turns out this PR doesn't actually change the behavior of anything since a couple lines down we turn `''` into undefined anyway.

PR #1855, opened 2021-11-22
Add highlights to div blocks in Code Editor

### LINKED ISSUE #1845: suggestion: background shading of blocks in editor
From https://www.reddit.com/r/homebrewery/comments/qum9t8/suggestion_would_love_if_the_editor_automatically/

> I sometimes find myself getting lost/confused in piles of text, especially when scrolling through a large brew. Having some sort of subtle highlight which makes tables and notes different from other text would be awesome for readability. Ideally, having different types of tables highlight with their own color would be awesome (stat blocks in yellow, notes in purple or green, tables in green or orange, etc.) but even just one color for all of them would be helpful.

image

Gazook89:
Attempts to fix #1845, raised originally on reddit, which requests that various elements in the code editor be given further visual distinction beyond just text color.  Adding a light background color behind the lines within divs/blocks was the suggestion solution.

This PR is my attempt at that.  Right now, there are still some parts of it that need work, but I think the bones are there.  It adds to the growing highlightCustomMarkdown() function, and runs a list of defined block types (currently just Note, Descriptive, and ToC) through every line containing `{{` to see if they exist....then if so, applies a matching class name.  The classes are styled in the editor.less stylesheet.  It continues to apply that class until either A) a closing `}}` occurs, or B) another block starts (`{{`).    If a block exists without the defined words, it applies a generic class "blockHighlight".

image

Current shortcomings include styles not continuing after the end of a block nested within another.  Also, the defined keywords should likely be matched at the beginning of the string (`{{note...` vs `{{red,note...`) to avoid conflicts as well as give a user some choice about what color is applied to the block in the editor.

I know in the issue #1845 this was possibly backburnered until after editor themes and other higher priority changes, so I'm totally fine with this PR being closed-- I just wanted to try it out, and get it written down.   If it works for the future, it can be reopened (or scrapped for parts).

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -9,6 +9,7 @@ const dedent = require('dedent-tabs').default;
 const CodeEditor = require('naturalcrit/codeEditor/codeEditor.jsx');
 const SnippetBar = require('./snippetbar/snippetbar.jsx');
 const MetadataEditor = require('./metadataEditor/metadataEditor.jsx');
+const { includes } = require('./snippetbar/snippets/snippets');
 
 const SNIPPETBAR_HEIGHT = 25;
 const DEFAULT_STYLE_TEXT = dedent`
@@ -113,7 +114,7 @@ const Editor = createClass({
 
 			let x = 0;
 			const blockTypes = ['note', 'descriptive', 'toc', 'monster'];
-			let blockClass;
+			let blockClass = [];
 
 			const lineNumbers = _.reduce(this.props.brew.text.split('\n'), (r, line, lineNumber)=>{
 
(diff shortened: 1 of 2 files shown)


client/homebrew/editor/editor.jsx
```diff
@@ -111,6 +111,10 @@ const Editor = createClass({
 			const customHighlights = codeMirror.getAllMarks().filter((mark)=>!mark.__isFold); //Don't undo code folding
 			for (let i=0;i<customHighlights.length;i++) customHighlights[i].clear();
 
+			let x = 0;
```
ericscheid, line 114: improve this variable name please

client/homebrew/editor/editor.jsx
```diff
@@ -162,7 +166,27 @@ const Editor = createClass({
...
+							} 
+						});
+						if(blockClass === null){ blockClass = 'blockHighlight'}
+					}
+					if(x>0){
+						codeMirror.addLineClass(lineNumber, 'background', blockClass);
+					}
+					if(line.trimLeft().startsWith('}}')){
+						x -= 1;
+						blockClass = null; // todo:  need to switch back to previous class if nested within another div
```
ericscheid, line 185: an array with push/pop is perfect for this type of tracking

Gazook89: That's exactly what I was thinking. Thanks for mentioning it so now I can be sure before working through that

client/homebrew/editor/editor.jsx
Gazook89: Ah yes, it's leftover from when the variable was much closer to the rest of the function and thus a little clearer in intent. I will fix. Was just glad I got all my console.logs out

client/homebrew/editor/editor.jsx
Gazook89: I think it can probably be refactored to something more concise, but committed a fix for nested divs (and also excluded spans at the beginning of a line being included).  Did not fix that 'x' variable yet.

Gazook89:
With latest commit:

image

Side note, this might have the added benefit of making it obvious when closing brackets are not included.

calculuschild:
I'm going to close this for now. Let's have further discussion on #1845 so we can lay out a specific set of behaviors we want this feature to have. This PR looks promising but I think we need to develop the idea a bit more first.

Issue #1907, opened 2021-12-19
Smarter code folding snippets?


ericscheid:
A suggested tweak - when pages are folded, limit the excerpted snippet/gloss to just the first heading through to the first `\n`. Otherwise it's just a wall of text.

Thus, for this brew it would be

```
↤ # Combat Maneuvers ↦
↤ # Combat Actions  ↦
↤ ## Movement ↦
↤ ## Defense ↦
↤ ## Special Targeting ↦
↤ # Footnotes on source rules ↦
↤ ### Parry Defence ↦
↤ ## Other options/wrinkles ↦
↤ ### Weapon Length and Positioning ↦
↤ ## Damage-type Maneuvers ↦
↤ ## No Quarter <small>Combat manevers - for fight... ↦
```

![image](https://user-images.githubusercontent.com/2605467/146691256-47510b85-909a-4395-9b10-00f12fa07bee.png)

ericscheid:
I'm looking at these lines...

https://github.com/naturalcrit/homebrewery/blob/aba2f58fc42efe088f9d4d721400910427d49b3b/shared/naturalcrit/codeEditor/codeEditor.jsx#L146-L162

ericscheid:
Hmm .. (given that a page might not commence with a heading) .. maybe at line 160 have something like

```
    if ( text[0] == '#' && text.indexOf('\n') )
        text = text.substring(text.indexOf('\n')-1);
```

That way it will either be an excerpt of whatever text, or will be just the heading text (including the leading `#`).

G-Ambatte:
I had a quick look at this - how do this look?


Folded:
![image](https://user-images.githubusercontent.com/19826659/192123715-69641c9a-9ed2-49cd-b685-b4e154e59958.png)

Brew Source:
```
# Combat Maneuvers
\page
# Combat Actions
\page
## Movement
\page
## Defense
\page
## Special Targeting
\page
# Footnotes on source rules
\page
### Parry Defence
\page
## Other options/wrinkles
\page
### Weapon Length and Positioning
\page
## Damage-type Maneuvers
\page
## No Quarter <small>Combat manuevers - for fighting</small>
\page





\page

This is a test page.
This is the second line.

\page

This is a test page.
This is the second line.

### Header later on


\page

```

Gazook89:
That is spot on to what I would prefer--- showing only the first header to appear on a page, unless no headers exist.  Would be eager to hear from others though.

[...]

Issue #1905, opened 2021-12-19
Active Line Highlight is more of a hinderance than anything else


JackWeighill:
The new line highlighting is far too bright and hides the cursor. Please either reduce opacity or change its color to a darker grey, preferably both.

calculuschild:
Thanks for the feedback! We are looking at changing this.

Out of curiosity, we set the color to a pretty faint blue which should already be darker than the rest of the white editor. Are you using a dark mode or something that is conflicting with this?

ericscheid:
I'm not seeing any garish colouring...

![image](https://user-images.githubusercontent.com/2605467/146691104-9397994d-1573-451f-a8d2-9dbd6421f81d.png)

JackWeighill:
> Thanks for the feedback! We are looking at changing this.
> 
> Out of curiosity, we set the color to a pretty faint blue which should already be darker than the rest of the white editor. Are you using a dark mode or something that is conflicting with this?

My mistake, I have a Stylus script that's so old I forgot that it even existed 😅

I've done a good bit of digging into Inspect Element and I cannot for the life of me find the name of the class/attribute for the colour of the Active Line Selection

calculuschild:
@JackWeighill The classname is `CodeMirror-activeline-background`.

JackWeighill:
Thanks!

PR #1913, opened 2021-12-20
Revert some codemirror tweaks, disable code folding in style tab

jeddai:
Disable code folding in style tab, disable active line highlight and whitespace visibility.

Issues: #1905, #1907


ericscheid:
Do we need to disable active line highlight now, given https://github.com/naturalcrit/homebrewery/issues/1905#issuecomment-998398811 ?

calculuschild:
@ericscheid There were some other complaints on Reddit as well.

PR #1914, opened 2021-12-20
Add discord link nav item

jeddai:
This is a permalink to the Discord of Many Things; it's something we agreed to add as a part of our partnership with them. 

Positioning is absolutely up for discussion, I just put it where I thought it made the most sense.


calculuschild:
The main navbar is already pretty tight on space (look at any /edit page for the most crowded one), so I'm hesitant to add another button there.

Perhaps better would be to add our social media links onto the home page demo document? Reddit, Github, and Discord?

jeddai:
Works for me!

PR #1929, opened 2021-12-29
Add discord, github, and reddit links to home page document

jeddai:
Re-started from the master branch to prevent including a bunch of commits that didn't need to be included.


client/homebrew/pages/homePage/welcome_msg.md
```diff
@@ -1,4 +1,28 @@
...
+}
+```
+
+<div id="header" class="wide">
 # The Homebrewery
+<div class="links" style="">
```
calculuschild, line 19: Don't need `style` if it's blank.

client/homebrew/pages/homePage/welcome_msg.md
```diff
@@ -1,4 +1,28 @@
...
+<a href="https://discord.gg/by3deKx" target="_blank" title="discord"><i class="domt"></i></a>
```
calculuschild, line 20: Is this icon given to you from DOMT? The icon i see on Discord is different (more colors on the dice).  Also, is it obvious that this is a discord link? Should we somehow indicate with a subtitle underneath what each icon is? Does DOMT want us to specify that they are not part of Homebrewery but just somebody we are cooperating with? Would it make more sense to link only to our specific channel and use the standard discord link?

I don't think we need to rely on the `<i>` syntax from FontAwesome  for our own custom icons. Let's just use a standard `image` tag.

Also, lets make sure all the icons are the same height.

calculuschild (changes requested):

client/homebrew/pages/homePage/welcome_msg.md
jeddai: > Is this icon given to you from DOMT? The icon i see on Discord is different (more colors on the dice).
I think I could get a png that has the current logo, but if we want it to be an svg then this is all they have.

> Also, is it obvious that this is a discord link? Should we somehow indicate with a subtitle underneath what each icon is?
Hmm I'm not sure on that one. I added the title attributes so if a user hovers over it it'll pop up as a tooltip but you're right, might be worth having a text blurb with details.

> Does DOMT want us to specify that they are not part of Homebrewery but just somebody we are cooperating with?
I don't think they have a preference, I'll leave that one up to you.

> Would it make more sense to link only to our specific channel and use the standard discord link?
That's the intent, but an invite link does not exist for that yet. We can update it once the channel exists.

> I don't think we need to rely on the <i> syntax from FontAwesome for our own custom icons. Let's just use a standard image tag.
Agreed if we're using the icon as an image, I'll make that change

> Also, lets make sure all the icons are the same height.
What's funny is they are, the DOMT icon just has that little bit at the bottom that hangs off and makes it look like they aren't. I'll play around with it with absolute positioning and get that looking better.

calculuschild: > I think I could get a png that has the current logo, but if we want it to be an svg then this is all they have.

I'm ok just using a PNG. The icon is small enough I don't think we need to worry about vector-level resolution.

> I don't think they have a preference, I'll leave that one up to you.

Let's add a short text at the end of the page to clarify:

*If you are looking for more 5e Homebrew resources check out r/UnearthedArcana and their list of useful resources here.   **The DiscordOfManyThings*** (however they style it) ***is another great resource to connect with fellow homebrewers for help and feedback.***

> What's funny is they are, the DOMT icon just has that little bit at the bottom that hangs off 

The DOMT one is visibly shorter on my screen:

image

calculuschild:
Oh another thought. We want to apply these changes to the `v3_preview` document as well, since that will take over as the new homepage at some point.

client/homebrew/pages/homePage/welcome_msg.md
jeddai: > I'm ok just using a PNG. The icon is small enough I don't think we need to worry about vector-level resolution.

I'll convert the svg to a png then!

> Let's add a short text at the end of the page to clarify

Okay!

> ...I think... I might prefer just a standard Discord icon. Would the DOMT people care?

That's fair, I'll ask them

jeddai: > The DOMT one is visibly shorter on my screen

I checked and it actually looks like I just didn't center it properly, whoopsy

jeddai:
Okay I'm like... 90% sure I have addressed everything. Should be good to test and play around with! The v3 doc was very full so I had to move things around the pages.

jeddai:
Ah thanks for that reversion, I missed reverting those when I committed. Just did it so I could use v3 locally again.

calculuschild (approved):

calculuschild:
Alright. I just did a bit of tweaking to the v3 text to get it to fit on one page again. Merging.

Thanks!

PR #1931, opened 2021-12-29
[NFC] Outline database connection into a separate file

AlexeySachkov:


server.js
```diff
@@ -77,16 +77,10 @@ const config = require('nconf')
...
-	{ retryWrites: false });
-mongoose.connection.on('error', (err)=>{
-	console.log('Error : Could not connect to a Mongo Database.');
-	console.log('        If you are running locally, make sure mongodb.exe is running.');
-	console.log(err);
-	throw 'Can not connect to Mongo';
-});
+// DB
+const DB = require('./server/db.js');
+// FIXME: consider implementing error hanlding for a Promise returned by connect
```
AlexeySachkov, line 82: Note: `node` already handles errors here, which looks like this:
```
$ NODE_ENV=local PORT=8888 MONGODB_URI=invalid node server.js
server on port:8888
Could not connect to a Mongo Database.
MongoParseError: Invalid connection string "invalid"
    at new ConnectionString (github.com\naturalcrit\homebrewery\node_modules\mongodb-connection-string-url\lib\index.js:68:19)
    at parseOptions (github.com\naturalcrit\homebrewery\node_modules\mongodb\lib\connection_string.js:213:17)
    at new MongoClient (github.com\naturalcrit\homebrewery\node_modules\mongodb\lib\mongo_client.js:62:63)
    at github.com\naturalcrit\homebrewery\node_modules\mongoose\lib\connection.js:779:16
    at new Promise (<anonymous>)
    at NativeConnection.Connection.openUri (github.com\naturalcrit\homebrewery\node_modules\mongoose\lib\connection.js:776:19)
    at github.com\naturalcrit\homebrewery\node_modules\mongoose\lib\index.js:332:10
    at promiseOrCallback (github.com\naturalcrit\homebrewery\node_modules\mongoose\lib\helpers\promiseOrCallback.js:10:12)
    at Mongoose._promiseOrCallback (github.com\naturalcrit\homebrewery\node_modules\mongoose\lib\index.js:1158:10)
    at Mongoose.connect (github.com\naturalcrit\homebrewery\node_modules\mongoose\lib\index.js:331:20)
If you are running locally, make sure mongodb.exe is running.
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "undefined".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

My comment is kind of asking about whether we want to have anything on top of that, like control over how error message is printed, control over actions which are done on error (second attempt to connect, for example), etc.

calculuschild:
Is there a particular reason we want the database connection in a separate file?

AlexeySachkov:
> Is there a particular reason we want the database connection in a separate file?

This PR can be viewed as a piece of code outlined from my previous #1183. As stated there, DB connectivity is moved to a separate file to be re-usable in tests. The change is inspired by naturalcrit/homebrewery@a0bc4fddf86fc00be1237f08852103527bd0cf7c.

Note: in #1183 I haven't reached the point where such reusability would actually be needed, but to me the change seems to be reasonable in terms of code cleanliness/modularity (even though I only save ~5 lines from `server.js`)

AlexeySachkov:
> Note: in #1183 I haven't reached the point where such reusability would actually be needed, but to me the change seems to be reasonable in terms of code cleanliness/modularity (even though I only save ~5 lines from server.js)

I was wrong, I actually reached that point in API tests: https://github.com/naturalcrit/homebrewery/pull/1183/commits/465f4a5236546eda38ceffd1f40e0a80b9872a1f

AlexeySachkov:
@calculuschild, do you have any further questions/comments? It would be great to have this merged, so I can proceed with further refactoring of `server.js` (#1934) with the final goal of introducing API tests for the project

server/db.js
```diff
@@ -0,0 +1,37 @@
...
+					// FIXME: do we need to pass 'error' to 'reject'?
+					return reject();
+				}
+
+				return resolve();
+			});
+
+	};
+
+	return new Promise(resolver);
```
calculuschild, line 31: In general we've been using await/async and `.catch()` rather than the traditional `resolve/reject` promise syntax. Could you change to that to keep it unified with our other code?

If you want, you can add a `.catch` here to handle the error, but we also use `express-async-handler` to make it a bit easier to send errors along to our middleware without it, if you can make that work.

calculuschild (changes requested):

server/db.js
AlexeySachkov: Refactored to use `async`/`await` in e84cd4fe8b589f687cd29746910864fe80bc57d3

calculuschild:
Looks good. I'll merge this now. Thanks for the work on it!

calculuschild (approved):

PR #1959, opened 2022-01-11
[NFC] Outline an express app into a separate module

### LINKED ISSUE #1934: `server.js` requires refactoring to be testable
This issue is to discuss commit https://github.com/naturalcrit/homebrewery/pull/1183/commits/a293f2e4e5693363c2aec712dc5e2a5555aa0ec2, which is a part of my testing initiative #1183.

The thing I would like to achieve is to be able programmatically instance an application without it consuming a port, to test it. For example, see documentation for [`supertest`](https://www.npmjs.com/package/supertest).

Note: the commit also does something with `app-module-path` and I suggest to leave that discussion for another issue/PR, because that change is not directly relevant to restructuring code to make it testable.

AlexeySachkov:
This is done in order to be able to re-use that app in API unit tests later.

This PR resolves #1934


AlexeySachkov:
Note: this is a first PR in a series, I plan to further refactor `server.js` into a several smaller modules, like:
- separate file for account middleware
- separate file for error handling middleware
- separate file for "static" pages like Home, v3_preview, FAQ, Changelog
- separate file for "dynamic" pages like New, Edit, Share, etc.

At the moment, I simply copied most of the contents from `server.js` into `app/server.js` and made it export an express app, so it can be re-used in unit testing later

calculuschild:
Ok so looks like literally all this does is make the express app into a module, right? And then server.js turns into simply a script that calls the db connection function, then starts the app?

AlexeySachkov:
> Ok so looks like literally all this does is make the express app into a module, right? And then server.js turns into simply a script that calls the db connection function, then starts the app?

Exactly, the app itself becomes re-usable, i.e. we can have a script, which tests it, we can have a script provides command line interface for an app (though it will require some additional work), etc.

calculuschild:
Seems easy enough, but making server.js into merely a wrapper around the app when server.js already is the app, makes me wonder if perhaps we could just turn server.js into a module directly? 


 Let me know if I'm missing something.

calculuschild:
So the test script would be something similar to the new `server.js` file, but instead of starting up a port it runs tests on the module.

AlexeySachkov:
> Seems easy enough, but making server.js into merely a wrapper around the app when server.js already is the app, makes me wonder if perhaps we could just turn server.js into a module directly?
> 
> Let me know if I'm missing something.

What I want to achieve is to avoid automatic listening of a port by an app in order to use it in environments where it is not needed. An example would be unit-test for API, which doesn't really need to consume a port, because everything can be emulated by existing testing frameworks. 

We can turn `server.js` into a module as well, but I would still like to have an entry point, which simply returns an express app without consuming a port for testing if we decide not to split `server.js` like I propose to do.

AlexeySachkov:
> So the test script would be something similar to the new `server.js` file, but instead of starting up a port it runs tests on the module.

Yes, it will look like:
```js
const request = require('supertest')
const app = require('server/app.js')
const DB = require('server/db.js');
const Homebrew = require('server/homebrew.model.js');

describe('/source/:id', () => {
  beforeAll(() => {
    DB.connect();
    const brew = new Homebrew.model({...});
    return brew.save();
  });

  test('able to get source of existing rew', () => {
    return request(app)
        .get('/source/shareid').send().expect(200)
        .then((response) => {
          expect(response.text).toBe('<code><pre style="white-space: pre-wrap;">This is text</pre></code>');
          expect(response.headers).toHaveProperty('content-type', 'text/html; charset=utf-8');
        });
  });

});
```

calculuschild:
Ok ok I see. Forgive me; I'm still figuring out a lot of this as I go, so I appreciate the knowledge and will probably have similar questions on your other PRs.

AlexeySachkov:
> Ok ok I see. Forgive me; I'm still figuring out a lot of this as I go, so I appreciate the knowledge and will probably have similar questions on your other PRs.

No problem, that's absolutely the same for me :)

AlexeySachkov:
One more comment before merging this: it would be great to test this on other OSes than Windows. I'm mostly sure that everything will be ok, but it won't hurt to double-check

calculuschild:
Going to mark everyone as a reviewer since I know we have some Mac and Linux people in the group.

jeddai:
I can check on both Mac and Linux unless someone else would like to!

AlexeySachkov:
> I can check on both Mac and Linux unless someone else would like to!

@jeddai, do you have any updates here?

jeddai:
Yep! I’m checking them today, very sorry, been dealing with a lot at work.

AlexeySachkov:
> Yep! I’m checking them today, very sorry, been dealing with a lot at work.

No worries, thanks a lot!

jeddai:
I did a smoke test of the application when running on Mac and Ubuntu and ran the tests on Mac and Ubuntu and everything seems good to go!

jeddai (approved):

calculuschild:
Thanks @jeddai for testing. Looks like it's all working. Merging!

Issue #1960, opened 2022-01-11
Config object is created several times through the codebase


AlexeySachkov:
We have the following piece of code copied at least three times:

https://github.com/naturalcrit/homebrewery/blob/279352377bbd533f1062af29f8328dc9a618c1ad/server.js#L74-L78

https://github.com/naturalcrit/homebrewery/blob/279352377bbd533f1062af29f8328dc9a618c1ad/server/googleActions.js#L6-L10

https://github.com/naturalcrit/homebrewery/blob/279352377bbd533f1062af29f8328dc9a618c1ad/server/token.js#L4-L8

This doesn't looks good and might lead to errors if we decide to update the way config is loaded.

What I would like to do is to load the config once in the entry point file `server.js` and then pass it as an argument to every module/function where it is used.

I'm submitting an issue do discuss such refactoring first, because it might complicate the code a bit: one extra argument will be passed around some code paths. Please let me know what do you think about the suggested approach

calculuschild:
Would your app test script also end up duplicating this since you plan to pass the config in as a parameter to the app?

G-Ambatte:
In a [recent PR on my NC fork](https://github.com/G-Ambatte/naturalcrit/tree/addLocalSupport), I had to pass the `NODE_ENV` config item from `server.js` down through multiple levels to get it to `account.actions.js`, in order for that sub-module to be able to identify the current operating environment and modify the returned cookie content to reflect. This would take that to the next logical step and pass the entire configuration, so any config item can be checked/confirmed/acted against in any sub-module.

AlexeySachkov:
> Would your app test script also end up duplicating this since you plan to pass the config in as a parameter to the app?

That's a good point, I haven't realized that I will need config there as well, because it will be mandatory for the app

> In a recent PR on my NC fork, I had to pass the NODE_ENV config item from server.js down through multiple levels to get it to account.actions.js, in order for that sub-module to be able to identify the current operating environment and modify the returned cookie content to reflect. This would take that to the next logical step and pass the entire configuration, so any config item can be checked/confirmed/acted against in any sub-module.

I did a local prototyping myself as well, but the result doesn't look fancy: there is quite a number of places where extra argument would be required together with some refactoring.

However, a bit later I realized that we could create a separate wrapper module around `nconf`, which would encapsulate all the options we pass to `nconf` there - going this way the config parameters will be stored in a single file, which can be directly included in every place its needed.

[...]

PR #1979, opened 2022-01-21
Add very basic infrastructure for HTTP testing

AlexeySachkov:
So far this is a minimal amount of changes, which are needed to start testing our server part of the app.

Note: these changes are not enough to test more complex things as add/edit brew APIs, because those require database connectivity to be present. Integrating database setup into the testing would be my next step

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,18 +33,21 @@
   "jest": {
     "modulePaths": [
       "mode_modules",
-      "shared"
-    ]
+      "shared",
+      "server"
+    ] 
   },
   "babel": {
     "presets": [
       "@babel/preset-env",
       "@babel/preset-react"
+    ],
+    "plugins": [
+      "@babel/plugin-transform-runtime"
     ]
   },
   "dependencies": {
     "@babel/core": "^7.16.7",
-    "@babel/plugin-transform-runtime": "^7.16.8",
     "@babel/preset-env": "^7.16.8",
     "@babel/preset-react": "^7.16.7",
     "body-parser": "^1.19.1",
@@ -80,8 +83,10 @@
     "vitreum": "git+https://git@github.com/calculuschild/vitreum.git"
   },
   "devDependencies": {
+    "@babel/plugin-transform-runtime": "^7.16.10",
     "eslint": "^8.7.0",
     "eslint-plugin-react": "^7.28.0",
-    "jest": "^27.4.5"
+    "jest": "^27.4.5",
+    "supertest": "^6.2.2"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,7 +49,8 @@
       "devDependencies": {
         "eslint": "^8.7.0",
         "eslint-plugin-react": "^7.28.0",
-        "jest": "^27.4.5"
+        "jest": "^27.4.5",
+        "supertest": "^6.2.2"
       },
       "engines": {
         "node": "16.11.x"
@@ -2981,6 +2982,12 @@
         "node": ">=8"
       }
     },
+    "node_modules/asap": {
+      "version": "2.0.6",
+      "resolved": "https://registry.npmjs.org/asap/-/asap-2.0.6.tgz",
+      "integrity": "sha1-5QNHYR1+aQlDIIu9r+vLwvuGbUY=",
+      "dev": true
+    },
     "node_modules/asn1.js": {
       "version": "5.4.1",
       "resolved": "https://registry.npmjs.org/asn1.js/-/asn1.js-5.4.1.tgz",
@@ -4188,9 +4195,9 @@
       "integrity": "sha1-4wOogrNCzD7oylE6eZmXNNqzriw="
     },
     "node_modules/cookiejar": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/cookiejar/-/cookiejar-2.1.2.tgz",
-      "integrity": "sha512-Mw+adcfzPxcPeI+0WlvRrr/3lGVO0bD75SxX6811cxSh1Wbxx7xZBGK1eVtDf6si8rg2lhnUjsVLMFMfbRIuwA=="
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/cookiejar/-/cookiejar-2.1.3.tgz",
+      "integrity": "sha512-JxbCBUdrfr6AQjOXrxoTvAMJO4HBTUIlBzslcJPAz+/KT8yk53fXun51u+RenNYvad/+Vc2DIz5o9UxlCDymFQ=="
     },
     "node_modules/copy-anything": {
       "version": "2.0.1",
@@ -4596,6 +4603,16 @@
       },
       "engines": {
         "node": ">=0.8.0"
+      }
+    },
+    "node_modules/dezalgo": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/dezalgo/-/dezalgo-1.0.3.tgz",
+      "integrity": "sha1-f3Qt4Gb8dIvI24IFad3c5Jvw1FY=",
+      "dev": true,
+      "dependencies": {
+        "asap": "^2.0.0",
+        "wrappy": "1"
       }
     },
     "node_modules/diff-sequences": {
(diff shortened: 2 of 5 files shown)


calculuschild:
Is this ready for review?

AlexeySachkov:
> Is this ready for review?

Yes, sure. I do plan to work more on testing, but that most likely will be through a separate PRs to keep them smaller and easier for review (plus, at the moment I don't have anything else ready).

This PR contains some very basic acceptance testing, I.e. now we won't be able to screw the server up in a way that home page won't be accessible, because it will be caught by tests.

Another finding is that our server app seems to be slow - those five tests take about 5 seconds to complete, which I think should be investigated to see if that is a test/environment problem or a real inefficiency of the app.

AlexeySachkov:
> Note: these changes are not enough to test more complex things as add/edit brew APIs, because those require database connectivity to be present. Integrating database setup into the testing would be my next step

To be precise here, my very next step will be to address #1960, because I will need config in tests to connect to database.

Regarding `package.json` conflicts: I will resolve them after I get review comments and/or approval, plus I will upload updated package.json.

calculuschild:
Ok, this seems fine to me. I do see the slowness on the tests. I wonder if it is related to the jest tests running on the "uncompiled" files, where every single test might be having to parse through the `require` tree and load all the assets. The response time on the actual server is about 11 ms on average.

AlexeySachkov:
> I do see the slowness on the tests. I wonder if it is related to the jest tests running on the "uncompiled" files?

Perhaps? The other test we have for markdown also has some `require` directives, but indeed less of them.
I had a guess that the slowness is related to `app.js` reading content of `welcom_msg`, `changelog` and other files, but even if I comment those lines, the test is still slow, so the reason is in somewhere else.

>  Is there some issue where every test might be having to parse through the require tree and load all the assets? I wouldn't think so since app.js is required at the first so they should all be using the same instance.... I'm not really sure.

I doubt that this is the case, my understanding is that every test file should only be loaded once and all tests from it are launched in a single run (even though different files seems to be processed in parallel).

> Ok, this seems fine to me so far if you want to update the packages.

Ok, let me then resolve merge conflicts and publish `package-lock.json` content. To me, those 5 seconds for tests look like nothing and I would accept anything under 5 minutes (in total).

However, we should still try to understand the root cause to be able to speed up testing: in my local workspace I've added a second test file with a single test and it takes about the same time to complete - that will eventually sum up to a significant amount of time. The fact that the second test file takes about the same time means that the slowdown is likely caused by some common phase in those tests, I will see if I would be able to do some profiling here

package.json
```diff
@@ -33,18 +33,21 @@
...
       "@babel/preset-env",
       "@babel/preset-react"
+    ],
+    "plugins": [
+      "@babel/plugin-transform-runtime"
     ]
   },
   "dependencies": {
     "@babel/core": "^7.16.7",
-    "@babel/plugin-transform-runtime": "^7.16.8",
```
AlexeySachkov, line 47: Oh, I haven't realize we already have this dependency. I will then leave it in a regular `dependencies` section, because I don't fully understand where else it is used besides testing.

calculuschild: It is used in the "build homebrewery" script.

AlexeySachkov:
> Ok, let me then resolve merge conflicts and publish package-lock.json content.

Done in 4fd085b

package.json
AlexeySachkov: Is "build homeberewery" script considered a `dev` thing?

AlexeySachkov:
> > Is there some issue where every test might be having to parse through the require tree and load all the assets? I wouldn't think so since app.js is required at the first so they should all be using the same instance.... I'm not really sure.
>
> I doubt that this is the case, my understanding is that every test file should only be loaded once and all tests from it are launched in a single run (even though different files seems to be processed in parallel).

Well, reading through [Jest blog](https://jestjs.io/blog/2016/03/11/javascript-unit-testing-performance), it actually seems to be the case. Jest by design re-creates all modules required for the each test (i.e. `it` function) in order for tests to have less side effects on each other. I will dig deeper to understand how to improve the situation for us, but I still don't think that it should block the commit

package.json
calculuschild: Eh.... I suppose we could say that. It is required for the production site to work, since it bundles the js and copies assets over to the proper directories for deployment, but the build script itself isnt actually included in the app.

calculuschild: Which reminds me. If you are planning to test serving of static assets like images, etc., I don't think Jest will be able to see them since the resulting "build" directory is not necessarily the same hierarchy as the source code directory.

AlexeySachkov: > Eh.... I suppose we could say that. It is required for the production site to work, since it bundles the js and copies assets over to the proper directories for deployment, but the build script itself isnt actually included in the app.

Ok, I don't understand how the deployment works, so I suggest that we leave everything as-is for now, just to make sure that there will be no regressions. We can re-arrange our dependencies later if that would be needed, through a separate PR.

> Which reminds me. If you are planning to test serving of static assets like images, etc., I don't think Jest will be able to see them since the resulting "build" directory is not necessarily the same hierarchy as the source code directory.

I've already encountered this issue with `robots.txt` route. I think we should probably make the root directory for static assets configurable somehow - besides the fact that it will help with testing, it could also be useful in local setups, where static assets directory is set to some user-friendly directory for easier assets management.

AlexeySachkov:
I've submitted #1982 to track bad performance of tests

package.json
calculuschild: > I think we should probably make the root directory for static assets configurable somehow

The problem is that it's not just the asset *location*, but the whole folder nesting structure. I.e., we might include specific graphics in a folder somewhere (like the "google Drive" icons) and those get `require()`ed in the appropriate file. But when bundling happens, the whole source folder structure is collapsed into a one-file bundle, which means any assets get similarly reordered/collapsed into a reduced file structure.

Is there any way you can set Jest to run tests off of the bundled app rather than the source code? That would potentially solve several issues:

1) All assets are readily available; we don't have to build a whole secondary structure for testing.
2) Testing is done on code that would actually go into production; a test might pass on the raw source but fail after bundling and we wouldn't catch it.
3) (possibly) Avoid the issues with slow tests?

AlexeySachkov: > Is there any way you can set Jest to run tests off of the bundled app (which resides in the /build folder) rather than the raw source code? That would potentially solve several issues:

I'm not entirely sure how this would help, because my understanding is that `build` directory mostly contains _client_ code - the server part says in place and the only benefit here would be in `shared` part, which is according to my understanding could be server-side rendered to speed up the client app.

Additionally, I think that most of HTTP testing will be focused on API part of the app, because that is where "business logic" resides. Regular user-visible routes will likely be tested in less details, i.e. I do not plan to check the exact html returned by server on `/edit` page - I will check for HTTP status code, response headers and at most I will check that brew source text is present on the page. How exactly it is rendered should be tested by corresponding tests for React component, which indeed could benefit from being launched on bundled version of the app.

I think the summary here is that we need more research and discussion about what exactly to test and how? And based on it we should be able to come up with tests implementation and particular technical solutions which would help tests. Like if we don't plan to check the exact html returned by `/edit` page, then we can mock the whole `Homebrewery` React component to simply print it props to reduce amount of unnecessary things performed by test and to simply checks for the essential things like what has been queried from the database and sent back to user.

calculuschild: > I've already encountered this issue with robots.txt route. I think we should probably make the root directory for static assets configurable somehow - besides the fact that it will help with testing...

Ah, wait, we define in `app.js` the static assets folder here:
https://github.com/naturalcrit/homebrewery/blob/b2ec0d4a0c2b6eedfeaff495b84aa8f3d1788a20/server/app.js#L65

So is the test is operating under a different `__dirname`? Is there a way to fix that somehow without having to pass in custom paths all the way through the app? I kind of prefer having the deployment paths built off of `__dirname` because it ensures the files are always in the right location relative to the server by default. Making that configurable trickles down to several other places in the code that would need to see that custom path to put files in that new location when building. I think that adds more complexity than I want for something that should be kind of fixed.

> I'm not entirely sure how this would help, because my understanding is that build directory mostly contains client code

Ah yeah. I was not thinking straight on that.

calculuschild:
I'm going to merge this, but I noticed that the last PR that split out `app.js` broke the config loading and I missed that. I'm assuming we can make the changes to fix that as part of #1981 .

package.json
AlexeySachkov: > So is the test is operating under a different __dirname? Is there a way to fix that somehow without having to pass in custom paths all the way through the app?

Yes, my understanding is that `jest` launched everything under some other directory. I will dig deeper into documentation to see if we can re-configure `jest` instead of modifying the source code

PR #1981, opened 2022-01-23
[NFC] Outline config creation into a separate module

### LINKED ISSUE #1960: Config object is created several times through the codebase

AlexeySachkov:
This is done in order to have config creation rules
unified in one place to avoid modifying them multiple times
if they change.

We already had 3 duplicated pieces of code initializing the
config and there will be more config uses in future tests.

This resolves #1960


calculuschild:
This looks good. I will give this a test run tomorrow and then merge it.

Only concern I have is we are starting to split out lots of small bits into lots of separate files. I don't have any issue with reducing code duplication, but I imagine getting the testing working will create a lot more of these small modules which starts to obfuscate the code.. @AlexeySachkov What are your thoughts on keeping some of these small modules (this, the Mongo DB connection, etc.) inside the relevant "parent" file and then exporting multiple modules from that file for reuse elsewhere? Not that this PR need to be redone, but it might be worth looking at later on if what once was a single `server.js` turns into 1 wrapper and 7 small helper modules, especially in cases where the only reason it is being broken up is to make testing work.

AlexeySachkov:
> especially in cases where the only reason it is being broken up is to make testing work.

At the moment HTTP tests I'm working on look more like end-to-end tests rather than unit-tests: they consume the whole app, perform a proper request and check the response. So, at least at the moment I don't see why I would split `server.js` for _testing_ purposes, simply because I don't have tests which would require this.

> . @AlexeySachkov What are your thoughts on keeping some of these small modules (this, the Mongo DB connection, etc.) inside the relevant "parent" file and then exporting multiple modules from that file for Jest to access? Not that this PR needs to be redone now, but it might be worth looking at later on if what once was a single server.js turns into 1 wrapper and 7 small helper modules

However, I've been thinking about splitting `server.js` into a smaller pieces for _readability_ purposes, i.e. `app.js` file is about 300 lines long with bunch of helper functions in there and more than 10 explicitly defined routes. Looking at it once more, I think that the most unattractive part for me is a bunch of those helper functions like `splitTextStyleAndMetadata` and "static" page routes which are nearly the same.

Anyway, I don't have a strong opinion on the topic, but my understanding of a node modules philosophy is that they should stay small, dedicated to particular task, isolated things, so my preference is to have more, but smaller files. I'm a bit horrified by the size of dependencies tree, but amount of files _within_ a single package (in this case - `homebrewery`) doesn't seem to concern me that much.

calculuschild:
> So, at least at the moment I don't see why I would split server.js for testing purposes, simply because I don't have tests which would require this.

We have already had to split server.js out into `app.js` and `db.js` to expose them to the test script. This is what I was referring to.

> I've been thinking about splitting server.js into a smaller pieces for readability purposes

This was my concern as well; separating key functions out into their own files makes understanding the code significantly more difficult because it requires jumping around between many more files to track the program logic flow. I'm going to veto for now splitting out helper functions into their own files if they are used exclusively in one area (i.e., if splitting them out does not help code-reuse).

> my understanding of a node modules philosophy is that they should stay small, dedicated to particular task, isolated things, so my preference is to have more, but smaller files

The sentiment is valid, but the "do one thing and do it well" can stop at the function level; I don't think each individual function needs a separate module in a separate file. It's pretty common to have one file export an array of related modules or a module with named components, especially when they are all support each other and are used in the same location. I'd rather have one file `helpers.js` with a bunch of 10-line modules `module.exports = { helper1 : helper1, helper2 : helper2...}` than a bunch of tiny 10-line files, especially when they have common dependencies and call each other (managing complex nested requirement loops becomes a headache to track, and the mental overhead of sifting through all the separate "boilerplate" each new file needs can be a lot).

jeddai:
Just going to comment on this PR since it's still open -- I think I missed something when testing https://github.com/naturalcrit/homebrewery/pull/1959. Enabling v3 isn't currently working for me locally and I think it has to do with the config in the app.js pointing to an invalid file location.

server/config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = require('nconf')
+    .argv()
+    .env({ lowerCase: true })
+    .file('environment', { file: `config/${process.env.NODE_ENV}.json` })
```
jeddai, line 4: I think we should use the __dirname syntax to get the current directory for the executing file, and not rely on the user running the project from the correct directory. While this currently works in this file, it's not really intuitive to the user as to where the config files are located in the project. 

Also, `app.js` needs to be updated to use this module, and despite using the same syntax here for the file location it does not work in `app.js`. The inconsistency concerns me, which is why I think __dirname is a better plan here.

```suggestion
    .file('environment', { file: `${__dirname}/../config/${process.env.NODE_ENV}.json` })
    .file('defaults', { file: `${__dirname}/../config/default.json` });
```

jeddai (changes requested):
`app.js` needs its config impl replaced with a reference to the `config.js` module still, but pending discussion on the other comment I left it functionally is a solid approach and works.

See my comment on `config.js` for a short discussion re: __dirname syntax

server/config.js
jeddai: Elaborating because I think I figured out why it works right now in `config.js` but not in `app.js`.

Since `app.js` calls `process.chdir(__dirname)` it ends up changing the path for its execution to the `server` sub-folder, in which there is no `config` directory. `config.js` does not do that, and since it's being required in `server.js` the `config` folder does exist. 

I don't think `chdir` is the way to go, I really think we should use `__dirname` in our static paths to non-js files.

calculuschild: Yes, this got broken in #1959 when the `config` initialization was left in `server.js` but `process.chdir(__dirname)` was moved out so it could no longer find the config files.

@AlexeySachkov I think @jeddai 's suggested fix is the way to go here.

AlexeySachkov: @jeddai, @calculuschild, Oops, sorry folks for breaking this. I was suspicious about this `process.chdir` line, but apparently not enough. I will prepare a fix

AlexeySachkov: Ok, I did the change locally, but it doesn't seem to work correctly:

```js
const path = require('path');

const defaultConfigPath = path.resolve(__dirname, '../config/default.json');

console.debug('path.resolve(__dirname, "../config/default.json") in config.js ' + defaultConfigPath);
console.debug('"${__dirname}/../config/default.json" in config.js ' + `${__dirname}/../config/default.json`);

module.exports = require('nconf')
    .argv()
    .env({ lowerCase: true })
    .file('environment', { file: `${__dirname}/../config/${process.env.NODE_ENV}.json` })
    .file('defaults', { file: defaultConfigPath });

const fs = require('fs');

fs.readFile(path.resolve(__dirname, '../config/default.json'), 'utf8', (err, data) => {
    if (err) throw err;
    console.log('default config file content read from config.js: ' + data);
  });
```

I'm launching the server out of `server/` sub-directory and this is what I see (`server.js` was explicitly tweaked to use port `8080` as a fallback):
```
$ node ../server.js 
path.resolve(__dirname, "../config/default.json") in config.js C:\Users\asachkov\github.com\naturalcrit\homebrewery\config\default.json
"${__dirname}/../config/default.json" in config.js C:\Users\asachkov\github.com\naturalcrit\homebrewery\server/../config/default.json
config.get(web_port) in server.js: undefined
server on port: 8080
default config file content read from config.js: {
        "secret" : "secret",
        "web_port" : 8000,
        "enable_v3" : true
}
```

The path produced by `path.resolve` is correct (path which uses `${__dirname}` uses different separators so I wasn't sure about it, but it doesn't work as well), the file is definitely accessible, but `config` object doesn't seem to be able to grab the data out of it? What am I missing here?

jeddai: Ah you're on Windows. I think `path.resolve` is going to be for sure required, in hindsight. Seems like it does some OS normalization and switches the `/` to `\` for Windows (I'm on Windows too, I just use it from within WSL so I didn't notice that).

I think this should do it:
```js
const path = require('path');

module.exports = require('nconf')
    .argv()
    .env({ lowerCase: true })
    .file('environment', { file: path.resolve(__dirname, `../config/${process.env.NODE_ENV}.json`) })
    .file('defaults', { file: path.resolve(__dirname, '../config/default.json') });
```

jeddai: As for the port issue, I tried to replicate it using the file as described in the comment above and was unsuccessful. Does what I did in this image generally seem like the same thing you did?

image

AlexeySachkov: I've just returned `config.js` to use plain relative paths and started a server from a root project directory, where everything is supposed to be working and it still doesn't.

I've also spend several minutes in debugger going through `nconf` code: it was definitely able to read the file correctly, but it by some reason can't get any values out of it and I have no idea why.

I will now try googling for the issue, but my fallback plan is to simply add `path.resolve` and hope that everything will work on your side, folks.

jeddai: I'm starting to wonder if it's a Windows thing... Which quite frankly would suck.

AlexeySachkov: This could be some local issue, probably I'm running different version of node (`v16.13.1`) and I've encountered some incompatibility with packages we use?

Right now, I'm loading `defaults.json` the way it were loaded, using relative path without `__dirname` or `path.resovle`; I'm launching the server from the root directory; I've saved `nconf` object into a local variable before storing it to `module.exports` in order to use it in `config.json` and this is what I get:

```
$ node server.js
config.js: config.get(web_port) 8000
server.js: config.get(web_port) undefined
server on port: 8080
```

Values loaded by config magically disappear when config object is accessed by `server.js`. I have no idea why

calculuschild: Does the nconfig object actually store the config values when it's initiated or does it just store like... a path to the config values? So when server.js tries to do config.get those paths are off by one folder?

jeddai: I've console logged it out before and it definitely seems like it's a path reference and not actual values.

calculuschild: Ok, so if we restore `process.chdir(__dirname)` to `server.js` so the "root" folder for everything is based on that, can we go back to `.file('environment', { file: \`config/${process.env.NODE_ENV}.json\` })` as an absolute path?

jeddai: Seems we can configure in-memory storage of the values https://www.npmjs.com/package/nconf#Storage-Engines:

> Memory
> A simple in-memory storage engine that stores a nested JSON representation of the configuration. To use this engine, just call .use() with the appropriate arguments. All calls to .get(), .set(), .clear(), .reset() methods are synchronous since we are only dealing with an in-memory object.
> ```
>nconf.use('memory');
> ```

calculuschild: If that does what it sounds like, it might work.

I think we should still restore `processs.chdir(__dirname)` to `server.js` so the root directory of the app is consistent when it's needed.

AlexeySachkov: > I've console logged it out before and it definitely seems like it's a path reference and not actual values

Hmm, I went on a quick 20 minute adventure in an out with a debugger and I definitely saw it reading a file within `.file()` call.

Adding `use('memory')` doesn't help me regardless of whether I add it before or after calling `.file()`, but what helps me is changing first `.file()` argument from `defaults` to `defaults2` - it seems like there are special cases for configurations named `defaults` and `overrides`, which effectively drop our values

jeddai: I think the issue there is that regardless of using `.use('memory')` it's still not a singleton, so each time it gets required by something it's going to be a new instance, created then.

AlexeySachkov: I think the following issue could be relevant: indexzero/nconf#142

AlexeySachkov: > I think the following issue could be relevant: indexzero/nconf#142

Nope, `grep` doesn't show anyone requiring `nconf` or importing from `nconf`

jeddai: Have you tried with the following nconf setup:
```js
const path = require('path');

module.exports = require('nconf')
    .argv()
    .env({ lowerCase: true })
    .file('environment', { file: path.resolve(__dirname, `../config/${process.env.NODE_ENV}.json`) })
    .file('defaults', { file: path.resolve(__dirname, '../config/default.json') });
```

AlexeySachkov: > Have you tried with the following nconf setup:

Yes, and it doesn't work.

My current guess is that I'm encountering [module caching caveat](https://nodejs.org/api/modules.html#module-caching-caveats) with some of our modules being re-resolved and therefore some of loaded values being overwritten/dropped. At least that is my current guess

AlexeySachkov: Ok, folks, I've pushed an update: 4e0ab4b39367cfba4361d57e7fea23e66ee816ef, in which I've applied `path.resolve` approach to load `.json` files as suggested by @jeddai. Also I've changed `defaults` to `default` to fix the problem I was experiencing.

With these changes I'm able to properly launch server from both root directory and from `server/` directory - it would be great to have confirmation that it works on your side as well, @jeddai, @calculuschild.

AlexeySachkov:
@calculuschild,

> I'm going to veto for now splitting out helper functions into their own files if they are used exclusively in one area (i.e., if splitting them out does not help code-reuse).

Sure, no problem. I wasn't planning on further splitting anyway

> The sentiment is valid, but the "do one thing and do it well" can stop at the function level; I don't think each individual function needs a separate module in a separate file. It's pretty common to have one file export an array of related modules or a module with named components, especially when they are all support each other and are used in the same location. I'd rather have one file helpers.js with a bunch of 10-line modules module.exports = { helper1 : helper1, helper2 : helper2...} than a bunch of tiny 10-line files, especially when they have common dependencies and call each other (managing complex nested requirement loops becomes a headache to track, and the mental overhead of sifting through all the separate "boilerplate" each new file needs can be a lot).

+1 here, a whole module for a single function is overkill in case we can actually group some functions locally together.

> I think we should still restore processs.chdir(__dirname) to server.js so the root directory of the app is based on the root folder of the project, rather than setting it in app.js which sets the root to the /server folder.

I would love to do that, but it seems like it requires path changes all over the place, so I would prefer to do that as a separate pull request

calculuschild:
> I would love to do that, but it seems like it requires path changes all over the place, so I would prefer to do that as a separate pull request

That makes sense to me. :+1:

server/config.js
jeddai: Hmmm that seems most likely. @calculuschild does the branch load the config correctly for you now?

calculuschild:
@AlexeySachkov @jeddai 

I pushed up a couple changes that I think fix everything without needing to rely on `path.resolve` everywhere. In short, the simple solution is:

1) Everything that uses `config` also relies on `app.js`, so let `app.js` set the working directory with ``process.chdir(`${__dirname}/..`);`` . This is the very first thing it does at the top of the file, making the root `/homebrewery/`.
2) Any subsequent calls to static files, including in `config.js`, now default to the project root (`/homebrewery/`) and work like a charm. Works fine no matter where the app is called from. (`server.js`, `../server.js from inside `/server`, etc.)
3) Testing also works fine, since when it includes `app.js` the process working directory gets updated. Testing now runs from the same root folder as the production environment automatically.

Separately, there was the broken `robots.txt`, which wasn't really related to the config files but was literally just an `off-by-one folder` issue. The simple solution would have just been to update ``res.sendFile(`${__dirname}/robots.txt`)`` to ``res.sendFile(`${__dirname}/../robots.txt`)``. `app.js` was moved one folder deeper, which meant `__dirname` for `app.js` was `/server`, and `server/robots.txt` didn't exist, but `server/../robots.txt` is fine.

However, we can go further since `process.cwd()` is updated at this point, and get rid of `__dirname` entirely with ``res.sendFile(`${process.cwd()}/robots.txt`);``, or ``res.sendFile(`robots.txt`, { root: process.cwd() });``. Second option looked cleaner to me than string concatenation.

For reference, since there seemed to be a lot of confusion on these:

* `__dirname` is the absolute path to the directory of the current file.
* `process.cwd()` is the current working directory of the node app. All file references use this directory as the relative "root".
* `process.chdir(path)` updates `process.cwd()` to the given path.
* `process.chdir(__dirname) then updates the working directory to the directory of the current file. This is now set for the *whole* app, which means every file can assume the same root directory for any paths.

AlexeySachkov:
@calculuschild, thanks!

Personally, still prefer `path.resolve` way, but I won't object against merging the PR, changes look good to me to merge.

PR #1993, opened 2022-01-28
Fix robots.txt route

AlexeySachkov:
The route was broken by #1959, because path to `robots.txt` became invalid with that PR.
Tests I added in #1979 were actually highlighting the issue, but I thought that this is a test issue and not a real problem.

This PR fixes that and enables the corresponding test. We can't just use `${__dirname}/../robots.txt`, because `express.js` under the hood checks for malicious strings and going "up" through `..` is considered dangerous - `403 Forbidden` was returned in this case.


calculuschild:
Resolved by #1981.

PR #1994, opened 2022-01-29
Use path.resolve instead of process.chdir

AlexeySachkov:
Resolving paths dynamically against the directory, where `server.js` is located
is a more reliable and flexible approach: with it `server.js` can be launched from
any directory and even if `app.js` is used directly without `server.js` (like it is done
in testing), it will still work, because the app has no assumptions about the
current working directory it was launched from.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -4,6 +4,7 @@ const jwt = require('jwt-simple');
 const express = require('express');
 const yaml = require('js-yaml');
 const app = express();
+const path = require('path');
 
 const homebrewApi = require('./homebrew.api.js');
 const GoogleActions = require('./googleActions.js');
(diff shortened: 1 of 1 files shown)


AlexeySachkov:
This is a follow-up from our [discussion](https://github.com/naturalcrit/homebrewery/pull/1981#discussion_r793972608) in #1981. @calculuschild: I've decided not to rely on `process.chdir` at all and instead dynamically resolve paths to static assets within `app.js`. Expecting particular working directory in `app.js` means that we will have to adjust it in every test, which doesn't seem like a good idea.

calculuschild:
>  Expecting particular working directory in app.js means that we will have to adjust it in every test, which doesn't seem like a good idea.

Expecting a particular working directory is desired outcome. Relying on a relative paths everywhere because we can't trust the where the root directory is is going to lead to more problems down the road I think, since it's a core part of how Node operates. If anything, tests should be adjusted as needed to work with the main product, not the other way around.

I'm going to take a look and see if I can get the config module PR #1981 working with `process.chdir`.

calculuschild:
Resolved by #1981.

PR #1995, opened 2022-01-30
Custom CSS syntax: `recolor: #FFAABB` to auto-calculate hue-rotate, etc.

calculuschild:
Dumb little experiment as part of making custom themes. I know there's not an issue for this, but I don't plan to merge this. Just wanted to share the code so I can get opinions on the idea and come back to it later in a better PR if it seems worth it.

I love the idea of being able to recolor assets for custom themes, but we always have to rely on `hue-rotate` or `brightness` filters to get the color we want, and it's not intuitive at all, especially for inexperienced users who just want to get their CSS to work and don't want to have to calculate what angle to rotate around a color wheel to turn a blue symbol orange.

This PR adds a custom CSS property `recolor` which takes a color (currently just a 6-character hex value) and then calculates the required `filter: hue-rotate() brightness() saturate()` values to turn an asset into that color, assuming the asset starts as mostly red `255, 0, 0`.

Only works in the curly div/span syntax since it was easiest to test there, but this could equally be added as a CSS post-process step to modify the user's style tab before injecting into the DOM.

Then, we simply update any graphics, etc. to be red, apply the filters as required to make them appear as normal in our themes, and then users can recolor them as they wish with `{{monster,frame,recolor:#008800.... }}` to turn the monster frame dark green, for example.

### DIFF
--- a/server.js
+++ b/server.js
@@ -1,6 +1,8 @@
 const DB = require('./server/db.js');
 const server = require('./server/app.js');
 
+process.chdir(__dirname);
+
 const config = require('nconf')
 	.argv()
 	.env({ lowerCase: true })
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -329,6 +329,66 @@ const voidTags = new Set([
 	'input', 'keygen', 'link', 'meta', 'param', 'source'
 ]);
 
+const recolor = (target)=>{
+	//First, extract the RGB components of the hex color notation.
+
+	var r = parseInt(target.substr(1,2), 16); // Grab the hex representation of red (chars 1-2) and convert to decimal (base 10).
+	var g = parseInt(target.substr(3,2), 16);
+	var b = parseInt(target.substr(5,2), 16);
+
+  r /= 255, g /= 255, b /= 255;
+  var max = Math.max(r, g, b), min = Math.min(r, g, b);
+  var h, s, l = (max + min) / 2;
+
+  if(max == min){
+      h = s = 0; // achromatic
+  }else{
+      var d = max - min;
+      s = l > 0.5 ? d / (2 - max - min) : d / (max + min);
+      switch(max){
+          case r: h = (g - b) / d + (g < b ? 6 : 0); break;
+          case g: h = (b - r) / d + 2; break;
+          case b: h = (r - g) / d + 4; break;
+      }
+      h /= 6;
+  }
+
+	// fake hue-rotate to see how much to post-brighten and saturate
+  let angle = h * 2 * Math.PI;
+	console.log({angle: angle});
+  let sin = Math.sin(angle);
+  let cos = Math.cos(angle);
+
+	let matrix = [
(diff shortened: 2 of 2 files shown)


5e-Cleric:
Just saw this, why not calculate the average color of the asset, and work from there to the final color?

ericscheid:
> Grabbing the average color is not as easy as that. Doing it pixel by pixel will be extremely slow, and doesn't work for svgs, 

Here is some js that gets average img color. Running it in the browser appears to be very fast - for the quite large `parchmentBackground.jpg` image it only took 155ms.

```javascript
function getAverageColorFromImage(url) {
  const t0 = performance.now();
  return new Promise((resolve, reject) => {
    const img = new Image();
    img.crossOrigin = "Anonymous";
    img.onload = function() {
      const canvas = document.createElement("canvas");
      const ctx = canvas.getContext("2d");
      canvas.width = this.width;
      canvas.height = this.height;
      ctx.drawImage(this, 0, 0);
      const imageData = ctx.getImageData(0, 0, canvas.width, canvas.height);
      const pixels = imageData.data;
      let r = 0, g = 0, b = 0;
      for (let i = 0; i < pixels.length; i += 4) {
        r += pixels[i];
        g += pixels[i+1];
        b += pixels[i+2];
      }
      const numPixels = pixels.length / 4;
      const avgR = Math.round(r / numPixels);
      const avgG = Math.round(g / numPixels);
      const avgB = Math.round(b / numPixels);
      const t1 = performance.now();
      const runTime = t1 - t0;
      resolve({
        avgColor: `rgb(${avgR}, ${avgG}, ${avgB})`,
        runTime,
        url
      });
    };
    img.onerror = reject;
    img.src = url;
  });
}
```

calculuschild:
> it only took 155ms.

keep in mind this would be happening with every keystroke. That would be an extremely laggy experience.

But even with an instantaneous algorithm, I think the other issues I mentioned would make auto-detection less useful.

5e-Cleric:
> keep in mind this would be happening with every keystroke. That would be an extremely laggy experience.

Well then what about making it a separate tool on naturalcrit.com?

calculuschild:
Do we really need an entire separate tool where a user could just as easily type in a starting color of their choice?

calculuschild:
@explainthispr

dbolack-ab:
> Do we really need an entire separate tool where a user could much more easily just type in the starting color they want?

I think, more practically, simply declare that images for this system are intended to be done in based a specific color ( I'd say black, but if the code needs it to be R, G, or B.... Blue ) and simplify the process.

5e-Cleric:
thiis would have been very useful to me yesterday, using a custom table frame that is supposed to change color, but oh well.

dbolack-ab:
> > it only took 155ms.
> 
> keep in mind this would be happening with every keystroke. That would be an extremely laggy experience.
> 

Can that be memoed?

5e-Cleric:
> > > it only took 155ms.
> > 
> > 
> > keep in mind this would be happening with every keystroke. That would be an extremely laggy experience.
> 
> Can that be memoed?

I'd go a step further and make a custom memoization so the function doesn't run if the color is _similar_.

Issue #1987, opened 2022-01-26
Document editing permissions


jeddai:
The edit endpoint is currently open to any user. This feature would enable users to share document editing permissions with other users, but it would be an opt-in setup instead of it being open with no option to disable it.

* Currently, the API is designed so that if another user accesses an edit url that does not belong to them they can still edit the document and it will add them as an author. 

<hr/>

* Moving forward, if you access an edit url and you are not one of the document's authors, you will get a 403 page that says you don't have edit access to that document. 
* If you as the author want other users to be able to edit some of your docs, you can add authors to a document via the metadata pane (using their username).
* A warning would say that collaborative editing is in an interesting state and to make sure to save the document offline somewhere
* Document version in the database would be incremented on each save
* If the document version being saved is lower than the document version in the database it would prevent the save and show an error

G-Ambatte:
> If you as the author want other users to be able to edit some of your docs, you can add authors to a document via the metadata pane (using their username).

There's some good conversation in the Gitter developer chat, but the short version is that adding an author should generate an invitation which must be accepted by the new author BEFORE edit access is provided. However, this then is open for abuse (e.g. harassing a particular user via invite bombing), so some form of invite blocking needs to be in place prior to this being open to the public.
I believe that this probably needs the UserInfo system in place to record the blocking preferences.

calculuschild:
> However, this then is open for abuse (e.g. harassing a particular user via invite bombing),

This can all be avoided by letting users handle the invites externally (Hey Joe, I added you as an author to the brew. Here's the edit link so you can accept: homebrewery.com/edit/buh). I think we can assume if users are collaborating on a brew, they also have some other means of contact. I.e., we don't need to "generate an invitation", but simply mark the user as "invited" so they have permission to access the edit link if they choose.

A spammer can't "invite bomb" if the Homebrewery doesn't send any messages.

G-Ambatte:
> mark the user as "invited" so they have permission to access the edit link if they choose

I like this approach.

> "Hey Joe, I added you as an author to the brew. Here's the edit link so you can accept: homebrewery.com/edit/buh"

We could probably launch a "Copy invitation text to clipboard?" window if we wanted to get really user friendly, so there's less opportunity for human error.

ericscheid:
Side note: if an author is added via invitation to a brew, their name should not yet appear in the div.brewItem.brewInfo until they accept that invitation. 

[...]

PR #2006, opened 2022-02-05
Prevent edits to documents when the logged in user is not the author

jeddai:
update api and front end to allow authors to be manually edited and prevent unwanted modifications to documents

Issues: #1987

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -160,7 +160,7 @@ const MetadataEditor = createClass({
 			</div>
 			*/}
 
-			<StringArrayEditor label='authors' valuePatterns={[/^$/]} canEdit={(value)=>value !== global.account?.username}
+			<StringArrayEditor label='authors' valuePatterns={[/^$/]} canEdit={(value)=>value !== global.account?.username && value !== this.props.metadata.authors[0]}
 				addPlaceholder='add username' editPlaceholder='input username'
 				values={this.props.metadata.authors}
 				onChange={(e)=>this.handleFieldChange('authors', e)}/>
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -160,8 +160,8 @@ const MetadataEditor = createClass({
 			</div>
 			*/}
 
-			<StringArrayEditor label='authors' valuePatterns={[/^$/]} canEdit={(value)=>value !== global.account?.username && value !== this.props.metadata.authors[0]}
-				addPlaceholder='add username' editPlaceholder='input username'
+			<StringArrayEditor label='authors' valuePatterns={[/^$/]} cannotEdit={[global.account?.username, this.props.metadata.authors[0]]}
+				placeholder='add username' unique={true}
 				values={this.props.metadata.authors}
 				onChange={(e)=>this.handleFieldChange('authors', e)}/>
 
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -160,7 +160,8 @@ const MetadataEditor = createClass({
 			</div>
 			*/}
 
-			<StringArrayEditor label='authors'
+			<StringArrayEditor label='authors' valuePatterns={[/^$/]} canEdit={(value)=>value !== global.account?.username}
+				addPlaceholder='add username' editPlaceholder='input username'
 				values={this.props.metadata.authors}
 				onChange={(e)=>this.handleFieldChange('authors', e)}/>
 
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -7,6 +7,8 @@ const request = require('superagent');
 
 const SYSTEMS = ['5e', '4e', '3.5e', 'Pathfinder'];
 
+const StringArrayEditor = require('./stringArrayEditor.jsx');
+
 const MetadataEditor = createClass({
 	displayName     : 'MetadataEditor',
 	getDefaultProps : function() {
@@ -26,9 +28,10 @@ const MetadataEditor = createClass({
 	},
 
 	handleFieldChange : function(name, e){
-		this.props.onChange(_.merge({}, this.props.metadata, {
+		this.props.onChange({
+			...this.props.metadata,
 			[name] : e.target.value
-		}));
+		});
 	},
 	handleSystem : function(system, e){
 		if(e.target.checked){
@@ -103,19 +106,6 @@ const MetadataEditor = createClass({
 		</div>;
 	},
 
-	renderAuthors : function(){
-		let text = 'None.';
-		if(this.props.metadata.authors && this.props.metadata.authors.length){
-			text = this.props.metadata.authors.join(', ');
-		}
-		return <div className='field authors'>
-			<label>authors</label>
-			<div className='value'>
-				{text}
-			</div>
-		</div>;
-	},
-
 	renderRenderOptions : function(){
 		if(!global.enable_v3) return;
 
@@ -170,7 +160,9 @@ const MetadataEditor = createClass({
 			</div>
 			*/}
 
-			{this.renderAuthors()}
+			<StringArrayEditor label='authors'
+				values={this.props.metadata.authors}
+				onChange={(e)=>this.handleFieldChange('authors', e)}/>
 
 			<div className='field systems'>
 				<label>systems</label>
--- a/client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
@@ -6,11 +6,13 @@ const StringArrayEditor = createClass({
 	displayName     : 'StringArrayEditor',
 	getDefaultProps : function() {
 		return {
-			label        : '',
-			values       : [],
-			valuePattern : null,
-			canEdit      : null,
-			onChange     : ()=>{}
+			label           : '',
+			values          : [],
+			valuePatterns   : null,
+			editPlaceholder : '',
+			addPlaceholder  : '',
+			canEdit         : null,
+			onChange        : ()=>{}
 		};
 	},
 
@@ -46,16 +48,21 @@ const StringArrayEditor = createClass({
 
 	addValue : function(value){
 		this.handleChange(_.uniq([...this.props.values, value]));
+		this.setState({
+			temporaryValue : ''
+		});
 	},
 
-	removeValue : function(value){
-		this.handleChange(this.props.values.filter((v)=>v !== value));
+	removeValue : function(index){
+		this.handleChange(this.props.values.filter((_, i)=>i !== index));
 	},
 
 	updateValue : function(value, index){
-		const values = this.props.values;
-		values[index] = value;
-		this.handleChange(values);
+		const valueContext = this.state.valueContext;
+		valueContext[index].value = value;
+		valueContext[index].editing = false;
+		this.handleChange(valueContext.map((context)=>context.value));
+		this.setState({ valueContext, updateValue: '' });
 	},
 
 	editValue : function(index){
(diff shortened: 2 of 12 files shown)


server/homebrew.api.js
```diff
@@ -91,7 +98,7 @@ const updateBrew = (req, res)=>{
 			brew.updatedAt = new Date();
 
 			if(req.account) {
-				brew.authors = _.uniq(_.concat(brew.authors, req.account.username));
+				brew.authors = _.uniq(_.concat(updateBrew.authors.filter((a)=>a.length > 0), req.account.username));
```
calculuschild, line 101: What is this doing? Just making sure you can't add authors with an empty string for the name?

Could this be handled in the UI so you can't add `''` as a username anyway?

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -170,7 +160,9 @@ const MetadataEditor = createClass({
 			</div>
 			*/}
 
-			{this.renderAuthors()}
+			<StringArrayEditor label='authors'
```
calculuschild, line 163: It is possible to delete the current users' name from the list. It comes back on page refresh, but perhaps we can add some property to StringArrayEditor to signal which strings should not be editable so the user's own name doesn't have the `x` button?

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,124 @@
...
+			return;
+		}
+		const valueContext = this.state.valueContext.map((context, i)=>{
+			context.editing = index === i;
+			return context;
+		});
+		this.setState({ valueContext, updateValue: this.props.values[index] });
+	},
+
+	handleValueInputKeyDown : function(event, index) {
```
calculuschild, line 72: Might be worth adding an arrow or checkmark or something as a "submit" button for users who don't guess to try Enter.

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,124 @@
...
+		if(event.key === 'Enter') {
```
calculuschild, line 73: Pressing Enter after editing an existing value should exit edit mode for that entry. 

Also, it seems to just create a new entry but the old entry reappears when you click out.

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+	closeEditor : function(index) {
+		const valueContext = this.state.valueContext;
+		valueContext[index].editing = false;
+		this.setState({ valueContext, updateValue: '' });
+	},
+
+	render : function() {
+		const valueElements = Object.values(this.state.valueContext).map((context, i)=>context.editing
+			? <input type='text' className='value' autoFocus
```
calculuschild, line 100: Can we add a default text with `placeholder="Add username"` or something?

server/homebrew.api.js
jeddai: Yeah we can handle it in the UI as well, but I don't want it to be possible on the API either.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
jeddai: That's a good idea -- need to figure out how to get the username into this element though

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
jeddai: Yep, I'll do that!

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
jeddai: Yep!

server/homebrew.api.js
calculuschild: That's fair. Just so the UI doesn't inply you can do something but the API does something else.

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
calculuschild: Clicking out as "cancel" is fine. The problem is editing and then pressing enter creates a *new* entry rather than updating the current entry, so when you click out the old entry you edited magically comes back.

jeddai: Hmm I haven't seen this behavior -- if you could send me some screenshots or a short video that would be great

client/homebrew/editor/metadataEditor/metadataEditor.jsx
jeddai: Yeah the issue is that I looked through the element tree and haven't quite figured out where the username is and isn't available yet

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
calculuschild: https://user-images.githubusercontent.com/5878534/152845031-fbd5f51d-c0ad-4768-908c-a2493f98f207.mp4

It might only occur when editing the first author in the list. The others seem to work fine.

jeddai: Oh that's super weird, yeah I'll figure out what's up there!

jeddai: Lmao it's because I was using the truthy operator on `index`, so when index is 0 it resolved to false and added an element instead of editing it.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -170,7 +160,10 @@ const MetadataEditor = createClass({
 			</div>
 			*/}
 
-			{this.renderAuthors()}
+			<StringArrayEditor label='authors' valuePatterns={[/^$/]} canEdit={(value)=>value !== global.account?.username && value !== this.props.metadata.authors[0]}
```
calculuschild, line 163: For `canEdit`, would it be any cleaner to simply provide an array of strings to check against rather than a whole function which essentially has those same strings hardcoded in?

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -128,11 +128,11 @@ const EditPage = createClass({
 	},
 
 	handleMetaChange : function(metadata){
+		console.log(metadata);
```
calculuschild, line 131: Remove before merging.

server/app.js
```diff
@@ -220,6 +220,10 @@ app.get('/user/:username', async (req, res, next)=>{
 app.get('/edit/:id', asyncHandler(async (req, res, next)=>{
 	res.header('Cache-Control', 'no-cache, no-store');	//reload the latest saved brew when pressing back button, not the cached version before save.
 	const brew = await getBrewFromId(req.params.id, 'edit');
+	if(brew.authors.length > 0 && !(!!req.account && brew.authors.includes(req.account?.username))) {
```
calculuschild, line 223: If you are just checking for the existence of a variable, I don't think you need to typecast into a boolean; `req.account` should automatically evaluate falsey if it's undefined, etc.

In any case you can simplify with optional chaining even further.

```suggestion
	if(brew.authors.length > 0 && !brew.authors.includes(req?.account?.username)) {

```

server/homebrew.api.js
```diff
@@ -81,6 +81,24 @@ const updateBrew = (req, res)=>{
...
+			}
+
+			if(brew.version > updateBrew.version) {
+				console.warn(`Document ${brew._id} had a version mismatch`);
+				res.setHeader('Content-Type', 'application/json');
+				return res.status(409).send(JSON.stringify({ message: `The brew has been changed on a different device. Please save your changes elsewhere, refresh, and try again.` }));
+			}
+
+			if(req.account) {
+				brew.authors = _.uniq(_.concat([!!brew.authors ? brew.authors[0] : undefined], updateBrew.authors, req.account.username)).filter((a)=>a !== undefined && a.length > 0);
```
calculuschild, line 98: If you are already filtering out undefined anyway, do we need `!!brew.authors ? brew.authors[0] : undefined` ? Can we reduce to `_.concat([brew?.authors?.[0]], updateBrew.authors, req.account.username)` ?

server/homebrew.api.js
```diff
@@ -81,6 +81,24 @@ const updateBrew = (req, res)=>{
 	HomebrewModel.get({ editId: req.params.id })
 		.then((brew)=>{
 			const updateBrew = excludePropsFromUpdate(req.body);
+
+			if(brew.authors.length > 0 && !(!!req.account && brew.authors.includes(req.account?.username))) {
+				console.warn(`User ${req.account?.username} is not an author on brew ${brew._id}`);
+				res.setHeader('Content-Type', 'application/json');
+				return res.status(403).send(JSON.stringify({ message: `You must be added as an author by the document owner to edit this brew!` }));
```
calculuschild, line 88: So we throw a 403 when the user tries to visit the edit page, as well as trying to save on that edit page. Do we need both? If they can't even get to the edit page do we need this second check here? Mostly I don't like the code duplication....

jeddai: I think the 403 is necessary in both places because they could have been on the page, lost authorship, then tried to save. 

If the code duplication is a problem I can pull that out into a function?

calculuschild: In that case, both retrieving a brew and updating a brew occur via this `homebrew.api` file, so I think it makes sense to emit the 403 from here on GET ing a brew rather than in `app.js`. Both updating and getting can use that function.

jeddai: I'm not seeing how `app.js` currently calls anything in `homebrew.api.js` for retrieving the brew on the edit url -- is there a specific spot you're referring to?

calculuschild: Ah you're right. `app.js` directly goes to `homebrewModel.js` rather than via the API....

Ok, I guess we can leave it as is for now since that looks like it will require a bigger rework (I think at some point we want all brew interaction with the database passing through the API for consistency).

I guess if you want to leave it as is and just comment on each that there is an identical function in the other file, or make it into a function either works.

jeddai: Sounds good! I'll pull it into a helper function somewhere.

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,139 @@
...
+		const valueContext = this.state.valueContext;
+		valueContext[index].editing = false;
+		this.setState({ valueContext, updateValue: '' });
+	},
+
+	render : function() {
+		const valueElements = Object.values(this.state.valueContext).map((context, i)=>context.editing
+			? <React.Fragment key={i}>
+				<div className='input-group'>
+					<div className='input-group'>
```
calculuschild, line 109: What's with the double-nested `<div className='input-group'>`?

client/homebrew/editor/metadataEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,139 @@
...
+				{!!this.props.canEdit && !this.props.canEdit(context.value) ? null : <div className='icon steel'><i className='fa fa-times fa-fw' onClick={(e)=>{ e.stopPropagation(); this.removeValue(i); }}/></div>}
+			</div>
+		);
+
+		return <div className='field values'>
+			<label>{this.props.label}</label>
+			<div className='list'>
+				{valueElements}
+				<div className='input-group'>
+					<input type='text' className={`value ${this.valueIsValid(this.state.temporaryValue) ? '' : 'invalid'}`} placeholder={this.props.addPlaceholder}
```
calculuschild, line 128: The input box is showing red by default because when empty, it has an invalid value. I think it should only show red *while* the user is editing it.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -151,8 +151,12 @@ const UserPage = createClass({
 			return (_.deburr(brew.title).toLowerCase().includes(testString)) ||
 			(_.deburr(brew.description).toLowerCase().includes(testString));
 		}) : this.props.brews;
-		return _.groupBy(brewCollection, (brew)=>{
-			return (brew.published ? 'published' : 'private');
+		return _.groupBy(brewCollection, ({ collaborating, published })=>{
```
calculuschild, line 154: The userPage has been changed with #1799 so this will need to be moved. Update this branch to master to see the changes.

jeddai:
This was handled by #2509, #2561, #2572, and #2575!

Issue #1798, opened 2021-11-03
Unify base page structure - ListPage


G-Ambatte:
#1423 was an attempt to unify the underlying editor page for `NewPage` and `EditPage`. However the rest of the project moved on before the PR was completed.

This issue is to record the idea - unifying the underlying page structure so that functions do not need to be modified in multiple places - so that it is not lost with the closure of #1423.

My intention is to close #1423 and open a new PR with a smaller, sleeker set of changes, implementing a minimal combined EditorPage to be called by NewPage and EditPage, which can be merged further through later PRs.
I also intend to push UserPage through an underlying ListPage, which can be used for any future pages that need to list multiple brews.

---

This issue is to track the creation of the ListPage base page.

G-Ambatte:
Updated the Issue description to reflect the separation of ListPage and EditorPage base pages.

Gazook89:
I'm going to reopen this Issue, for now, but maybe a new issue is better.  I'm happy to go with whichever direction makes most sense....

I don't think this is issue is totally resolved, if taking it's title to the maximum extent:  All pages should have one base page upon which their specific functions are built.  It seems like right now we only have the "brew list" and "account page" set to use one common base...but nothing else is on that common base.   Even if "new" and "edit" and "share" and maybe others have slightly different sets of buttons in their navbars, they can still rest on that same basic layout, rather than each laying out their navbars separately.

A minor suggestion is to rename the `uiPage.jsx` to `baseLayout.jsx` or something similarly descriptive of the function of that component.  I think `uiPage.jsx` falls short in describing the purpose of that file (everything is "UI", and it's not really a "Page" either).

calculuschild:
Look at #1800 which is the sub-issue specifically for unifying EditPage, NewPage, and HomePage, and the accompanying PR #1801 which I would love to see restarted and completed.

Gazook89:
_This is sort of an incomplete thought but I have to go to bed and likely toy with the code a bit later.  I'm just hitting submit so I can come back to it later._  

-----

Is this really complete though?  I see this structure right now:

`userPage.jsx` --> `listPage.jsx`

`userPage` has it's own navbar render method which it then passes to `listPage`.   `userPage` is basically doing the same thing as what `uiPage.jsx` does, and I think we can unify what that "top level" component is since both `userPage` and `uiPage` are basically containers for Navbar + Page.

Right now I envision these changes:
[...]

Issue #1844, opened 2021-11-19
link whole .brewItem card to /edit or /share?


ericscheid:
From #240

> The whole card should be either a link to the Edit Page if you are the author, or to the Share page if not the author...still retain the little pop up/:hover element that shows the current 4 options, but anticipate the users click based on authorship when clicking anywhere else on the card.

>> The common action is to either edit-if-author or view-share-if-visitor. Make that the easiest.

G-Ambatte:
We should be able to attach an `onClick` method to the BrewItem which will then contains the logic to redirect appropriately; my initial feeling is that this should be a relatively simple change which should improve the User Page user experience.

ericscheid:
Better to attach an event listener to a higher container, and from there detect if the element clicked was the brewItem (and not the individual links in the sidebar.

````
document.querySelectorAll('.published,.unpublished')
.forEach( 
  item => { 
    item.addEventListener( 'click', 
      (event) => {
        // do the hard work here
      }
  } );
````

The "hard work" would be 
1. determining if the click landed inside element with ancestor `<div class="links">` (if so just return true and let the brwser handle the click); and failing that
2. determine if the click was on an element with ancestor `<div class="brewItem">`, then find the appropriate link in the `.links` div inside that `.brewItem`.

With item 2, it would be handy if each of the `<a>` elements had an appropriate identifying class, like this:

```
<div class="links">
  <a href="/share/mumble" class="share" ...>...</a>
  <a href="/edit/mumble" class="edit" ...>...</a>
  <a href="/download/mumble" class="download" ...>...</a>
  <a class="delete">...</a>
</div>
```
(This would be preferable to doing `document.querySelector('[href^="/edit"]'), as class="edit" is deliberate and explicit, while `href^="edit"` might be subject to change for reasons completely unrelated to this click handler. Thus: safer.)

While we're at it, the two divs currently found via `document.querySelectorAll('.published,.unpublished')` could maybe have a "collection" class also applied (in anticipation of future efforts of having folders/views/collections). Alternatively, the event listener could simply be attached to `.phb` since we're only handling those clicks with `event.target.closest('.brewItem')` (and thus the interactive elements within `.phb .sort-container` can still have their clicks passed to them by default).

I will have the `// hard work here` bit figured out soon. Fiddling about in the console now...

**Comment by ericscheid** [comment]
So, something like this, but optimised/linted/etc

```
document.querySelectorAll('.published,.unpublished')
.forEach( 
[...]

PR #2007, opened 2022-02-06
Make brewItems stand out a bit.

calculuschild:
Realized I had this on my local repo for a long time now. Small style tweak on the `user` page. #1798 reminded me I had this.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
@@ -17,6 +17,8 @@
 	-webkit-column-break-inside : avoid;
 	page-break-inside           : avoid;
 	break-inside                : avoid;
+	box-shadow                  : 0px 4px 5px 0px #333;
+	background-color            : #cab2802e;
 	.text {
 		min-height : 54px;
 		h4{


G-Ambatte:
image

Looks good to me!

Gazook89:
It likely isn't the focus of this PR, but mentioning it just in case you want to do something beyond just a simple CSS change:  Issue #1844 -- turn the card into a button itself.  

Also, if you have two cards side by side, and one has more content/is taller than the other, it does this:

image

I think it would be better to use a `flex` arrangement, and set `flex-direction` to column.  it will remove the gaps.   Of course, then the "grid" with the aligned tops will be gone, which might be more annoying.  In which case, setting it to `grid`, or something else might be better.  Haven't had time to play with it, and gotta go feed a baby.

Gazook89 (approved):
I left a comment in the 'conversation' as I didn't realize that review comments were separate.  But as a comment on these specific changes to box-shadow and background-color, I have nothing to add.  Looks good.

calculuschild:
> It likely isn't the focus of this PR, but mentioning it just in case you want to do something beyond just a simple CSS change: Issue #1844 -- turn the card into a button itself.

I'll leave that as a separate PR. Thanks for looking guys!

PR #2019, opened 2022-02-09
update api and frontend to allow for google drive document stubs

jeddai:


calculuschild:
Hey @jeddai. I know you aren't going to like this, but after spending a good amount of time looking at it, this PR has so much going on and multiple areas undergoing large refactors, I'm struggling to follow everything to make sure I'm not missing any potential issues. I see at least three major refactors going on here that as your title says are "API updates" in preparation for stubs, but not part of stubs themselves. Could you possibly split out a few smaller PRs so we can more easily handle those refactors first?

1) Moving "transfer to/from Google" logic out of `New/Edit` pages and into `homebrew.api`.
2) Making the Service Account the primary method of accessing Google Brews rather than the API key.
3) Refactoring `homebrew.api` to use a single path regardless of brew type, rather than requiring separate paths for Mongo and Google brews. (`/update/:id` and `/updateGoogle/:id`)

Each of those is significant enough that I'm worried about introducing hidden errors but changing all three at once makes tracking that very difficult.

PR #2026, opened 2022-02-15
Update Google interactions to prefer the service account where viable

jeddai:
Commits
* update googleActions and related files to use service-level auth where viable

Iterative approach to changes required for Google document stubs, part 1. See issue #1838 

This only changes the GoogleActions so that they will primarily use the google service account loaded by the application for interacting with drive-based documents.

Notes:
* Users who are not the main author will not see the brew on their user page
* Deletion and Creation are the two pieces that still require a valid oauth token

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -25,7 +25,7 @@ const getBrewFromId = asyncHandler(async (id, accessType)=>{
 	if(id.length > 12) {
 		const googleId = id.slice(0, -12);
 		id             = id.slice(-12);
-		brew = await GoogleActions.readFileMetadata(config.get('google_api_key'), googleId, id, accessType);
+		brew = await GoogleActions.readFileMetadata(googleId, id, accessType);
 	} else {
 		brew = await HomebrewModel.get(accessType == 'edit' ? { editId: id } : { shareId: id });
 		brew = brew.toObject(); // Convert MongoDB object to standard Javascript Object
--- a/server/app.js
+++ b/server/app.js
@@ -25,7 +25,7 @@ const getBrewFromId = asyncHandler(async (id, accessType)=>{
 	if(id.length > 12) {
 		const googleId = id.slice(0, -12);
 		id             = id.slice(-12);
-		brew = await GoogleActions.readFileMetadata(googleId, id, accessType);
+		brew = await GoogleActions.getGoogleBrew(googleId, id, accessType);
 	} else {
 		brew = await HomebrewModel.get(accessType == 'edit' ? { editId: id } : { shareId: id });
 		brew = brew.toObject(); // Convert MongoDB object to standard Javascript Object
--- a/server/app.js
+++ b/server/app.js
@@ -200,13 +200,16 @@ app.get('/user/:username', async (req, res, next)=>{
 	});
 
 	if(ownAccount && req?.account?.googleId){
-		const googleBrews = await GoogleActions.listGoogleBrews(req, res)
-		.catch((err)=>{
-			console.error(err);
-		});
+		const auth = await GoogleActions.authCheck(req.account, res);
+		let googleBrews = await GoogleActions.listGoogleBrews(auth)
+			.catch((err)=>{
+				console.error(err);
+			});
 
-		if(googleBrews)
+		if(googleBrews) {
+			googleBrews = googleBrews.map((brew)=>({ ...brew, authors: [req.account.username] }));
 			brews = _.concat(brews, googleBrews);
+		}
 	}
 
 	req.brews = _.map(brews, (brew)=>{
--- a/server/googleActions.js
+++ b/server/googleActions.js
@@ -12,10 +12,7 @@ let serviceAuth;
 try {
 	serviceAuth = google.auth.fromJSON(keys);
 	serviceAuth.scopes = [
-		'https://www.googleapis.com/auth/drive',
-		'https://www.googleapis.com/auth/drive.appdata',
-		'https://www.googleapis.com/auth/drive.file',
-		'https://www.googleapis.com/auth/drive.metadata'
+		'https://www.googleapis.com/auth/drive'
 	];
 } catch (err) {
 	console.warn(err);
@@ -94,15 +91,8 @@ const GoogleActions = {
 		return folderId;
 	},
 
-	listGoogleBrews : async (req, res)=>{
-
-		const oAuth2Client = GoogleActions.authCheck(req.account, res);
-
-		//TODO: Change to service account to allow non-owners to view published files.
-		// Requires a driveId parameter in the drive.files.list command
-		// Then remove the `auth` parameter from the drive object initialization
-
-		const drive = google.drive({ version: 'v3', auth: oAuth2Client });
+	listGoogleBrews : async (auth)=>{
+		const drive = google.drive({ version: 'v3', auth });
 
 		const obj = await drive.files.list({
 			pageSize : 1000,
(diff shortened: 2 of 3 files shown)


calculuschild:
Ok, so essentially we have 3 different Authorization types (for my own notes here):

- **API_Key** Has limited access but doesn't require login. Can view other users' files created with HB, but not edit them.
- **Service Account** The Homebrewery's "Google Account". Has full edit access to all files created with HB as long as the files have the appropriate "write" permissions applied.
- **User Account** Has full access to all files created with HB as long as the files have the appropriate "write" permissions applied. Difference is User account is needed to apply the "write" permissions in the first place on a new file. We could theoretically use User Accounts in place of Service Accounts, but then people not logged in with Google would have no access.

So.... API_key shouldn't be needed anymore since it can't edit. Tasks should be split by:

#### User Account - *Things only the original user should be able to do*
`getGoogleFolder` - Locates the users' HB folder, or if not found, creates a new one.
`newGoogleBrew` - Creates a new brew in the users' HB folder, then applies write permissions so others can edit it via HB. We may want to change the permissions from `anyone` to only our Service Account email, although a quick test with this didn't work, and it didn't work in the past either for unknown reasons.
`listGoogleBrews` - Finds all documents a user owns that were created via HB. For a user, this is his own creations. A co-author would not see that file in the list. The Service Account would see no files at all, since it doesn't own any files. We can add a "driveId" to list files from a specific drive, or `q : 'folderID' in parents` to list files from a specific folder to view files that are not owned but shared to the user, in which case the Service Account *can* see those files. Thus: viewing brews on a different user's profile page requires either storing their driveId or folderId somewhere, or Ids of the individual files (in stubs). Viewing co-authored files that you are not the original owner of, won't work with the `driveId` or `folderId` route because that requires users storing lists of other people's drives to search, but stubs would still work. Conclusion: Don't use Service Account here because we won't be storing `driveId`s when stubs does the job. Keep User Accounts so un-stubbed brews are still visible and can be stubbed when visited.
`deleteGoogleBrew` - Unless.... we need to automatically handle brew ownership transfer down the line.

#### Service Account - *Things other users should be able to do*
`existGoogleBrew` - Just checks if a brew exists, and only used before updating. Probably not needed; error handling for missing files can happen in `updateGoogleBrews` itself.
`updateGoogleBrew` - Yep, needed for co-authors.
`readFileMetadata` - Needed to view brews that you don't own. Prolly should rename to `getGoogleBrew`.
`increaseView` - Needed so strangers can update the view count.

server/googleActions.js
```diff
@@ -81,13 +96,11 @@ const GoogleActions = {
 
 	listGoogleBrews : async (req, res)=>{
 
-		oAuth2Client = GoogleActions.authCheck(req.account, res);
+		const oAuth2Client = GoogleActions.authCheck(req.account, res);
```
calculuschild, line 99: `newGoogleBrew` has `auth` passed in externally from `homebrew.api` which uses `authCheck()` to generate the token. 

`listGoogleBrews` and `deleteGoogleBrew` still have the odd method of passing in the `req/res` objects instead and then running `authCheck` in here. Can we update them all to have the `authCheck` outside and oAuth token passed in? Or perhaps just pass in the `req.account` object? For consistency. Seems like errors can be thrown here, but actually adding errors to the to the `res` object should be happening in the api path.

I think making these consistent will help the next steps in the stubs update be cleaner.

server/googleActions.js
```diff
@@ -81,13 +96,11 @@ const GoogleActions = {
...
 
 		//TODO: Change to service account to allow non-owners to view published files.
 		// Requires a driveId parameter in the drive.files.list command
-		// const keys = JSON.parse(config.get('service_account'));
-		// const auth = google.auth.fromJSON(keys);
-		// auth.scopes = ['https://www.googleapis.com/auth/drive'];
+		// Then remove the `auth` parameter from the drive object initialization
```
calculuschild, line 103: Lets delete this. With Stubs coming we don't need to make this a Service Account (see my rant under `listGoogleBrews` on my [other comment](https://github.com/naturalcrit/homebrewery/pull/2026#issuecomment-1042019886))

server/googleActions.js
```diff
@@ -129,8 +142,8 @@ const GoogleActions = {
 	  return brews;
 	},
 
-	existsGoogleBrew : async (auth, id)=>{
-		const drive = google.drive({ version: 'v3', auth: auth });
+	existsGoogleBrew : async (id)=>{
```
calculuschild, line 145: I think this whole function can be deleted. We only use it to check if a brew exists before submitting an update, but `updateGoogleBrew` should already be handling that with an error response if we try to update a brew that doesn't exist.

calculuschild (changes requested):

server/googleActions.js
```diff
@@ -5,7 +5,22 @@ const { nanoid } = require('nanoid');
 const token = require('./token.js');
 const config = require('./config.js');
 
-//let oAuth2Client;
+const keys = typeof(config.get('service_account')) == 'string' ?
+	JSON.parse(config.get('service_account')) :
+	config.get('service_account');
+let serviceAuth;
+try {
```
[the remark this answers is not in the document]
jeddai: The line that loads the google credentials will fail if they don't exist, I put the try with the warning in place so devs and others who try to run it without the google creds won't run into issues.

server/googleActions.js
jeddai: Yep I can do that and have the auth passed in and have it throw errors where applicable. FWIW I'm really thinking the auth check should be done as middleware and applied to the request for subsequent middlewares, but I'll do that in a separate PR.

server/googleActions.js
```diff
@@ -5,7 +5,22 @@ const { nanoid } = require('nanoid');
...
+	serviceAuth = google.auth.fromJSON(keys);
+	serviceAuth.scopes = [
+		'https://www.googleapis.com/auth/drive',
+		'https://www.googleapis.com/auth/drive.appdata',
+		'https://www.googleapis.com/auth/drive.file',
+		'https://www.googleapis.com/auth/drive.metadata'
```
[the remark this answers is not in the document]
jeddai: Done!

server/app.js
```diff
@@ -200,13 +200,16 @@ app.get('/user/:username', async (req, res, next)=>{
...
-
-		if(googleBrews)
+		const auth = await GoogleActions.authCheck(req.account, res);
+		let googleBrews = await GoogleActions.listGoogleBrews(auth)
+			.catch((err)=>{
+				console.error(err);
+			});
+
+		if(googleBrews) {
+			googleBrews = googleBrews.map((brew)=>({ ...brew, authors: [req.account.username] }));
```
calculuschild, line 210:
```suggestion
```

I don't think this is the place to be adding the user to the authors. This should probably be handled on brew update or somewhere? Let's handle this later on when the stubs are more integrated.

server/googleActions.js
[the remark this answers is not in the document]
calculuschild: OK. Can we add a custom note to the error here then? Like "Google Service Account not set up properly in config" or something.

server/googleActions.js
```diff
@@ -248,9 +233,8 @@ const GoogleActions = {
 		return newHomebrew;
 	},
 
-	readFileMetadata : async (auth, id, accessId, accessType)=>{
-
-		const drive = google.drive({ version: 'v3', auth: auth });
+	readFileMetadata : async (id, accessId, accessType)=>{
```
calculuschild, line 236: Let's rename this function to `getGoogleBrew`.

server/app.js
jeddai: I did that because that's effectively what it was doing but it was doing it in the listGoogleBrews method using the same request member variable.

calculuschild: Mmmmmmmmmmmmmmmk. I'll keep it for now then but let's see if we can move this into the stub logic later on.

jeddai: Oh yeah it will not exist anywhere once stubs are in place

server/googleActions.js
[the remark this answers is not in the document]
jeddai: Yeah!

calculuschild (approved):

PR #2036, opened 2022-02-20
Add HTTP tests for /source route

AlexeySachkov:
This PR introduces a few tests for `/source` route, which check that the app is able to correctly retrieve brew sources and send them back to user. It also introduces a very basic mock for `GoogleActions` module so we can run HTTP tests without actual access to Google Drive servers.


tests/routes/source.test.js
```diff
@@ -0,0 +1,100 @@
...
+		return app.get('/source')
+            .send()
+            .expect(404);
+	});
+
+	// FIXME: add tests for retrieving a Google brew source
+
+	afterAll(()=>{
+		return Homebrew.model.deleteMany();
+	});
```
calculuschild, line 95: Um.... this deletes all brews in the database, not just the tested ones. Just ran it on my local drive and sure enough, everything was deleted. We definitely don't want this running on the live server.

AlexeySachkov: Good thing that you have spotted that. Actually, I would expect tests to be connected to a different DB, because they are launched under "test" node environment. Probably I don't fully understand the API to work with Mongo here.

I suggest we block this PR from an accidental merge by explicitly requesting changes

calculuschild (changes requested):
Test needs fixing to not delete the live database.

tests/routes/source.test.js
AlexeySachkov: @calculuschild , sorry that it took so long for me to respond here, but I finally figured it out: what you saw is indeed happening. The reason for that is that even in tests we by default connect to the main database.

In 6a2581e8d306feb67f51223255fe6bdafecd5da9 I added `config/test.json` to override database in test environment. 

However, there is still caveat: `jest` sets `NODE_ENV=test` _only_ if `NODE_ENV` is not set yet ([link to docs](https://jestjs.io/docs/environment-variables)). So, in local environment where you likely have `NODE_ENV=local`, it won't work.

I wonder what would be the preferred way of handling that? I doubt that ignoring database cleanup in local environment is a good idea, because someone who has local database with their brews won't be happy to see them all removed just by (possibly accidentally) running tests. We could update `package.json` to include `NODE_ENV=test` to all test-related commands, but I'm concerned that we might miss some, because there are 4 of them already.

dbolack-ab:
Do the two new commits fix the requested change?

calculuschild:
Still not fixed.

dbolack-ab:
Do we need this route?

calculuschild:
> Do we need this route?

I don't know how often `/source` gets used, but its been around for years at this point. I don't think we need to get rid of it given we just spent a lot of effort adding `/css`.

5e-Cleric:
I'd say we can close this, and maybe start from scratch, what is the issue? we should log it as an issue.

dbolack-ab:
> I'd say we can close this, and maybe start from scratch, what is the issue? we should log it as an issue.

Makes sense to me.

PR #2047, opened 2022-03-02
[WIP] Account Page

### LINKED ISSUE #2043: [UI Feature] Add diagnostic information to account menu
Commonly reported issues in the subreddit and here on GitHub revolve around users being unaware whether their account is connected to Google, or that their Google refresh token is expired.
I propose that we add an additional item to the new Account dropdown menu (added in PR #2015) to allow the user to display some of the information, decoded from the JWT cookie, in order to assist with diagnostics, self or otherwise.

Specifically, I suggest we show the following information:

- the account username (jwt.username)
- the token issue date (jwt.issued)
- if the account is linked to Google (jwt.googleId ? 'YES' : 'NO')
- if the account is linked to Google, time remaining until a log in refresh is required (jwt.googleId ? 'Time remaining to Google log in required: moment(jwt.issued).add(1,'y').toNow()' : '')
- possibly the shareId of the current brew, if one is loaded

This could currently be achieved through a simple `alert()` box, although we may also wish to create an information window that is more in line with the Homebrewery UI. It may be worth considering shifting the REPORT ISSUE button to the new Diagnostics window, and including the diagnostic information in the automatically generated Issue Report template.

I believe that such a feature would help reduce some of the current issue reports, and hopefully aid in the swift resolution of future issues.

G-Ambatte:
On completion, this PR resolves #2043.

Adds a new page `/account` which displays user information from the JWT token.

Images from current state of the PR:

Account menu:
image

Account Info page:
image

image

image

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -142,7 +142,7 @@ app.get('/migrate', async (req, res, next)=>{
 //Account Information page
 app.get('/account', async (req, res, next)=>{
 	let text = dedent`
-	#Account
+	# Account Details
 	
 	Not logged in.`;
 
@@ -169,16 +169,19 @@ app.get('/account', async (req, res, next)=>{
 		}
 
 		text = dedent`
-	# Account
-	
-	### Info:
-
-	Username : ${req.account.username}  
-	Logged in at: ${req.account.issued}  
-	Linked to Google? ${req.account.googleId ? 'YES' : 'NO'}  
-	${req.account.googleId ? `Google auth check: ${auth.credentials.access_token ? 'YES' : 'NO'}` : ''}  
-	${req.account.googleId ? `Google file list count: ${files ? files.length : 'NO'}` : ''}  
-	`;
+	# Account Details
+	:
+	{{wide,centered,text-align:center,border-color:#58180D,border-width:2px,border-style:solid,border-radius:10px,padding:10px
+	### Info
+	:
+	|||
+	|:-|:-|
+	|Username|${req.account.username}|
+	|Logged in at|${req.account.issued}|
+	|Linked to Google|${req.account.googleId ? 'YES' : 'NO'}|
+	${req.account.googleId ? `|Google auth check: ${auth.credentials.access_token ? '|YES|' : '|NO|'}` : ''}
+	${req.account.googleId ? `|Google file list count: ${files ? `|${files.length}|` : '|NO|'}` : ''}
+	}}`;
 	}
 
 	const brew = {
--- a/server/app.js
+++ b/server/app.js
@@ -8,6 +8,7 @@ const express = require('express');
 const yaml = require('js-yaml');
 const app = express();
 const config = require('./config.js');
+const dedent = require('dedent-tabs').default;
 
 const homebrewApi = require('./homebrew.api.js');
 const GoogleActions = require('./googleActions.js');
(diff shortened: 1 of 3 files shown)


G-Ambatte:
Attempts to implement Google permissions checking.

Signed into Homebrewery only:
image

Signed into Homebrewery only but account IS linked to Google:
image

Signed into Google but no permissions granted:
image

Signed into Google with permissions granted:
image

G-Ambatte:
Added a little styling to the info panel.

image

calculuschild:
I think we might need to agree on exactly what information we want to display here, and ensure it is all as clear as possible to the users what it means. For instance. "Google auth check" probably means nothing to most users, but "Google drive accessible" might be more obvious? Actually, there's probably still a better way to say that.

(Also, we might want to fancy up the styling a bit more with some icons and a less "spreadsheet" feel with the table styling. Make is as un-intimidating and not-confusing as possible. @Gazook89 might have some ideas.)

Eventually this page might be where we place password management, account deletion, etc., so even if we keep the PHB style, it might benefit from some obvious "Homebrewery" interface icons, much like we do in the user brews page with the black button panels and icons everywhere, etc.

image

G-Ambatte:
1. The PR started out as "what account information can be displayed that might be useful for troubleshooting" and grew from there; there is probably a considerable amount that can be added to clarify what each item is for the uninitiated.
2. Styling was originally kept as minimal as possible for brevity.
3. This needs to become it's own `/pages/account` page, which will probably link in to the `basePages/uiPage` PR.

It's currently on the back burner for me at the moment, but I think there's a reasonably clear direction to progress this PR once I can pick it back up again.

calculuschild:
1) Ok, let's stick with what you already have, and just work on naming them in the most "user-friendly" way so it's obvious what users are looking at. We can add more info later.
2) That's fine. I just don't want it to be confused with "a brew". We can probably borrow some styling from the User Page just to make it *slightly* more "Hombrewery UI".
3) Thumbs up.

jeddai:
I'll review in ~4-5 hrs

jeddai:
Is it intended that the account page doesn't have the `help` and `new brew` nav items?

server/app.js
```diff
@@ -138,6 +139,59 @@ app.get('/migrate', async (req, res, next)=>{
...
+	:
+	|||
+	|:-|:-|
+	|Username|${req.account.username}|
+	|Logged in at|${req.account.issued}|
+	|Linked to Google|${req.account.googleId ? 'YES' : 'NO'}|
+	${req.account.googleId ? `|Google auth check: ${auth.credentials.access_token ? '|YES|' : '|NO|'}` : ''}
+	${req.account.googleId ? `|Google file list count: ${files ? `|${files.length}|` : '|NO|'}` : ''}
+	}}`;
+	}
```
jeddai, line 185: I'm not certain how I feel about this templating vs writing a React template that accepts props. Was there a reason for choosing this strategy over React+route props?

G-Ambatte: As WIP, finding out what data was available to be presented, and what it represented in human terms was my first concern; presentation was a secondary concern.
A proper template should be created that is visually distinct from normal Homebrewery brews that ideally can be re-used for future UI pages - as `basePages/uiPage/uiPage.jsx`, probably with an associated `basePages/uiPage/uiPage.less`.

We should create a separate PR for creating a reusable UI page template that accepts props, which we can then populate with this data. This Draft PR was mostly a quick proof of concept, and once the scope widened, it stalled out.

jeddai: Oh you commented on that in the gitter chat! Okay sounds good!

G-Ambatte: Oh hey, I already did that, creating #2049.

calculuschild:
Superseded by #2161

Issue #1982, opened 2022-01-23
HTTP tests are pretty slow


AlexeySachkov:
#1979 introduces a basic HTTP testing and it those tests work very slowly: 5 basic tests to check that "200 OK" is returned from a particular route take ~5 seconds to complete.

5s is not so long, but once we add more tests, that slowness would be significant and annoying.

Several guesses were made in the PR:
1. `app.js` slowly loads files like `welcome_msg.md`, `changelog.md`, etc.
2. `jest` is using non-minified packages, which are huge
3. `jest` is re-creating all modules for each test (`it` function) to prevent one test affecting another

I've checked for (1) by commenting out files loading - nothing has changed.

According to the [official docs](https://jestjs.io/docs/next/configuration#resetmodules-boolean), modules are not being reset after each test by default.

`--verbose` mode shows the following data:
```
 PASS  tests/routes/static-pages.test.js (5.57 s)
  Tests for static pages
    √ Home page works (392 ms)
    √ Home page v3 works (24 ms)
    √ Changelog page works (10 ms)
    √ FAQ page works (9 ms)
    ○ skipped robots.txt works
```

So, the tests are pretty quick by themselves (except for the first one) and they take less than half a second to finish, whilst the whole suite took more than 5 seconds to finish. That data indeed suggests that something like (2) is happening, i.e. we are wasting time somewhere at initialization/startup phase.

The following experiment:
```javascript
var startTime = performance.now();

// Mimic https responses to avoid being redirected all the time
const app = supertest.agent(require('app.js').app)
    .set('X-Forwarded-Proto', 'https');

var endTime = performance.now();

console.debug(`Initialization of the app and supertest took ${endTime - startTime} milliseconds`);
```

Shows the following results:
```
  console.debug
    Initialization of the app and supertest took 2992.700200021267 milliseconds

      at Object.<anonymous> (tests/routes/static-pages.test.js:11:9)
```

So, it took ~3 seconds to create an `app` instance and ~0.5 seconds to complete tests, which leaves ~1.5 more seconds which were spent somewhere else (`require('supertest')` only takes ~0.25s)

5e-Cleric:
Can i get feedback on this? @calculuschild test run pretty smoothly considering how many there are, was any of this fixed? Or was the issue exaggerated?

calculuschild:
This is specifically the `tests/routes/static-pages.test.js ` which haven't been touched in a while.

[...]

PR #2052, opened 2022-03-07
[LOCAL ONLY] Add passwordless login

### LINKED ISSUE #269: Login on locally running version does not work
I run Homebrewery locally without any issues except when I try to login. I get re-directed to the NaturalCrit.com login page. Should the login process be self-contained? I login into the Natural Crit page and I get re-directed to my local copy but the login has failed. How do I create logins locally?

### Additional Details

**Share Link** :

or

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

PASTE BREW CODE HERE

</pre></code></details>

G-Ambatte:
This PR adds the ability to log in without a password on local installations only. It also implements a new global object `config` which - at the time of this PR - contains only the NODE_ENV (in property `environment`) and JWT token encoding secret key (in property `secret`).

This PR resolves #269.


G-Ambatte:
@AlexeySachkov do you have any idea why this PR will only pass testing if the timeout of the first test is increased to more than 5s? It looks like it completed in just under 12s (https://app.circleci.com/pipelines/github/G-Ambatte/homebrewery/701/workflows/fde4f457-e058-48c5-8186-060061d9eca5/jobs/700)

I tried splitting out the test stage from the build stage (see https://app.circleci.com/pipelines/github/G-Ambatte/homebrewery?branch=experimental-CI_update) but even so, the first of the route tests takes about 5 or 6s to complete, while the rest complete in literally milliseconds.

Any thoughts? Is this related to the PR at all, or is this generalized testing weirdness???

AlexeySachkov:
> @AlexeySachkov do you have any idea why this PR will only pass testing if the timeout of the first test is increased to more than 5s? It looks like it completed in just under 12s (https://app.circleci.com/pipelines/github/G-Ambatte/homebrewery/701/workflows/fde4f457-e058-48c5-8186-060061d9eca5/jobs/700)
> 
> I tried splitting out the test stage from the build stage (see https://app.circleci.com/pipelines/github/G-Ambatte/homebrewery?branch=experimental-CI_update) but even so, the first of the route tests takes about 5 or 6s to complete, while the rest complete in literally milliseconds.
> 
> Any thoughts? Is this related to the PR at all, or is this generalized testing weirdness???

I think it is generalized testing weirdness which has little to do with your PR, see #1982. It seems like tests takes forever to be initialized/set up and I don't know the exact reason for that. One of the reasons could be that modules loading is slow under `jest`, but I really doubt that your changes caused 2x slowdown - CI machine is likely a non-physical machine, which means that its performance could be unstable as well.

I think it should be ok to increase the timeout, but I suggest we do that globally through [jest config](https://jestjs.io/docs/configuration#testtimeout-number) in `package.json`:
https://github.com/naturalcrit/homebrewery/blob/c9935fa45ccc502ec0e454ee090471a9c88be2aa/package.json#L33-L39

G-Ambatte:
> > @AlexeySachkov do you have any idea why this PR will only pass testing if the timeout of the first test is increased to more than 5s? It looks like it completed in just under 12s (https://app.circleci.com/pipelines/github/G-Ambatte/homebrewery/701/workflows/fde4f457-e058-48c5-8186-060061d9eca5/jobs/700)
> > I tried splitting out the test stage from the build stage (see https://app.circleci.com/pipelines/github/G-Ambatte/homebrewery?branch=experimental-CI_update) but even so, the first of the route tests takes about 5 or 6s to complete, while the rest complete in literally milliseconds.
> > Any thoughts? Is this related to the PR at all, or is this generalized testing weirdness???
> 
> I think it is generalized testing weirdness which has little to do with your PR, see #1982. It seems like tests takes forever to be initialized/set up and I don't know the exact reason for that. One of the reasons could be that modules loading is slow under `jest`, but I really doubt that your changes caused 2x slowdown - CI machine is likely a non-physical machine, which means that its performance could be unstable as well.
> 
> I think it should be ok to increase the timeout, but I suggest we do that globally through [jest config](https://jestjs.io/docs/configuration#testtimeout-number) in `package.json`:
> 
> https://github.com/naturalcrit/homebrewery/blob/c9935fa45ccc502ec0e454ee090471a9c88be2aa/package.json#L33-L39

Cool, I'll put it into a separate PR, probably tomorrow at this stage

G-Ambatte:
I have removed the test timeout changes from this PR and raised #2094 to apply the timeout change globally as suggested. However, as the test still takes >5s to complete, it is now failing again. I am extremely confident that this PR will pass the tests once #2094 is merged.

server/app.js
```diff
@@ -262,14 +262,26 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
...
+	// Create configuration object
+	const configuration = {
+		local       : false,
+		environment : config.get('node_env')
+	};
+	// Add local only items to configuration object
+	const localEnvironments = config.get('local_environments');
+	if(localEnvironments.includes(configuration.environment)){
+		configuration.local = true;
+		configuration.secret = config.get('secret');
```
calculuschild, line 274: We really shouldn't be passing `secret` along to the client browser. Even in this case where it's purely for local logins, I am wary of potential abuses of this in the future.

Instead, the `localLogin` function in `account.jsx` should send a request to the server. See the Naturalcrit Repo for an example: https://github.com/naturalcrit/naturalcrit/blob/6e62368f807596dcafcbd5bcad5181bce2982455/client/naturalcrit/account.actions.js#L5-L15

Even if we don't require a password, the login should still go through the server instead of exposing key variables to the browser.

server/app.js
```diff
@@ -262,14 +262,26 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
...
 	const props = {
 		version     : require('./../package.json').version,
 		url         : req.originalUrl,
 		brew        : req.brew,
 		brews       : req.brews,
 		googleBrews : req.googleBrews,
 		account     : req.account,
-		enable_v3   : config.get('enable_v3')
+		enable_v3   : config.get('enable_v3'),
+		config      : configuration
```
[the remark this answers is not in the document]
G-Ambatte: My first thought was anything that gets pulled via `config.get()` - the only exception to this being `enable_v3` as I didn't want to mess with existing functionality.
However this could easily be extended to user configuration; the user's preferred default renderer, theme, etc. However while this change would lay the groundwork for such features, they are well outside the scope of this PR.

server/app.js
G-Ambatte: This should now be implemented in today's changes.

client/homebrew/navbar/account.navitem.jsx
```diff
@@ -28,7 +29,29 @@ const Account = createClass({
 		};
 	},
 
+	localLogin : async function(){
+		const username = prompt('Enter username:');
+		if(!username) {return;};
```
calculuschild, line 34:
```suggestion
		if(!username) {return;}
```

client/homebrew/navbar/account.navitem.jsx
```diff
@@ -28,7 +29,29 @@ const Account = createClass({
...
+		const expiry = new Date;
+		expiry.setFullYear(expiry.getFullYear() + 1);
+
+		const token = await request.post('/login')
+				.send({ username })
+				.then((response)=>{
+					return response.body;
+				})
+				.catch((err)=>{
+					return null;
```
calculuschild, line 45:
```suggestion
					console.warn(err);
```

G-Ambatte: Committed both suggested changes, and resolved the new merge conflict.

server/app.js
```diff
@@ -259,9 +259,36 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
 	return next();
 }));
 
+const localEnvironments = config.get('local_environments');
+const nodeEnv = config.get('node_env');
```
calculuschild, line 263:
```suggestion
const nodeEnv = config.get('node_env');
const isLocalEnvironment = config.get('local_environments').includes(nodeEnv);
```

server/app.js
```diff
@@ -259,9 +259,36 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
 	return next();
 }));
 
+const localEnvironments = config.get('local_environments');
+const nodeEnv = config.get('node_env');
+// Login
+app.post('/login', (req, res)=>{
+	// Local only
+	if(!localEnvironments.includes(nodeEnv)){
```
calculuschild, line 267:
```suggestion
	if(!isLocalEnvironment){

```

server/app.js
```diff
@@ -259,9 +259,36 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
...
 //Render the page
 const templateFn = require('./../client/template.js');
 app.use((req, res)=>{
+	// Create configuration object
+	const configuration = {
+		local       : false,
+		environment : nodeEnv
+	};
+	// Add local only items to configuration object
+	if(localEnvironments.includes(nodeEnv)){
```
calculuschild, line 290:
```suggestion
	if(isLocalEnvironment){

```

server/app.js
```diff
@@ -259,9 +259,36 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
...
+		configuration.local = true;
+	};
```
calculuschild, line 292:
```suggestion
	const configuration = {
		local       : isLocalEnvironment,
		environment : nodeEnv
	};
```

server/app.js
```diff
@@ -259,9 +259,33 @@ app.get('/print/:id', asyncHandler(async (req, res, next)=>{
 	return next();
 }));
 
+const nodeEnv = config.get('node_env');
+const isLocalEnvironment = config.get('local_environments').includes(nodeEnv);
+// Login
+app.post('/login', (req, res)=>{
```
calculuschild, line 265: What do you think about renaming this route to `/localLogin` or something?

G-Ambatte: I've just been experimenting with renaming the route `/local/login` and wrapping it a `isLocalEnvironment` check - so the route shouldn't even get loaded in Production.

G-Ambatte: Pushed the commit now

calculuschild:
Nice.... This is looking pretty good. As @jeddai said, we can look later at getting Google drive integration but that can be a separate PR. I think this is ready to merge?

G-Ambatte:
I think we're good!

This one has been a bit of a journey, not least because it's in the top five oldest open issues.

calculuschild (approved):

PR #2055, opened 2022-03-09
Add divider reset on window resize #2053

### LINKED ISSUE #2053: [UI] Reset divider position on browser window resize
Some devices - particularly mobile - can have the divider position change rapidly, effectively losing the divider. This is most noticeable when the switching between portrait and landscape modes on the device; this is further aggravated by the divider not dragging nicely on touch devices (again, particularly mobile).

Suggested solution: Have the divider position reset to 50% of screen width on browser window resize.

G-Ambatte:
This PR resets the divider position to 50% whenever a browser window resize event is detected.

This should resolve #2053.

### DIFF
--- a/shared/naturalcrit/splitPane/splitPane.jsx
+++ b/shared/naturalcrit/splitPane/splitPane.jsx
@@ -10,7 +10,6 @@ const SplitPane = createClass({
 		return {
 			storageKey   : 'naturalcrit-pane-split',
 			onDragFinish : function(){} //fires when dragging
-
 		};
 	},
 	getInitialState : function() {
@@ -26,6 +25,17 @@ const SplitPane = createClass({
 				size : paneSize
 			});
 		}
+		window.addEventListener('resize', this.resetSize);
+	},
+
+	componentWillUnmount : function() {
+		window.removeEventListener('resize', this.resetSize);
+	},
+
+	resetSize : function() {
+		this.setState({
+			size : window.innerWidth / 2
+		});
 	},
 
 	handleUp : function(){
--- a/shared/naturalcrit/splitPane/splitPane.jsx
+++ b/shared/naturalcrit/splitPane/splitPane.jsx
@@ -26,6 +26,7 @@ const SplitPane = createClass({
 		if(paneSize){
 			this.setState({
 				size        : paneSize,
+				dividerSize : paneSize,
 				screenWidth : window.innerWidth
 			});
 		}
@@ -37,19 +38,24 @@ const SplitPane = createClass({
 	},
 
 	changeSize : function() {
-		const oldWidth = this.state.screenWidth;
 		const oldLoc = this.state.size;
-		const newLoc = this.limitPosition(window.innerWidth * (oldLoc / oldWidth));
+		const oldWidth = this.state.screenWidth;
+		let newLoc = oldLoc;
+		// Allow divider to increase in size to original position
+		if(window.innerWidth > oldWidth) {
+			newLoc = Math.min(oldLoc * (window.innerWidth / this.state.screenWidth), this.state.dividerSize);
+		}
+		// Limit current position to between 10% and 90% of visible space
+		newLoc = this.limitPosition(newLoc, 0.1*(window.innerWidth-13), 0.9*(window.innerWidth-13));
+
 		this.setState({
 			size        : newLoc,
 			screenWidth : window.innerWidth
 		});
 	},
 
-	limitPosition : function(x) {
-		const minWidth = 1;
-		const maxWidth = window.innerWidth - 13;
-		const result = Math.min(maxWidth, Math.max(minWidth, x));
+	limitPosition : function(x, min = 1, max = window.innerWidth - 13) {
+		const result = Math.round(Math.min(max, Math.max(min, x)));
 		return result;
 	},
 
@@ -71,7 +77,8 @@ const SplitPane = createClass({
 
 		const newSize = this.limitPosition(e.pageX);
 		this.setState({
-			size : newSize
+			size        : newSize,
+			dividerSize : newSize
 		});
 	},
 	/*
--- a/shared/naturalcrit/splitPane/splitPane.jsx
+++ b/shared/naturalcrit/splitPane/splitPane.jsx
@@ -12,12 +12,14 @@ const SplitPane = createClass({
 			onDragFinish : function(){} //fires when dragging
 		};
 	},
+
 	getInitialState : function() {
 		return {
 			size       : null,
 			isDragging : false
 		};
 	},
+
 	componentDidMount : function() {
 		const paneSize = window.localStorage.getItem(this.props.storageKey);
 		if(paneSize){
--- a/shared/naturalcrit/splitPane/splitPane.jsx
+++ b/shared/naturalcrit/splitPane/splitPane.jsx
@@ -15,42 +15,36 @@ const SplitPane = createClass({
 
 	getInitialState : function() {
 		return {
-			size        : null,
-			screenWidth : 0,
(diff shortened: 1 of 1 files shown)


Gazook89:
just commenting to say i support this approach to resetting the divider.  I do still think that it should be possible to prevent the divider from ever disappearing, but with the divider being the same width as the scrollbar it might actually be a hassle?  Also, this approach efficiently works with mobile devices.  

I have nothing of importance to comment on about the actual code, though.

calculuschild:
It works... but I don't love the idea of my divider position suddenly snapping out of place every time I resize. I'm picturing this thought process:

> "Alright, I got my divider positioned just where I want it at the 2/3 position. Ah, I guess I should stretch the window a little taller so I can see more of the editor. What!? Why did the divider jump back to the middle!? I'd better report this new bug!"

We already tried to handle this issue once before with #1760 , however we only went part-way by preventing the user from dragging the divider beyond the window edge. What we missed, was the user dragging the window edge beyond the divider (or some other action causing the window to shrink past the divider).

So, instead of jumping to exactly 50%.... perhaps we can calculate the current percent position and scale that position? So if it's at 80% and you resize, the position is still 80% of whatever the new size is? Plus or minus whatever pixel values to make sure the divider never goes below the minimum/maximum position?

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -26,6 +25,18 @@ const SplitPane = createClass({
 				size : paneSize
 			});
 		}
+		window.addEventListener('resize', this.resetSize);
```
calculuschild, line 28: This should happen in the `componentDidMount` step instead.

G-Ambatte: It was - however the lack of a blank line between `getInitialState` and `componentDidMount` was causing the GitHub diff tool to only show the top of the block of text, which was `getInitialState`.

calculuschild: Oh hah. I'm dumb. Viewing this on mobile makes it even worse.

G-Ambatte:
Percentage based divider positioning on browser window resizing in working in this PR - at the very least, it should keep the pane divider in the visible area.
I suspect that there is still value in a reset function, but this can probably live in the Account or Settings page, which I think has already been suggested elsewhere.

ericscheid:
My 2c .. I typically will drag the divider so I have at least full width of the rendered page in the preview pane (which varies according to browser zoom and whether I have A4 css code in place). Having to jiggle the preview pane left-to-right to see the preview is just awful.

That said, I will also sometimes drag the divider to have a narrower editor pane, simply because I don't want to have very long lines of paragraph text to scan properly. I do the opposite though if the brew text is some chonky table I'm editing (and thus do want to see an entire table row on just one line).

Never is my thought process "ah, for aesthetics I want the divider at 33% / 66%".

I would be happy if the divider generally stayed put, only shifting if the resized window is *narrower* (possibly even with a margin to allow for scroll bars etc- thus a window 800px wide would clamp the divider to 750px).

calculuschild:
> Never is my thought process "ah, for aesthetics I want the divider at 33% / 66%".

I didn't literally mean choosing a position based on percent; that was just an example. I could rephrase that to be:

> "Alright, I got my divider positioned just where I want it **with the preview panel fully visible**. Ah, I guess I should stretch the window a little taller so I can see more of the editor. What!? Why did the divider jump back to the middle!? I'd better report this new bug!"

I guess we have 4 options for how to respond to resizing :

1. Do not adjust the divider at all. (current behavior, possible to size the window smaller than the divider position)
2. Force the divider into a predefined position (initial behavior of this PR)
3. Adjust the divider to maintain the current percent (this PR currently)
4. Keep the editor pane at its current size, expanding/shrinking the preview pane. (@ericscheid 's solution)
5. Keep the preview pane at its current size, expanding/shrinking the editor pane.

@G-Ambatte Already has a `limit` function set up to ensure the min/max position is within some distance from the window edge, and that should work for any of these approaches.

I'm fine changing it to behavior 4 if that makes more sense to people. The only concern I have is this possible case:

> I have my editor panel just at the size I want it, nice and big so I can see the whole table. (minutes later...) Ah, let me snap the window to the side of the desktop for a moment so I can see some other document side-by-side (window shrinks to smaller than the previous editor panel width, so only the editor is visible and the divider is at the right edge of the window). And now un-snap! Oh, my editor window is still small. Now I have to *laboriously drag it back out* so I can see the whole table again.

Not that it's even really an issue. Just that keeping the current percent "saves" your position regardless of how much you grow and shrink the page, while keeping the editor panel width loses your position as you shrink the window too far.

Maybe if we go that route, we could store the last "user-set" divider position. Then when resizing the window, it will shrink if needed, but as long as the user doesn't touch the divider again, resizing the window will allow the divider to stretch back to the previously-set position.

Gazook89:
Conversations like this make me glad this is a hobby for me, not my job :)

calculuschild:
@G-Ambatte What do you think? Does it make more sense to attempt to preserve the editor panel size rather than preserve the percent ratio?

G-Ambatte:
I think percentage based divider positioning makes sense **IF** the elements between the divider also scale - so reducing the size of the browser window reduces the size of all elements, so that they stay the same size _relative to each other._

Where the elements do NOT resize/zoom/whatever as the external window changes dimensions, I think the most intuitive solution is to give priority to an element to remain at its original size while the other elements shrink and adjust around it. As the primary purpose of the EditPage is to access and view the Editor, it would make sense that this would be the priority element.

---

I think the best approach right now would be to implement option 4 (static editor size), try it out, and if we hate it, change to option 3 (percentage) with element scaling.

calculuschild:
Option 4 then. Can you make that work @G-Ambatte ?

G-Ambatte:
I'll try to put something together tonight - in short, the divider position will only be changed if the new maximum divider position (`innerWidth - 13`, from memory) is less than the current divider position, in which case it will be reduced to the new maximum.

I suspect that there may be value in allowing the divider position to increase to the original position (possibly stored in `this.props.size`?) but I would need to think about that some more to determine how I would implement such a feature. I'll worry about that later, when I'm actually in front of a development machine.

G-Ambatte:
Damn - I thought I had pushed a commit with this working as described before I went off an a tangent and added other features, but apparently not.
I'll revert and push the base commit later today.

G-Ambatte:
I've just pushed a commit that does the following:
- during window resize, divider position is limited to between 10% and 90% of visible horizontal space, to eliminate "The divider has disappeared!" issues
- if you shrink the window, making the editor smaller, then increase the window again, the divider can return to its original position, but will not grow larger

I started to add buttons to allow divider reset functionality, but I feel that this is sufficiently outside the original scope of this Issue/PR that a new one should be created.

calculuschild:
Went through and made some small changes:
- renamed the variables/functions for clarity
- simplified the window resizing logic
- added divider position limiting in `componentDidMount`. Noticed you could shrink the window really small, but since that doesn't update the users' last set divider position in `localStorage`, refreshing the page will cause the divider to reload back to it's position off screen.

@G-Ambatte Any final comments? I think your method works fine so I'm ready to merge this.

G-Ambatte:
> @G-Ambatte Any final comments? I think your method works fine so I'm ready to merge this.

Looks good to me!

I think the next development on the divider bar is to create a way to set it's position - max left, max right, center - but having multiple options implies multiple buttons which implies a toolbar which I believe is more than sufficient scope creep to be a separate PR entirely.

calculuschild:
Thanks @G-Ambatte  for this PR!

PR #2056, opened 2022-03-09
Reset divider position on log out

### LINKED ISSUE #2054: [UI] Reset divider position on logout
This issue is to track the addition of code to reset the divider position when a signed in user logs out.

This is related to but separate from #2053.

G-Ambatte:
This PR resets the divider position when the user logs out by clearing the local storage item `naturalcrit-pane-split`.

This resolves #2054.

### DIFF
--- a/client/homebrew/navbar/account.navitem.jsx
+++ b/client/homebrew/navbar/account.navitem.jsx
@@ -20,6 +20,9 @@ const Account = createClass({
 
 	handleLogout : function(){
 		if(confirm('Are you sure you want to log out?')) {
+			// Reset divider position
+			window.localStorage.removeItem('naturalcrit-pane-split');
+			// Clear login cookie
 			document.cookie = `nc_session=;expires=Thu, 01 Jan 1970 00:00:01 GMT;path=/;samesite=lax;${window.domain ? `domain=${window.domain}` : ''}`;
 			window.location = '/';
 		};


Gazook89:
Would it be worth making any `/new` document reset the slider as well?  Create a new document, the slider is reset?

I mention it because if a user does not have an account, and runs into this issue, they won't be able to log out.   Now, if your other related PR goes through with the 'reset on window resize', there is another method of accomplishing a reset.  And to me this concern i'm raising here feels a little like a real 'fringe' concern....but raising it just in case.

G-Ambatte:
This was the method discussed with @calculuschild last night. The reset on browser resize was something I thought of while driving this morning, as it seems that some users do prefer to create while offline.
I suspect that reset divider position on `/new` is probably less likely to be an issue - I like to have my divider at about 33% of my main monitor (16:9 widescreen), but if I throw it to my small second monitor (4:3) then 50% is a better position.

I suspect that the final solution may be to implement UserInfo and add divider position as a user-specific editor setting, although in my case, it varies from machine to machine and even monitor to monitor. That's not to say that we couldn't implement a percentage based divider position preference with multiple settings for the browser window's horizontal width... Just that developing such a system is probably going to be a long way down the priority list.

Gazook89:
> I suspect that the final solution may be to implement UserInfo

This wouldn't help if you aren't a registered user, though.

Agreed about `/new`, though; I don't think I thought that through.  What I had in mind would require presume that divider position is tracked per brew, not user, which wouldn't make any sense.  So forget that.

calculuschild:
Ok, so if #2055 goes through, do we still feel like we need this solution? (If we prevent the issue from occurring in the first place?)

G-Ambatte:
I feel like divider location can probably be considered a user specific decision, and thus resetting it - and any other user specific decisions like theme or whatever else we come up with in the future - should probably be reset to the defaults on logout. However that line of thinking revolves around getting those user specific decisions stored in an internal storage mechanism and applied on user log in, which is a much larger issue that the one currently in question.

calculuschild:
Ok, I will go ahead and merge this. You are right that it makes sense to kind of "reset" everything back to defaults on logout.

PR #2062, opened 2022-03-11
Remove post title when opening issue on Reddit; add leading questions

Gazook89:
This PR removes the automatic addition of *[Issue] Describe Your Issue Here* from new reddit posts created from the HB link in the navbar.  Users will need to enter their own title, rather than being able to just click *Submit* without adding a descriptive title since this is a required field on reddit.

This PR also adds default text to the main text area of a post in an effort to get more descriptive issues and anticipate some initial questions:

image

I think the first 3 questions are common starting points to helping many users (though not all).  Others questions could be asked up front as well....possibly in such a way that Automod could pickup on the responses and offer some troubleshooting tips right out the gate.  But that might be overkill/annoying.

### DIFF
--- a/client/homebrew/navbar/help.navitem.jsx
+++ b/client/homebrew/navbar/help.navitem.jsx
@@ -1,6 +1,7 @@
 const React = require('react');
 const createClass = require('create-react-class');
 const _ = require('lodash');
+const dedent = require('dedent-tabs').default;
 
 const Nav = require('naturalcrit/nav/nav.jsx');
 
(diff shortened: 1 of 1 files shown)


client/homebrew/navbar/help.navitem.jsx
```diff
@@ -10,7 +10,7 @@ module.exports = function(props){
 			need help?
 		</Nav.item>
 		<Nav.item color='red' icon='fas fa-fw fa-bug'
-			href={`https://www.reddit.com/r/homebrewery/submit?selftext=true&title=${encodeURIComponent('[Issue] Describe Your Issue Here')}`}
+			href={`https://www.reddit.com/r/homebrewery/submit?selftext=true&text=Browser%28s%29%3A++%0D%0AOperating+System%3A++%0D%0ALegacy+or+v3+renderer%3F%3A++%0D%0AIssue%3A++`}
```
calculuschild, line 13: can we use `encodeURIComponent` to make this more legible?

Gazook89: Yeah, I wasn't sure how to get the line breaks in there and wasn't 100% confident on dedent so skipped it.  But I should have at least tried rather than use an online URL encoder.  

Will back with that in a bit (hopefully).

Gazook89: Okay this is done with b1932dc

Gazook89:
So with bold text, is that going to work in New Reddit? I had done the same, but found that when it created the post it just showed the asterisks because by default new reddits comment field doesn't use markdown. But maybe I was wrong about that.

calculuschild:
Hm, I didn't think to check that. I'll look and revert that if it seems broken.

Gazook89:
Hm, it works for me now?  Maybe I had missed an asterisk or something earlier....I'm assuming at some level reddit is automagically determining it is markdown.  Whatever, glad it works.

calculuschild:
Cool 👍

PR #2065, opened 2022-03-12
Small changes to theme swapper

Gazook89:
Forewarning:  Not sure i've ever opened a PR on another branch that isn't master/main and isn't my own, so let me know if I got this wrong.

I was looking at this branch before jumping in on a theme and a few things caught my eye, addressed in this PR:

1. A simple typo on the theme dropdown when in Legacy
2. The theme texture `mask-image` is applied where it can mask the text of the theme name if metadata panel is small enough (and doesn't even really need to be that small).
3. Opening the dropdown menu pushes the subsequent metadata panel items down, until the dropdown is closed.

Sorry it's 3 different items in one PR, they just seemed small enough to bundle together and seems likely that if you want to accept them, you can just change it on your side without doing a merge.

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -151,7 +151,7 @@ const MetadataEditor = createClass({
 			dropdown =
 				<Nav.dropdown className='disabled' trigger='disabled'>
 					<div>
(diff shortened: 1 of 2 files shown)


calculuschild:
Thanks for these fixes. Yes, this will just apply onto the other branch, so it works as you would expect.

Instead of having the graphic shrink and stretch, leading to a distorted image, I just have the image position shift out of the way when the panel is too narrow so as not to cover the text.

PR #2085, opened 2022-03-22
Unset `:first-line` small-caps rule with *Remove Drop Cap* snippet

Gazook89:
This PR simply adds a bit to the v3 "Remove Drop Cap" snippet that also removes the `:first-line` rule, so that the small-caps goes away, too.   Questions about the capitalized letters in the first line seem to come up occasionally in the subreddit (some assume it is a glitch). 

And in the DOMT server today they made a pinned post about it, which is what spurred this PR:

image

If for some reason someone wants no drop-cap, but retain the small-caps, it should be fairly evident how to do that (delete the relevant rule).


calculuschild:
Easy enough. Merged!

PR #2089, opened 2022-03-24
Update app to use a single endpoint for google and non-google stored brews

jeddai:
I'm thinking maybe the Homebrew API methods that are <action>LocalBrew and <action>GoogleBrew might could be moved out of the api file and put in their own file, but that's really the only thing I'm still thinking about. Otherwise I think this is good to go!

See #2019


server/homebrew.api.js
```diff
@@ -41,154 +41,195 @@ const excludePropsFromUpdate = (brew)=>{
...
 
-const newGoogleBrew = async (req, res, next)=>{
-	let oAuth2Client;
+const updateBrew = async (req, res)=>{
+	let brew = excludePropsFromUpdate(req.body);
+	const { transferToGoogle, transferFromGoogle } = req.query;
 
-	try {	oAuth2Client = GoogleActions.authCheck(req.account, res); } catch (err) { return res.status(err.status).send(err.message); }
+	let saved;
+	if(!!brew.googleId && transferFromGoogle) {
```
AlexeySachkov, line 112: Excuse me for my ignorance, but what does `!!brew.googleId` mean in JavaScript? I understand what it means in C, but not in JS.

My concern is that if it is some kind of a tricky rare construct, then it might be better to refactor it to make the code more readable

calculuschild: Doesn't Javascript already implicitly cast all variables to booleans when evaluated in an `if` statement? I.e,. `if(brew.googleId)` would evaluate false for all of those examples you listed anyway.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -200,73 +200,20 @@ const EditPage = createClass({
...
+			.catch((err)=>{
+				console.log('Error Updating Local Brew');
+				this.setState({ errors: err });
+			});
+
+		this.savedBrew = res.body;
+		if(this.state.saveGoogle && transfer) {
+			history.replaceState(null, null, `/edit/${this.savedBrew.googleId}${this.savedBrew.editId}`);
+		} else if(transfer) {
+			history.replaceState(null, null, `/edit/${this.savedBrew.editId}`);
```
calculuschild, line 216:
```suggestion
		if(transfer) {
			history.replaceState(null, null, `/edit/${this.savedBrew.googleId ?? ''}${this.savedBrew.editId}`);
```

client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -157,45 +157,24 @@ const NewPage = createClass({
...
-				localStorage.removeItem(STYLEKEY);
-				localStorage.removeItem(METAKEY);
-				window.location = `/edit/${brew.editId}`;
-			});
-		}
+		brew = res.body;
+		localStorage.removeItem(BREWKEY);
+		localStorage.removeItem(STYLEKEY);
+		localStorage.removeItem(METAKEY);
+		window.location = `/edit/${brew.googleId ? brew.googleId : ''}${brew.editId}`;
```
calculuschild, line 177:
```suggestion
		window.location = `/edit/${brew.googleId ?? ''}${brew.editId}`;
```

calculuschild:
Ok, so.... this all looks to be in order? I haven't tested it yet but all your changes make sense. Just those couple small suggestions above.

Also, make sure your NPM version is updated. The `package-lock.json` file in this PR has changed to "version 1 lockfile" which is older.

server/homebrew.api.js
jeddai: Removed both uses of the `!!`

client/homebrew/pages/editPage/editPage.jsx
jeddai: Done!

client/homebrew/pages/newPage/newPage.jsx
jeddai: Done!

jeddai:
Switched my node version to 16.14.0, seems to use the package-lock v2 now. Should be good for testing!

G-Ambatte:
Unifying brew handling to a single interface should simplify the basePage editorPage PR (#1800/#1801) as well - handling saving was one of the points of significant complexity.

calculuschild:
This all seems to work after testing. I'll go ahead and merge this. 🤞 

One small edge case that I'll make a separate issue for: since this keeps the same IDs between Google Drive/ Homebrewery when transferring the file, it's possible to end up with a `duplicate ID` error from Mongo if you transfer from Google (Google Brew lands in the Trash), then visit the old link to that trashed file/restore the trashed file so it shows up in your user page, attempt to transfer that file back to the Homebrewery as well. The error however manifests as "your Google Credentials are expired" popup. I am doubtful it will ever come up but thought I'd document it.

PR #2090, opened 2022-03-24
Add UserPage filtering to/from URL

### LINKED ISSUE #1858: update window.location when using filters on user page
Now we have a simple text filter on the user page, it sure would be helpful if that could persist through a reload or history traversal. Additionally, it would be useful for the window.location to reflect same, such that we now have a custom link we can share.

a filtered user page

- [ ] update window.history via `history.replaceState()` when search box typed into
- [ ] user page detects `?q=criteria` in browser.location on page load: stuff that into the search box, do the search
- [ ] also do this for tags, when we get around to having tags

### LINKED ISSUE #2023: "Search" or "Filter" ?
We have used two names for one thing in the user interface. This is not a problem now, but could lead to problems down the track when we start adding more functionality of similar-but-not nature, and we want to refer to one of those but not get it confused with the other.

I would suggest we be consistent now to refer to the widget on the user page page (and future public pages) into which the user can enter criteria (currently just a string) to dynamically update the present view (i.e. hide non-matching brews) .. refer to it as "Filter".

Later, when a facility which is "enter criteria, submit request, wait for server to respond" could be termed a "search". (Such a thing would likely be used for "Search all published brews", and not be dynamic.)

So .. "filter" is dynamic in-browser, "search" is command-and-response involving server.

image

G-Ambatte:
This PR adds the ability to specify a `filter` parameter to the `/user` URL, which causes the results to be filtered by the passed parameter.
While typing into the text box does not update the URL, pressing the Enter key causes the browser's current URL to update to reflect the contents of that input field in the filter parameter. Pressing Enter on an empty input field results in the filter parameter being removed from the URL.

These changes will resolve #1858.

---

I also changed one piece of verbiage from "search" to "filter", which resolves #2023. The only place that the term "search" is used is in FontAwesome class names or the names of built-in functions, neither of which we are able to change.

---

EDIT: After feedback from @ericscheid, I have changed the `window.location.replace` to `history.replaceState`. This allowed the URL to update without causing a page reload, so I have removed the URL update on pressing Enter and switched to updating the URL live.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -77,13 +77,8 @@ const ListPage = createClass({
 		this.setState({
 			filterString : e.target.value,
 		});
+		this.updateUrl(e.target.value);
 		return;
-	},
-
-	handleKeys : function(e){
-		if(e.key === 'Enter') {
-			this.updateUrl(e.target.value);
-		  }
 	},
 
 	updateUrl : function(filterTerm){
@@ -93,7 +88,7 @@ const ListPage = createClass({
 		urlParams.set('filter', filterTerm);
 		if(!filterTerm) urlParams.delete('filter');
 		url.search = urlParams;
-		window.location.replace(url.href);
+		window.history.replaceState(null, null, url);
 	},
 
 	renderFilterOption : function(){
@@ -105,8 +100,6 @@ const ListPage = createClass({
 					autoFocus={true}
 					placeholder='filter title/description'
 					onChange={this.handleFilterTextChange}
-					onFocus={(e)=>{e.target.select();}}
-					onKeyDown={(e)=>{this.handleKeys(e);}}
 					value={this.state.filterString}
 				/>
 			</label>
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -84,9 +84,12 @@ const ListPage = createClass({
 	updateUrl : function(filterTerm){
 		const url = new URL(window.location.href);
 		const urlParams = new URLSearchParams(url.search);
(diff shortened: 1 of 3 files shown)


client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -74,19 +75,39 @@ const ListPage = createClass({
...
+	},
+
+	updateUrl : function(filterTerm){
+		const url = new URL(window.location.href);
+		const urlParams = new URLSearchParams(url.search);
+		if(urlParams.get('filter') == filterTerm) return;
+		urlParams.set('filter', filterTerm);
+		if(!filterTerm) urlParams.delete('filter');
+		url.search = urlParams;
+		window.location.replace(url.href);
```
ericscheid, line 96: IIRC, updating `window.location` will cause a page reload (which might reset other widgets on the page, such as sort order).
 
What you might want here is instead to use `history.replaceState()`.
https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState

G-Ambatte: In that case, I can probably remove the 'onKeyDown ()=> if(keyDown=Enter) updateUrl()` stuff and just do it live with `history.replaceState()`... Hang on, I'm just going to quickly rewrite half of this PR...

G-Ambatte: I vastly prefer the performance of `history.replaceState` to `window.location.replace`.

The URL now updates live.

G-Ambatte: A small thing, but I like that `history.replaceState` permits the use of a URL object, where `window.location.replace` needed to be passed `url.href` specifically.

calculuschild (approved):
Good work! Looks great.

PR #2091, opened 2022-03-25
Remove calls to deprecated `componentWillMount`

### LINKED ISSUE #1936: Need to fix `comopnentWillMount` is deprecated warning
When launching local Homebrewery instance I see the following warning:

```
Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move code from componentWillMount to componentDidMount (preferred in most cases) or the constructor.

Please update the following components: Homebrewery
    in Homebrewery
```

I made a local experiment with replacing `componentWillMount` with `componentDidMount` and it seems to work, but I barely tested it, so I can't be sure

G-Ambatte:
This PR removes the deprecated `componentWillMount` call from `homebrew.jsx`, shifting the declarations in that function to `getInitialState` instead.

This PR resolves #1936.

### DIFF
--- a/client/homebrew/homebrew.jsx
+++ b/client/homebrew/homebrew.jsx
@@ -32,11 +32,14 @@ const Homebrew = createClass({
 			}
 		};
 	},
-	componentDidMount : function() {
-		global.account = this.props.account;
+
+	getInitialState : function(){
 		global.version = this.props.version;
+		global.account = this.props.account;
 		global.enable_v3 = this.props.enable_v3;
+		return {};
 	},
+
 	render : function (){
 		return (
 			<Router location={this.props.url}>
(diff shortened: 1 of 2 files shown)


calculuschild:
Huh. I guess `getInitialState()` (or `getDefaultProps()`) are the only other candidates for setting up the component before it mounts.

image

I don't like using `getInitialState` for anything other than setting up the `state`, but looks like we don't have much choice....

I'll give it a test and then merge.

G-Ambatte:
> I don't like using getInitialState for anything other than setting up the state, but looks like we don't have much choice....

It does seem like there should be a better way to set data up before the Render fires... and it looks like `ComponentWillMount` was it, but it's been/being taken away.

¯\\\_(ツ)_/¯

calculuschild:
Alright, seems good. Merging!

PR #2095, opened 2022-03-26
[WIP] UserInfo

G-Ambatte:
Initial pass at UserInfo.

- updateActivity : Updates to the `lastActivity` property to the app middleware; the update function checks if the user's lastActivity property is a different date to today before applying the update, so each user should only have `lastActivity` update once in every 24 hour period. Currently hooked into the Express app middleware as well as API GET method.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -10,6 +10,7 @@ const app = express();
 const config = require('./config.js');
 
 const homebrewApi = require('./homebrew.api.js');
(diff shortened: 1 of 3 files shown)


G-Ambatte:
This branch was broken beyond my ability to repair; I have gone full scorched earth and deleted it. I will open a new PR with a branch built from the current `master`.

PR #2114, opened 2022-03-31
Google Drive document stubs

jeddai:


jeddai:
Known issues as of 3/30/2022 23:36 CST:
- User list page shows stubbed google drive brews twice

calculuschild:
At a quick glance this looks good so far. Haven't dug deeply into it yet but I think this is on the right track.

However, while we're doing this huge shuffling of data around, I wanted to list out the metadata properties and which storage space they should belong, because I don't think *every property* needs to be part of the stub. In general, we want to make sure catastrophic failure of the Homebrewery keeps any google brews intact, so any vital data that makes up the brew content should remain in Google. Any property that should be community-searchable should remain in the HB Stub. Then there are a few properties that don't really serve either purpose, but I would lean toward keeping stubs as a minimal search index as much as possible:

#### HB Stub
- googleId
- shareId
- editId
- authors
- **owner? (new)**
- tags
- systems
- published
- **social media thumbnail (new)**

#### Google Drive
- shareId (to redirect to stub)
- editId (to redirect to stub)
- isStubbed
- text
- title
- description
- views
- renderer
- **theme (new)**
- version
- createdAt
- updatedAt
- pageCount

For instance, we need to know if a brew is published in order to search for it, and if HB dies, users probably don't mind as if the `published` status is lost. However knowing what renderer/theme it was built on will be important in rebuilding the document later.

`version` and `pageCount` I don't think are necessarily searchable parameters that need to be in the Stub, but aren't really part of the brew content either. In this case I personally prefer keeping the stub smaller and dumping those into Google Drive.

Any differing thoughts or opinions? I can update the list as needed.

server/googleActions.js
```diff
@@ -185,7 +185,10 @@ const GoogleActions = {
 				'title'     : brew.title,
 				'views'     : '0',
 				'pageCount' : brew.pageCount,
-				'renderer'  : brew.renderer || 'legacy'
+				'renderer'  : brew.renderer || 'legacy',
+				'isStubbed' : true,
+				'createdAt' : new Date().toISOString(),
```
calculuschild, line 190: Same thing here. We already utilize the built-in `obj.data.createdTime` here from the Google Drive file.

server/homebrew.api.js
```diff
@@ -6,13 +6,59 @@ const GoogleActions = require('./googleActions.js');
...
+	return async (req, res, next)=>{
+		const { brew } = req;
+
+		if(!brew) {
+			const id = req.params.id, googleId = req.body?.googleId;
+			const found = await findBrew(id, accessType, googleId);
+			req.brew = accessType !== 'edit' && found.toObject ? found.toObject() : found;
+		}
+
+		!!next ? next() : undefined;
```
calculuschild, line 58: I don't think we need to worry about `next()` not existing. Express.js should know how to handle it if we just call `next()`.

server/homebrew.api.js
```diff
@@ -6,13 +6,59 @@ const GoogleActions = require('./googleActions.js');
...
+			if(googleId) {
+				console.warn(`Unable to find document stub for ${accessType}Id ${id}`);
+			} else {
+				console.warn(err);
+			}
+		});
+
+	let googleBrew;
+	if(googleId || brew?.googleId) {
+		googleBrew = await GoogleActions.getGoogleBrew(googleId || brew?.googleId, id, accessType)
```
calculuschild, line 35:
```suggestion
	if(googleId || brew?.googleId) {
		let googleBrew = await GoogleActions.getGoogleBrew(googleId || brew?.googleId, id, accessType)
```

server/homebrew.api.js
```diff
@@ -89,148 +135,176 @@ const newBrew = async (req, res)=>{
...
 
 router.post('/api', asyncHandler(newBrew));
-router.put('/api/:id', asyncHandler(updateBrew));
-router.put('/api/update/:id', asyncHandler(updateBrew));
-router.delete('/api/:id', asyncHandler(deleteBrew));
-router.get('/api/remove/:id', asyncHandler(deleteBrew));
-
-module.exports = router;
+router.put('/api/:id', asyncHandler(getBrew('edit')), asyncHandler(updateBrew));
+router.put('/api/update/:id', asyncHandler(getBrew('edit')), asyncHandler(updateBrew));
```
calculuschild, line 303: Help me understand why we need to call `getBrew` on all of these api calls? Isn't the brew already provided from the client in `req` by the time we get to this point? Unless there's something changed in this PR resulting in cases requiring re-fetching brews before we handle them?

jeddai: It is, but in all of these we have to fetch the instance of the brew from the database in order to properly save it, since save is a method on the Mongoose object. Since all of them involve stubbing now, that has to be done in all routes here.

server/homebrew.api.js
```diff
@@ -6,13 +6,59 @@ const GoogleActions = require('./googleActions.js');
...
+	return brew;
+};
+
+const getBrew = (accessType)=>{
+	return async (req, res, next)=>{
+		const { brew } = req;
+
+		if(!brew) {
+			const id = req.params.id, googleId = req.body?.googleId;
+			const found = await findBrew(id, accessType, googleId);
```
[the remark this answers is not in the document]
jeddai: Nope, moving that now!

server/homebrew.api.js
```diff
@@ -6,13 +6,59 @@ const GoogleActions = require('./googleActions.js');
...
+		});
+
+	let googleBrew;
+	if(googleId || brew?.googleId) {
+		googleBrew = await GoogleActions.getGoogleBrew(googleId || brew?.googleId, id, accessType)
+			.catch((err)=>{
+				console.warn(err);
+			});
+		if(!brew && !googleBrew) throw 'Brew not found in database or Google Drive';
+		brew = brew ? _.merge(brew, googleBrew) : googleBrew;
```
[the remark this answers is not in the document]
jeddai: In this case `googleBrew` does have a possibility of not existing. If the call to `GoogleActions.getGoogleBrew` throws an exception, it will be caught by the `.catch` method above. Since that's a different method scope, I have it writing the error to the log and ending the scope, which is the same as doing a return there.

The check there can have googleBrew as undefined, and since we've already looked for the brew in the db and also couldn't find it, the brew is unknown to us so we throw the exception.

I've attached a console example:

image

server/googleActions.js
```diff
@@ -141,13 +141,14 @@ const GoogleActions = {
...
-					pageCount : brew.pageCount,
-					systems   : brew.systems.join()
+					'title'     : brew.title,
+					'shareId'   : brew.shareId || nanoid(12),
+					'editId'    : brew.editId || nanoid(12),
+					'views'     : brew.views,
+					'pageCount' : brew.pageCount,
+					'renderer'  : brew.renderer || 'legacy',
+					'isStubbed' : true,
+					'updatedAt' : new Date().toISOString(),
```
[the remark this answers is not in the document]
jeddai: Ah I misremembered pulling that out of the properties

server/homebrew.api.js
calculuschild: Right, we need the mongo object. Gotcha. Ok.

server/homebrew.api.js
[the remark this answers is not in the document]
calculuschild: Mostly there's just a lot of checking for existence of brew/googleBrew and I was trying to cut down on any redundant checks for legibility. If this is the easiest way to avoid an error then let's keep it. I'll take another look and see if there's any other cleanups we can make.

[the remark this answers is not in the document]
jeddai: Yeah that's entirely fair and if you do find anything please let me know! 

I think I may have added a few checks exactly like the above now in the deletion, update, and new brew methods. I also adjusted the return types of some of the google action methods to account for that. Some that were previously void now return the object id or just `true`.

server/homebrew.api.js
```diff
@@ -6,13 +6,52 @@ const GoogleActions = require('./googleActions.js');
...
+		const { brew } = req;
+
+		if(!brew) {
+			let id = req.params.id, googleId = req.body?.googleId;
+
+			if(id.length > 12) {
+				googleId = id.slice(0, -12);
+				id = id.slice(-12);
+			}
+			let found = await HomebrewModel.get(accessType === 'edit' ? { editId: id } : { shareId: id })
```
calculuschild, line 28: Can we change `found` to `stub` or something?

jeddai: Sure!

server/homebrew.api.js
```diff
@@ -6,13 +6,52 @@ const GoogleActions = require('./googleActions.js');
...
+					}
+				});
+
+			if(googleId || found?.googleId) {
+				const googleBrew = await GoogleActions.getGoogleBrew(googleId || found?.googleId, id, accessType)
+					.catch((err)=>{
+						console.warn(err);
+					});
+				if(!found && !googleBrew) throw 'Brew not found in database or Google Drive';
+				found = found ? Object.assign(excludeStubProps(found), excludeGoogleProps(googleBrew)) : googleBrew;
```
calculuschild, line 43: What happens if we find a stub but the associated Google Brew is missing?

jeddai: Good call, I think right now it would try to load the page but would be missing some fields like text, etc...

I need to make sure it accounts for that, in that case do you think it makes sense to have the response result in a 500?

server/homebrew.api.js
```diff
@@ -6,13 +6,52 @@ const GoogleActions = require('./googleActions.js');
...
+			} else if(!found) {
+				throw 'Brew not found in database';
+			}
```
calculuschild, line 46: Instead of `else if`, does this work if we instead treat it as a catch-all for any missing brew (including from database)?

```suggestion
			}
			
			if(!found) {
				throw 'Brew not found';
			}
```

In essence, if we just cleanly divide this up into 4 segments:

1) Convert IDs to Google IDs as needed
2) Check for a stub (or full brew) -> No stub: Warn: "Brew not found in Database" or "Stub not found for Google Brew"
3) If Google ID, check for Google Brew -> No Google Brew: Warn: "Brew not found in Google Drive"
   1) Merge Google Brew into Stub if needed
4) If no brew, throw error

jeddai: Yeah that's a solid plan

server/homebrew.api.js
calculuschild: For Google Drive we can probably grab whatever error code Google sends and respond with that. In case it's just due to user not logged in/ etc so we can provide an appropriate instruction to the user.

server/homebrew.api.js
```diff
@@ -33,19 +72,36 @@ const MAX_TITLE_LENGTH = 100;
...
 	for (const prop of propsToExclude) {
 		delete brew[prop];
 	}
 	return brew;
 };
 
+const excludeStubProps = (brew)=>{
+	const propsToExclude = ['text', 'textBin', 'renderer', 'pageCount', 'version'];
+	for (const prop of propsToExclude) {
+		brew[prop] = undefined;
```
calculuschild, line 91: How does `= undefined` differ from `delete` we use elsewhere?

jeddai: I found that `delete` doesn't work on Mongoose object instances, whereas ` = undefined` did.

calculuschild: Perhaps we can do the `to object` conversion right after retrieving a Mongo brew in `getBrew` so we can be sure the result of `getBrew` is always normal Javascript object we can manipulate? Instead of carrying around an unknown object type that we have to keep cloning or checking the type?

Then in the case we actually need a mongo object to "save", etc., we just do a simple `mongo.find(editId)` or whatever right where we need it.

jeddai: I'm not _opposed_ to that plan, but I will say that the only case in which it returns a mongoose object is when it's fetched for editing. In all other cases it does the `toObject` call. I feel like the switch to setting values to undefined is small enough that the benefit of not having to make a whole additional call to the database to end up updating it is worth it. 

Happy to impl your suggestion though if it seems problematic enough though

calculuschild: Not just for the delete thing. I think it will be helpful to guarunteed the object type coming out of that function. Especially in the future if we make more changes, I don't want to leave a trap behind for a future developer.

calculuschild: I think that's basically what we had before anyway. Rather than preaching each action with getBrew, we just called mongo.find inline when needed.

calculuschild: TLDR; `getBrew` gets the brew data to send to the user (maybe should be `getBrewForUser` or something). If we just need a mongo object so we can save, we don't need the whole process of "retrieve the brew, unzip the compressed brew text from `textBin`, get the brew from Google Drive, etc. This also means we can do similar to #2154 in the `find` by adding a "fields" parameter so the mongo object we retrieve for saving is only the `_id` or something, which should greatly reduce the amount of database/server bandwidth for large brews, instead of grabbing the entire 5MB document every time the autosave triggers, unzipping it, and then zipping the new version and sending a whole new 5MB back to the database.

server/homebrew.api.js
```diff
@@ -33,19 +72,36 @@ const MAX_TITLE_LENGTH = 100;
...
+	}
+	return brew;
+};
+
+const excludeGoogleProps = (brew)=>{
+	const modified = brew.toObject ? brew.toObject() : brew;
```
calculuschild, line 97: Is this a symptom of not knowing if `brew` is a normal object or a Mongo document?

jeddai: Yep -- google brews are normal objects, stub/HB brews are mongoose. I'm pulling that out with my updates

server/homebrew.api.js
```diff
@@ -55,28 +111,12 @@ const beforeNewSave = (account, brew)=>{
...
-	saved = saved.toObject();
-	saved.gDrive = false;
-	return saved;
-};
-
 const newGoogleBrew = async (account, brew, res)=>{
 	const oAuth2Client = GoogleActions.authCheck(account, res);
 
-	return await GoogleActions.newGoogleBrew(oAuth2Client, brew);
+	const newBrew = excludeGoogleProps(_.clone(brew));
```
calculuschild, line 117: Why cloning here? Another symptom of Mongoose docs? Or are we worried about affecting the input brew? I don't see anything here where we need to preserve a separate original copy of the brew.

If we have a reason we need to preserve the input brew, perhaps the `_.clone` should occur inside `excludeGoogleProps` instead of having to pass in a clone each time.

server/homebrew.api.js
```diff
@@ -89,148 +129,177 @@ const newBrew = async (req, res)=>{
...
+	saved = await newHomebrew.save()
+		.catch((err)=>{
+			console.error(err, err.toString(), err.stack);
+			throw `Error while creating new brew, ${err.toString()}`;
+		});
 	if(!saved) return;
+	saved = saved.toObject();
+
 	return res.status(200).send(saved);
 };
```
calculuschild, line 167: It's not letting me do a suggested edit, but moving the extra `if(transferToGoogle)` checks up:

```
	let googleId, saved;
	if(transferToGoogle) {
	//Save brew to Google Drive; newHomebrew will just be a stub
		googleId = await newGoogleBrew(req.account, newHomebrew, res)
			.catch((err)=>{
				res.status(err?.status || err?.response?.status || 500).send(err?.message || err);
			});
		if(!googleId) {
			res.status(500).send('Unable to save document to Google Drive');
			return;
		}
		newHomebrew.googleId = googleId;
		excludeStubProps(newHomebrew);
	} else {
	//Prep brew for Mongo; newHomebrew will be the full brew
		// Compress brew text to binary before saving
		newHomebrew.textBin = zlib.deflateRawSync(newHomebrew.text);
		// Delete the non-binary text field since it's not needed anymore
		newHomebrew.text = undefined;
	}

	saved = await newHomebrew.save()
		.catch((err)=>{
			console.error(err, err.toString(), err.stack);
			throw `Error while creating new brew, ${err.toString()}`;
		});
	if(!saved) return;
	saved = saved.toObject();
	return res.status(200).send(saved);
};
```

server/homebrew.api.js
jeddai: Good call, it should go inside the exclude method.

server/homebrew.api.js
```diff
@@ -89,148 +129,177 @@ const newBrew = async (req, res)=>{
...
+		const deleted = await deleteGoogleBrew(req.account, brew.googleId, brew.editId, res)
 			.catch((err)=>{
 				console.error(err);
-				res.status(500).send(err);
+				res.status(err?.status || err?.response?.status || 500).send(err.message || err);
 			});
-		if(!saved) return;
+		if(!deleted) {
+			if(res.headersSent) {
+				res.status(500).send('Unable to delete brew from Google');
```
calculuschild, line 183: Somehow this seems redundant with the line above where we are already sending an error in the `catch`. In fact, should we move all the error reporting from here into `deleteGoogleBrew` so any call to `deleteGoogleBrew` will automatically have errors handled?

Keeping the `if` is fine so we can `return`.

server/homebrew.api.js
```diff
@@ -89,148 +129,177 @@ const newBrew = async (req, res)=>{
...
-		if(!saved) return;
-
-		await deleteLocalBrew(req.account, brew.editId)
-			.catch((err)=>{
-				console.error(err);
-				res.status(err.status).send(err.message);
-			});
+		if(!brew.googleId) {
+			if(res.headersSent) {
+				res.status(500).send('Unable to save brew to Google');
```
calculuschild, line 197: Same thing here. Seems like we could move all the `res.send` errors into `newGoogleBrew` to reduce the number of places we need to have duplicated error handling?

server/homebrew.api.js
```diff
@@ -89,148 +129,177 @@ const newBrew = async (req, res)=>{
...
-	}
-};
-
-const deleteLocalBrew = async (account, id)=>{
-	const brew = await HomebrewModel.findOne({ editId: id });
-	if(!brew) {
-		throw { status: 404, message: 'Can not find homebrew with that id' };
+		if(!googleDeleted) {
+			if(!res.headersSent) {
+				res.status(500).send('Unable to delete brew from Google');
```
calculuschild, line 264: And here.

server/homebrew.api.js
jeddai: So this would be the workflow if I'm understanding correctly:
- Retrieves the brew during `getBrew` and converts it to a standard object
- Updates the object and deletes values where applicable for the current iteration
- Fetch the Mongo object
- `Object.assign(stub, updated)` or a `_.merge` call of some kind to join the values from the updated object to the Mongo object

In most cases that works fine, but when transferring to Google it will be deleting the text and textBin from the updated object, meaning the object won't have those values, so the assign/merge will not change the values on the Mongo stub. 

For that case we would then need to run `brew.<x> = undefined` where x is the GDrive stored props, which just brings us back to the method as it is now.

server/homebrew.api.js
jeddai: This in particular is a good example of why I prefer try/catch instead of `.catch`. If we used try/catch in these cases and didn't have the `.catch` then at the end of the catch we could return.

I'm not really fond of passing the response around between files, partially because it makes understanding where things are happening more difficult. 

Considering the repetitive nature of this particular if statement, we could write a middleware piece that sends a 500 if no headers have been sent that runs after all other middlewares.

calculuschild: At the very bottom of `api.js` we already have an error-handling middleware that does this for any uncaught errors. Is that what you had in mind?

server/homebrew.api.js
calculuschild: The flow is basically what I was thinking, yes.

Ok, we can leave `brew.x = undefined` for managing the stubs in this special case then. However, the other parts (`getBrew` returns a standard object, and for saving we fetch the Mongo object with the minimal number of fields to reduce data)

server/homebrew.api.js
jeddai: _Partially_ that, but it's more a "I'm the last middleware and no headers have been sent back to the client so I'm going to send something so it doesn't time out" instead of sending a particular error passed into the middleware

calculuschild: Do you think the existing Middleware can be updated to handle errors in the way you want?

jeddai: The existing middleware expects an error as the first piece passed in, whereas the middleware I'm intending to write would not. It would simply handle any cases where the backend failed to send headers for some reason. While that _shouldn't_ ever happen, if it did we would end up with request timeouts.

server/homebrew.api.js
jeddai: How would you recommend applying the values from the standard object to the Mongoose object?

server/homebrew.api.js
calculuschild: If not a fan of passing the `res` object around, now that I think about it, I think our existing error handler middleware should be able to `send` the error as long as we just leave it uncaught or rethrow here.

So, instead of re-catching the error at each level (unless we need to add info to the error), we should be able to create the appropriate error object inside `newGoogleBrew`, throw it, and then just let the middleware catch it for final sending?

server/homebrew.api.js
jeddai: So I know as soon as we start working with array types more `merge` will be a problem -- it merges arrays recursively, which won't work for tags and authors. The same issue applies for undefined, for example,
```js
const x = _.merge({ tags: ['tag1', 'tag2'] }, { tags: ['tag1'] })
x // { a: ['tag1', 'tag2'] }

const y = _.merge({ a: 'test' }, { a: undefined })
y // { a: 'test' }
```
I'll try the spread operator today and see if it copies the Mongoose methods.
```js
const z = { ...{a: 'test'}, ...{a: undefined} }
z // { a: undefined }
```

calculuschild: If not `_.merge` there should be some similar approach.

I guess we don't need to merge properties as much as overwrite them, no? Does `_.assign` do what we need? By the time the JS object gets to this point, can we assume it looks exactly how the final saved object should look, minus ids and stuff? So if we previously had `tags: [tag1, tag2, tag99]` and now the brew has `tags: [tag2, tag300]` we would want to just fully overwrite `tags` to be the new array, instead of merging the arrays.

jeddai: Sorry I've been away for a bit, I tried to make this work but both using `new HomebrewModel(brew)` and `_.assign(await HomebrewModel.find({ _id: req.brew.id }), brew)` the HomebrewModel gets _very_ upset and errors out saying it's a duplicate key on `_id`. I don't think this is going to be a viable change without some sneaky js work, and I'd much prefer no sneaky js work and `toObject` called in a few extra places. Thoughts?

jeddai: Never mind! I was not performing the find correctly. 😬

calculuschild: Can you point out which lines are causing the conflict?

But as a guess, I'm thinking in the singular `toObject` we could clear the `_id` parameter from the resulting JSObject (its only useful as a unique identifier for Mongo. We dont really use it), and when we need to look up a document in the database, we just use the editID or shareID. No need to try preserving the original mongo ID between transfers, etc. Just our edit/share IDs.

calculuschild: Lol. Alright. Ask away if you run into something else.

jeddai:
Okay this should be good to review again!

server/homebrew.api.js
```diff
@@ -89,148 +136,151 @@ const newBrew = async (req, res)=>{
...
-				console.error(err);
-				return res.status(500).send('Error while saving');
-			});
-
-		brew = _.merge(dbBrew, brew);
-		brew.text = mergeBrewText(brew);
+		if(!updated) return;
+	}
 
+	if(brew.googleId) {
```
calculuschild, line 201: Can this `if` be combined with the one above?

jeddai: Well the one above is an else-if, I'd have to call `excludeStubProps` in both ifs, and then move the else here into the else above. If you're cool with the duplication then sure

server/homebrew.api.js
```diff
@@ -6,13 +7,57 @@ const GoogleActions = require('./googleActions.js');
...
 const asyncHandler = require('express-async-handler');
+const { nanoid } = require('nanoid');
 
 // const getTopBrews = (cb) => {
 // 	HomebrewModel.find().sort({ views: -1 }).limit(5).exec(function(err, brews) {
 // 		cb(brews);
 // 	});
 // };
 
+const getBrew = (accessType)=>{
```
calculuschild, line 18: I don't think I have any more major changes to add (still need to do more testing to make sure it's not broken anywhere). Only thing is I think these longer functions could benefit from some comments marking out each of the major steps since there's a lot of checking `googleId`, which may not be obvious. For instance, comments for:

1) Parse Brew ID
2) Search HB database for Brew (or Stub)
3) Search GDrive for Brew if needed

jeddai: Yeah adding comments is a great plan, I'll do that!

server/homebrew.api.js
```diff
@@ -89,148 +136,151 @@ const newBrew = async (req, res)=>{
 
 	beforeNewSave(req.account, brew);
 
-	let saved;
+	const newHomebrew = new HomebrewModel(brew);
+	newHomebrew.editId = nanoid(12);
+	newHomebrew.shareId = nanoid(12);
+
+	let googleId, saved;
```
calculuschild, line 143: Another good place for a comment. Since `transferToGoogle` in this case actually means "save to google" rather than transferring out of the HB database.

1) Save Brew to Google Drive and prep a Stub
2) Save Brew (or Stub) to HB database

server/homebrew.api.js
```diff
@@ -89,148 +136,151 @@ const newBrew = async (req, res)=>{
...
 };
 
 router.post('/api', asyncHandler(newBrew));
-router.put('/api/:id', asyncHandler(updateBrew));
-router.put('/api/update/:id', asyncHandler(updateBrew));
-router.delete('/api/:id', asyncHandler(deleteBrew));
-router.get('/api/remove/:id', asyncHandler(deleteBrew));
-
-module.exports = router;
+router.put('/api/:id', asyncHandler(getBrew('edit')), asyncHandler(updateBrew));
```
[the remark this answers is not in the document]
jeddai: `getBrew` performs the fetch from both the database and Google and combining those objects. Since at the time we don't know if the user is transferring to/from google or if the doc is google or not I think it's pretty important there. I can look into removing the outer `if(!brew)` piece though

server/homebrew.api.js
jeddai: I mean I can always change the query param name too

jeddai: transferToGoogle made sense before, but saveToGoogle makes more sense now.

server/homebrew.api.js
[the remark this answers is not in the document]
calculuschild: Well, for `update`, we hit the outer `if(!brew)` anyway and don't retrieve anything, don't we? Since we are already sending in the updated brew from the client?

server/homebrew.api.js
calculuschild: But still add comments :)

server/homebrew.api.js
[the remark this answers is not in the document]
jeddai: So for `update`, the incoming brew is actually on `req.body`, whereas `getBrew` is referring to `req.brew`, which is set by either this middleware, or earlier middlewares. I'm pretty sure I had put that outer if in because there's a couple spots where it hits it because an earlier middleware had already set `req.brew`. I'm wondering if it may have been a dev artifact though, and I might be able to remove it. I'll play around with it and let you know

server/homebrew.api.js
```diff
@@ -89,148 +136,151 @@ const newBrew = async (req, res)=>{
...
+	let brew = _.assign(req.brew, excludePropsFromUpdate(req.body));
+	brew.text = mergeBrewText(brew);
 	const { transferToGoogle, transferFromGoogle } = req.query;
 
-	let saved;
 	if(brew.googleId && transferFromGoogle) {
-		beforeNewSave(req.account, brew);
-
-		saved = await newLocalBrew(brew)
+		const deleted = await deleteGoogleBrew(req.account, brew.googleId, brew.editId, res)
```
calculuschild, line 177: Oh, one thing here: When transferring, we want to make sure we successfully save the brew to the target location before we delete (or stub) the original.

server/homebrew.api.js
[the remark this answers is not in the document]
calculuschild: Ok, so let me step through this:

1) User sends `req.brew` = updated brew
2) We `getBrew()`, so `req.body` = old brew
3) We immediately overwrite the old brew with the updated brew with `let brew = _.assign(req.brew, excludePropsFromUpdate(req.body));` The content of the old brew is lost at this point, but we now have a Mongo object.
4) We retrieve the old brew a second time with `HomebrewModel.findOne()` and immediately overwrite it again, and we still have a Mongo object.
5) We save the brew.

Seems like we can skip steps 2 and 3. Especially since `getBrew` does all the extra steps of looking for Google brews and stuff.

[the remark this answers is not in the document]
jeddai: The `if (!brew)` is in fact a development artifact, removing it!

server/homebrew.api.js
jeddai: Done!

server/homebrew.api.js
jeddai: Done!

server/homebrew.api.js
jeddai: Done!

server/homebrew.api.js
```diff
@@ -55,182 +122,184 @@ const beforeNewSave = (account, brew)=>{
...
-	const { transferToGoogle, transferFromGoogle } = req.query;
-
-	let saved;
-	if(brew.googleId && transferFromGoogle) {
-		beforeNewSave(req.account, brew);
+	// Initialize brew from request and body, destructure query params, set a constant for the google id, and set the initial value for the after-save method
+	let brew = _.assign(req.brew, excludePropsFromUpdate(req.body));
+	const { saveToGoogle, removeFromGoogle } = req.query;
+	const googleId = brew.googleId;
+	let afterSave = async ()=>true;
```
jeddai, line 180: This is where we initially set `afterSave`

server/homebrew.api.js
```diff
@@ -55,182 +122,184 @@ const beforeNewSave = (account, brew)=>{
...
-		await deleteGoogleBrew(req.account, `${brew.googleId}${brew.editId}`, res)
+	if(brew.googleId && removeFromGoogle) {
+		// If the google id exists and we're removing it from google, set afterSave to delete the google brew and mark the brew's google id as undefined
+		afterSave = async ()=>{
+			return await deleteGoogleBrew(req.account, googleId, brew.editId, res)
+				.catch((err)=>{
+					console.error(err);
+					res.status(err?.status || err?.response?.status || 500).send(err.message || err);
+				});
+		};
```
jeddai, line 192: Here we update `afterSave` to delete the google brew

server/homebrew.api.js
```diff
@@ -55,182 +122,184 @@ const beforeNewSave = (account, brew)=>{
...
+	// Save the database brew
+	const saved = await brew.save()
+		.catch((err)=>{
+			console.error(err);
+			res.status(err.status || 500).send(err.message || 'Unable to save brew to Homebrewery database');
+		});
 	if(!saved) return;
+	// Call and wait for afterSave to complete
+	const after = await afterSave();
+	if(!after) return;
```
jeddai, line 248: Lastly, we call `afterSave` after the brew has been saved in the database.

calculuschild:
Testing notes:

#### Bugs:
- "Share" link from the navigation bar still points to the old "full Google ID" URL. Same with the "Share -> Post to Reddit" button
- Editing an old Google Brew and saving should update the browser URL to just the EditID
- Nobody can delete HB brews. Returns 500 server error and does nothing.
- Co-author cannot delete Google Brew. Returns 500 error and does nothing.
- Owner *can* delete google brew, but the stub remains, and owner remains as an author. Cannot delete the stub.
  - When deleting a co-authored brew, should:
  1) If brew is stored on deleting user's GDrive, Move Google Brew to HB so other authors retain access.
  2) Remove user from authors list.
  - When deleting a single-authored brew, should:
  1) Delete brew *and* stub, if any.

#### Error Handling:
- Fake EditID: "Brew not found in Homebrewery database or Google Drive" :white_check_mark:
- EditID too short: "Brew not found in Homebrewery database or Google Drive" :white_check_mark:
- Fake ShareID: "Brew not found in Homebrewery database or Google Drive" :white_check_mark:
- ShareID too short: "Brew not found in Homebrewery database or Google Drive" :white_check_mark:
- Fake GoogleID: Error log from Google, 404 :white_check_mark:
- GoogleID too short: Error log from Google, 404 :white_check_mark:
- GoogleID but shareID doesn't match: "Share ID does not match" :white_check_mark:
- GoogleID but editID doesn't match: "Edit ID does not match" :white_check_mark:


#### Other notes:
- User's can successfully co-author a Google brew 
  - Brew shows up under both accounts
  - Both users can edit
  - Can't transfer in/out of Google Brew if not signed in with Google, but can still edit
  - Co-author *can* transfer the Google Brew to HB, and then transfer it to his *own* Google Drive. All authors continue to have access.
    - Perhaps the dialog popup should change slightly in this case to indicate you are moving the file out of someone elses' Google Drive.
  - Co-author user page does not show GDrive icon on Google brews they do not own
  - Co-author *does* see GDrive icon when editing a brew they do not own
    - Perhaps we could add a third GDrive icon (GDrive with a little "person" icon or something to show it's on Google drive but another owner.)

image


googleDriveShared

jeddai:
Alrighty!
- All the deletion issues should be resolved
- The link issues should be resolved
- Editing will result in the url being updated

Do you think it might make sense to include the authorship-related changes in a future PR?

server/app.js
```diff
@@ -190,7 +190,7 @@ app.get('/user/:username', async (req, res, next)=>{
 			});
 
 		for (const brew of brews) {
-			const match = googleBrews.findIndex((b)=>b.editId === brew.editId);
+			const match = googleBrews?.findIndex((b)=>b.editId === brew.editId) ?? -1;
```
calculuschild, line 193: Doesn't `findIndex()` already return `-1` if it doesn't find a match?

If it's a case of handing when `googleBrews = empty`, how about we instead just add a check to exit the loop early if `googleBrews` is empty? Probably more readable and if there are lots of brews we might save some time.

Also, can we safely just pre-filter this loop so the only brews we loop over have `brew.googleId`, i.e., we already know they are stubs?

server/homebrew.api.js
```diff
@@ -257,17 +257,22 @@ const deleteGoogleBrew = async (account, id, editId, res)=>{
...
-	if(brew.googleId) {
-		const googleDeleted = await deleteGoogleBrew(account, brew.googleId, brew.editId, res)
+	let brew = req.brew;
+	const { googleId, editId } = brew;
+	const account = req.account;
+	const isOwner = account && (brew.authors.length === 0 || brew.authors[0] === account.username);
+	console.log(brew.authors, googleId, account.username);
+	let afterSave = async ()=>true;
+	if(googleId && isOwner) {
+		afterSave = async ()=>await deleteGoogleBrew(account, googleId, editId, res)
```
calculuschild, line 267: Can we add a comment here? Just to be consistent with the comments below that help outline each of the cases we have to handle.

server/homebrew.api.js
```diff
@@ -282,13 +287,21 @@ const deleteBrew = async (req, res)=>{
 					throw { status: 500, message: 'Error while removing' };
 				});
 		} else {
+			if(googleId && isOwner) {
```
calculuschild, line 290: Add a comment here too explaining this case of moving the Google brew to HB so other authors retain access.

server/homebrew.api.js
calculuschild: Hmmm.... Trying to work through a clearer branching flow here. While the `afterSave()` approach works, it does scramble the order around a bit to be a little harder to read. What if we simply set a flag `let shouldDeleteFromGoogle = true` or something? Then I think we can rearrange the logic a bit for legibility:

```
If not on HB (Google Brew only):

  -> Flag to delete from Google

Else (must be on HB as a full brew or stub):

  If current user is only author:

    -> Delete from HB
    -> Flag to delete from Google (if stub)

  Else (multiple authors):

    If current user is owner of Google Brew:

      -> Update stub to full brew
      -> Flag to delete from Google

    -> Remove current user as author
    -> Save brew to HB with updated author list

    After save successful:

      If flag to delete from Google:

        -> Delete from Google
```

server/app.js
```diff
@@ -189,19 +189,19 @@ app.get('/user/:username', async (req, res, next)=>{
...
+			for (const brew of brews) {
+				const match = googleBrews.findIndex((b)=>b.editId === brew.editId);
+				if(match !== -1) {
+					brew.googleId = googleBrews[match].googleId;
+					brew.stubbed = true;
+					brew.pageCount = googleBrews[match].pageCount;
+					brew.renderer = googleBrews[match].renderer;
+					brew.version = googleBrews[match].version;
+					googleBrews.splice(match, 1);
+				}
```
calculuschild, line 202: Sorry, for this I meant something like:

```
for (const brew of brews.filter(b=>b.googleId)) {  // pre-filter so we only loop over stubs
	if(!googleBrews?.length) //break out of the loop once we've reached all the googleBrews
		break;
	const match = googleBrews.findIndex((b)=>b.editId === brew.editId);
	if(match !== -1) {
		brew.googleId = googleBrews[match].googleId;
		brew.stubbed = true;
		brew.pageCount = googleBrews[match].pageCount;
		brew.renderer = googleBrews[match].renderer;
		brew.version = googleBrews[match].version;
		googleBrews.splice(match, 1);
	}
}
```

jeddai: Sorry I didn't notice the pre-filtering bit initially -- check it out now. And I actually moved the for loop into the if statement that was happening right after it, which validates the googleBrews exist and have a length of greater than one. Doesn't make sense to call the for loop at all if those two conditions aren't met.

calculuschild: Ok I see it now. This looks fine. 👍

calculuschild:
Alright. I'll give it one last test run and then probably merge it up tomorrow. Hooray!

calculuschild:
Some remaining issues:

- Creating a new Google Brew, or editing an existing Google Brew, appears as two documents on the User page. One with the Google Drive icon and the "long" link, and a HB brew with the "short" link. They both link to the same file though, and saving on the "long link" version will update the URL in the address bar back to the "short link. However saving does not remove the duplicate entry on the user page.
  - We want the one entry to show the Google Drive icon but also

- Owner cannot delete brew from Google. 500 Error. The "stub" entry disappears from the user page (and is removed from the Mongo database), but the Google Drive file remains, and still appears on the User page with the "long link". Editing and saving the brew restores the stub with the same short link as before. The result is the same whether you try to delete the "stub" copy or the "long link" copy.

server/homebrew.api.js
```diff
@@ -55,182 +122,197 @@ const beforeNewSave = (account, brew)=>{
...
+					throw { status: 500, message: err };
+				});
+		}
 	}
-
-	if(brew.authors.length === 0) {
-		// Delete brew if there are no authors left
-		await brew.remove()
+	if(deleteGoogleBrew) {
+		const deleted = await deleteGoogleBrew(account, googleId, editId, res)
```
calculuschild, line 298: Lol... Variable name reuse:

```
TypeError: deleteGoogleBrew is not a function
    at deleteBrew (C:\Users\Trevor\Documents\GitHub\homebrewery\server\homebrew.api.js:298:25)
```

server/app.js
```diff
@@ -219,58 +189,71 @@ app.get('/user/:username', async (req, res, next)=>{
 				console.error(err);
 			});
 
-		if(googleBrews) {
+		if(googleBrews && googleBrews.length > 0) {
+			for (const brew of brews.filter((brew)=>brew.googleId)) {
```
calculuschild, line 193: Need to include `googleId` in the `fields` array above. Otherwise we are not retrieving `googleId` for any brews, and we can't compare stubs to google brews. Hence we end up with both the stub and the google brew appearing in the user page.

jeddai: Ahh I missed that projection, my bad

jeddai: Should be fixed now!

server/homebrew.api.js
jeddai: Fixed it!

calculuschild:
Ok.... all seems to be working and I don't see any other issues. I'm going to pull the trigger on this and merge it up. Great work @jeddai !

PR #2127, opened 2022-04-11
[V3] Fix note render to pdf

lucastucious:
As mentionned in #1569, note cannot be selected in pdf. 

So here is a fix


lucastucious:
Fixed it for monster and descriptive block too

calculuschild:
Thanks @lucastucious ! Looks like Mac finally released a version of Preview that doesn't have the box-shadow bug, so we can finally apply this and re-enable text selection.

PR #2129, opened 2022-04-13
Fix meta tags

G-Ambatte:
This PR fixes the unexpected behaviour of the default meta tag values. It also changes the default type to article from website, based on initial feedback.


client/template.js
```diff
@@ -11,10 +11,10 @@ module.exports = async(name, title = '', props = {})=>{
 		<link rel="icon" href="/assets/homebrew/favicon.ico" type="image/x-icon" />
 		<meta property="og:title" content="${props.brew?.title ?? 'Homebrewery - Untitled Brew'}">
 		<meta property="og:url" content="${HOMEBREWERY_PUBLIC_URL}/${props.brew?.shareId ? `share/${props.brew.shareId}` : ''}">
-		<meta property="og:image" content="${props.brew?.thumbnail ?? `${HOMEBREWERY_PUBLIC_URL}/thumbnail.png`}">
-		<meta property="og:description" content="${props.brew?.description ?? 'No description.'}">
+		<meta property="og:image" content="${props.brew?.thumbnail ? props.brew?.thumbnail : `${HOMEBREWERY_PUBLIC_URL}/thumbnail.png`}">
```
calculuschild, line 14: Why the change here? The logic should be identical.

G-Ambatte: It looks like switching to `||` fixed the issue as well. I've just pushed a PR using that operator instead.

PR #2132, opened 2022-04-14
[WIP] SRD Monster snippets

### LINKED ISSUE #300: SRD Monster list in Homebrewery
stolksdorf,

I have created a web page (javascript) which displays each of the SRD monsters in Markdown format.

http://miniworld.com/adnd/generators/monsterListBrewery.html

If you want to grab the code and add it to your app, you are more than welcome, though obviously you would need to adjust the selection to something that is more fitting to your app.

Note: some monsters need cleanup and I will try to get around to them (they still have HTMl mixed in).

Vlos

G-Ambatte:
While perusing the oldest open issues, I came across #300 - SRD Monsters. A little Googling returned [this GitHub gist](https://gist.github.com/tkfu/9819e4ac6d529e225e9fc58b358c3479) by [tkfu](https://gist.github.com/tkfu), which is the SRD monsters in one large JSON object.

I have wrapped this in a random selection function and built it into a standard monster block. The formatting of the Traits and Actions fields required a little tweaking of the `p+p` format, so I added the `srd` class to the monster block and added the required styling to the `5ePhb.style.less`, adjacent to the existing `monster` styling blocks.

Given the size of this PR - the data file is nearly 600kb alone - I thought it best to let everyone see the state of it as it currently stands.
With some additional work, I believe that this PR will resolve #300.

---

Snippet Menu:
image

Result:
image


ericscheid:
Looks great.

Highly confident that the very next request would be for a way to pick a specific SRD monster (instead of random).

Perhaps a very simple popup dialog with a) text input, b) list of matches, c) an I'm feeling lucky button.

G-Ambatte:
Yes - I was looking at how to do that but that level of interaction isn't something the snippet gen functions have done previously, so it's treading new ground. So I thought I should let some other eyes scan over it as it stands, before I dig into the snippet generator any deeper.

G-Ambatte:
So I _WAS_ just going to leave it at that...

- Shifted functions to separate SRD module
- Fix random SRD monster function to select different monsters each time
- Add function to select by monster name
  - Currently operates by user typing in desired monster's name
  - Monster names are exposed via the `monsterNames` function, so it should be possible to implement some sort of selection
    - 327 monsters currently in the list

G-Ambatte:
Whoooops... Somehow completely left Legendary Actions off the list, so they weren't getting processed at all. Added now.

image

G-Ambatte:
Select SRD Monster:

Menu:
image

Input box:
image

Error message:
image


Multiple results found alert:
image

Result in brew:
image

themes/5ePhb.style.less
```diff
@@ -452,6 +452,11 @@ body {
 		.useColumns(0.96, @fillMode: balance);
 	}
 
+	//SRD Monster Blocks
+	.monster.srd p+p{
+		text-indent: 0em;
```
ericscheid, line 457: don't need to specify units if value is zero

Gazook89:
I wonder if this type of thing should be in a new snippet menu, like "Generators"?   Or perhaps this could just replace the current stat block generator (once it more closely matches the existing output in the Traits/Actions areas).

themes/5ePhb.style.less
calculuschild: If SRD monsters need special treatment here, it seems like we should just fix the original monster styling to match. We should be able to make them fully compatible so we don't have to maintain two different statblock styles.

Gazook89:
>  How far do we go beyond being just a "text editor" and becoming a "database/generator?" When do we move something like this to a separate "app"

two more thoughts on this:   

1)  Is this something where we just reach out to existing generators (like tetra-cube) and give them the updated syntax for v3 and let them handle it?  Stay out of the 'content' business entirely
2) Or as you suggest as a possibility, a separate tool in the "naturalcrit family of products"....`https://generators.naturalcrit.com/`.   Of course, that means a whole new site and interface....

themes/5ePhb.style.less
G-Ambatte: In the `monster` block generated by the existing snippet, each **Trait** and **Action** is separated by a `\n:\n`, whereas in the scraped monster data, the items are only separated by the closing and opening tags of a `<p>` block. So the styling fix is really only because of the limitations of the scraped data.
A little processing of the data - making **Traits**, **Actions**, and **Legendary Actions** into lists, for example, rather than just a block of HTML encoded text - would likely resolve this.

G-Ambatte:
>     2. Or as you suggest as a possibility, a separate tool in the "naturalcrit family of products"....`https://generators.naturalcrit.com/`.   Of course, that means a whole new site and interface....

I like the idea of this: if we were to create a `/data/new` endpoint (POST an object to `/data/new` containing `{title: 'something', text: 'something else', metadata: '{ renderer: legacy }' }` and get redirected to https://homebrewery.naturalcrit.com/new, preloaded with the data from the object), then the SRD Monster Generator program (and/or similar projects) could serve as an standalone example of how to create a new brew via the data API.
There are logistical questions about hosting and subdomains and the like which can probably only be answered by @calculuschild, but a lot of this needs to stop being hypothetical and become actual code before answers are actually useful.

G-Ambatte:
> a separate tool in the "naturalcrit family of products"....https://generators.naturalcrit.com/

I believe that in the interest of making sure that Homebrewery remains focused on being an editor, it would be best to shift this development out of Homebrewery and into either it's own separate project, or as an aspect of the NaturalCrit project. However this new project would ideally link to Homebrewery; ideally by opening a Homebrewery new page which has been pre-populated with the appropriate SRD monster data for the users to alter and save to their accounts as they see fit.

I'll mark the PR as blocked and leave it open for now, just in case there's any further useful discussion to be had.

G-Ambatte:
Another option would be to use the DND5eAPI.com API to retrieve the SRD data. Whether this would reduce it to the point of being snippet suitable again is another question entirely... I suspect that the underlying point, that we keep the focus of HB on the editor, is still entirely valid.

ericscheid:
See also #1400 
Particularly comments about [importing data from third party APIs](https://github.com/naturalcrit/homebrewery/discussions/1400#discussioncomment-3640459)

dbolack-ab:
Another vote here for moving this out of a static file and into a mini-tool - even if we keep a menu item for it.

5e-Cleric:
Would seem weird to me to create a tool for this, speciall with https://tetra-cube.com/dnd/dnd-statblock.html being Homebrewery compatible.

Gazook89:
@G-Ambatte is it okay to close this PR, with the discussion seemingly coming to the conclusion that this should be a NaturalCrit tool outside of Homebrewery if it ever comes to fruition?  If so, you might consider opening an issue on NaturalCrit repo that just links to this PR so the idea isn't lost to time.

5e-Cleric:
Closing as suggested in favor of an issue both in NaturalCrit https://github.com/naturalcrit/naturalcrit/issues/64 and here: https://github.com/naturalcrit/homebrewery/issues/300

PR #2154, opened 2022-04-27
Add fields to Mongoose brew query

### LINKED ISSUE #2147: User Page retrieves whole brews instead of details only
We need to update the user page to not retrieve the entire brew contents when filling out the `brewItem` boxes. Turns out some of our users like to embed giant base64 images into their text, and when they make 50 copies of that brew, suddenly the HB is trying to fetch 550 MB of data all at once. This kills the crab.

(and crashes the HB due to low memory, and hangs the Mongo database while the search tries to complete, eventually also crashing due to low memory)

https://github.com/naturalcrit/homebrewery/blob/156abe3b9e820b3a680e0cfa6dd4252903a62cee/server/homebrew.model.js#L55-L66

We just use projection to filter out which fields we actually want to fetch:

```

const project: { title : 1, description : 1 ....}

Homebrew.find(query, {projection: project}).lean().exec((err, brews)=>{ //lean() converts results to JSObjects
			if(err) return reject('Can not find brew');
			return resolve(brews);
		});
```

G-Ambatte:
This PR adds a `fields` parameter to the `get` and `getByUser` methods, allowing only specific fields to be returned. This defaults to NULL, meaning that all fields are returned.

It also adds a `fields` parameter to the function call in `app.js` for UserPage, which limits the returned data to exclude the `text` and `textBin` fields from the returned data.

This will resolve #2147.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -194,7 +194,18 @@ app.get('/download/:id', asyncHandler(async (req, res)=>{
 app.get('/user/:username', async (req, res, next)=>{
 	const ownAccount = req.account && (req.account.username == req.params.username);
 
-	const fields = 'title pageCount description authors views shareId editId createdAt updatedAt lastViewed';
+	const fields = [
+		'title',
+		'pageCount',
+		'description',
+		'authors',
+		'views',
+		'shareId',
+		'editId',
+		'createdAt',
+		'updatedAt',
+		'lastViewed'
+	];
 
 	let brews = await HomebrewModel.getByUser(req.params.username, ownAccount, fields)
 	.catch((err)=>{
--- a/server/app.js
+++ b/server/app.js
@@ -194,7 +194,9 @@ app.get('/download/:id', asyncHandler(async (req, res)=>{
 app.get('/user/:username', async (req, res, next)=>{
 	const ownAccount = req.account && (req.account.username == req.params.username);
 
-	let brews = await HomebrewModel.getByUser(req.params.username, ownAccount)
+	const fields = 'title pageCount description authors views shareId editId createdAt updatedAt lastViewed';
+
+	let brews = await HomebrewModel.getByUser(req.params.username, ownAccount, fields)
 	.catch((err)=>{
 		console.log(err);
 	});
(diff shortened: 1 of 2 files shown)


server/app.js
```diff
@@ -194,7 +194,9 @@ app.get('/download/:id', asyncHandler(async (req, res)=>{
 app.get('/user/:username', async (req, res, next)=>{
 	const ownAccount = req.account && (req.account.username == req.params.username);
 
-	let brews = await HomebrewModel.getByUser(req.params.username, ownAccount)
+	const fields = 'title pageCount description authors views shareId editId createdAt updatedAt lastViewed';
```
calculuschild, line 197: Let's use the "array" syntax for this, to make it easier to add/remove values to `fields`  in the future.

See: https://mongoosejs.com/docs/api.html#query_Query-select

```
// include a and b, exclude other fields
query.select('a b');
// Equivalent syntaxes:
query.select(['a', 'b']);
```

calculuschild: Otherwise I think this all looks good. 👍

G-Ambatte: It's easy enough as a text string to just have it be `null` when not used, and then everything works as it has previously - I'll need to take a deeper dive into the `select` function to switch to that method.
Another option would be to use `fields = ['title','description'].join(' ');` to continue using the same parameter & function as it currently does.

calculuschild: It should work just the same according to the Mongoose docs. Null should also work. No need to change anything else.

G-Ambatte: Looks like the Mongoose select function can use a text string or list without any difficulty, so it made switching to the select function much easier. The latest commit should be using the new select function and a list of fields to return.

calculuschild: Oh, LOL. I see your confusion. The code I linked to wasn't so you would use the Select functions I was just showing you the part of the docs that says you can use an array instead of a string. I apologize.

Can you please return it back to how you had it (not using Select), but just use an array?

G-Ambatte: I've reverted to the find parameter method rather than the select function in the latest commit - to my mind, including it in the original function call *should* be more efficient than generating a full set of results and filtering it in a subsequent function.
I've no indication that this is actually the case, but all other things being equivalent, that's my preference.

G-Ambatte:
I've reverted to the parameter method in this latest commit, keeping the array of strings.

calculuschild:
Cool. Looks good to me then!

PR #2162, opened 2022-05-03
Brew & Source Navigation

### LINKED ISSUE #1756: Feature Request: Better navigation for large brews
If you have a brew with a lot of pages it's easy to lose track, what would help a lot would be a way to jump to certain places easily. Some possible ways would be:

- Having a page display like [ 5 / 31 ] like most PDF readers where you can change the first number to jump to the specified rendered page.
- Being able to click on the rendered headers to jump to the place where they are in the code.
- Having a mini ToC or book mark toggle that let's you jump to headers in both code and rendered version of the brew.

But anything that makes navigation easier would be appreciated. Maybe even a way to cascade pages to hide them temporarily in the code.

G-Ambatte:
Resolves #1756

This PR adds the `brewJump` and `sourceJump` functions to the CodeEditor, allowing the Brew Preview window and the source code editor to be scrolled to a specific position.

- `brewJump` uses `getElementById('pX').scrollIntoView()` (where X is the page number) to scroll the Brew preview iFrame to the top of the page.

- `sourceJump` uses `codeEditor.setCursorPosition(X,1)` (where X is the line number) to scroll the Editor window to the desired line number.

Currently the PR adds a new `editorTool` button to the SnippetBar (active on Text and Style tabs only) which will shift the Brew Preview and the Editor position to the top of their respective windows.


calculuschild:
The original idea back with stolksdorf was to add a couple buttons to the divider bar, similar to what Overleaf does:

image

Right arrow moves the renderer to match the editor cursor, left arrow moves the editor to match the renderer. But rather than try to match to an exact paragraph, etc., we just jump to the top of the corresponding page, for simplicity.

G-Ambatte:
I'm trying to get it working in the Snippet Bar in the first instance, then I'll progress to divider bar buttons.

G-Ambatte:
Current state:

image

Divider bar buttons: nonfunctional but look nice.
Snippet bar buttons: functional.

ericscheid:
Could you scale them down a bit, to be in proportion with the other UI icons in proximity?

calculuschild:
The snippet bar buttons don't seem to do anything for me. Its just dependent on the cursor position, right? I haven't looked at the code yet.

Divider buttons look good. Maybe we can color them the same as the snippet buttons too? Quick mockup:

image

G-Ambatte:
Moving the brew is dependent on the cursor position - it counts how many `\pages` there are between the start of the brew and the cursor position, then scrolls the preview iFrame to show `#p{that number}`

Moving the source is dependent on the preview scroll position - it finds how far down the brew renderer has been scrolled, then iterates through each page element, adding the height until it reaches or exceeds the current scroll position. The brew text is then parsed until it reaches that number of lines containing `\page` and the cursor position is set to `{that number}, 1`.

---

There are probably better/smarter ways to go about it, so feel free to point them out.

G-Ambatte:
> Divider buttons look good. Maybe we can color them the same as the snippet buttons too? Quick mockup:
> 
> image

Adjusted the background-color and added a box-shadow:
image

calculuschild:
> Moving the brew is dependent on the cursor position - it counts how many `\pages` there are between the start of the brew and the cursor position, then scrolls the preview iFrame to show `#p{that number}`
> 
> Moving the source is dependent on the preview scroll position - it finds how far down the brew renderer has been scrolled, then iterates through each page element, adding the height until it reaches or exceeds the current scroll position. The brew text is then parsed until it reaches that number of lines containing `\page` and the cursor position is set to `{that number}, 1`.

Hmm... that didn't seem to work for me. I tested this on a brew with just single lines for each page:

```
test 1
\page
test 2
\page
test 3
\page
...
```

No matter what page the renderer is at, clicking the **left** button moves the cursor up to `test 1`. No matter where the cursor is, clicking **right** does nothing. The **up** button seems to work.

I will have to try again and see if I am missing one of the newer commits or something.

G-Ambatte:
Both snippetbar functions should be working as of last night's commit (https://github.com/naturalcrit/homebrewery/pull/2162/commits/af3a403971b343f5e9cbccbf97c38754d9476b3d); the only commit since then was a styling change.

Gazook89:
One thing styling-wise I notice is that the top button in the divider overlaps with the scroll up button (when there is a scrollbar in the editor).  Maybe they should be moved down just a tad.  Or, move them all the way to the bottom but not overlapping the scroll-down button.  

Maybe consider hiding them until :hover over divider bar, so that they aren't overlapping text or preview when you are actually typing:

```css
#divider .jumpButtons {
    opacity:0;
    transition: all .2s ease;
}

#divider:hover .jumpButtons {
    opacity: 1;
    transition: all .2s ease;
}
```

And hotkeys for this would be good.....?  I don't know if arrow keys can be used in hotkeys, but if so I would think `ctrl` `left arrow` and `ctrl` `right arrow` to indicate jump source and jump preview, respectively.  Include `cmd` for macs.   I just don't know if all computers have arrow keys or not  (obviously not all have numpads, though).

ericscheid:
> Moving the source is dependent on the preview scroll position - it finds how far down the brew renderer has been scrolled, then iterates through each page element, adding the height until it reaches or exceeds the current scroll position. The brew text is then parsed until it reaches that number of lines containing `\page` and the cursor position is set to `{that number}, 1`.

Instead of `adding height etc` (which is problematic because of interpage guttering, `.page#pN` custom sizing, etc) .. you can do something using [getBoundingClientRect()](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) (e.g. https://stackoverflow.com/a/7557433/1601921)

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +174,39 @@ const Editor = createClass({
 		}
 	},
 
-	brewJump : function(){
-		const currentPage = this.getCurrentPage();
-		window.location.hash = `p${currentPage}`;
+	brewJump : function(targetPage=this.getCurrentPage()){
+		if(!window || this.isMeta()) return;
+		window.frames['BrewRenderer'].contentDocument.getElementById(`p${targetPage}`).scrollIntoView({ behaviour: 'auto', block: 'start' });
```
calculuschild, line 179:
```suggestion
		window.frames['BrewRenderer'].contentDocument.getElementById(`p${targetPage}`).scrollIntoView({ behavior: 'auto', block: 'start' });
```

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -60,10 +60,35 @@ const Snippetbar = createClass({
 		});
 	},
 
+	toTop : function(){
+		this.props.sourceJump(0);
+		this.props.brewJump(1);
+	},
+
+	moveBrew : function(){
```
calculuschild, line 68: If these functions are just wrappers to call another function with no change, we probably don't need this and can call `this.props.brewJump()` directly from the button.

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +174,39 @@ const Editor = createClass({
...
+		// const hashPage = (page != 1) ? `p${page}` : '';
+		// window.location.hash = hashPage;
+	},
+
+	sourceJump : function(targetLine=null){
+		if(this.isText() || this.isStyle()) {
+			if(targetLine == null) {
+				const brewPosition = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0).scrollTop;
```
calculuschild, line 187: We do a similar calculation for this in `the `handleScroll()` function in `brewRenderer.jsx` but it's actually pretty simple:

```
const target = e.target;
viewablePageNumber = Math.floor(target.scrollTop / target.scrollHeight * prevState.pages.length)
```

Where `e` is the `brewRenderer` DIV element.

client/homebrew/editor/editor.jsx
```diff
@@ -96,7 +96,7 @@ const Editor = createClass({
 	},
 
 	getCurrentPage : function(){
-		const lines = this.props.brew.text.split('\n').slice(0, this.cursorPosition.line + 1);
+		const lines = this.props.brew.text.split('\n').slice(0, this.refs.codeEditor.getCursorPosition().line + 1);
 		return _.reduce(lines, (r, line)=>{
 			if(line.indexOf('\\page') !== -1) r++;
```
calculuschild, line 101: Make sure to account for Legacy and V3 line splitting, since they are slightly different.

```
if((this.props.renderer == 'legacy' && line.includes('\\page')) ||
   (this.props.renderer == 'V3'     && line.match(/^\\page$/)))
{
...
}
```

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +174,39 @@ const Editor = createClass({
...
+			if(targetLine == null) {
+				const brewPosition = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0).scrollTop;
+				const pageCollection = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('page');
+
+				let currentPage = 0;
+				let currentPagePosition = 0;
+				while (currentPagePosition < brewPosition && currentPage < pageCollection.length) {
+					currentPagePosition = currentPagePosition + pageCollection[currentPage].scrollHeight;
+					currentPage++;
+				};
```
calculuschild, line 195: We do a similar calculation for this in the `handleScroll()` function in `brewRenderer.jsx` but it's actually pretty simple (excerpted and modified for legibility):

```
const target = e.target;
const currentPage = Math.floor(target.scrollTop / target.scrollHeight * prevState.pages.length)
```

Where `e` is the `brewRenderer` DIV element.

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +174,39 @@ const Editor = createClass({
...
-		window.location.hash = `p${currentPage}`;
+	brewJump : function(targetPage=this.getCurrentPage()){
+		if(!window || this.isMeta()) return;
+		window.frames['BrewRenderer'].contentDocument.getElementById(`p${targetPage}`).scrollIntoView({ behaviour: 'auto', block: 'start' });
+		// const hashPage = (page != 1) ? `p${page}` : '';
+		// window.location.hash = hashPage;
+	},
+
+	sourceJump : function(targetLine=null){
+		if(this.isText() || this.isStyle()) {
```
calculuschild, line 185: Probably don't want to jump the source if we are in the Style panel.

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: They are wrappers at the moment; if further development doesn't discover a need for any additional functionality here, then my plan is to call it directly.

client/homebrew/editor/editor.jsx
G-Ambatte: As I understand it, `Math.floor(target.scrollTop / target.scrollHeight * pages.length)` finds the fraction of the BrewRenderer height and multiplies it by the number of pages, truncating the result to zero decimal places. This approach then would be inaccurate for pages of differing sizes - for example, if some pages were in portrait and others in landscape.
This method should (bearing in mind @ericscheid's comment about using getBoundingRect() instead) more accurately track the top of each page.

But, with that said, how often do users change the page size halfway through the brew? The `Math.floor()` method has been accurate enough for a long time now, but these functions will shine a light on any inaccuracy.

client/homebrew/editor/editor.jsx
G-Ambatte: This should now be included in the latest commit.

G-Ambatte:
I think I've misunderstood `getBoundingClientRect()`... If I understand it correctly now, we should be able to simply iterate through the `page` collection and return the first element with a `.top` position that is higher than zero.

G-Ambatte:
I've changed the function to use `getBoundingClientRect().bottom`; the page is considered visible if more than half of it is currently on the screen - that is, if the `bottom` is more than half the current BrewRenderer `height`.

client/homebrew/editor/editor.jsx
calculuschild: Good point about potentially different page sizes. I do like the `boundingRect` approach better.

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +178,42 @@ const Editor = createClass({
 		}
 	},
 
-	brewJump : function(){
-		const currentPage = this.getCurrentPage();
-		window.location.hash = `p${currentPage}`;
+	brewJump : function(targetPage=this.getCurrentPage()){
+		if(!window || this.isMeta()) return;
```
calculuschild, line 182: Perhaps the buttons should not be present at all in the Meta tab (and `left` and `right` should not appear in the Style tab). Instead of checking here and having a nonfunctional button.

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +178,42 @@ const Editor = createClass({
...
+		// console.log(`Scroll to: p${targetPage}`);
+		window.frames['BrewRenderer'].contentDocument.getElementById(`p${targetPage}`).scrollIntoView({ behavior: 'smooth', block: 'start' });
+		// const hashPage = (page != 1) ? `p${page}` : '';
+		// window.location.hash = hashPage;
+	},
+
+	sourceJump : function(targetLine=null){
+		if(this.isText() || (this.isStyle() && targetLine == 0)) {
```
calculuschild, line 190: Same thing, maybe `left` and `right` should not appear (or be greyed out) in the Style tab instead of checking here.

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +178,42 @@ const Editor = createClass({
...
+				targetLine = 0;
+
+				const pageCollection = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('page');
+				const brewRendererHeight = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0).getBoundingClientRect().height;
+
+				let currentPage = 1;
+				for (const page of pageCollection) {
+					if(page.getBoundingClientRect().bottom > (brewRendererHeight / 2)) {
+						const pageId = page.id;
+						currentPage = parseInt(pageId.slice(1, pageId.length)) || 1;
```
calculuschild, line 201:
```suggestion
						currentPage = parseInt(page.id.slice(1)) || 1;
```

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +178,42 @@ const Editor = createClass({
...
+					}
+				}
+
+				if(currentPage > 1){
+					const textString = this.props.brew.text.split('\\page').slice(0, currentPage-1).join('\\page');
+					const textPosition = textString.length;
+					const lineCount = textString.slice(0, textPosition).split('\n').length;
+
+					targetLine = lineCount;
+				}
```
calculuschild, line 212: We need to handle the difference between Legacy and V3 `\page` markers again. I might suggest replicating the logic from the `getCurrentPage()` function you made: split the text into lines, and scan through each until the total number of `\page` found matches `currentPage`. Using the same logic also exposes an opportunity for cleanup of duplicate code in the future, where currently it's not obvious that these are doing essentially the same thing.

calculuschild:
The only other thing I can think of is getting the divider buttons functional. I don't know if we *need* the left/right toolbar buttons once those are working (two sets of buttons almost right next to each other), but that's up for debate. Otherwise I think it's good to go.

client/homebrew/editor/editor.jsx
G-Ambatte: Looks like we can just pass regex to `split()` so it shouldn't be difficult to implement; splitting this to a separate pagination function that can be re-used elsewhere is probably worthwhile for DRY eliminations.

The `currentPage > 1` check was more about it skipping the unnecessary calculations - if current page in the Brew Renderer has been determined to be 1 (or less) then the target line to move to should just remain at it's initial value of 0.

calculuschild: > The currentPage > 1 check was more about it skipping the unnecessary calculations...

That's what I figured. But I think in this case the slight improvement in code simplicity is worth the slight chance of being able to skip a calculation.

> splitting this to a separate pagination function that can be re-used elsewhere is probably worthwhile for DRY eliminations.

It would be nice too (in a different PR probably) to have the pagination common between the Editor and BrewRenderer components. As you know already, we do this `currentPage` calculation in `brewRenderer` so we can draw the page count box and handle partial brew rendering.

G-Ambatte:
> The only other thing I can think of is getting the divider buttons functional. I don't know if we _need_ the left/right toolbar buttons once those are working (two sets of buttons that do the same thing almost right next to each other), but that's up for debate. Otherwise I think it's good to go.

From what I've been reading, I think we'll need to pass the functions as `refs` to the SplitPane component. I believe that we already do this for the CodeEditor component; if I'm correct, then there is already a working example in the codebase that I can crib from.

calculuschild:
Would it make sense to move these functions up to the splitPane component then, instead of passing it up from the editor?

G-Ambatte:
Probably - it generally seems easier to pass down, rather than up. At first, it made sense to me that they should exist at the Editor level, although thinking about it now, the SplitPane level does make more sense, as the function effectively touches both sides of the divider.

On a different note, it appears that Alt-Left and Alt-Right key shortcuts are already hooked by CodeMirror by default, which duplicates the functionality of `Home` and `End` keys. These would seem to be prime targets then to instead be used for match source to brew (left) and match brew to source (right).

G-Ambatte:
It might make more sense to go up one level further, to the EditPage itself, as at this level we have access to both the SplitPane component and the brew text.

calculuschild:
Hmmmm.... Moving up to `editPage` could work, but then we need to duplicate that into `/new` again.

**Edit:** Ahhh... it actually might be easier to just update some prop on the child when the button is clicked, and then in the `editor` look at `componentDidUpdate()` and if that value changed, run either Left or Right. Example from [StackOverflow](https://stackoverflow.com/questions/51322103/trigger-a-function-that-is-inside-the-child-component-from-the-parent):

> To trigger a function on a child component from parent, you can pass a property that changes when some action happens on the parent component. This will trigger a re-render (in most cases, for more info please take a look at [shouldComponentUpdate](https://reactjs.org/docs/react-component.html#shouldcomponentupdate)) in child component. You can check the property changes and then do what you need to do in child component.

```
class Parent extends React.Component {
  state = {
    someParameter: 'someInitialValue',
  }
  someFunction = (text) => {
    this.setState({ someParameter: 'someValue' });
  }
  render () {
    return(
      <Child someProperty={this.state.someParameter} />
    )
  }
}

class Child extends React.Component {
  someFunction = (text) => {
    console.log('Message from parent: ', text);
  }
  componentDidUpdate(prevProps, prevState, snapshot) {
    // Check if the suplied props is changed
    if(prevProps.someProperty !== this.props.someProperty) {
      // run the function with the suplied new property
      this.someFunction(this.props.someProperty);
    }
  }
  render() {
    return(
      {/* ... */}
    )
  }
}
```

G-Ambatte:
Of course... Yes, that'll be much easier.

client/homebrew/editor/editor.jsx
```diff
@@ -174,9 +178,40 @@ const Editor = createClass({
...
+
+				let currentPage = 1;
+				for (const page of pageCollection) {
+					if(page.getBoundingClientRect().bottom > (brewRendererHeight / 2)) {
+						currentPage = parseInt(page.id.slice(1)) || 1;
+						break;
+					}
+				}
+
+				const textSplit = this.props.renderer == 'V3' ? '\n\\page\n' : '\\page';
```
calculuschild, line 205:
```suggestion
				const textSplit = this.props.renderer == 'V3' ? `/^\\page$/gm` : '\\page';
```

client/homebrew/editor/editor.jsx
G-Ambatte: SnippetBar buttons have been removed; Divider buttons only appear in Content Editor.

client/homebrew/editor/editor.jsx
G-Ambatte: SnippetBar buttons removed; Divider buttons only appear in Content Editor.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -83,20 +90,58 @@ const SplitPane = createClass({
...
 	},
 
 	render : function(){
 		return <div className='splitPane' onMouseMove={this.handleMove} onMouseUp={this.handleUp}>
-			<Pane ref='pane1' width={this.state.currentDividerPos}>{this.props.children[0]}</Pane>
+			<Pane
+				ref='pane1'
+				width={this.state.currentDividerPos}
+			>
+				{React.cloneElement(this.props.children[0], {
```
calculuschild, line 139: Can you clarify to me why we need to have `cloneElement` here?

client/homebrew/editor/editor.jsx
```diff
@@ -191,6 +232,7 @@ const Editor = createClass({
 
 	renderEditor : function(){
 		if(this.isText()){
+			this.props.setMoveArrows(true);
```
calculuschild, line 235: I think it's generally discouraged to update the state inside of Rendering, since updated state is what triggers a Render. It's possible to end up in an infinite loop this way.

Is it possible to call this inside of the `handleViewChange()` maybe?

shared/naturalcrit/splitPane/splitPane.jsx
G-Ambatte: As I understand it, the `this.props.children[0]` is immutable, so we need to clone it in order to modify it to add the extra props.

client/homebrew/editor/editor.jsx
G-Ambatte: Just moved the logic to `handleViewChange`, should be in the next commit which will be pushed shortly.

shared/naturalcrit/splitPane/splitPane.jsx
calculuschild: Hm. I looked it up and that seems to be the case. How strange that they would do it that way....

G-Ambatte:
This seems to be functional now, there was an issue with the button position not updating when the divider was being moved due to a window resize but I'm fairly certain that it's resolved now.

client/homebrew/editor/editor.jsx
G-Ambatte: There was a small edge case where the previous logic resulted in the source position for page 1 resulting in line 1, rather than line 0. The current logic checks to see if there are any new lines `\n` in the slice of text between the start and the current page position; if not, then the line is set to 0.
But it occurs to me that this is functionally the same as the `currentPage > 1` check.

shared/naturalcrit/splitPane/splitPane.jsx
G-Ambatte: I am struck by the thought that we could move the Pagination functions to SplitPane and use a similar method to append the pagination function as a prop to both the Editor and the BrewRenderer.
That probably falls outside the scope of this particular PR, so I'm going to raise the deduplication as a separate issue.

G-Ambatte: Raised as #2198.

G-Ambatte: > Hm. I looked it up and that seems to be the case. How strange that they would do it that way....

I wonder if we couldn't do something like:

```
getDefaultState : function() {
    children: this.props.children
}
```

then modify `this.state.children` and finally pass it as `this.state.children[0]` rather than `this.props.children`.

calculuschild:
This *seems* to be working pretty well. A couple of nitpicks that we can address now or in a different PR, depending on your interest, @G-Ambatte :

- Perhaps we can add a tooltip for the buttons when hovering
- With "sourceJump", the editor does not scroll `\page` to the top of the view when jumping from an early page to a later page. Instead the view only moves enough so the top of `\page` is visible at the bottom of the screen. Probably would be nice to scroll the cursor to the top of the editor page, similar to how `brewJump` scrolls the brewPreview to the top of the page.
- Would be nice to have the text editor scroll smoothly into place like the brewRenderer does.
- Would be nice to have a visual cue for the text editor on "sourceJump", like flashing the line the cursor jumped to or something to help the user pinpoint.
- Would be nice to have "no action" button clicks (view/editor is already in position) still show something visually. For instance, have the brewRenderer bounce slightly back into position. Having the text editor flash the current line as above would work for the editor.

G-Ambatte:
General:
- [ ] Add a tooltip for the buttons when hovering
  - This should be simple to add to the existing LESS file.

Source Editor:
- [ ] Scroll `\page` to the top of the view
- [ ] Scroll smoothly into place (see `cm.scrollIntoView()` maybe?)
  - These first two points maybe solved in the same function. Further investigation to follow.
- [ ] Flash target line in Source editor

Brew Preview:
- [ ] Bounce brew when movement not required
  - If we incorporated an animation effect, scrolling down slightly in the direction opposite the movement, then scroll slightly past the desired position before rolling back, then scrolling to the current page would activate this animation and cause a slight bounce effect. How possible this is in reality and the difficulties of actual implementation have not yet been investigated.

G-Ambatte:
> - [ ] Scroll `\page` to the top of the view
I've attempted to add the `cm/addon/scroll/scrollpastend` add on to CodeMirror; this should allow for the target line to always be scrolled to the top of the editor window. However, despite having added this in the same way as the other addons, it appears to have no effect. Further investigation to follow.

> - [ ] Scroll smoothly into place (see `cm.scrollIntoView()` maybe?)
Test run using `cm.scrollIntoView()` did not result in an appreciable difference - the scroll is very fast. Currently investigating `cm.scrollIntoView` options, which seem to be very limited.

> - [ ] Flash target line in Source editor

I've added a class to the target line, and styling to the LESS file to make it visually distinct. This class is then removed when the editor is next interacted with, extending the existing function that removes custom classes from CM lines. I imagine we could make this an animation using keyframes, etc in the LESS, but at this point, I'm happy to leave that for later development if desired.

calculuschild:
I think I got everything working (except the Previewer bouncing, but I didn't see an easy way to do that). I think it's probably in a good spot to send out? Any objections/comments?

https://user-images.githubusercontent.com/5878534/171939271-b52ff169-db9f-4860-bf41-cffab9f062bf.mp4

G-Ambatte:
Brew bounce added.

client/homebrew/editor/editor.jsx
```diff
@@ -189,7 +190,28 @@ const Editor = createClass({
 	brewJump : function(targetPage=this.getCurrentPage()){
 		if(!window || this.isMeta()) return;
 		// console.log(`Scroll to: p${targetPage}`);
-		window.frames['BrewRenderer'].contentDocument.getElementById(`p${targetPage}`).scrollIntoView({ behavior: 'smooth', block: 'start' });
+		const currentPos = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer')[0].scrollTop;
```
calculuschild, line 193: Since you access `window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer')[0]` multiple times, should it be stored in a `const` ?

client/homebrew/editor/editor.jsx
```diff
@@ -189,7 +190,28 @@ const Editor = createClass({
...
+		const targetPos = window.frames['BrewRenderer'].contentDocument.getElementById(`p${targetPage}`).getBoundingClientRect().top;
+		const interimPos = targetPos >= 0 ? -30 : 30;
+
+		const bounceDelay = 100;
+		const scrollDelay = 500;
+
+		if(!this.state.isBrewScrolling) {
+			this.setState({
+				isBrewScrolling : true
+			});
```
calculuschild, line 203: Do we need the state handling at all? I disabled this check and didn't notice any functional change.

If we don't need it, that would simplify this somewhat.

calculuschild:
> Brew bounce added.

Wow! That's a really clever way of doing this. It feels really nice. Great work!

client/homebrew/editor/editor.jsx
G-Ambatte: It stops the function being spammed - without it, you can rapidly click the button to reverse up the page, before returning to the correct spot when you stop clicking it.

client/homebrew/editor/editor.jsx
G-Ambatte: That's a good idea; I've made the change, tested, and as it's working fine, pushed it through in a fresh commit now.

client/homebrew/editor/editor.jsx
calculuschild: Ah. In that case, it might be a little cleaner to use Lodash's `_.debounce()` function wrapper. See https://lodash.com/docs/4.17.15#debounce

Any function you wrap inside, like `_.debounce(jumpEditor, 150)` will only execute every 150 ms at most. Probably with the `'leading': true,` option so it triggers on the first click of a clicking spam, rather than the last click.

G-Ambatte: Hmmm... I removed the state checks, wrapped the animation function in a `_.debounce` call... and now it's not executing anything in that block at all. If I recall correctly, it's also used in the editPage.jsx, I'm about to jump over there to check out a working example.

G-Ambatte: I've mimicked the `editPage.jsx` usage of `_.debounce`, which has it functional again... however I'm now back at the original issue of needing a state to indicate that the animation is in progress, during which `debounce.cancel` should be called instead.

...I think I've got it now, using `throttle` instead of `debounce`. I'll push a commit shortly with the update.

client/homebrew/editor/snippetbar/snippetbar.less
```diff
@@ -46,6 +46,30 @@
 					color : black;
 				}
 			}
+			&.toTop{
```
calculuschild, line 49: I guess we don't need these styles anymore with the snippetbar buttons removed.

client/homebrew/editor/editor.jsx
calculuschild: Awesome!

G-Ambatte: There was some weird behavior, it looks better now with a slight change to the `throttle` options.

client/homebrew/editor/snippetbar/snippetbar.less
G-Ambatte: Agreed; pruning them now.

client/homebrew/editor/editor.jsx
G-Ambatte: Just fell down a rabbit hole; there are requests for a callback on completion of `scrollIntoView` going back at least six years; I'm pretty sure I saw some related issues that are closer to nine. It would make it a lot easier to lock out additional calls to the jump function if they'd been implemented.

G-Ambatte:
I think this is now in a very good position to merge.

Final functionality wrap up:
- **Brew Movement:** The current page number is determined by splitting the source text from the first character to the current cursor position, then the number of `\\page` (or `/^\\page$/` for v3) is counted (min 1). The current scroll position and the position of the target element (which has an ID of `p` + the previous determined page number) are determined; the target element position is relative, so if it is positive (down), a negative (up) interim scroll position is set, and vice versa. The brew preview is then scrolled to this interim position, and a timer set to scroll to the target position 100ms later.
The entire function is throttled (via Lodash `throttle`) so that the animation/movement cannot be called more than once every 500ms.

- **Source Movement:** The current page is determined by iterating through each page element and breaking out of the loop as soon as the bottom of a page is more than half the current display height of the brew renderer. The source text is then split on page breaks, reduced to the determined number of pages, then reassembled before being split once again on line breaks. The length of the resulting collection determines the line number to scroll to.
The current scroll position and target scroll positions are then determined and the the CodeMirror editor is scrolled vertically using `setInterval`, as the built in CodeMirror scroll options result in instantaneous movement. When the target position is reached, the custom class `sourceMoveFlash` is applied to the target line in order to highlight it in the editor.

- **Updating Elements from SplitPane:** The Editor element is applied as a child of SplitPane on each page that features an editor; in order to add the movement methods to this element universally - that is, without having to edit every single page - the SplitPane render method was changed to clone the immutable Editor prop (`React.cloneElement(this.props.children[0], { newPropsToAdd })`) so that the divider button display update and Brew + Source movement functions could be added.

calculuschild:
\*chef's kiss\* 👨‍🍳👌

Issue #2215, opened 2022-06-13
Add "spellcasting ability" to the Class Features snippet


calculuschild:
The small bit of formatting for "spellcasting ability" seems to trip people up. We can include it into the Class Features snippet:

<pre><code>
## Class Features
As a fletcher, you gain the following class features
#### Hit Points

**Hit Dice:**                    :: 1d10 per fletcher level
**Hit Points at 1st Level:**     :: 10 + your Constitution modifier
**Hit Points at Higher Levels:** :: 1d10 (or 6) + your Constitution modifier per fletcher level after 1st

#### Proficiencies

**Armor:**   :: Medium armor, Shields, Heavy armor
**Weapons:** :: Martial weapons, Squeegee
**Tools:**   :: Artian's tools, Thieve's tools

**Saving Throws:** :: Intelligence, Dexerity
**Skills:**        :: Choose two from Perception, Deception, Nature, Acrobatics

<b>#### Spellcasting Ability

{{text-align:center
**Spell save DC** = 8 + your proficiency bonus + your Charisma modifier
:
**Spell attack modifier** = your proficiency bonus + your Charisma modifier  
}}</b>
</code></pre>

jlgraves:
Okay, I'm pretty new at this - but this seemed like a nice place to get my feet wet contributing.

I've made this change to both `classfeature.gen.js` on the `snippets` and `snippetsLegacy` folders.   They seem to work on my end; I added some randomization similar to the other elements.

If this looks OK, I can submit a pull request from my fork. 


```
return dedent`
		## Class Features
		As a ${classname}, you gain the following class features
		#### Hit Points

		**Hit Dice:**                    :: 1d${hitDie} per ${classname} level
		**Hit Points at 1st Level:**     :: ${hitDie} + your Constitution modifier
		**Hit Points at Higher Levels:** :: 1d${hitDie} (or ${hitDie/2 + 1}) + your Constitution modifier per ${classname} level after 1st

		#### Proficiencies

		**Armor:**   :: ${_.sampleSize(['Light armor', 'Medium armor', 'Heavy armor', 'Shields'], _.random(0, 3)).join(', ') || 'None'}
		**Weapons:** :: ${_.sampleSize(['Squeegee', 'Rubber Chicken', 'Simple weapons', 'Martial weapons'], _.random(0, 2)).join(', ') || 'None'}
[...]

PR #2223, opened 2022-06-18
Implements issue #2215 - spell-casting ability

jlgraves:
This pull is two commits which implements a commonly confusing formatting issue in #2215 .
* One commit is for the legacy version of the snippet
* One commit is for the modern version of the snippet
The formatting is slightly different in the two versions, but the content is there in both.

 I checked `$ npm run verify` and it passed, so this should be good to go!


client/homebrew/editor/snippetbar/snippetsLegacy/classfeature.gen.js
```diff
@@ -32,6 +33,11 @@ module.exports = function(classname){
 		`- **Saving Throws:** ${_.sampleSize(abilityList, 2).join(', ')}`,
 		`- **Skills:** Choose two from ${_.sampleSize(skillList, _.random(4, 6)).join(', ')}`,
 		'',
+		'#### Spellcasting Ability',
+		`**Spell save DC** = ${_.sample([6, 8, 10])} + your proficiency bonus + your ${spellSkill} modifier`,
+		'',
+		`**Spell attack modifier** = your proficiency bonus + your ${spellSkill} modifier  `,
+		'',
```
calculuschild, line 40: Great work on this PR! Just one thing: The text here needs to be centered in the Legacy version too. In the V3 version we do it with `{{text-align:center`. Here we will need to resort to a raw HTML div: `<div style=text-align:center>`

If you add that for me then I'll merge this PR.

calculuschild (changes requested):

client/homebrew/editor/snippetbar/snippetsLegacy/classfeature.gen.js
jlgraves: Okay, I added that change and made a couple small typo fixes at the same time.

calculuschild (approved):

calculuschild:
Thanks @jlgraves! I'll merge this now. Might not go live quite yet as we finish up some other issues but it will be in the next version.

PR #2224, opened 2022-06-18
react-router-dom 5->6 upgrade

jeddai:
* Update react-router-dom to 6.3.0
* Add WithRoute function component wrapper

Instead of converting all our route class components to function components, I made a function component wrapper called `WithRoute`. As far as I can tell it's working great, any and all thoughts/feedback would be appreciated!


client/homebrew/homebrew.jsx
```diff
@@ -12,6 +13,21 @@ const NewPage = require('./pages/newPage/newPage.jsx');
 //const ErrorPage = require('./pages/errorPage/errorPage.jsx');
 const PrintPage = require('./pages/printPage/printPage.jsx');
 
+const WithRoute = (props)=>{ //(Element, additionalPropsFn)=>{
+	const params = useParams();
+	const location = useLocation();
+	let additionalProps = {};
+	if(props.additionalPropsFn) {
+		additionalProps = props.additionalPropsFn({ props, params, location });
+	}
```
calculuschild, line 22: Looks like react-router-dom v6' now has a built-in way to handle URL query parameters using the new [useSearchParams](https://reactrouter.com/docs/en/v6/hooks/use-search-params) hook. I think that means we can probably get rid of `additionalPropsFn` to handle that and just do a one-liner `const queryParams = useSearchParams();`

This also means we can get rid of the `query-string` dependency.

client/homebrew/homebrew.jsx
```diff
@@ -12,6 +13,21 @@ const NewPage = require('./pages/newPage/newPage.jsx');
...
+	const Element = props.el;
+	return <Element {...{
+		...props,
+		el                : undefined,
+		additionalPropsFn : undefined
+	}} {...params} {...additionalProps}/>;
```
calculuschild, line 28: If we remove `additionalPropsFn` and use `queryParams` as I noted above, I think we can simplify this to (something like):

```suggestion
	const Element = props.el;
	props.el = undefined;
	return <Element {...props, ...params, ...queryParams}/>;
```

calculuschild (changes requested):
Last think I'm noticing: We are getting errors on the server for "route not found". We might need to add a little extra something here or in the backend:

> No routes matched location "/assets/homebrew/favicon.ico"

> No routes matched location "/randomaddress" (should land on home page if no match)

client/homebrew/homebrew.jsx
jeddai: The curly braces will have to be one set per variable we're destructuring, just as a result of how jsx handles it. 

I did originally try to modify `props` like what you have in your suggestion and React said no -- it's apparently immutable.

jeddai:
So I fixed the favicon one, as for the random base url I think the existing app routes you to the new page. I think we should consider adding a proper 404 page instead maybe?

calculuschild:
We can look at a 404 page in a later PR, since I think that will end up being a bigger effort beyond just getting this dependency updated.

Until then, I just don't want the Homebrewery to land at a blank page when a bad URL is entered. We should be able to get away with something real simple like a wildcard route that just redirects to the homepage.

jeddai:
Done!

calculuschild (approved):

calculuschild:
Looks great! Merging this now. 👍

Issue #758, opened 2018-09-17
[FEATURE]: Add folders/tags to organize brews...


ctgilley:
## Description

I've got a lot of brews, and I imagine others do as well.  It would be great to be able to organize these into folders that I can drill into for specific brews.  For example, I might want to have one folder for each campaign that I'm running, or one to hold my custom magic items, or another to hold my custom races and classes.

ericscheid:
I would suggest tags instead of folders.

ctgilley:
Sure, tags would be fine -- any way to structure/filter/organize.

jeddai:
Discussion from #697:
> > Side note: the homebrew data model and UI does include a "tags" field, but it is commented out and undeveloped.
> > There is interest in tags for organising reasons too. #758
>
> I was just thinking about that; I have about a dozen brews that I made to help out someone in the subreddit; I've taken to giving them a title that starts with "Example:" so that they appear in one place in the list, but it'd be even better if I could dump them in an "Examples" folder where I don't actually have to look at them.

@G-Ambatte what if we handled the document filtering via tags, but also addressed the visibility concern by making the user page a dashboard of sorts? 
1. Set up documents to allow for tags, 
2. Add a section to the user page that allows users to filter their documents based on the tags they have on their documents.
    * This would contain a unique list of all the tags present on a user's documents
3. When you click a tag, it filters the user's documents so only documents with that tag are showing
4. When you change which tags/sort method are being used, that gets stored in local storage
5. When you visit that page, it defaults to whatever is in local storage

jeddai:
Smart folders would be easily doable using the tag system, thanks to NASA from the Discord of Many Things for this idea.

ericscheid:
A question on "folders" (aka views) .. would authors want a folder where some of the brews are for everyone to view, and other brews only for the author to view (e.g. DM only, vs all players). And of these player-viewable brews, can some of them not be marked as "Published" (i.e. discoverable via the homebrewery public content search facilities).

I would want to a) create a container for a purpose (e.g. Friday Game Group), b) toss in all the relevant brews, c) mark some as me-only brews, d) have a link I can send to all my players. And maybe have sub-folders there too.

This can _kinda_ be done via tags, there are definite overlaps. But also not. Food for thought.

ericscheid:
> Smart folders would be easily doable using the tag system, thanks to NASA from the Discord of Many Things for this idea.

See also #1858

calculuschild:
[...]

PR #2233, opened 2022-06-25
Document tags

jeddai:
When run this locally, make sure you set `MIGRATE=true` in your terminal. You should see `Successfully updated all brews to new schema definition` in the run console. Subsequent runs should say `No brews to update`. These messages correspond to updating the schema definition for `tags` to be a string array instead of a string.

From the metadata editor, you should be able to do the following:
* Add a tag by typing in the tag input and pressing `Enter` or clicking the check icon
* Update a tag by clicking on the text, changing the value, and pressing `Enter` or clicking the check icon
* Close the tag update box by pressing `Enter` to save or `Escape` to cancel (alternatively, clicking the check to save or `X` to cancel)
* Delete a tag by clicking the `X` within the tag badge

Tags should be saved to the document in the database when the debounce hits or on a manual save.

#758


client/homebrew/editor/editor.jsx
```diff
@@ -299,7 +299,10 @@ const Editor = createClass({
 					style={{ display: 'none' }}
 					rerenderParent={this.rerenderParent} />
 				<MetadataEditor
-					metadata={this.props.brew}
+					metadata={({
+						...this.props.brew,
+						tags : typeof this.props.brew.tags === 'string' ? [] : this.props.brew.tags
+					})}
```
calculuschild, line 305: I wonder if we can have this type checking handled upon receiving the brew to begin with. Make sure we have the `brew` in a good format right from the start instead of deep in here where it's kind of hidden. Especially where we might want to access `tags` from elsewhere and don't want to have to do this type check every time.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -45,9 +47,10 @@ const MetadataEditor = createClass({
 	},
 
 	handleFieldChange : function(name, e){
-		this.props.onChange(_.merge({}, this.props.metadata, {
+		this.props.onChange({
+			...this.props.metadata,
```
calculuschild, line 51: I do like this better. There's no functional difference you've noticed doing it with the spread operator vs. `merge` with an empty object?

calculuschild (changes requested):

calculuschild:
@jeddai Sorry this took *so long* to review. I want to push the string editor a bit more and look for bugs, but this looks pretty good.

client/homebrew/editor/editor.jsx
jeddai: So handle it on the backend when it sets req.brew?

jeddai:
So sorry I dropped off the face of the planet for a bit there, wrapping up these changes now 😬

calculuschild:
No problem. I've been fairly absent myself.

jeddai:
Should be updated now!

client/homebrew/editor/editor.jsx
jeddai: Done!

calculuschild:
@G-Ambatte @Gazook89 @RKuerten  Perhaps one of you guys could bug test this? I've reviewed the code a while ago but I'm down in the trenches of my final PhD project and am struggling to find time to dedicate to this for a few weeks. When this was first made into an old PR there were just some edge cases from inputting weird text and deleting things in weird orders etc. I think @jeddai fixed most of those but if anybody can find a moment to poke at it just to be sure before I merge it.

calculuschild:
Issue: If I try to add a duplicate tag, it highlights in red and won't let me save. All good. However, if I try to edit an *existing* tag, it counts itself in the duplicates. That is, if I start editing a tag but then change my mind, I am forced to change the tag anyway because it stays highlighted in red, and the checkmark is gone, and pressing ENTER does nothing until the tag is changed from its saved value. I *can* click the `x` which ends the editing, but it's the same `x` used for deleting.

- Currently-edited tag should not be counted in duplicate checking
- `x` button when editing should have the same behavior as when not editing, or they should be different buttons.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,138 @@
...
+				value,
+				editing : false
+			})) : [],
+			temporaryValue : '',
+			updateValue    : ''
+		};
+	},
+
+	componentDidUpdate : function(prevProps) {
+		if(!_.eq(this.props, prevProps)) {
```
calculuschild, line 31: I think by definition `componentDidUpdate` is triggered to run whenever `props` is updated, so `this.props != prevProps` will always be true.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -0,0 +1,138 @@
...
+			this.setState({
+				valueContext : !!this.props.values ? this.props.values.map((newValue)=>({
```
calculuschild, line 33: I don't think Javascript needs `!!` to typecast to boolean. `this.props.values` will already be evaluated as a boolean because of the ternary operator.

calculuschild (approved):

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
jeddai: It also triggers when the state is updated, and since the contents of this `if` are running `setState` it runs until the call stack is exceeded if the `if` statement is removed.

calculuschild: Ah, right right. In that case can we have it just check whether `this.props.values` has changed?

calculuschild:
Looks like you got those items fixed. Any last thing you want to do on this PR before I merge?

Issue #2234, opened 2022-06-25
How to globally disable first line indent?


Waldleufer:
## Description

Dear fabulous homebrewery developers,

When dealing with centred text blocks, imagine a poem, or song lyrics, for example, I find it really disturbing if the lines have an "overhang" to one side. 
I already found #636 where you describe a way to manually disable it by adding a `\` at the end of every paragraph. 

## Requested Feature
Could you maybe also add a CSS snippet that disables the new line indentation for a whole paragraph?

That would be really appreciated!

### Desired behaviour
Without first line indent:
![image](https://user-images.githubusercontent.com/22003632/175766859-6d6134bc-b69e-4bb9-997d-3f77b00c19f6.png)

### Actual behaviour
With the line indent (current default):
The lines 2, 3 and 4 have an "overhang" to the right.
![image](https://user-images.githubusercontent.com/22003632/175766925-89ae396c-76a7-408c-bd2d-92acc88ee1b8.png)

The combination of centred text and line indentation just doesn't work as expected, as you can see in this minimal example.

** Share Link**:
https://homebrewery.naturalcrit.com/share/1iiJh2tCevmD7hYfqtgxzsIFuv60f1IbVowPOzmuKi6Z6

**V3 Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>
{{wide
{{refrain 
  Sing us a song, you're the piano man
	
  Sing us a song tonight
	
  Well, we're all in the mood for a melody
	
  And you've got us feelin' alright
}}

{{refrain 
  Sing us a song, you're the piano man\
  Sing us a song tonight\
  Well, we're all in the mood for a melody\
  And you've got us feelin' alright
}}
}}

Styling:

```css
/* A4 Page Size */
.page{
	width  : 210mm;
	height : 296.8mm;
	text-align: center;
	font-size: 1.4em;
}
.refrain {
	background: linear-gradient(-8deg, #5A5AAD, #323299, #5A5AAD, #000080, #5A5AAD, #5A5AAD, #5A5AAD);
	font-weight: bold;
	-webkit-background-clip: text;
	color: transparent;
}

.refrain::first-letter {
	font-family: SolberaImitationRemake;
	font-size: 1.3cm;
	background: linear-gradient(-8deg, #5A5AAD, #323299, #5A5AAD, #000080, #5A5AAD, #5A5AAD, #5A5AAD);
	line-height: 0.5em;
	font-variant: small-caps;
	font-weight: normal;
[...]

PR #2236, opened 2022-06-27
Remove Indents: Proposing a style snippet

Waldleufer:
As discussed in #2234 

There is much room for new snippets, and maybe others can benefit from this as well?

😉

The target branch is probably not the one you would want to merge this into, feel free to change it appropriately. 

Cheers,
Waldleufer

### DIFF
--- a/client/homebrew/editor/snippetbar/snippets/snippets.js
+++ b/client/homebrew/editor/snippetbar/snippets/snippets.js
@@ -115,6 +115,14 @@ module.exports = [
 						}\n\n`
 			},
 			{
+				name : 'Remove Indents',
+				icon : 'fas fa-bars',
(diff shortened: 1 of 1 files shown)


Waldleufer:
Also, feel free to just close this PR if you don't want to include this snippet.

calculuschild:
Closing this, as we already have several methods of ignoring indents using Markdown directly:

```
paragraph 1\ <- End with a backslash, next paragraph will not be indented

paragraph 2
```

```
paragraph 1   <- End with two spaces, next paragraph will not be indented

paragraph 2
```

```
paragraph 1

paragraph 2 :: <- Including two `::` in the line causes it to be rendered as a <dd>, which is not indented.
```

PR #2256, opened 2022-07-14
Add SVG for Notification & Render Warning dismissal action

### LINKED ISSUE #1965: [LOCAL]:  Cannot close notifications when FontAwesome not available
As reported via Gitter, the icons that are used to close the notifications and warnings are loaded from FontAwesome. When FontAwesome is not available for any reason, these icons do not appear and thus cannot be clicked, meaning that the notifications and warnings cannot be closed, and the brew Preview is covered.

To ensure that the FontAwesome icons are always available, Homebrewery might switch to using the locally hosted package - [Font Awesome - Free NPM package](https://www.npmjs.com/package/@fortawesome/fontawesome-free).
It would also seem like an appropriate time to review any other web hosted fonts, such as Open Sans, hosted by Google Fonts. While initial investigation into Google Fonts did not reveal a WOFF version of Open Sans, the font does have a permissive license and a WOFF version was located at [FontSquirrel](https://www.fontsquirrel.com/fonts/open-sans).

G-Ambatte:
This PR resolves #1965.

The primary impact of this change is to improve usability of Homebrewery on systems that are completely isolated from the Internet; in this case, specifically unable to load the FontAwesome icons across the web.

This PR replaces the X used as the Dismiss icon on notifications and render warnings windows with a SVG, so these windows may still be closed even when the FontAwesome online library is unable to be reached.


G-Ambatte:
Recent reports from Gazook89 are that the notification close icon is also missing when FontAwesome is blocked, which appears to be happening on some local installs due to a cross origin error. This PR should resolve that as well, as - to my understanding - it is essentially the same issue.

calculuschild:
Hm.... Is there some chance this can be solved instead by just adding FontAwesome to our repo, but only using the local copy if the CDN download fails? Then we can get the whole icon set working for local users (or users who can't access the CDN), but 99% of everyone else can benefit from the CDN?

G-Ambatte:
Absolutely - adding FontAwesome to the repo would fix it. However, from memory, FontAwesome adds some ~14MB to the repo, while adding only the `cross` SVG as a special case adds only a few hundred bytes.

Gazook89:
I think you can get a subset of the FA library. I think.

calculuschild:
> However, from memory, FontAwesome adds some ~14MB to the repo,

I don't think it's quite that big (I remember it being quite large, but it's just a font file and come CSS... right? Since we don't use the SVG version? We should check) But to clarify, my thinking is there should be a way to make that a fallback so 99% of users still get their copy from the CDN, while only local/blocked users have to get the repo copy.

G-Ambatte:
The instructions at https://fontawesome.com/docs/web/setup/packages for installing an offline line are to install this package - https://www.npmjs.com/package/@fortawesome/fontawesome-free - which shows an unpacked size of 17.6MB.

If we can get a slimmed down version - just the icons that we actually use - that I imagine that would be much smaller.

calculuschild:
I don't think we need the whole package. If we just look at the CSS file we include and the font files it uses, it's only like 120 kb.

https://fontawesome.com/docs/web/setup/host-yourself/webfonts hints how this can work. I.e., download the package, but we only need the CSS file and any relevant font files. I suspect the majority of the file size in the package comes from including SVGs of each glyph, which I think tend to be larger than the font files we need.

calculuschild:
Closing this in favor of just self-hosting the whole font file instead of manually picking out individual SVGs.

PR #2290, opened 2022-08-27
Make V3 default setting in /new, and homepage

calculuschild:
Makes V3 the default setting.

- /new will default to use V3
- Homepage uses the existing `/v3_preview` text
- Metadata panel now links to the old homepage at `/legacy`
- Updated the "notice" popup to indicate the change.


jeddai (approved):
Tested it and it looks good!

PR #2297, opened 2022-08-28
update WithRoute to handle query params correctly

jeddai:
Please test on the brews page by filtering and refreshing -- the search should remain after refresh.
Also test that trying to print will automatically open both the page and dialog.


calculuschild:
Seems to be all working! Thanks for the quick fix!

PR #2300, opened 2022-08-29
[UserPage] Collapse Brew Groups

### LINKED ISSUE #1797: [UserPage] Change order of Published/Private Brews
As mentioned in [this Reddit comment](https://www.reddit.com/r/homebrewery/comments/qlh2hn/bug_sorting_by_views_not_really_sorting_by/hj4p8jr), some users would like to be able to re-order the groups on the User Page, as they do not edit Published brews, which are listed at the top of the page. Having the ability to push this group to the bottom of the page would improve their user experience.

G-Ambatte:
This PR resolves #1797.

- mousing over headers changes mouse to a pointer
- clicking headers collapses/expands sections
- collapse state is saved to local storage on page exit
- collapse state is loaded from local storage on page load
- default state (when local storage keys don't exist) is visible

### DIFF
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -6,27 +6,57 @@ const moment      = require('moment');
 
 const BrewItem    = require('./brewItem/brewItem.jsx');
 
+const USERPAGE_KEY_PREFIX = 'HOMEBREWERY-LISTPAGE-VISIBILITY';
+
 const ListPage = createClass({
 	displayName     : 'ListPage',
 	getDefaultProps : function() {
 		return {
 			brewCollection : [
 				{
-					title : '',
-					class : '',
-					brews : []
+					title   : '',
+					class   : '',
+					brews   : [],
+					visible : true
 				}
 			],
 			navItems : <></>
 		};
 	},
 	getInitialState : function() {
+		let brewCollection = [];
+
+		if(typeof window !== 'undefined') {
+			brewCollection = this.props.brewCollection.map((brewGroup)=>{
+				const localVisibility = localStorage.getItem(`${USERPAGE_KEY_PREFIX}-${brewGroup.class}`) ?? 'true';
+				if(brewGroup.visible != (localVisibility=='true')) {
+					brewGroup.visible = (localVisibility=='true');
+				};
+				return brewGroup;
+			});
+		}
+
 		return {
-			sortType     : 'alpha',
-			sortDir      : 'asc',
-			filterString : this.props.query?.filter || '',
-			query        : this.props.query
+			sortType       : 'alpha',
+			sortDir        : 'asc',
+			filterString   : this.props.query?.filter || '',
+			query          : this.props.query,
+			brewCollection : brewCollection
 		};
+	},
+	componentDidMount : function() {
+		// SAVE TO LOCAL STORAGE WHEN LEAVING PAGE
+		window.onbeforeunload = this.saveToLocalStorage;
+	},
+
+	componentWillUnmount : function() {
+		window.onbeforeunload = function(){};
+	},
+
+	saveToLocalStorage : function() {
+		this.state.brewCollection.map((brewGroup)=>{
+			localStorage.setItem(`${USERPAGE_KEY_PREFIX}-${brewGroup.class}`, `${brewGroup.visible}`);
+		});
 	},
 
 	renderBrews : function(brews){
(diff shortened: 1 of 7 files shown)


client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -6,29 +6,66 @@ const moment      = require('moment');
...
 			],
 			navItems : <></>
 		};
 	},
 	getInitialState : function() {
+		// HIDE ALL GROUPS UNTIL LOADED
+		const brewCollection = this.props.brewCollection.map((brewGroup)=>{
+			brewGroup.visible = false;
+			return brewGroup;
+		});
```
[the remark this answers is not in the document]
G-Ambatte: Agreed - I'll clean it up shortly.

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -6,29 +6,66 @@ const moment      = require('moment');
...
 
+	componentDidMount : function() {
+		// SAVE TO LOCAL STORAGE WHEN LEAVING PAGE
+		window.onbeforeunload = this.saveToLocalStorage;
+
+		// LOAD FROM LOCAL STORAGE
+		if(typeof window !== 'undefined') {
+			const brewCollection = this.props.brewCollection.map((brewGroup)=>{
+				const localVisibility = (localStorage.getItem(`${USERPAGE_KEY_PREFIX}-${brewGroup.class}`) ?? 'true')=='true';
+				brewGroup.visible = (brewGroup.visible != localVisibility ? localVisibility : brewGroup.visible);
```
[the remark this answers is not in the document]
G-Ambatte: Looking at this now, it's "if brewGroup.visible doesn't equal localVisibility, set it to localVisibility, but if it does then leave it alone" which always resolves to "set brewGroup.visible to localVisibility". I'll clean it up shortly.

G-Ambatte:
I thought I was committing only one set of changes, however, it was actually both. The latest commit resolves both the unnecessary `visible` prop and the unnecessary logic around setting `brewGroup.visible` in componentDidMount.

client/homebrew/pages/basePages/listPage/listPage.less
```diff
@@ -74,4 +74,27 @@
...
+		&.inactive {
+			color: #707070;
+		}
+		&.inactive::before {
+			content: '\f105';
+			font-family: 'Font Awesome 5 Free';
+			font-weight: 900;
+			font-size: 0.6cm;
+    		padding-right: 0.5em;
+		}
```
calculuschild, line 98: Possible opportunity to reduce duplicated CSS?

```suggestion
		&.active {
			color: #58180D;
		}
		&.inactive {
			color: #707070;
			
		}
		&.active::before, &.inactive::before {
			font-family: 'Font Awesome 5 Free';
			font-weight: 900;
			font-size: 0.6cm;
    			padding-right: 0.5em;
		}
		&.active::before {
			content: '\f107';
		}
		&.inactive::before {
			content: '\f105';
		}
```

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -31,9 +31,10 @@ const UserPage = createClass({
 
 		const brewCollection = [
 			{
-				title : `${usernameWithS} published brews`,
-				class : 'published',
-				brews : brews.published
+				title   : `${usernameWithS} published brews`,
+				class   : 'published',
+				brews   : brews.published,
+				visible : true
```
calculuschild, line 37: Can we remove this `visible : true` as well?

jeddai:
Ditto on the other comments, otherwise I tested this and it works great. Do we have any concerns about someone logging out and back in as a different user but seeing the user page collapse status of the previous user?

G-Ambatte:
> Ditto on the other comments, otherwise I tested this and it works great. Do we have any concerns about someone logging out and back in as a different user but seeing the user page collapse status of the previous user?

My first iteration appended the current username to the local storage configuration item, so you only got your own setting. However, I realized that leaking someone's username was far worse than leaking their brew visibility setting, so I removed that concatenation.

Could probably use a hashed version of the username instead, to achieve the same effect.

calculuschild:
Let's keep it as-is for now. We can consider this later on as we get more into personal user account settings.

client/homebrew/pages/userPage/userPage.jsx
G-Ambatte: Should be removed in the latest commit.

calculuschild:
Looks good! 💯 🎉 👉😎👉

PR #2318, opened 2022-09-01
[UserPage] Add tags to UserPage BrewItem

### LINKED ISSUE #2317: [UserPage] Add Brew tags to BrewItem
Now that brews are capable of being tagged, we should add the brew's tags to the User Page BrewItem.

G-Ambatte:
This PR resolves #2317.

This PR adds a new Tags line to the UserPage BrewItem, which only displays if the current Brew has tags.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
@@ -104,6 +104,13 @@ const BrewItem = createClass({
 			</div>
 			<hr />
 			<div className='info'>
+				{brew.tags ? <>
+					<span title={`Tags:\n${brew.tags.join('\n')}`}>
+						<i className='fas fa-tags'/> {brew.tags.join(', ')}
+					</span>
+					<br />
+				</> : <></>
+				}
 				<span title={`Authors:\n${brew.authors?.join('\n')}`}>
 					<i className='fas fa-user'/> {brew.authors?.join(', ')}
 				</span>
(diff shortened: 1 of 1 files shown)


client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
```diff
@@ -104,6 +104,13 @@ const BrewItem = createClass({
 			</div>
 			<hr />
 			<div className='info'>
+				{brew.tags ? <>
+					<span title={`Tags:\n${brew.tags.join('\n')}`}>
```
ericscheid, line 108: Could we also put  `className='tags'` onto this `<span/>` ?

G-Ambatte: Makes sense, probably `className='brewTags'` so it won't clash with `className='systemTags'`, which will be coming Soon™ in a different PR.

G-Ambatte: Should be in the latest commit now.

calculuschild:
I'm going to merge this as-is since it solves the issue, but I think it might be worth another PR to add a similar styling for each tag as it appears in the Metadata panel. Otherwise it can be hard to tell where one tag ends and another begins, especially with multi-word tags:

image

ericscheid:
Over on GM Binder I'm typically seeing just 1–4 tags per brew. Admittedly, their tags are constrained to a known set, users can't add their own tags willy-nilly.

Nonetheless, there always has to be that one guy...
image

The comma delimited list could at least be marked up with spans for each tag - that'd give use lots of flexibiity to rejigger things via css (including user side custom css. (The actual comma inserted with a `.brewTags span:not(:last-child):after {content: ','}`)

(I might even be tempted to argue the tags list should be semantically coded as an `<ul>` and `<li>` but it's late)

ericscheid:
I would though do the styling less obnoxiously, like this, maybe even toned down more

image

(and maybe with the tag-icon in front)

PR #2320, opened 2022-09-01
[UserPage] Add tags to filter

### LINKED ISSUE #2319: [UserPage] Add Brew tags to UserPage filter
Now that Brews are capable of being tagged, we should expand the UserPage functionality to include the tag contents in the filter search.

G-Ambatte:
This PR resolves #2319.

This PR expands the brew filtering to include tags, while leaving open the ability to extend the filtering functionality further in the future.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -141,9 +141,18 @@ const ListPage = createClass({
 
 	getSortedBrews : function(brews){
 		const testString = _.deburr(this.state.filterString).toLowerCase();
+
+		const checkString = (stringToTest)=>{
+			return (_.deburr(stringToTest).toLowerCase().includes(testString));
+		};
+
 		brews = _.filter(brews, (brew)=>{
-			return (_.deburr(brew.title).toLowerCase().includes(testString)) ||
-			(_.deburr(brew.description).toLowerCase().includes(testString));
+			const brewStrings = []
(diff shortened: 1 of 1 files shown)


client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -183,9 +183,18 @@ const ListPage = createClass({
 
 	getSortedBrews : function(brews){
 		const testString = _.deburr(this.state.filterString).toLowerCase();
+
```
calculuschild, line 186: This works, but the triple-nesting of functions seems more complicated than necessary.

How about this:

```
	getSortedBrews : function(brews){
		const testString = _.deburr(this.state.filterString).toLowerCase();

		brews = _.filter(brews, (brew)=>{
			const brewStrings = _.deburr([
				brew.title,
				brew.description,
				brew.tags].join('\n')
				.toLowerCase());

			return brewStrings.includes(testString);
		});
		return _.orderBy(brews, (brew)=>{ return this.sortBrewOrder(brew); }, this.state.sortDir);
	}
```

calculuschild (changes requested):

client/homebrew/pages/basePages/listPage/listPage.jsx
G-Ambatte: Looks good to me! I've pushed that through in the latest commit.

calculuschild (approved):

PR #2330, opened 2022-09-03
[UserPage] Add system tags to brew item

### LINKED ISSUE #1842: show brew system tags on user page brew items
From #240

> I know there has been talk of tagging/grouping, but in the meantime it would seem that the selected game system for a brew could be represented on the .brewItems. I envision either small color coded labels below the current .info div listing the system and other tags, or using the game system to determine the entirety of each .brewItem card's background color.

- [ ] output an extra line in the `.brewItem div.info` to show the system tags (if any)
  - `<span title="System: 5e">5e</span>` or similar
- [ ] provide css styling as desired.

G-Ambatte:
This PR resolves #1842.

This PR adds the systems array to the information retrieved for each brew on the UserPage, and adds styling to display it.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
@@ -104,7 +104,7 @@ const BrewItem = createClass({
 			</div>
 			<hr />
 			<div className='info'>
-				{brew.tags ? <>
+				{([].concat(brew.tags)[0]!='') ? <>
 					<span className='brewTags' title={`Tags:\n${brew.tags.join('\n')}`}>
 						<i className='fas fa-tags'/> {brew.tags.join(', ')}
 					</span>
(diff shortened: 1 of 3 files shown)


G-Ambatte:
Currently, the PR uses the D20 icon for the Systems tags:

image

However, a D6 is also available in the FontAwesome icons:

image

Any other suggestions for FA icons to use would be welcome.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
```diff
@@ -104,7 +104,7 @@ const BrewItem = createClass({
 			</div>
 			<hr />
 			<div className='info'>
-				{brew.tags ? <>
+				{([].concat(brew.tags)[0]!='') ? <>
```
calculuschild, line 107: @G-Ambatte Can you remind me what this is doing?

calculuschild:
This PR is fine, but I want to have a bit more discussion on handling `system` tags before merging this. I'll add some comments to #1842

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
G-Ambatte: As `brew.tags` has previously existed as an empty string, when it is converted to an array, `tags[0]` is `''`. When tags are actually set, `tags[0]` contains actual content. So we can't just check for the existence of tags with `tags.length > 0`, we need to check that `tags[0]` is not equal to an empty string - but still taking into account that `tags` might still be a string, having not yet been converted to an array yet.

calculuschild:
Alright, I am going to close this for now. Let's plan to eventually just fold Systems into the Tags.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
calculuschild: Ok. Can you add this fix to a new PR? I'm going to close this one because I think we are going to remove Systems anyway and merge into Tags, but this line seems like an important fix on its own.

PR #2332, opened 2022-09-03
[UserPage] Show Brew's Meta thumbnail on UserPage

### LINKED ISSUE #2331: [UserPage] Display thumbnail when set
When a Brew has a thumbnail set in the metadata, it should be displayed on the UserPage in that Brew's BrewItem.

G-Ambatte:
This PR resolves #2331.

This PR adds the thumbnail metadata to the information passed to the UserPage, and adds new elements to the BrewItem when that information is present. 

Currently the result looks like this:

image


client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
```diff
@@ -98,6 +98,12 @@ const BrewItem = createClass({
 		const dateFormatString = 'YYYY-MM-DD HH:mm:ss';
 
 		return <div className='brewItem'>
+			{brew.thumbnail ?
+				<div className='thumbnail'>
+					<img className='thumbnailImage' src={brew.thumbnail} ></img>
+				</div> :
+				<></>
```
calculuschild, line 105: Do we need this empty fragment?

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
```diff
@@ -19,6 +19,20 @@
...
+		height: 100px;
+		border: 2px solid #58180D;
+		border-radius: 8px;
+		float: left;
+		margin-right: 5px;
+		display: flex;
+		img.thumbnailImage {
+			max-width: 100%;
+			max-height: 100%;
+			margin: auto;
```
calculuschild, line 33: Similar behavior but grows the image to fill the whole box and crops the excess instead of shrinking to fit. Might look a little cleaner instead of the empty space inside the box.

```suggestion
		img.thumbnailImage {
			width : 100%;
			height : 100%;
			object-fit : cover;
			border-radius : inherit;
```

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
```diff
@@ -19,6 +19,21 @@
 	break-inside                : avoid;
 	box-shadow                  : 0px 4px 5px 0px #333;
 	background-color            : #cab2802e;
+	.thumbnail  {
+		width: 80px;
+		height: 100px;
```
calculuschild, line 24: What do you think about making this a square? 80px x 80px?  My guess the average image is closer to a square, but I actually have no idea.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
```diff
@@ -19,6 +19,21 @@
...
+		border: 2px solid #58180D;
+		border-radius: 8px;
+		float: left;
+		margin-right: 5px;
+		display: flex;
+		img.thumbnailImage {
```
calculuschild, line 30: There is some slight positioning weirdness where the image is not quite filling the whole box. I'm not sure why, but there must be some tweak to make that fit perfectly. 
image

ericscheid: This can be fixed by removing the `border` and `border-radius` from the container div, and applying instead directly on the image (assuming `object-fit: cover;` is also being used)

Gazook89:
In my opinion, image should be on the right side of the brew item. Otherwise the titles are not aligned between items with images and not

Gazook89:
Here is a quick mockup of an alternate idea....again, no local install so just did this in devtools for now:

image

```css
element {
	background-image: url(https://i.imgur.com/YPV3Txc.jpg);
	position: absolute;
	width: 150px;
	height: 100%;
	top: 0;
	right: 0;
	z-index: -1;
	background-size: cover;
	mask-image: linear-gradient(80deg, #0000 20%, #050 40%);
	opacity: 50%;
}
```

Will it work for 100% of images?  probably not...the efficacy will depend largely on the composition of the image used.

calculuschild:
Not a bad idea. How does it look with a light image?

Gazook89:
> @Gazook89 not a bad idea. How does it look with a light image?

Here is a pretty light image, also demonstrates the use of an image with a white background:
image

I think it looks *good enough*.  One issue is that the fade removes the head of the creature.  This is the danger of any clipping, though-- one must take account of the image's composition.  In my mind this is an *annoyance* but one that is easily overlooked for now.  Some day I think the entire user page could be reworked to something besides a two column scroll-a-thon and when that happens the images could be incorporated better-- but that is a different Issue.

Perhaps a default image of the HB logo could be used, as well, if no thumbnail is present...or maybe that is unwanted clutter.

---
Here is the html node in case anyone wants to quickly copy/paste into devtools again:
```html
<div style="background-image: url(https://www.aidedd.org/dnd/images/bone-devil.jpg);position: absolute;width: 150px;height: 100%;top: 0;right: 0;z-index: -1;background-size: cover;mask-image: linear-gradient(80deg, #0000 20%, #050 40%);opacity: 50%;"></div>
```

calculuschild:
@G-Ambatte What do you think? Do you want to try applying @Gazook89 's style and move the image to the right?

G-Ambatte:
> @G-Ambatte What do you think? Do you want to try applying @Gazook89 's style and move the image to the right?

I'm more than happy to implement the design; I'll try to push through an update to the PR later tonight.

G-Ambatte:
Here's a screengrab of the updated style:

image

G-Ambatte:
Some examples with no description text:

image

calculuschild:
This looks good to me. My only remaining thought is perhaps we should set a max-height on the image so it is guaranteed to fit into the smallest `brewItem` height.

Gazook89:
> max-height

If the max height is set to the smallest brewItem height, say 200px or whatever it is, then if the brewItem height is larger it is likely that there will be an unflattering straight edge at the bottom of the image, since it won't expand to fill the brewItem.

`background-size` could be set to `contain` instead?

G-Ambatte:
Switched to `contain`, added `no-repeat` and positioned to `top right`. Those same four brews:

image

image

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
calculuschild: I think you can do something like 

```
{brew.thumbnail && <div ....
```

So it draws the div if true, and just doesn't do anything if false.

calculuschild:
Noice. 🎉 💯 🥇

Issue #2325, opened 2022-09-03
Stubs that point to externally-deleted Google Brews cannot be deleted


calculuschild:
1) Make a Google brew (a stub is created).
2) Delete the brew directly from your google drive (The stub remains).
3) Open user page, find stub
4) Try to open, will get a "brew not found" error from Google.
5) Try to delete, nothing happens.

##### Main Task:
- [ ] Allow deleting stubs that point to a missing Google Brew. Currently tries to `get` Google Brew, fails with a 404, then gives up.

##### Possible Added Tasks:
- [ ] When visiting user page, auto-delete any brewless stubs.
  - We already get both stubs from HB and brews from Google at the same time so it wouldn't be hard.
  -  **Possibly dangerous... might be a way to accidentally delete stubs for brews that are temporarily out of access (expired Google Credentials.... high traffic?) The Google brew itself would be safe, just un-stubbed.**
  - May be possible to be smart about this and only delete stubs when we know for sure we got a good response from Google (received a list of at least one Google Brew, etc.)

calculuschild:
Fixed by #2345

PR #2345, opened 2022-09-07
update deletion to delete brews when the google brew cannot be found

jeddai:
#2325

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -267,11 +267,12 @@ const deleteGoogleBrew = async (account, id, editId, res)=>{
 };
 
 const deleteBrew = async (req, res, next)=>{
+	// Delete an orphaned stub if its Google brew doesn't exist
 	try {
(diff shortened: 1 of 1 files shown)


server/homebrew.api.js
```diff
@@ -259,7 +266,16 @@ const deleteGoogleBrew = async (account, id, editId, res)=>{
...
+		req.brew = await getBrew('edit')(req, res, next);
+	} catch (e) {
+		const { id, googleId } = getId(req);
+		console.warn(`No google brew found for id ${googleId}, the stub will be deleted.`);
+		await HomebrewModel.deleteOne({ editId: id });
+		return next();
+	}
+
 	let brew = req.brew;
 	const { googleId, editId } = brew;
```
calculuschild, line 280: @jeddai I'm getting an error that it can't destructure `brew` because it's undefined. It won't let me delete *any* brew now. Can you investigate?

calculuschild:
Thanks @jeddai !

PR #2367, opened 2022-09-15
Add different Issue forms

Gazook89:
Probably jumping the gun here, but here is my take on utilizing Github Issue Forms.   Issue Forms are detailed [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser), and here is a generic example:

image

This is all meant to address the large number of "saving" issues that get reported.  Obviously it's best to just not have those issues, but when they do come up they seem to come in droves.  Also, they are typically reported by first-time github users with little additional information or follow-up.  This form automatically applies a "saving" label to the issue, which could be used in further automation with Github Actions (such as auto-closing with comment after a certain number of days of no activity).

But further, this PR includes the Saving Issue Form, Generic Issue Form, and a Feature Request Form.  Additionally, when creating a new Issue there will be a step that asks which type of Issue you need to create **AND** it will offer links to the subreddit and Discord of Many Things.  This may help direct users to better places for simple questions/answers.   Finally, as currently set in this PR, a normal Issue can be opened as well (with the current markdown template).

One thing I didn't include but might be helpful are dropdown menus (multiple selections are possible) that let a user indicate which page the issue occurred on, or what *category* of feature request they have.

Another thing we can do is add another option in the Issue Form chooser  (config.yml) that opens a reddit DM to calculuschild if seeking password help.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -5,7 +5,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        We'd love to hear your idea!  Please be sure to search the current Issues for any duplicate requests.
+        "We'd love to hear your idea!  Please be sure to [search the current Issues](https://github.com/naturalcrit/homebrewery/issues) for any duplicate requests."
   - type: textarea
     id: user-request
     attributes:
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,20 @@
+name: Feature Request
+description: Have an idea to improve the Homebrewery? Let us know!
+labels: ["feature request"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        We'd love to hear your idea!  Please be sure to search the current Issues for any duplicate requests.
+  - type: textarea
+    id: user-request
+    attributes:
+      label: "Your idea:"
+      description: The best feature requests provide an explanation of the current issue and then an explanation of how it could be improved.  Screenshots/images can be pasted right in as well!
+    validations:
+      required: true
+  - type: checkboxes
+    id: terms
+    options:
+      - label: I have searched the Issues tracker for any duplicate requests and found none.
+        required: true
--- a/.github/ISSUE_TEMPLATE/general_issue.yml
+++ b/.github/ISSUE_TEMPLATE/general_issue.yml
@@ -3,8 +3,7 @@ description: Report an issue unrelated to Saving
 body:
   - type: markdown
     attributes:
-      value: |
-        Please include as much information as possible.
+      value: Please include as much information as possible.
   - type: checkboxes
     id: renderer
     attributes:
@@ -16,17 +15,30 @@ body:
       validations:
         required: true
   - type: dropdown
-  id: operating-system
-  attributes:
-    label: Operating System
-    description: Which OS were you using when the issue occurred?
-    options:
-      - label: Windows
-      - label: MacOS
-      - label: Linux
-      - label: other
-    validations:
-      required: true
+    id: browser
+    attributes:
+      label: Browser
+      description: Which browser were you using when the issue occurred?
+      options:
+        - label: Chrome
+        - label: Firefox
+        - label: Edge
+        - label: Safari
+        - label: other
+      validations:
+        required: true
+  - type: dropdown
+    id: operating-system
+    attributes:
+      label: Operating System
+      description: Which OS were you using when the issue occurred?
+      options:
+        - label: Windows
+        - label: MacOS
+        - label: Linux
+        - label: other
+      validations:
+        required: true
   - type: textarea
     id: user-description
     attributes:
--- a/.github/ISSUE_TEMPLATE/general_issue.yml
+++ b/.github/ISSUE_TEMPLATE/general_issue.yml
@@ -1,0 +1,42 @@
+name: General Issue
+description: Report an issue unrelated to Saving
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please include as much information as possible.
+  - type: checkboxes
+    id: renderer
+    attributes:
+      label: Renderer
+      description: Which renderer does this issue occur on?  If you are unsure, you can check the renderer in the Metadata Editor.
+      options:
+        - label: Legacy
+        - label: v3
+      validations:
+        required: true
+  - type: dropdown
+  id: operating-system
+  attributes:
+    label: Operating System
+    description: Which OS were you using when the issue occurred?
+    options:
+      - label: Windows
+      - label: MacOS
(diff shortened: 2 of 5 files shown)


.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,20 @@
+name: Feature Request
+description: Have an idea to improve the Homebrewery? Let us know!
+labels: ["feature request"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        We'd love to hear your idea!  Please be sure to search the current Issues for any duplicate requests.
```
calculuschild, line 8: Is it possible to make `current Issues` a link to the issues page?

.github/ISSUE_TEMPLATE/general_issue.yml
```diff
@@ -0,0 +1,42 @@
...
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Please include as much information as possible.
+  - type: checkboxes
+    id: renderer
+    attributes:
+      label: Renderer
+      description: Which renderer does this issue occur on?  If you are unsure, you can check the renderer in the Metadata Editor.
```
calculuschild, line 12: Can we add something like "(click the 🛈 button at the top of the text editor)"? Lots of users have no idea there is a metadata panel.

Also, the button popup calls it "Properties", so we should call it that here.

.github/ISSUE_TEMPLATE/save_issue.yml
```diff
@@ -0,0 +1,22 @@
+name: Saving Issue
```
calculuschild, line 1: Can we change the link in the error popup to point directly to this issue template somehow? Currently it automatically pastes the error into the issue body, but it likely needs tweaking to fill the appropriate text field in this new format.

https://github.com/naturalcrit/homebrewery/blob/d20bb15204f563e850605905a0d6cfe251a43d04/client/homebrew/pages/editPage/editPage.jsx#L325

.github/ISSUE_TEMPLATE/feature_request.yml
Gazook89: I think so...it seems like regular markdown should work, so i've done that.  I'm not sure how to test this PR without it being on the live repo, so might need to push it live and see what happens....

.github/ISSUE_TEMPLATE/general_issue.yml
Gazook89: done

.github/ISSUE_TEMPLATE/save_issue.yml
```diff
@@ -0,0 +1,22 @@
...
+title: "Save Error: "
+labels: ["saving"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Woops, sorry there was an issue Saving.  Here are some possible quick solutions to try:
+        1. Refreshing your Google credentials in Homebrewery by signing out, and back in (they expire after one year).
+        2. Waiting a few minutes and trying again
+        3. Checking the Issues in Github or the /r/homebrewery subreddit for answers to similar issues
```
calculuschild, line 12: If the user has gotten to this page due to the error popup, option 1 will not be the case (we have a specific error popup that already tells them this.)

Option 2 may temporarily avoid the issue, but we still want them to report it, since it's usually a sign we need to fix something (something crashed the server or we are mishandling their data somehow).

Option 3 is similar; if they have gotten here through the error popup, it's because it's a problem we haven't solved yet and reddit may be unlikely to help, so we still want to make sure they report it.

I'm not sure the best way to phrase this, but we don't want to discourage users from reporting issues if they are arriving from that error message.

.github/ISSUE_TEMPLATE/save_issue.yml
Gazook89: I can look into it.   But, with this PR there are two types of issue templates, so it'd have to determine which type of issue is being reported.  That might not be *that* difficult, but it's more of a *thing*.

calculuschild: If it's coming from that line in the code we know it's a saving issue, if that helps.

.github/ISSUE_TEMPLATE/save_issue.yml
Gazook89: Perhaps the whole list can be moved to the bottom of the form, as a sort of "thanks for reporting, take a look at these further steps in the meantime"

.github/ISSUE_TEMPLATE/save_issue.yml
Gazook89: ah okay, I didn't look at the greater context of that line.  Yes, that would make it much easier.  Will look at tomorrow*

\**not guaranteed to be tomorrow*

calculuschild: This comment shows how you can format a link to auto-fill parts of the issue form: https://github.com/community/community/discussions/15477#discussioncomment-2793106

.github/ISSUE_TEMPLATE/save_issue.yml
calculuschild: Yeah, I think that makes sense.

.github/ISSUE_TEMPLATE/feature_request.yml
calculuschild: Right, I'm not seeing any way to preview what these will look like either. Even in their demo video they're like "so lets commit this now so we can see what it looks like." It does seem to use markdown though, so you're right I think this should work...

.github/ISSUE_TEMPLATE/save_issue.yml
Gazook89: https://stackoverflow.com/questions/34146618/pre-populate-the-github-new-issue-form-using-the-querystring

Gazook89:
This might be getting out of scope of the PR, but just a thought:

Could the Report Bug button in HB somehow offer the choice of reporting in Github or Reddit?  And then pre-populate desired fields using info from props?

- In Github, it would automatically open the General Issue form and grab OS, Browser, and Renderer to fill those fields.
- In Reddit, it could populate those same categories just in the post.

I just don't know if for the github side of things if we want to choose 'General Issue' form automatically or leave it to the user to pick from the template chooser.

calculuschild:
> Could the Report Bug button in HB somehow offer the choice of reporting in Github or Reddit? And then pre-populate desired fields using info from props?

Perhaps. Maybe we just need another button so "report issue" goes to github and "ask the community" goes to reddit or something. But this can be a separate PR.

> I just don't know if for the github side of things if we want to choose 'General Issue' form automatically or leave it to the user to pick from the template chooser.

calculuschild:
Ok, I think this is good enough to merge. Let's see what it looks like!

PR #2372, opened 2022-09-15
Reposition Thumbnail Preview in Metadata Editor

Gazook89:
This moves the thumbnail preview to be alongside both the Description and Thumbnail URL inputs.  Currently, alongside the tags, it can obscure the tag badges if enough of them are added.

In this PR the thumbnail also shifts down below the URL input if the width is reduced to a certain point.

If this PR sits for too long unmerged, I may also take another stab at making the labels shift above the fields at a certain breakpoint.

gif

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.less
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.less
@@ -80,12 +80,11 @@
 		justify-self: center;
 		width: 80px;
 		height: min-content;
-		border: 2px solid white;
 		flex: 1 1;
 		max-height: 115px;
 		aspect-ratio: 1 / 1;
 		object-fit: contain;
-		background-color: cadetblue;
+		background-color: #AAA;
 	}
 
 	.systems.field .value{
--- a/client/homebrew/editor/metadataEditor/metadataEditor.less
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.less
@@ -8,6 +8,7 @@
 	padding          : 25px;
 	background-color : #999;
 	height           : calc(100vh - 54px); // 54px is the height of the navbar + snippet bar.  probably a better way to dynamic get this.
+	overflow-y       : auto;
 
 	& > div {
 		margin-bottom: 10px;
--- a/client/homebrew/editor/metadataEditor/metadataEditor.less
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.less
@@ -9,19 +9,28 @@
 	background-color : #999;
 	height           : calc(100vh - 54px); // 54px is the height of the navbar + snippet bar.  probably a better way to dynamic get this.
 
-	display: flex;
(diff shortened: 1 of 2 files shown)


calculuschild:
Is this ready for review?

client/homebrew/editor/metadataEditor/metadataEditor.less
```diff
@@ -43,22 +61,33 @@
...
+		position: relative;
+		justify-self: center;
+		width: 80px;
+		height: min-content;
+		border: 2px solid white;
+		flex: 1 1;
+		max-height: 115px;
+		aspect-ratio: 1 / 1;
+		object-fit: contain;
+		background-color: cadetblue;
```
calculuschild, line 88: Something weird happens with this background color when the panel is very narrow. This container stretches out on the left and right and looks a bit odd. Any way we can just maintain the aspect ratio?

image

Gazook89: oh i actually kinda like it.   I don't like to mix left-aligned content (the labels and input fields) with centered content (the thumbnail).    It seems odd and empty.

This is entirely an aesthetic opinion, so I will go whichever way the wind blows.

calculuschild: @jeddai @G-Ambatte @ericscheid opinions?

jeddai: I think I'd actually argue for a middle ground of keep it left-aligned, but the alignment should be off of the leftmost position of the field. So the second part of the image where we see the left hand of the thumbnail lines up with the left hand of the input above it.

Gazook89: To what Jeddai says--- I think that might be tough because the thumbnail preview is on the same 'row' as the description and URL fields and just flexes to that second row.  I'm not sure it'd be possible to have both:

- right-situated of the description/URL fields at wider screen size, and
- a blank space to the left when flexed to the bottom.

I could change the color of that background square or reduce it's saturation to make it less 'eye-popping'.  Though I like the bright splashes of color that are present elsewhere in the UI.  Additionally maybe removing the white border helps?

This PR is not terribly important so it can simmer for a bit until more opinions come in.

calculuschild: @Gazook89  I have two main concerns:

- The white border seems weird where it encompasses all this blank space. It's especially weird when the image is not square, so the border doesn't quite touch the image. Maybe removing the border would help as you say.

image

- The turquoise color clashes very badly with some other colors a user might choose:

image

Maybe just another neutral gray?

image

I don't know. Everything else about this PR is fine; just something about the styling here is not quite jiving with me.

calculuschild:
Yeah... this is acceptable. I'll merge it. :100:

Issue #1870, opened 2021-12-03
Snippet Options Interface


jeddai:
## Terminology:
| Term | Definition |
|:---|:---|
| [gutter](https://codemirror.net/demo/marker.html) | a CodeMirror concept -- examples would be the line numbers gutter or code folding gutter |
| [line widget](https://codemirror.net/demo/widget.html) | a CodeMirror concept that is a widget either above or below a line of code in the CodeMirror instance |
| snippet | HB's integrated snippets, such as images, class tables, monsters, etc... |
| widget | a small utility that performs a function |

## Description

While discussing things with the DOMT team, one of the biggest topics that came up was snippet options. The overall concept stemmed from there being a solid number of users who are very unfamiliar with HTML/CSS, and having a UI to edit snippets would be really beneficial to them.

The DOMT team was initially thinking something akin to GMB's snippet modal, but this would allow for after-insertion editing as well as initial editing (the editor could be set up to open itself when a snippet is inserted).

### Thoughts
Some ideas for what this could look like:
* A gutter or line widget that when clicked will create a line widget that will allow for UI-based editing of the snippet
* Examples snippet options that could be made editable
  * LInk URLs/text
  * Inserting/removing frame/decoration/wide/other classes for known div types (monster, classTable, artist, note, etc...)
  * Values for vertical and horizontal spacing
  * Image height, width, position, URL, and alt text
  * Watercolor image selector and color picker
* In chatting with @calculuschild, we realized it could be implemented in such a way that different document themes _(tba)_ have different widgets available to them, and users could create their own widgets. A configurable approach would be ideal here, despite that being a bit more complicated from a programming perspective.

Working proof of concept: https://github.com/jeddai/homebrewery/pull/1
Videos: [monster/table frame](https://cdn.discordapp.com/attachments/910644083737894933/915254945245200444/Screen_Recording_2021-11-30_at_8.54.41_AM.mov), [artist credit position](https://cdn.discordapp.com/attachments/910644083737894933/915442666475515935/Screen_Recording_2021-11-30_at_9.20.37_PM.mov)

### Important Notes
[...]

PR #2373, opened 2022-09-16
Watercolor image mask

calculuschild:
Thanks to /u/FlamableConcrete for creating and providing the image masks found at https://watercolors.giantsoup.com/. They have graciously donated their entire library of masks for free use on the Homebrewery (partly because the server costs for hosting those images is becoming expensive).

- [x] Get masking to work
- [x] Allow arbitrary rotation and positioning
- [ ] Mask shapes (96ppi):
  - [x] Straight Edges (x8 so far)
  - [x] Corners (internal) - may need to be manually extended to not cut off when rotated
  - [ ] Corners (external) - may need to be manually extended to not cut off when rotated
  - [ ] Holes
  - [ ] Stripes

This PR adds a snippet that creates a "watercolor Image Mask" using curly braces:

```
{{imageMask3,--rotation:5,--offset:4cm,revealImage,
  ![](https://assets1.ignimgs.com/2019/05/29/dndmobile-br-1559158957902_160w.jpg?width=1280)
}}
```
image

The class `imageMask` with the accompanying number selects which mask to load. Class `top`, `left`, `right`, or `bottom` will rotate the mask and position the image accordingly. `revealImage` adds some transparency to the mask and draws a checkerboard pattern behind the user image as a tool to help with image positioning in case it is hard to see behind the mask.

image

The image can be positioned around the page and sized as expected using curly injectors.

___

### Question 1:

 We have LOTS of masks available (like 100+), because /flamableconcrete made unique ones for each direction.
image

Given that we can just rotate them around with CSS, I am just merging all of the separate sets into one. However, they are all at different heights and with slightly different skews without any real pattern. **Should we:**

- Adjust every mask to be approximately the same height, giving the user full customizability to choose any mask, and then positioning the mask up or down with CSS as needed to fit their image? We can arrange it so that `left:50%` aligns the border of the mask right at the 50% point on the page. A widget interface might make this even easier https://github.com/naturalcrit/homebrewery/issues/1870, https://cdn.discordapp.com/attachments/910644083737894933/915254945245200444/Screen_Recording_2021-11-30_at_8.54.41_AM.mov.
- Leave the masks as-is, but sort them by number from short to tall? This would reduce the amount of user steps, but limit the shapes available at a given height. We have a lot so it might not be a big deal.

If we go with the first option, perhaps we sort them by number anyway based on the "tilt" of the mask, from left-leaning to right-leaning, although they don't all fit that categorization either.... Maybe random is best and we eventually add more of a GUI to select the mask you want.

### Question 2:

I have currently only implemented a few "edge" masks. What syntax should we use to handle the other two types: corners, and "center" cutouts? `imageMaskCorner`/`imageMaskEdge`? `top-left`, `top,left` ? 

### Question 3:

These images are all about 100kb. I'm going to try to convert them all to webp (~40% smaller than PNG so far) to save some space, but wonder how much compression/resolution we can accept? The PHB background is already a quite blurry JPG, but that doesn't matter so much because it's mostly one color in the background. In a similar way, all the high-detail areas on these masks are going to be blended with the more detailed user image anyway, so we might get away reducing the resolution to... 72ppi or lower with some amount of lossy compression. The user image should still look crisp underneath..
___

Obviously any other feedback or opinions are welcomed.


Gazook89:
> Question 1

I know flamable had lots of different "themes" in his collection since they weren't being used as masks originally, but just overlayed images for each different D&D book theme.  Since they are being used as masks now and will work with any page background, the collection we need should be significantly smaller, right?  

And I suggest the first option for this question.  All masks should be normalized in such a way that the mask is larger than a page size (and keep in mind that we should think of the largest page dimensions such as for A4).  Any single mask should be able to mask the *entire* page by itself, without even the water color edge showing.  Then, the mask can be shifted left or right to slowly reveal the watercolor edge.  

> This would reduce the amount of user steps,

Option two likely wouldn't actually reduce the number of steps, because users would have to flip through a bunch of options to find something that will work.  


---

> Question 2

Not sure on this.  Some things like corners can just be a single corner, but offer UI to rotate it (a slider) so it can fit into any corner.  Same with the normal "single edge" masks....one set of masks, and a rotation slider so it can be top-to-bottom, bottom-to-top, left-to-right, right-to-left, and any angle in between.

---

> Question 3

I haven't used flamable's masks in a long time but i seem to recall them already being a bit pixely--- but it really has been a long time and could be wrong.  When I was working on my side project [framer](https://gazook89.github.io/framer/) (it never got finished and please don't look under the hood) I created a bunch of watercolor masks myself since I couldn't use flamable's due to cross-origin issues on FF.

---

I think ultimately it would be good to curate a new list of snippets that is pre-formed page layouts.  Choose "text left, image right" from a dropdown and it spits out an image mask and single column page.   This snippet menu would include the existing Cover Page, ToC, and a new "credits" page as well as "full page image", "full spread image", etc.

calculuschild:
> Since they are being used as masks now and will work with any page background, the collection we need should be significantly smaller, right?

Even just using the masks, we have about 30 for each "edge". After rotating them all to the same orientation, that gives us like 114 total "edge" masks.

One vote for normalizing the images. ✅ Yep you describe exactly what I had in mind.

> ...a rotation slider so it can be... any angle in between.

Fortunately the current implementation uses a variable `--rotation` that the user can access directly if they want. I would probably wait to reveal this until we have a slider because I forsee some positioning weirdness we might want to iron out first.  For example, the `top, left, right, bottom` classes can be made to work with the images "normalized in such a way that the mask is larger than a page size" pretty easily via appropriate margins to shift the watercolor edge starts at position "0" . Arbitrary rotation of such an image that is larger than the actual page might give unexpected results based on where the mask center is.... I need to look into it.

>  ...i seem to recall them already being a bit pixely

They are a bit, at 96 ppi. However, to me this is mostly obvious only on the parchment texture; the transparency of the cutout borders makes it less apparent in those regions. If you have time to test this PR out I'd be curious to see if it seems as pixely as you remember.

> I think ultimately it would be good to curate a new list of snippets that is pre-formed page layouts. 

I agree, but I think that is a separate project from this PR. We can have a whole suite of "pre-made pages" as you say that just spit out a big group of snippets in a specific layout.

calculuschild:
@Gazook89 I've updated this one to now support arbitrary rotation and up/down motion. Use `--rotation:X` to set any angle in degrees, and `--offset:Xcm` to shift the pattern up/down from center. This currently only has 8 flat-edge images that I made into a seamless repeating pattern so they can take advantage of `mask-repeat` and don't get so pixelated but keep a smaller file size.

Still thinking about how to format corner images since they won't extend naturally with mask-repeat....

calculuschild:
@Gazook89 I added a snippet for some corner images. They don't work well when arbitrarily rotated, but can be shifted up and down, and using `top,left`, etc. will flip them in X or Y as needed to select any of the 4 corners.

themes/V3/Blank/style.less
```diff
@@ -167,6 +167,103 @@ body {
...
 	.watercolor12 { --wc : @watercolor12; }
 
+	/* Image Masks */
+	[class*="imageMask"] {
+		position              : absolute;
+		height                : 200%;
+		width                 : 200%;
+		left                  : 50%;
+		bottom                : 50%;
+		--rotation            : 0;
```
jeddai, line 177: When using `--rotation` with non-cardinal degrees I'm getting an interesting visual issue. Is there a way to resolve this I've missed?

image

themes/V3/Blank/style.less
```diff
@@ -167,6 +167,103 @@ body {
...
+		--scaleX              : 1;
+		--scaleY              : 1;
+		-webkit-mask-image    : var(--wc), var(--revealer);
+		-webkit-mask-repeat   : repeat-x;
+		-webkit-mask-size     : 50%; //Scale only X to fit page width, leave height at aspect ratio, designed to hang off the edge
+		-webkit-mask-position : 50% calc(50% - var(--offset));
+		mask-image            : var(--wc);
+		mask-repeat           : repeat-x;
+		mask-size             : 50%;
+		mask-position         : 50% calc(50% - var(--offset));
```
jeddai, line 189: I'd suggest adjusting offset to include an `--offset-x` and `--offset-y` so it can be moved either direction

jeddai (commented):
> I have currently only implemented a few "edge" masks. What syntax should we use to handle the other two types: corners, and "center" cutouts? imageMaskCorner/imageMaskEdge? top-left, top,left ?

I think it makes the most sense to do `imageMaskEdge` and have it use `top`, `left`, etc... but then have `imageMaskCorner` and it doesn't have top, left, or anything. I think it should be expected that users would set the rotation for corners.

Also I'm having a lot of trouble getting corner image masks to work at all... Not sure what's up, happy to post a pic if you'd like.

themes/V3/Blank/style.less
calculuschild: They should be tiled horizontally (designed to be seamless) so they extend infinitely to the sides. It seems this isn't kicking in anymore.... I'll investigate...

themes/V3/Blank/style.less
calculuschild: Following my above comment, this *should* be tiled infinitely in the X direction (if I can figure out why it broke), so offset-x wouldn't do much except throw off the center of rotation. Offset lets you move the edge toward or away from the center, and rotate... rotates. Between those two you can already position the edge anywhere on the page.

calculuschild:
> > I have currently only implemented a few "edge" masks. What syntax should we use to handle the other two types: corners, and "center" cutouts? imageMaskCorner/imageMaskEdge? top-left, top,left ?
> 
> I think it makes the most sense to do `imageMaskEdge` and have it use `top`, `left`, etc... but then have `imageMaskCorner` and it doesn't have top, left, or anything. I think it should be expected that users would set the rotation for corners.
> 
> Also I'm having a lot of trouble getting corner image masks to work at all... Not sure what's up, happy to post a pic if you'd like.

Yes please post a pic. It's possible Chrome has broken this since I last looked at it.

themes/V3/Blank/style.less
jeddai: Gotcha that addresses the concern then!

jeddai:
image

calculuschild:
> image

Is this still happening after I made my fix to the widths being overwritten?

jeddai:
Nope the corners function now!

Issue #2357, opened 2022-09-12
Clarify reason for invalid tag input


calculuschild:
Tags can become invalid due to a number of reasons: disallowed characters, exceeded character limit, duplicate of another tag, etc.

Currently, the tag simply turns red when this occurs. It would be handy to have a small popup label to indicate the specific issue that is causing an invalid tag.

Along with this, the URL input for thumbnails should have a similar popup for too-long URLs. See #2356

Gazook89:
I am tinkering with an approach that would utilize a single "error" area in the meta panel which would display thumbnail url errors and tagging errors (and any other areas).  80% of the reason is because I don't personally want to tackle positioning errors around tags.  But also I think it will appear less cluttered if there is an expected and consistent space for errors in that small panel.

This is a *thumbnail too long* error  (my length rule is currently just set to 5 characters for testing):

<img width="565" alt="image" src="https://user-images.githubusercontent.com/58999374/190540705-5df64283-5520-47c4-9ed4-2641d6cc3218.png">

Alternative:
<img width="565" alt="image" src="https://user-images.githubusercontent.com/58999374/190541075-8426b7b8-4656-4834-8a3e-d93a33825d29.png">

calculuschild:
> I am tinkering with an approach that would utilize a single "error" area

I like it. I think your alternative approach (overlapping might be best. Then we don't leave a wasted empty space at the top or have to worry about the UI shifting around to accommodate error messages of different lengths.

Gazook89:
If someone pastes an url that is too long into the box, do we want the text to still be entered but not used?   

Or prevent it from being in the input box entirely?   (this is my preference, because i already have it accomplished).

> alternative approach (overlapping might be best.

noted

calculuschild:
I'd allow it to be pasted in so they can see a glimpse of what they have copied. I can imagine users complaining "it gives me an error that the URL is too long, but the text box is empty!"

This also mimics the behavior of the tags. When an invalid input is entered the box just turns red but doesn't block the user from interacting with the page.

ericscheid:
In chrome, pasting a long url into an `<input maxlength="125">` will result in the too-long-url being truncated .. and since the permissable length is likely *much* longer than the input field would display this fact of truncation will not be obvious.

Same thing with Safari and Firefox. All three on MacOS.

dbolack-ab:
Has this been resolved with recent tag improvements?

5e-Cleric:
No it has not.

PR #2374, opened 2022-09-18
Metadata field validation

### LINKED ISSUE #2356: Add character limit to Thumbnail URL
We've already had some people try to store whole Base64 images in the URL which massively increases storage size and bottlenecks loading and saving of brews, potentially leading to full crashes of the server. The current limit is 124 but due to the max size of Google Drive file properties (the Homebrewery will crash to an error page), but this is pretty restrictive and occurs commonly in standard URLs. Hows about... 256 characters for starters?

Gazook89:
EDIT::: This is going to be superseded by https://github.com/Gazook89/homebrewery/tree/Field-Validation-System-2 which is just a fresh start using the information gained from the conversation here.

This is to fix #2356 by providing a validation error for too many characters in the thumbnail URL field.  It is setup to provide a regex-based method of setting field requirements that will generate an error bar above the editor.  Hopefully this means it can be used on any text fields with any regex.

Upon review by others, I'm bracing myself for someone to say "*oh, why didn't you just.....*" with a much simpler solution.

\*\*NOTE: with this draft, the PR has silly rules for easy testing:

1. Description field must have no more than 5 characters
2. URL field must have no more than 5 characters
3. URL field must start with "W".

image

image

Some issues:

- [ ] This prevents write to the `props.metadata` if errors exist  (even for valid fields)....this doesn't bother me much, let me know what you folks think.
- [ ] I have a max character count set to 5, using regex `/^.{0,5}$/` but it seems like it's allowing 6 characters to write to props.  I'm sure this is some **state** or React thing I don't understand?
- [ ] Should there be a way of minimizing these errors so they are not in the way?  I don't think dismissing them entirely is good, but a button to reduce in size, or after a short period of time?

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -52,7 +52,7 @@ const MetadataEditor = createClass({
 	},
 
 	handleFieldChange : function(name, inputRules, e){
-		console.log(this.state.errs)
+		// check if each field meets the requirements set for that field
 		for (const err of inputRules){
 			if(new RegExp(err.reg).test(e.target.value)){
 				this.setState((prevState)=>({
@@ -68,12 +68,15 @@ const MetadataEditor = createClass({
 			}
 		}
 
+		if(this.state.errs.length == 0){
+			this.props.onChange({
+				...this.props.metadata,
+				[name] : e.target.value
+			});
 
+		}
 
-		this.props.onChange({
-			...this.props.metadata,
-			[name] : e.target.value
-		});
+		console.log(this.props.metadata);
 	},
 
 	handleSystem : function(system, e){
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -244,7 +244,7 @@ const MetadataEditor = createClass({
 	render : function(){
 		return <div className='metadataEditor'>
 
-			{this.renderError()}
+			{this.renderValidationNotice()}
 			<div className='field title'>
 				<label>title</label>
 				<input type='text' className='value'
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -242,7 +242,15 @@ const MetadataEditor = createClass({
 					value={this.props.metadata.thumbnail}
 					placeholder='my.thumbnail.url'
 					className='value'
-					onChange={(e)=>this.handleFieldChange('thumbnail', e)} />
+					onChange={
+						(e)=>{
+							if(e.target.value.length > 5){
+								console.log('too long');
+							} else {
+								this.handleFieldChange('thumbnail', e)
+							}
+						}
+					} />
 				<button className='display' onClick={this.toggleThumbnailDisplay}>
 					<i className={`fas fa-caret-${this.state.showThumbnail ? 'right' : 'left'}`} />
 				</button>
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -36,7 +36,7 @@ const MetadataEditor = createClass({
 	getInitialState : function(){
 		return {
 			showThumbnail : true,
-			errs: []
+			errs : []
 		};
 	},
 
@@ -51,12 +51,31 @@ const MetadataEditor = createClass({
 		return <img className='thumbnail-preview' src={this.props.metadata.thumbnail || homebreweryThumbnail}></img>;
 	},
 
-	handleFieldChange : function(name, e){
+	handleFieldChange : function(name, inputRules, e){
+		console.log(this.state.errs)
+		for (const err of inputRules){
+			if(new RegExp(err.reg).test(e.target.value)){
+				this.setState((prevState)=>({
+					errs : prevState.errs.filter(function(o){
+						return (o.field !== name || o.type !== err.type);
+					})
+				}));
+			} else {
+				err.field = name;
+				if(_.findIndex(this.state.errs, { 'field': err.field, 'type': err.type }) < 0){
+					this.setState((prevState)=>({ errs: [...prevState.errs, err] }));
+				}
+			}
+		}
+
+
+
 		this.props.onChange({
 			...this.props.metadata,
 			[name] : e.target.value
 		});
 	},
+
 	handleSystem : function(system, e){
 		if(e.target.checked){
 			this.props.metadata.systems.push(system);
@@ -102,20 +121,6 @@ const MetadataEditor = createClass({
 			});
 	},
 
-	checkValid : function(err, e){
-		if(new RegExp(err.reg).test(e.target.value)){
-			this.setState((prevState)=>({
-				errs : prevState.errs.filter(function(o){
-					return (o.field !== err.field || o.type !== err.type);
-				})
-			}));
-		} else {
-			if(_.findIndex(this.state.errs, { 'field': err.field, 'type': err.type })){
-				this.setState((prevState)=>({ errs: [...prevState.errs, err] }));
-			}
-		}
-
-	},
 
 	renderValidationNotice : function(){
 		if(this.state.errs.length == 0) return;
@@ -265,13 +270,13 @@ const MetadataEditor = createClass({
 				<label>description</label>
 				<textarea value={this.props.metadata.description} className='value'
 					onChange={(e)=>{
-						this.checkValid({
-							field : 'description',
-							type  : 'maxLength',
-							reg   : '^.{0,5}$',
-							msg   : 'Max URL length of 5 characters.'
-						}, e);
-						this.handleFieldChange('description', e);
+						this.handleFieldChange('description', [
+							{
+								type : 'maxLength',
+								reg  : '^.{0,5}$',
+								msg  : 'Max description length of 5 characters.'
+							}
+						], e);
 					}} />
 			</div>
 			<div className='field thumbnail'>
@@ -281,13 +286,18 @@ const MetadataEditor = createClass({
 					placeholder='my.thumbnail.url'
 					className='value'
 					onChange={(e)=>{
-						this.checkValid({
-							field : 'thumbnail',
-							type  : 'maxLength',
-							reg   : '^.{0,5}$',
-							msg   : 'Max URL length of 256 characters.'
-						}, e);
-						this.handleFieldChange('thumbnail', e);
+						this.handleFieldChange('thumbnail', [
+							{
+								type : 'maxLength',
+								reg  : '^.{0,5}$',
+								msg  : 'Max URL length of 5 characters.'
+							},
+							{
+								type : 'wStart',
+								reg  : '^W',
+								msg  : 'URL must start with W'
+							}
+						], e);
 					}
 					 } />
 				<button className='display' onClick={this.toggleThumbnailDisplay}>
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -28,7 +28,6 @@ const MetadataEditor = createClass({
 				systems     : [],
 				renderer    : 'legacy',
 				theme       : '5ePHB',
-				validation  : [],
 			},
 			onChange : ()=>{}
 		};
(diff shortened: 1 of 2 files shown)


Gazook89:
Oh, and I won't be bummed if there is a totally different direction to bring this-- as with all my PR's, it's a learning thing first and foremost.

calculuschild:
@jeddai What do you think about this one? Would this be possible to integrate with your Tags validation to help with #2357 ?

Gazook89:
I can comment the code better too if needed (or I hope I can) especially if others need to grab from it. I’m not sure how it would integrate with the tag component but also I haven’t looked at that at all.

jeddai:
Form validation is an area I have a lot of opinions on so I'm going to apologize in advance...

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -234,15 +272,37 @@ const MetadataEditor = createClass({
...
+							{
+								type : 'maxLength',
+								reg  : '^.{0,5}$',
+								msg  : 'Max URL length of 5 characters.'
+							},
+							{
+								type : 'wStart',
+								reg  : '^W',
+								msg  : 'URL must start with W'
+							}
```
jeddai, line 302: Please keep in mind that my experience with form validation comes from Angular, so a lot of my opinions stem from its built-in reactive forms design. 

* [Reactive Forms](https://angular.io/guide/reactive-forms)
* [Validation Functions](https://angular.io/guide/form-validation#validator-functions)

**Strong suggestion:** Generally speaking, I think it makes a lot of sense to store all of the field validation rules in a single place. In Angular's reactive forms, that would be on the form control itself, but with React since there isn't exactly a paradigm or class types for that I would lean more towards a constant object that exists outside the scope of the React component.

The primary reason for this is if for some reason we needed to manually trigger the change event for a field, we would need to include that field's validation rules. This could lead to duplication of those rules or we could just forget to send them or send the wrong ones which could end up with weird scenarios. If the rules are all in one object, we don't need to pass them around and the `handleFieldChange` method can just always check for validation rules for each field.

**Suggestion:** Consider using a structure similar to Angular's form validation functions instead of a pre-defined object with regex as the validator. The reason for this is primarily that regex is more difficult to understand than standard JS. As an example, something like the `maxLength` regex above can be replaced by `value.length <= 5`.

**Suggestion:** Update the `errs` "typed" array to be an object with the string list of errors. Ex:
```js
errs: {
    description: ['Max description length of 5 characters'],
    thumbnail: [],
}
```
I describe it as a "typed" array because it loosely expects the array's objects to conform to a specific "type" in its current design.

<hr/>

Here's a full example of all of what I'm suggesting!

Validators:
```js
const validators = {
    thumbnail: [
        (value)=>{
            return value?.length > 5 ? 'Max URL length of 5 characters' : null
        },
        (value)=>{
            return (value ?? '')[0] !== 'W' ? 'URL must start with W' : null
        }
    ]
}
```

Then the `handleFieldChange` function could be something like this:

```js
        const inputRules = validators[name] ?? [];
        // Call each input rule with the input's value, filter out falsy ones
		const errors = inputRules.map(rule => rule(e.target.value)).filter(Boolean)

        // Set the error state to the previous error state plus the current field's errors or undefined
        this.setState((prevState)=>({
		    errs : {
                ...(prevState.errs ?? {})
                [name]: errors.length > 0 ? errors : undefined
        }))

        // If the state has any truthy errors, don't send the update
        if (Object.values(this.state.errs ?? {}).filter(Boolean).length === 0) {
            this.props.onChange({
			    ...this.props.metadata,
			    [name] : e.target.value
		    });
        }
    }
```

jeddai:
> @jeddai What do you think about this one? Would this be possible to integrate with your Tags validation to help with #2357 ?

When it comes to integrating with the tags validation, that one was a tad different since it was designed to be extensible from that component, and this would be a step above that effectively. I may also be misunderstanding what you're suggesting.

jeddai:
Something else I'll generally comment on is I definitely would argue for having the errors be shown underneath the field(s) that are causing the error(s) as appropriate.

Some examples in existing libraries:
* [Bootstrap](https://getbootstrap.com/docs/4.0/components/forms/#server-side)
* [Angular Material](https://material.angular.io/components/input/examples#input-errors)

calculuschild:
> When it comes to integrating with the tags validation, that one was a tad different since it was designed to be extensible from that component, and this would be a step above that effectively. I may also be misunderstanding what you're suggesting.

 The current tags validation prevents updates and turns the tag red, but it doesn't display the reason why it's invalid. I'm mostly thinking if we are going to add a validation and error output to the metadata panel, we might want to have the tags share the same system.

jeddai:
Ah yes I'd definitely agree.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
Gazook89: So i've read through the above and it makes sense to me.  I agree that setting the rules outside of the inputs is best and had thought about nesting those validators inside the props properties themselves but I have no idea if that would have worked/made sense.  

Is there any value in moving the validators object to a file outside of editPage.js?  Currently the only place to change the metadata is on that page, but perhaps in the future that would change, and having the validators as it's own file so it can be imported wherever would help?   For example, being allowed to change the title or published status from the User Page.

Also, should I just copy/paste your code into the appropriate spots on this PR or close this PR in favor of your own PR?

jeddai: I put the code purely to be an example bc I felt my comment wasn't entirely clear on its own. I wrote them straight in the GitHub editor and didn't test them, so I definitely wouldn't recommend just pasting it in but they are definitely a good template off of which you could make changes.

calculuschild:
This might need to be handled between @Gazook89 and @jeddai due to the overlap between components. Maybe we need to build up a checklist of what needs to be tweaked here. 

Just based on the above comments:

- [ ] Move error to display below the selected element (is there even room?)
- [ ] Create a central "validators" object
- [ ] Get Tags to use the same core error system

Gazook89:
> Move error to display below the selected element (is there even room?)

I had the same thought about spacing when I was considering where to place the error, which is why i went with an overlay.  I understand the spacing needed for an error message appearing underneath a field when filling out a field once then submitting to never seeing it again, but with a set of fields like this I don't like that much spacing.  But perhaps it's not a big deal to increase that row spacing so that errors show below (with large enough font size).  

If continuing with an overlay, there should be a way to dismiss each error message individually, though.  I don't think it's likely, but if there are enough error messages that you can't see or click into the text box, that would be bad.

calculuschild:
@Gazook89 Let's see if we can finish this PR out. Perhaps let's start with the central validator object as @jeddai suggested https://github.com/naturalcrit/homebrewery/pull/2374#discussion_r973665731

ericscheid:
Also, is there a debounce on the fields?

That is, to avoid hammering someone's image server for a url that is increasing one character at a time?

https://homebrewery.naturalcrit.com/t
https://homebrewery.naturalcrit.com/th
https://homebrewery.naturalcrit.com/thu
https://homebrewery.naturalcrit.com/thum
https://homebrewery.naturalcrit.com/thumb
https://homebrewery.naturalcrit.com/thumbn
https://homebrewery.naturalcrit.com/thumbna
https://homebrewery.naturalcrit.com/thumbnai
https://homebrewery.naturalcrit.com/thumbnail
https://homebrewery.naturalcrit.com/thumbnail.
https://homebrewery.naturalcrit.com/thumbnail.p
https://homebrewery.naturalcrit.com/thumbnail.pn
https://homebrewery.naturalcrit.com/thumbnail.png

G-Ambatte:
I don't believe that there currently is, but it would be a sensible thing to do. Lodash's debounce() should be perfect for exactly this function - https://lodash.com/docs/4.17.15#debounce

Gazook89:
Gitter Conversation:
> calculuschild
I think under field. It takes up some space, but I think it will be more clear to the user. We can potentially "animate" the appearance of the error to push the other elements down so they don't take up space until an error occurs.
A popup could work as well as long as it pops up at the site of the error instead of just one universal banner. My only concern is it might cover up the other fields which could be obnoxious.

> Gazook89
if doing errors under the associated field i will likely increase the spacing between rows just a bit. I think its best avoid animating those things if possible, and instead try our best to make the err Message as succinct as possible. If we end up going with larger messages or throwing multiple errors on a single field, maybe animating a shift down is the only way to go

> calculuschild
Yeah. I'm hoping most errors will be just a few words.

>Gazook89
or, pop a red X button on the field (either right on the input or on the label to the left) with a tooltip on hover?

client/homebrew/editor/metadataEditor/metadataEditor.jsx
Gazook89: So I'm back at this, and feel close to just getting the basic stuff down (without the changes to UI/actually displaying errors), but bumping into a frustrating issue.  

First, I pretty much copied what you have above but I'm pretty sure there was a missing comma and closing `}` in the setState portion.  I'm pretty sure I fixed it correctly but mentioning it in case I didn't.   Second, I moved the validation rules into it's own file-- still in the Metadata Editor folder just for now.

[Here is the branch I just pushed to GH](https://github.com/Gazook89/homebrewery/tree/Field-Validation-System-2)

**Current Problem:**  Any text entered specifically into the Thumbnail URL field results in a console error about changing a "Uncontrolled Input" into a "Controlled Input".  I understand the difference, but I don't see how the "thumbnail url" field is any different from the "Title" field for which I do not get any error.  Neither do I get an error when typing into the Description field.   I feel like all three fields are setup exactly the same, yet I get different results....?

calculuschild:
> or, pop a red X button on the field (either right on the input or on the label to the left) with a tooltip on hover?

Let's just show the error directly. Eliminate any extra steps for the user to see what is wrong.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
Gazook89: I think I need to change the field `value` to `defaultValue={this.props.metadata.thumbnail}` and so on for each field.  This populates the fields with the appropriate saved values initially, but thereafter does not set the value to match the changing props.    Without this change, it's going in circles...the value is set to the prop value, and the prop is set by the value, and once we added a bit of logic to prevent writing to the prop suddenly the value couldn't be changed either.

*(right?)*

Gazook89:
> Let's just show the error directly. 
Okay

### New question 1)
should we prevent users from navigating away from the properties editor if there are invalid values? 

### New question 2)
If there is an invalid input, should we invalidate the entire *value* and set the prop to `null` or whatever it's typical default is?  Or do just take whatever the last valid value was and save it to props?  

For example, let's say the rule for thumbnail is "must be less than 5 characters".  If they type "beezlebub", do we chuck out their entry and just set the `prop.metadata.thumbnail = null`?

Or should the entirety of "beezlebub" be allowed to be entered, but only "beez" be saved to props (and then if they navigate out of properties editor and back into it, the field now says "beez")?

calculuschild:
1) Let the user navigate away at any time.

2) we should only be saving properties when the brew itself saves. As in, say the  value is "Satan" from the last time they edited it. Valid because it's <= 5 characters. They then go in and start typing, until it becomes "Beelzebub", let go of the keyboard, and then 2500 ms pass, triggering auto save. Since the field is currently invalid, we just don't update that field. I.e., it should stay as "Satan" if they refresh the page now, rather than a partial entry of "Beelz".

This means if the user navigates away from the properties menu, then goes back, it should refresh with the last saved values, not the invalid value they left behind.

That's my opinion anyway.

Gazook89:
👍 on both answers.  

One more:  

right now the setup is "Allow multiple rules per field".  This is beneficial to me because each rule is easier to read and write....less complicated regex / expressions.  Also, it allows more specific error messaging for whatever rule is being broken.

If a field has multiple rules broken  ("doesn't start with W and is also longer than 5 characters"), should both error messages be visible?  Will this crowd the small space available?  Or just display one error at a time, and once that is fixed show the next error?  Perhaps this could look something like this:  `URL must start with "W". [+1]`  changing the +1 to whatever number of errors.

calculuschild:
Hmm.... Perhaps just one error at a time... @jeddai ?

jeddai:
I like that last solution you proposed @Gazook89, of using the built-in invalidity piece to display multiple errors at once if there is more than one. 

I might suggest not numbering them, but bulleting them should be good!

Gazook89:
I'm going to close this after opening a new effort on PR #2488 which is current with the upstream/main.  I'll bring over open discussion points.

PR #2377, opened 2022-09-19
GitHub forms tweaks

### LINKED ISSUE #2376: Tweaks to issue forms
A few issues with #2367 

- [ ] "validations" needs to go up one level to be right under "type" instead of under "attributes"
- [ ] Some other required fields are missing from the "Feature Request" template
- [ ] Remove the default value from `title` in the `Save Issue` template. Currently it auto-populates with `Save Error:` which users will probably ignore and we'll end up with every save issue with the same title. Leave it blank so we can require users to put some thought into it with the "validations" setting.
- [ ] Add "validations" to the `Save Issue` template.
- [ ] Link from `editPage.jsx` doesn't quite work.
  - [ ] Change `labels` from "saving" to "Saving"
  - [ ] Change `template` from "template=save_issue.yaml" to "template=save_issue.**yml**"

Gazook89:
Here is the PR with the tweaks to fix #2376 

- [x]  "validations" needs to go up one level to be right under "type" instead of under "attributes"
- [x]  Some other required fields are missing from the "Feature Request" template
- [x]  Remove the default value from `title` in the `Save Issue` template. Currently it auto-populates with `Save Error:` which users will probably ignore and we'll end up with every save issue with the same title. Leave it blank so we can require users to put some thought into it with the "validations" setting.
- [x]  Add "validations" to the `Save Issue` template.
- [x]  Link from `editPage.jsx` doesn't quite work.
- [x]  Change `labels` from "saving" to "Saving"
- [x]  Change `template` from "template=save_issue.yaml" to "template=save_issue.**yml**"

### DIFF
--- a/.github/ISSUE_TEMPLATE/general_issue.yml
+++ b/.github/ISSUE_TEMPLATE/general_issue.yml
@@ -12,8 +12,8 @@ body:
       options:
         - label: Legacy
         - label: v3
-      validations:
-        required: true
+    validations:
+      required: true
   - type: dropdown
     id: browser
     attributes:
@@ -37,8 +37,8 @@ body:
         - label: MacOS
         - label: Linux
         - label: other
-      validations:
-        required: true
+    validations:
+      required: true
   - type: textarea
     id: user-description
     attributes:
--- a/.github/ISSUE_TEMPLATE/general_issue.yml
+++ b/.github/ISSUE_TEMPLATE/general_issue.yml
@@ -25,8 +25,8 @@ body:
         - label: Edge
         - label: Safari
         - label: other
-      validations:
-        required: true
+    validations:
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/general_issue.yml
```diff
@@ -25,8 +25,8 @@ body:
         - label: Edge
         - label: Safari
         - label: other
-      validations:
-        required: true
+    validations:
+      required: true
   - type: dropdown
```
calculuschild, line 30: There are two other spots in this file that need `validations` moved to the correct spot.

Gazook89: dang, done.

Issue #2071, opened 2022-03-14
Awkward flow in navbar when brew title is long


Gazook89:
This is a minor thing that's bothered me for awhile, sorry to open an issue on it:   When a brew title is long, and goes to two lines, the navbar height increases but the nav items come out looking odd.  

<img width="1252" alt="image" src="https://user-images.githubusercontent.com/58999374/158091661-672c359f-c045-4dbb-865c-53fb96ec61c1.png">

A simple solution is to make the nav items flexboxes, and vertically center the fontawesome icons.  It looks like this:

<img width="1252" alt="two line title" title="two line title" src="https://user-images.githubusercontent.com/58999374/158091494-ccea6aec-ef7d-4c1d-8886-89c82a8ab871.png">

And with single line title:
<img width="1252" alt="single line title" title="single line title" src="https://user-images.githubusercontent.com/58999374/158091592-06455408-85bb-415a-abb9-f0ada35d0ec9.png">

The CSS changed is this:
```css
nav .navItem {
	-webkit-transition: background-color .25s ease;
	-moz-transition: background-color .25s ease;
	-o-transition: background-color .25s ease;
	transition: background-color .25s ease;
	padding: 8px 12px;
	cursor: pointer;
	background-color: #333;
	font-size: 10px;
	font-weight: 800;
	color: white;
	text-decoration: none;
	text-transform: uppercase;
	line-height: 13px;
	display: flex;  /* added these three items */
	align-items: center;
	justify-content: center;
}
```

I can create a PR quickly if interested.  I looked at it in both Chrome and FF.

Gazook89:
The navbar can still go to two rows, though, if the window is narrow.  So I think my suggested flexbox option is still appropriate even if `ellipsis` is added.   

I'm not sure I've seen the save/saving... thing happen.  Right now just playing with it (in the production app) I can get my title really close to the 'full' space, and any changes trigger "Save Now", then a second passes before "Saving..." and then to "Saved.".  At that final step, the cell doesn't get any smaller.   I might just suck at reproducing it though.   But it could be fixed by setting a `min-width` on that item?  Or if it's a flex item, `flex: 0 0 50px` (exactly pixel count TBD).

Gazook89:
Also I see now in my screenshots that the `saved` item is not centered (or at least doesn't appear to be, though it might centered in whatever div it is in).  Obviously that can be fixed.

calculuschild:
[...]

Issue #2295, opened 2022-08-28
[UserPage] Minor Fixes - Checklist


G-Ambatte:
User page changes:

- [x] show current tags in BrewItem (mimic current author listing functionality?) (#2317 / #2318)
- [x] expand filter to include tags (#2319 / #2320)
- [x] add URL parameters for sort by and direction ( #2315 / #2316)
- [x] save sort by & direction to local storage; load and use when page loaded with no URL parameters present (#2326 / #2327)
- [x] loading page with filter URLParameter does not apply filter to page (#2297)
- [x] ~~show current system tags (#1842 / #2330)~~ 
- [x] rework sorting options, possibly add page count to sorting options (#1796 / #2379)
- [x] remove autofocus from User Page (#2121 / #2305)
- [x] add section collapsing and/or reordering ( #1797 /  #2300)
- [x] show thumbnail from metadata (if not blank) ( #2331 /  #2332)
- [x] fix centering of SortContainer when scroll bar hidden ( #2303 / #2304; follow up #2352 / #2360 & #2382)
- [x] switch to v3 CSS (#2301 / #2302)
- [x] smarten possessive apostrophe in `usernameWithS` (replace `'` with `&rsquo;`) (#2328 / #2329)

ericscheid:
- [ ] show thumbnail from metadata (if not blank)

ericscheid:
- [ ] fix the centering of the sortcontainer

<img width="938" alt="image" src="https://user-images.githubusercontent.com/2605467/187063453-83773f01-9fbc-449b-8245-7020de753f28.png">

The current CSS has `left: calc(50vw - 408px)` .. but this fails if the scroll bar on OSX is set to hide.

Another strategy is needed.

ericscheid:
- [ ] Use v3 renderer

Just for sanity sake.

<img width="678" alt="image" src="https://user-images.githubusercontent.com/2605467/187063708-a810795d-acf6-41ad-8b4c-9771cfa9a29d.png">

G-Ambatte:
> * [ ]  Use v3 renderer

It should be trivial to switch over to the v3 CSS - I'll have a quick look at it later today/tonight.

G-Ambatte:
Add all of these to the main checklist.

G-Ambatte:
> smarten possessive apostrophe in `usernameWithS`

Added to the list.

Gazook89:
[...]

PR #2379, opened 2022-09-19
Change User Page search bar to Divs

### LINKED ISSUE #1675: tweak position of sort-container on user page
The `sort-container` on the user page isn't properly centered.
image

The position is determined by `.userPage .sort-container { left: calc(50vw - 408px); `.

Change this to `.userPage .sort-container { left: calc(50vw - 400px); `.

image

Gazook89:
Currently the User Page search/filter bar is rendered as a table which I think could be improved.  Using divs in a flexbox is a bit more flexible if more options are added later or screen size changes.

Also, as part of the UI, I wanted to change it to feel more cohesive with the existing UI rather than blending with the page content.  Structurally, in the HTML, I also moved the sort bar out of the '.page' div.

image

fixes #1675

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
@@ -96,7 +96,7 @@ const BrewItem = createClass({
 	render : function(){
 		const brew = this.props.brew;
 		if(Array.isArray(brew.tags)) {               // temporary fix until dud tags are cleaned
-			brew.tags = brew.tags?.filter(tag => tag); //remove tags that are empty strings
+			brew.tags = brew.tags?.filter((tag)=>tag); //remove tags that are empty strings
 		}
 		const dateFormatString = 'YYYY-MM-DD HH:mm:ss';
 
@@ -112,7 +112,7 @@ const BrewItem = createClass({
 					<div className='brewTags' title={`Tags:\n${brew.tags.join('\n')}`}>
 						<i className='fas fa-tags'/>
 						{brew.tags.map((tag, idx)=>{
-							let matches = tag.match(/^(?:([^:]+):)?([^:]+)$/);
+							const matches = tag.match(/^(?:([^:]+):)?([^:]+)$/);
 							return <span className={matches[1]}>{matches[2]}</span>;
 						})}
 					</div>
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -121,14 +121,8 @@ const ListPage = createClass({
 			>
 				{`${sortTitle}`}
 			</button>
-			{this.state.sortType == sortValue ? 
-				<button
-					onClick={this.handleSortDirChange}
-					className='sortDir'
-				>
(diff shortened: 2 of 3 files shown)


Gazook89:
Just going to update this comment as i goof around with some CSS.  I can't show hover states etc but I think once an overall 'active' state is settled on it'll be a lot easier to get the hover/inactive states worked out.  And sorry some of them are not scaled or dimensioned the same.

1. image
2. image
3. image
4. image

G-Ambatte:
I've added this PR to the #2295 UserPage Minor Fixes list. I think it looks good as part of the nav bar; I'll try to take a proper look at it later tonight.

ericscheid:
> Just going to update this comment as i goof around with some CSS. 
> 
> 4. image

I prefer this styling over the others, only without such eye-catching colouring — the sort options are optional and supporting to the list page experience.

Also, the color contrast of the llight grey on the darker grey is very poor.
https://coolors.co/contrast-checker/888888-555555

Gazook89:
> Also, the color contrast of the llight grey on the darker grey is very poor.

So i was wondering about this-- i know it is low contrast, but do you want it to be high contrast if it's not the selected option?  I suppose it could just be white, leaving the background coloring the sole distinguishing style for active.  

"the color is too contrasting and the gray is not contrasting enough!"  :)

ericscheid:
So .. contrast/coloring more like this
image

(except rendered as a browser wide bar, not a div floating over the page content area, and using proper sort-direction indicators)

- "SORT BY:" label is `#ffffff` on `#555555` ([7.46, Very Good](https://coolors.co/contrast-checker/ffffff-555555))
- selected sort option is `#dddddd` on `#333333` ([5.49, Good](https://coolors.co/contrast-checker/dddddd-555555))
- not-selected sort option is [`#cccccc` on `#555555` ([4.64, Good](https://coolors.co/contrast-checker/cccccc-555555))

G-Ambatte:
Suggestion of a slight functional addition - clicking the currently selected item will change the sort direction, rather than the current functionality (of changing state to the same state as it is currently in; i.e. functionally, doing nothing).

image

Might be different enough (functionality change, rather than rearranging UI) that it deserves it's own PR.

---

I nudged a couple of things and added some `&:hover` styling to `listPage.less` (text turns white and border visible on hover), resulting in this:

image

image

But we can always move away from the color to something less visually distracting.

calculuschild:
I think @ericscheid 's color scheme looks really clean :+1: I like it.

@G-Ambatte I agree clicking the same category again should change the sort direction. I think that's pretty standard functionality. Should be easy enough to just add to this PR.

Gazook89:
Okay, I went with @ericscheid's suggestions.  Here is what it looks like:

image

Note, there is a hover state now as well.  Also, I went with the fontawesome 'sort-up' and 'sort-down' icons....they aren't *full-size* carets, but rather a 'top and bottom'.  At first I thought they were too small, but now that clicking on the sort title has the same function as clicking on the caret, I think it doesn't matter.

G-Ambatte:
On an almost completely unrelated note, would a similar rework of the nav bar resolve #2071?

Gazook89:
> On an almost completely unrelated note, would a similar rework of the nav bar resolve #2071?

I have a branch for this as well from a long time ago. I just have to look over it again and see where it is at

calculuschild:
Alright, I have no other comment on this. Looks good!

PR #2380, opened 2022-09-20
Make Reddit Issue Template a List

Gazook89:
Currently, issues sent to reddit from HB look like this:

image

This PR changes this to a list, so each item is on it's own line.  I fudged this when we originally added a reddit template.

I also thought about trying to get the renderer to auto-populate into the template, but that's pretty much specific to just the editPage and probably not worth it at this point.


calculuschild:
Looks good! :+1:

PR #2382, opened 2022-09-21
Scroll behavior correction

### LINKED ISSUE #1881: Suggestion: scrollable/shorter recent brews dropdown menu
Suggestion: recent brews dropdown with it's own scroll function

Why? because i am experiencing a chrome bug, that when i put my mouse over the right part of the recent brews button, the dropdown opens, and it is larger than the page, so it creates another scroll bar at the right of the page, therefore moving everything, including the bottom, which suddently i'm not over.

All that activates and deactivates the hover effect, and weirds the page until you move your mouse.

[video](https://imgur.com/a/DDihHRf)

it is a minor inconvenience, but easily sorted by either decreasing the number of recent brews, or making the dropdown menu scrollable by itself.

Gazook89:
My last PR #2360  created a problem (i think, please double check if you can) where it doubled up the scrollbars on pages with iframes....so most of them.

This PR attempts to fix that error by narrowing the specificity of the rule.  

There was also a `overflow-y: overlay;` leftover that I could remove as well.

Finally, it sets a scrollbar style on the Recent Items scrollbar.  However, it only does so for firefox.  I read up on the CSS for Chrome scrollbars and just couldn't get it work-- maybe someone else knows what is up:

```css
.dropdown::-webkit-scrollbar-thumb {
  background : #333;
```

I can strip away the styling stuff and just leave the fix for the double scrollbar, too, if desired.  In any case, this should finally close #1881 .

### DIFF
--- a/client/homebrew/homebrew.less
+++ b/client/homebrew/homebrew.less
@@ -16,8 +16,5 @@
 		&.listPage .content {
 			overflow-y : scroll;
 		}
-		// &.editPage .content, &.sharePage .content, &.newPage .content {
-		// 	overflow-y : hidden;
-		// }
 	}
 }
--- a/client/homebrew/homebrew.less
+++ b/client/homebrew/homebrew.less
@@ -11,7 +11,13 @@
 			position   : relative;
 			height     : calc(~"100% - 29px"); //Navbar height
 			flex       : auto;
+			overflow-y : hidden;
+		}
+		&.listPage .content {
 			overflow-y : scroll;
 		}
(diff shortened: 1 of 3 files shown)


G-Ambatte:
Looks like it works fine in both Windows Chrome & Firefox. The `sort-container` position is off-center in both browsers, but combined with the rework of that element in #2379, I think this is fine.

I also tested the Recently Viewed/Edited scrollbar, which also appears to work as intended.

calculuschild:
Alright, I'm gonna merge this and take your word for it :+1: This is small enough I don't see it possibly breaking anything else.... 👀

PR #2383, opened 2022-09-21
Prevent text selection in editor when dragging splitPane divider.

### LINKED ISSUE #1632: Dragging editor pane divider can select non-visible text
Say you have a big brew.

You're editing one section. Your cursor is in that section.

You want to resize the editor/preview panes, so you drag the divider.

Result: a big chunk of text below the visible editor could end up being selected.

Problem: typing anything will obliterate that new selection.

Gazook89:
Hoping this fixes #1632, preventing the selection of text in the editor when the split pane divider is moved.  I've tested it a bunch and haven't seen any selections made, but someone else should try it a bunch to make sure...

Also, I noted that someone else has commented out a `unFocus` method that I think had the same goal.  I uncommented it and tried it, and was still able to get a selection.  I commented it out again and left it in this branch for now, but I can remove it with another commit before this gets merged (if it does).

### DIFF
--- a/shared/naturalcrit/splitPane/splitPane.jsx
+++ b/shared/naturalcrit/splitPane/splitPane.jsx
@@ -69,7 +69,8 @@ const SplitPane = createClass({
 		this.setState({ isDragging: false });
 	},
 
-	handleDown : function(){
+	handleDown : function(e){
+		e.preventDefault();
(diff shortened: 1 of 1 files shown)


calculuschild:
Ok, I can't seem to replicate the error either so I assume this is good enough. I'll merge it.

PR #2384, opened 2022-09-21
Add ability to toggle the Auto Save

### LINKED ISSUE #1546: turn off auto-save?
> Well, I've done it. I'm on a laptop, but I use a wireless mouse. While I was mousing around, my hand brushed my touchpad (I use both sometimes), accidentally highlighting a whole lot of text. One "f" later and most of my homebrew is gone, and of course Homebrewery saved almost immediately.
>
> It's my fault, I'm aware, and I suppose I should be backing up in other places, but the easier solution for me (by far) is turning off the auto save. I've never had a problem saving on my own, even in the midst of late-night RPG games.
> 
> Is there a way to do this?

Auto-save is a very handy thing, but ideally it should be paired with version-history. Which we don't have. And probably won't until someone pays for the massive inflation of storage space requirements =(

Since it's not always obvious when something like this happen (e.g. cats sits on kbd while user away getting cawfee) then having an option to turn off auto-save isn't totally unreasonable. 

It could be a simple toggle thing in the Editor UI, next to the save button. 

Tagging as complicated as any solution should be more than simply "turn off auto-save".

If this implemented, it should also throw a reminder every now and then (or even a non-modal titchy dropdown warning notification "You haven't saved for a while. Save now, or continue living dangerously?" — not unlike the titchy dropdown warning for "gdrive is borking, try logging in again")

Gazook89:
This PR resolves #1546 , adding an ability to turn the auto-save off.   Setting this as a DRAFT for now since it doesn't quite hit every box in the task list.   

1. There is no visual indicator of how to do this toggle, yet.  Right now, you just click on that section of the navbar.  I'm all for ideas here.
2. Possibly occasional pop-ups are needed to remind folks that auto-save is off.
3. Also, a decision about whether this setting should persist across sessions in localStorage.  If it does persist, I need to do something so that it still tries to save when the window closes.

The different states:

### Brew is Saved, auto-save is on:
image

### Brew is Saved, auto-save is off:
image

### Brew is *not* Saved, auto-save is off:
image

And then there are the normal 'brew is not saved, auto-save is on' which looks the same as above, and the 'saving...' which flashes by too fast for me to screenshot.

### DIFF
--- a/client/homebrew/navbar/navbar.less
+++ b/client/homebrew/navbar/navbar.less
@@ -55,6 +55,13 @@
 		text-align     : center;
 		text-transform : initial;
 	}
+	.save-menu .navItem i.fa-power-off {
+		color : red;
+		&.active {
+			color : rgb(0, 182, 52);
+			filter : drop-shadow(0 0 2px rgba(0, 182, 52, 0.765))
+		}
+	}
 	.patreon.navItem{
 		border-left  : 1px solid #666;
 		border-right : 1px solid #666;
--- a/client/homebrew/navbar/navbar.less
+++ b/client/homebrew/navbar/navbar.less
@@ -55,11 +55,16 @@
 		text-align     : center;
 		text-transform : initial;
 	}
-	.save-menu .navItem i.fa-power-off {
-		color : red;
-		&.active {
-			color : rgb(0, 182, 52);
-			filter : drop-shadow(0 0 2px rgba(0, 182, 52, 0.765))
+	.save-menu {
+		.dropdown {
+			z-index: 1000;
+		}
+		.navItem i.fa-power-off {
+			color : red;
+			&.active {
+				color : rgb(0, 182, 52);
+				filter : drop-shadow(0 0 2px rgba(0, 182, 52, 0.765))
+			}
 		}
 	}
 	.patreon.navItem{
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -64,7 +64,8 @@ const EditPage = createClass({
 			htmlErrors             : Markdown.validate(this.props.brew.text),
 			url                    : '',
 			autoSave               : true,
-			autoSaveWarning        : false
+			autoSaveWarning        : false,
+			unsavedTime            : new Date()
 		};
 	},
 	savedBrew : null,
@@ -73,6 +74,7 @@ const EditPage = createClass({
 		this.setState({
 			url : window.location.href
 		});
+
 
 		this.savedBrew = JSON.parse(JSON.stringify(this.props.brew)); //Deep copy
 
@@ -230,8 +232,9 @@ const EditPage = createClass({
 				editId 	 : this.savedBrew.editId,
 				shareId  : this.savedBrew.shareId
 			},
-			isPending : false,
-			isSaving  : false,
+			isPending   : false,
+			isSaving    : false,
+			unsavedTime : new Date()
 		}));
 	},
 
@@ -339,13 +342,14 @@ const EditPage = createClass({
 		}
 
 		if(this.state.autoSaveWarning){
+			console.log(this.state.unsavedTime);
 			setTimeout(()=>this.setState({ autoSaveWarning: false }), 4000);
 			this.setAutosaveWarning();
 
 			return <Nav.item className='save error' icon='fas fa-exclamation-circle'>
 			Reminder...
 				<div className='errorContainer'>
-					Autosave has been turned off, and you haven't saved for {Math.round((new Date() - new Date(this.props.brew.updatedAt)) / 1000 / 60)} minutes.
+					Autosave has been turned off, and you haven't saved for {Math.round((new Date() - this.state.unsavedTime) / 1000 / 60)} minutes.
 				</div>
 			</Nav.item>;
 		}
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -62,7 +62,8 @@ const EditPage = createClass({
 			confirmGoogleTransfer  : false,
 			errors                 : null,
 			htmlErrors             : Markdown.validate(this.props.brew.text),
-			url                    : ''
+			url                    : '',
+			autoSave               : true
 		};
 	},
 	savedBrew : null,
@@ -142,7 +143,12 @@ const EditPage = createClass({
 		return !_.isEqual(this.state.brew, this.savedBrew);
 	},
 
+	toggleAutoSave : function(){
+		this.setState((prevState)=>({ autoSave: !prevState.autoSave }));
+	},
+
 	trySave : function(){
+		if(!this.state.autoSave){ console.log('Auto-save is now off.'); return };
 		if(!this.debounceSave) this.debounceSave = _.debounce(this.save, SAVE_TIMEOUT);
 		if(this.hasChanges()){
 			this.debounceSave();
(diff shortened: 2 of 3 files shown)


calculuschild:
> 1. There is no visual indicator of how to do this toggle, yet.  Right now, you just click on that section of the navbar.  I'm all for ideas here.

Perhaps just another dropdown? Similar to the `SHARE`  button 
image

Maybe saying "Autosave: ☑️" with a check/cross depending on the status, kind of like the sort direction indicator on the updated userpage.

> 2. Possibly occasional pop-ups are needed to remind folks that auto-save is off.

Yeah.... Maybe if we can set a 20-minute timer (using (setInterval)[https://developer.mozilla.org/en-US/docs/Web/API/setInterval] or similar?) that resets every time a save occurs or the popup is closed. "Autosave has been turned off, and you haven't saved for 20 minutes" I imagine this can just reuse the error message popup. Also when first loading the edit page, should perhaps show a popup that just says "autosave is off" and then fades away after 4 seconds.

> 3. Also, a decision about whether this setting should persist across sessions in localStorage.  If it does persist, I need to do something so that it still tries to save when the window closes.

Yeah... probably should persist into localStorage. We might want to ask to save before closing though. If I deliberately turn autosave off, it's probably because I'm fiddling with something I don't actually want saved. It would be a nasty surprise if I scrambled my brew, said "nah, I'll just close and reopen the page to get back to how it was before", and instead, it just saves my scramble.

calculuschild:
Ooh fancy

Gazook89:
> pop-up when closing window and autosave is off

HB already pops a browser alert if closing the window with unsaved changes.  I think this is a browser thing. Either way, both Chrome and FF do this, with slightly different messages:

image

calculuschild:
A couple things I noticed:

- When I first opened a brew, autosave was disabled by default. We should make sure it is turned on by default (opt-out rather than opt-in).
- When turning on auto-save, it should trigger a save immediately since it doesn't seem to trigger the autosave until the user starts typing again.   ("I turned on auto-save and it didn't save my existing changes")

ericscheid:
> > pop-up when closing window and autosave is off
> 
> HB already pops a browser alert if closing the window with unsaved changes. I think this is a browser thing. Either way, both Chrome and FF do this, with slightly different messages:

This is in our code, triggered by the `onbeforeunload` event. Browsers by default don't pop any warning when unloading windows with unsaved changes. [Chrome has moved away from facilitating custom messages](https://chromestatus.com/feature/5349061406228480) in browser UI widgets )for obvious security reasons) and so you'll only get a generic message.

https://github.com/naturalcrit/homebrewery/blob/ce7c3dfb8aad89f66164c9f8cd1b03f1f6c1c5d9/client/homebrew/pages/editPage/editPage.jsx#L78-L82

There are other mentions of `onbeforeunload` too.

Gazook89:
Yeah sorry, I meant the message that either browser gives is determined by the browser…neither is saying “You have unsaved changes!”

calculuschild:
In that case I guess it's already handled by the browser as well as it can be.

Gazook89:
9bb1344 addresses the issues @calculuschild pointed out:

1. The default for autoSave is now `true`
2. When toggling auto-save to 'on' it now runs `trySave()` automatically.  This means there is a 3 second debounce.
3. When a save is a result of Auto-save the message displayed in the navbar is "auto-saved."    There is one tricky part....If your brew is saved already, and you toggle auto-save "off", the message quickly changes to "saved.".   This might give the impression that the brew *just* saved as a result of your toggle, but in fact it is just showing the saved status which is unchanged.   It sort of makes my mind think that I just saved my brew when trying to turn off autosave, but of course this isn't true:  With autosave previously on, it only makes since that the status would be saved.

So I think pretty much all that is left is some sort of alert on a timer reminding users to save.  The existing "error" pop-up was my original goal, but now I'm thinking about changing the "saved status" item itself....similar to what happens when there is a save error and it turns to "oops!".  I dunno, going to think about it.

calculuschild:
What is the purpose of moving the autosave check out of trySave? Doesn't every call to trySave need the check anyway?

Gazook89:
> What is the purpose of moving the autosave check out of trySave? Doesn't every call to trySave need the check anyway (and thus should just be built-in to the function)?

I am having a hard time thinking about the setTimeout() method to create an 'error' pop-up every ~15 minutes, and one thing that was bothering me was the constant attempts to `trySave()` and failing, rather than just *not* trying to save at all if autosave is off.

I'm not 100% sold on it, or even 30%, but at least for now I wanted to try this other direction so I could stop thinking about it.

calculuschild:
Ok. If its on the way to solving the timer issue that's fine, and we can reorganize later once that's solved. I was mistakenly assuming this was an arbitrary refactor that didn't seem to do anything.

Gazook89:
Okay I have a timed pop-up appearing at a set interval:

image

> set a 20-minute timer ....  "Autosave has been turned off, and you haven't saved for 20 minutes"

20 minute interval, with the above wording?

I still have a couple of things to wrap up such as popping an alert on load.

G-Ambatte:
Lodash's `throttle` function might be useful - I used `{leading:true, trailing:false}` for the `brewJump` function to scroll a given page into view, but I suspect the reverse might work for executing a function - like popping a notification - after a given amount of time, while simultaneously blocking any other calls to that function before the timer expires.

Gazook89:
Okay so I sort of redid all my work after switching back and forth between setTimeout, setInterval and debounce.  I think this is where things stand right now:

- [x] Save Autosave setting to localStorage
- [x] If reloading page with Autosave off, pop notification for 4 seconds
- [x] Display warning notification every 20 minutes (currently 15 seconds for testing) if autosave is off.*
- [x] Turn off the warning notification timer when toggling autosave on.
- [x] Run `trySave()` when toggling autosave on.
- [x] Pop browser alert when closing the window with unsaved changes.
- [x] Display the Save dropdown menu *over* the warning notification
- [x] Set "saved." message to reflect either manual save or autosave.

Still needed:
- [ ] The warning notification fires even if there are no unsaved changes, which likely isn't necessary.
- [ ] Fix the "you haven't saved in *x* minutes" so that it displays only the number of minutes from either the last save, *or from when the brew was opened* in the case that the brew hasn't been saved since open.  Otherwise it's easy to open a brew that has a `updatedAt` property of 5 months ago.

Gazook89:
I'm pushed this into a normal PR at this point.  There is one last bullet point that I think probably needs to get done, but my brain stopped working on it a few days ago.  I'd like to get more input from ya'll on what *is* there before tackling that last piece, in case what I already have done needs a total rework.

And @calculuschild , if doing Hacktoberfest, can this be counted (and does HB need to register)?

calculuschild:
The HB is registered for Hacktobwrfest, so this should count as long as it gets approved during this month. I think you also need to register yourself on their site if so remember correctly.

calculuschild:
> I'd like to get more input from ya'll on what is there before tackling that last piece

@Gazook89 I think this all looks great. Go ahead and finish up the last piece when you're ready and we can merge it!

calculuschild (commented):
> I'd like to get more input from ya'll on what is there before tackling that last piece

@Gazook89 I think this all looks great. Go ahead and finish up the last piece when you're ready and we can merge it!

Gazook89:
@calculuschild Okay I think I got that last thing (*only display 'off' warning if changes exist*).  

I also changed the time between warnings to 15 minutes.

I think this is ready to go, at least according to my list.  Hopefully the code is neat enough-- it feels a little bit like a lot for the simple task it is doing.

calculuschild:
Cool, I'm finally going to merge this! Great work! 💯 🎉

Gazook89:
One last comment on this even though it's closed--- I think I could have utilized Moment.js for the "no save since x minutes ago" portion, just the same as the recent items dropdown items.  I only just tonight found out about moment.js and considered it, but didn't want to add another library....but now i see we already use it.   

Oh well, this works as-is.

PR #2387, opened 2022-09-21
Change Monster Stat background color

### LINKED ISSUE #2123: New Framed Monster Statblock Background Color?
I *think* this is the right place to post this, but I was messing around with a PHB theme, and I found that **#F2E5B5** seems a bit more accurate (and to me, more pleasing to look at) than the current **#EEDBAB**. It has slightly less contrast and is a little more parchment-y, I think. 

This is obviously pretty low priority, and only my opinion, but I figured I'd mention it.

Gazook89:
This resolves #2123 with a tiny change to the monster stat block background color as suggested by @Kaiburr-kath-hound, for v3 only.

Sidenote, is there a good way to refresh phb css so that the changes go through quickly?  I thought i remember a configuration i could change that would quickly refresh the CSS after each change, but can't remember now.

Before:
image

After:
image


calculuschild:
⭐ ⭐ ⭐ ⭐ ⭐ ⭐ ⭐

PR #2388, opened 2022-09-21
Add renderer to pageInfo

### LINKED ISSUE #1928: Add some indicator of renderer version outside of Meta panel
It might be nice to have some visual cue as to the renderer version that is more apparent than the checkbox in the metadata panel.  I was thinking that maybe the "text editor" menu in the snippet menu could read "V3 Text Editor" or "Legacy Text Editor"?

Or just "V3" or "Legacy" written anywhere on the screen.

This idea came up when trying to troubleshoot another users brew from a posted screenshot, and just wondering which editor they were using.  Based on the fact they had some html tags, I'm assuming it's Legacy, but that may not necessarily be the case.  If the screenshot doesn't happen to include wherever this renderer indicator is located, then i'll be out of luck anyway, but it would have a higher chance of being included than the metadata panel.

I also think it would just be more clear to users as they work.

Gazook89:
Resolves #1928 by adding a field in the pageInfo for the renderer.  No additional functionality.

image

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -109,7 +109,12 @@ const BrewRenderer = createClass({
 
 	renderPageInfo : function(){
 		return <div className='pageInfo' ref='main'>
-			{this.state.viewablePageNumber + 1} / {this.state.pages.length}
+			<div>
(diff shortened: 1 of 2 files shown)


calculuschild:
Noice. 👍 :fire: :star: :sparkles: ✔️

PR #2392, opened 2022-09-24
[Editor] Prevent unfolding code in Editor when adding a Snippet

### LINKED ISSUE #2135: Inserting snippets unfolds code
Pretty much what the title says, I found that inserting any kind of snippet(images, spells, tables...) from the snippet menus unfolds all the code, which is quite annoying, and I don't think it was intended.

Happens in both unsaved and saved brews in both V3 and Legacy mode.

G-Ambatte:
This PR resolves #2135.

After investigating Gazook89's draft PR  #2391, I was unable to automate the re-folding of previously folded code blocks in the Editor. However, on comparing the `handleInject` function to the shortcuts for page and column breaks, I realized that the `handleInject` function was operating in a very different fashion, causing the document to effectively be regenerated, losing the folded marks.
By moving this function out of the `editor.jsx` to `codeEditor.jsx` and using the same method (`replaceSelection`) as the shortcut keys, folded code segments are no longer unfolded when a snippet is injected.

At this stage, I have included an optional `overwrite` parameter to overwrite the existing selection, but the `handleInject` function is currently called with this set to `false` in order to mimic the previous functionality.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -80,19 +80,7 @@ const Editor = createClass({
 	},
 
 	handleInject : function(injectText){
-		let text;
-		if(this.isText())  text = this.props.brew.text;
-		if(this.isStyle()) text = this.props.brew.style ?? DEFAULT_STYLE_TEXT;
-
-		const lines = text.split('\n');
-		const cursorPos = this.refs.codeEditor.getCursorPosition();
-		lines[cursorPos.line] = splice(lines[cursorPos.line], cursorPos.ch, injectText);
-
-		const injectLines = injectText.split('\n');
-		this.refs.codeEditor.setCursorPosition(cursorPos.line + injectLines.length, cursorPos.ch  + injectLines[injectLines.length - 1].length);
-
-		if(this.isText())  this.props.onTextChange(lines.join('\n'));
-		if(this.isStyle()) this.props.onStyleChange(lines.join('\n'));
+		this.refs.codeEditor?.injectText(injectText);
 	},
 
 	handleViewChange : function(newView){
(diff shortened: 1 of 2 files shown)


Gazook89 (approved):
Looks good, but one step further (into a possible new PR) would be to store the folds with the brew, so that the fold states could be preserved when refreshing the brew or loading it anew?

I think the original intent of this PR, to fix that unfolding issue, is achieved in any case.

G-Ambatte:
I think there is potentially value in storing fold information; however I think that would be a very different problem to solve than what this PR has been about.

Gazook89:
### Folded Preview Content

While you are in the folding functions, do you want to visit this issue #1907 concerning what appears in the folded 'preview' of the page? Displaying just the first header on the page would be ideal for me.  Even if the first lines are not a header I think it should find the first header-- it's a better descriptor of "what" is on the page than just a random line of text or code.   And if no header exists on the page, resort to just whatever is in the first line.

-----

### Folded Preview Styling

 At the same time, I think it's a good idea to rethink the styling of the folded preview text-- bold bright blue text for the preview text draws a lot of visual attention to text that likely should be taking a backseat, since it is folded.  Changing it to regular weight and gray would be my recommendation.

calculuschild (changes requested):
This seems to work great! The only comment I have is the cursor position gets lost after injecting. Normally we are able to keep typing after the inject but now have to click back onto the editor.

I also like that the injected text gets highlighted, now. I don't know if setting a cursor position would remove that.

G-Ambatte:
It's either/or, unfortunately. There's a post-replacement option which is currently set to `around`, which selects the newly inserted text. Changing this to `end` should place the cursor after the inserted snippet.

I can push a change using that option so we can do what the preferred behavior is.

G-Ambatte:
I've pushed a change to the post-insert selection now, so it uses the same parameter as inserting a page or column break. It's a bit different when inserting a snippet, as it doesn't (currently) use a hot key, so the user needs to click back into the Editor anyway.

calculuschild (approved):

PR #2396, opened 2022-09-25
[Editor] Change folded code text generation

### LINKED ISSUE #1907: Smarter code folding snippets?

G-Ambatte:
This PR will resolve #1907.

This PR modifies the text generated when code is folded, using the following rules:
1. Use the first header in the folded section. At this point, a `h6` at the top of the fold will be used over a later `h1`.
2. If no headers exist, use the text of the first non-empty line.
3. If no non-empty lines exist, generate text "Lines {start line}-{end line}"

---
Example output:

Folded:
image

Brew Source:
```
# Combat Maneuvers
\page
# Combat Actions
\page
## Movement
\page
## Defense
\page
## Special Targeting
\page
# Footnotes on source rules
\page
### Parry Defence
\page
## Other options/wrinkles
\page
### Weapon Length and Positioning
\page
## Damage-type Maneuvers
\page
## No Quarter <small>Combat manuevers - for fighting</small>
\page





\page

This is a test page.
This is the second line.

\page

This is a test page.
This is the second line.

### Header later on


\page

```

### DIFF
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -356,23 +356,19 @@ const CodeEditor = createClass({
 				let currentLine = from.line;
 				const maxLength = 50;
 
-				let headerText = '';
-				let otherText = '';
+				let foldPreviewText = '';
 				while (currentLine <= to.line && text.length <= maxLength) {
 					const currentText = this.codeMirror.getLine(currentLine);
 					currentLine++;
-					if(!currentText) {
-						continue;
-					}
 					if(currentText[0] == '#'){
-						headerText = currentText;
+						foldPreviewText = currentText;
 						break;
 					}
-					if(!otherText && currentText != '\n') {
-						otherText = currentText;
+					if(!foldPreviewText && currentText != '\n') {
+						foldPreviewText = currentText;
 					}
 				}
-				text = headerText || otherText || `Lines ${from.line+1}-${to.line+1}`;
+				text = foldPreviewText || `Lines ${from.line+1}-${to.line+1}`;
 
 				text = text.trim();
 				if(text.length > maxLength)
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -355,12 +355,24 @@ const CodeEditor = createClass({
 				let text = '';
 				let currentLine = from.line;
 				const maxLength = 50;
+
+				let headerText = '';
+				let otherText = '';
 				while (currentLine <= to.line && text.length <= maxLength) {
-					text += this.codeMirror.getLine(currentLine);
-					if(currentLine < to.line)
-						text += ' ';
-					currentLine += 1;
+					const currentText = this.codeMirror.getLine(currentLine);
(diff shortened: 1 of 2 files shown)


G-Ambatte:
As per the comments on #1907, I've added a `color` attribute to the LESS for the fold-marker, changing it from blue to grey.

In this example, I've folded all of the code, then re-opened one to make sure that the heading (and it's color) is visible.

image

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -355,12 +355,24 @@ const CodeEditor = createClass({
 				let text = '';
 				let currentLine = from.line;
 				const maxLength = 50;
+
+				let headerText = '';
+				let otherText = '';
```
calculuschild, line 360: Seems like we don't need both of these since we only use one or the other. Change to `let foldPreviewText = ''` or something?

If we hit a non-header, we store it temporarily, but then keep looking for a header until we find one and we can overwrite this same variable.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -355,12 +355,24 @@ const CodeEditor = createClass({
...
 				while (currentLine <= to.line && text.length <= maxLength) {
-					text += this.codeMirror.getLine(currentLine);
-					if(currentLine < to.line)
-						text += ' ';
-					currentLine += 1;
+					const currentText = this.codeMirror.getLine(currentLine);
+					currentLine++;
+					if(!currentText) {
+						continue;
+					}
```
calculuschild, line 366: When `!currentText` we are at the end of the loop anyway, right? Is this needed?

calculuschild (changes requested):

shared/naturalcrit/codeEditor/codeEditor.jsx
G-Ambatte: Applied in the latest iteration

shared/naturalcrit/codeEditor/codeEditor.jsx
G-Ambatte: Initially it was to skip out if the string was blank, I've tested it and it does not appear to be necessary.

calculuschild:
This seems fine to me now. Any objection to merging @G-Ambatte ?

G-Ambatte:
I've converted to an actual PR instead of a Draft, so it should be ready to merge!

calculuschild:
:cook: 🤌

calculuschild (approved):

PR #2406, opened 2022-09-28
Add sanitization of Style content

### LINKED ISSUE #1437: Sanitize style tab contents
It's possible to execute javascript via the style tab, because we aren't running it through the sanitizer that the main markdown text does. This will need to be fixed before we allow importing custom themes from other brews.

G-Ambatte:
This PR resolves #1437.

This PR adds a sanitization function to the BrewRenderer, which removes any `<script>` or `</script>` tags, in exactly the same way that the Markdown renderer already sanitizes the brew content.


G-Ambatte:
A point of discussion - we could also move the brew content sanitization to the same point in the process, allowing this function to be used for both the user submitted content of both the Style and Content editors.

Gazook89:
Can you correct the spelling in the original “sanatize” function as well?

calculuschild:
@G-Ambatte Yes, let's move sanitization to the same place (in the brewrenderer, before passing to Marked).

calculuschild:
LGTM! Thanks @G-Ambatte! 💥 ✨ 🏅

PR #2417, opened 2022-10-05
Add smarter footer shortcut/snippet

### LINKED ISSUE #2289: [FEATURE] Footnote Auto - Automatically use preceding h1 tag for text
Suggested in the subreddit recently was the idea that using an additional tag - for example, adding the `auto` class to the existing `footnote` element could trigger the use of the contents of the nearest preceding `<h1></h1>` tags.

I believe this functionality would need to be built in to the BrewRenderer, and in some ways may border on Brew Variables, in that it also replaces arbitrary value from the Brew text with a value that is defined elsewhere in that text.

G-Ambatte:
This PR resolves #2289.

This PR adds a new shortcut to the Code Editor, which generates a footer with text based on the nearest preceding header of a user determined H-level, an automatically numbered page number, and a page break,

The PR also adds a V3 base snippet with static footnote text.


G-Ambatte:
From Gazook89 on Gitter:

> [This PR] should just replace the current page number snippet. Currently the snippet just drops 'fancy text here' into that footer space, meaning the user has to click into it to change it. With your PR, it will automatically have something more relevant than "fancy text here" and still take just as many steps to change it if needed.

calculuschild (commented):
The idea behind this PR is fine (generate the footer text from the page contents), although I think it just needs to be a small update to the existing footer snippet instead of a whole bunch of hotkeys.

I would recommend the following approach:

- No hotkeys
- Update the `execute()` function in `snippetBar.jsx` to include the cursor position as a function parameter. This may mean having to pass in the cursor position from `editor.jsx` similar to how we pass down the `brew` into the `<SnippetBar>` component, seen here:

https://github.com/naturalcrit/homebrewery/blob/ce2298ddd05555528f2153e9eac5078749461ff9/client/homebrew/editor/editor.jsx#L316-L326

- Move the `newPageWithFooter()` function into its own file inside `themes/V3/5ePHB/snippets`
- Remove the generation of `\page` -- just make the footer
- Remove the footer generation from the two existing "page number" snippets in `V3/PHB/snippets.js`
- Add a single new snippet that just adds your new footer

5e-Cleric:
A little note, the footnote in the original books relates to the part of the book, so it would make sense to link this with #2654 , the part covers, and go fetch that title, wdyt?

G-Ambatte:
The function now lives as a Snippet, when clicked it defaults to finding the immediately preceding H1 header and injects that header's text into a footer block.
I've also added a sub-snippet menu that allows for the selection of other headers (H1-H6).

All of the shortcut keys have been removed from the editor.

5e-Cleric:
> A little note, the footnote in the original books relates to the part of the book, so it would make sense to link this with #2654 , the part covers, and go fetch that title, wdyt?

I have to remark this now that it is merged, this pr should not link to the nearest h1, if we are looking for accurateness.

Gazook89:
It should link to whatever we use as the header in our Part page snippets. I think…assuming that is what the books do. 

So when adding a Part page, I assume the header type is h1.

5e-Cleric:
I opened the book to show you what i meant, and discovered i was completely wrong, the PHB does use the last h1, my bad.

themes/V3/Blank/snippets.js
```diff
@@ -21,6 +21,11 @@ module.exports = [
 				icon : 'fas fa-file-alt',
 				gen  : '\n\\page\n'
 			},
+			{
+				name : 'Page Break with Footer',
+				icon : 'fas fa-file',
+				gen  : '\n{{footnote PART 1 | SECTION NAME}}\n{{pageNumber,auto}}\n\n\\page\n'
+			},
```
calculuschild, line 28: This new snippet seems a little redundant, given that we already have `New Page` and `Page Number` which will do the same thing.

How about just this set of snippets:

- New Page
- Page Number (remove the footer part of this snippet)
- Auto-incrementing page number (remove the footer part)
- Footer (new in this PR)

Then we don't have so much overlap in these snippets.

themes/V3/5ePHB/snippets.js
```diff
@@ -29,6 +30,49 @@ module.exports = [
...
+			{
+				name        : 'Footer',
+				icon        : 'fas fa-shoe-prints',
+				gen         : FooterGen.createFooter,
+				subsnippets : [
+					{
+						name   : 'Footer from H1',
+						icon   : 'fas fa-dice-one',
+						gen    : FooterGen.createFooter,
+						params : { headerSize: 1 }
```
calculuschild, line 42: Could I suggest this format? Then no need to pass around a params object all the way to the editor and back.

```suggestion
						gen    : FooterGen.createFooter(1),
```

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -165,7 +167,7 @@ const SnippetGroup = createClass({
 	},
 	handleSnippetClick : function(e, snippet){
 		e.stopPropagation();
-		this.props.onSnippetClick(execute(snippet.gen, this.props.brew));
+		this.props.onSnippetClick(execute(snippet.gen, this.props.brew, { ...snippet.params, cursorPos: this.props.cursorPos }));
```
calculuschild, line 170: Can we simplify this to remove the `params`? See my comment in `snippets.js`.

```suggestion
		this.props.onSnippetClick(execute(snippet.gen, this.props.brew, this.props.cursorPos));
```

Along with that, maybe even just pass in all of props if you think future snippets would need access to other snippet properties.

```suggestion
		this.props.onSnippetClick(execute(snippet.gen, this.props));
```

calculuschild (changes requested):

ericscheid:
> If not copied from the nearest h1, what should the text be generated from?

I assume this discussion has the snippet checking the existing brew and looking for the previous `h1` and copying the text from that. Which means if the text of the `h1` subsequently changes then the footer text will no longer be in sync?

For future reference: I'll just drop a x-ref to the [Variables in Brews discussion](https://github.com/naturalcrit/homebrewery/discussions/1400) which might offer a (future) solution.

No action required.

ericscheid:
> If not copied from the nearest h1, what should the text be generated from?

Does this new snippet code clean/sanitise the `h1` text, on the offchance it contains inline html or markup?

G-Ambatte:
> I assume this discussion has the snippet checking the existing brew and looking for the previous h1 and copying the text from that. Which means if the text of the h1 subsequently changes then the footer text will no longer be in sync?

Correct - the current implementation is static, so updating the header text does not update the footer.

> Does this new snippet code clean/sanitise the h1 text

No - the text is copied exactly as it exists between the `# ` and the `\n`.

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: I suspect dropping `this.props` in as a parameter may be overkill for this particular function - but, that said, there's no telling what a future function might require, so it is probably the best way to handle it.

themes/V3/5ePHB/snippets.js
G-Ambatte: I didn't think I could actually use `FooterGen.createFooter(1)` due to the parameters, but I think I just came up with a solution. I will experiment and update the PR if my madness turns out to be genius.

G-Ambatte: I'll push a new commit in a moment, but I've changed the generation function to return a function that takes a props parameter. By passing a parameter (headerSize) to the generation function, it returns a footer function that searches for first instance of that many `#` followed by a ` ` at the start of a preceding line.
It took a bit to wrap my brain around how to pass the `headerSize` parameter from the generation function into the returned function, but it looks to be working in my dev branch now.

G-Ambatte: I've also shifted to passing the entire `this.props` rather than `this.props.brew` and `this.props.cursorPos` as separate parameters. Hopefully having access to the entirety of the props will be useful future snippets.

G-Ambatte:
> No - the text is copied exactly as it exists between the # and the \n.

The latest commit performs a `.trim()` on the retrieved text. Anything more than that is probably best left as an exercise for the end user.

themes/V3/Blank/snippets.js
G-Ambatte: I'll shift these to the Blank theme and shuffle them into better positions in the next commit - hopefully in the next couple of hours.

G-Ambatte: Done!

G-Ambatte:
It occurs to me today that instead of using Regex to find the appropriate header, we could instead use the same approach used [here](https://github.com/naturalcrit/homebrewery/blob/9202f9c8ebed7703f0ea10d1b2e6617c01fe3bb7/server/homebrew.api.js#L106) - by parsing the brew text through the Markdown Lexer and then searching through the tokens for the most recent header (of the user-selected level), we can extract the token contents to the footer.
The main advantage I can think of for this approach is that it might also identify `<h1>HEADER</h1>` as a valid header, which the current approach would not.

G-Ambatte:
> it might also identify <h1>HEADER</h1> as a valid header

This doesn't look to be the case - from the Marked.JS demo page, `<h1>HEADER</h1>` is parsed as a HTML block, not a heading block. However, headings created by code like:

```
Heading
=
```

would be identified as valid headings using the Markdown Lexer approach, while the current approach can only recognize headings created using multiple `#` at the start of the line.

calculuschild:
This is a good point. I think the current function is good enough that I can work on merging it as-is, but feel free to make that change if I don't merge it first. Otherwise a new PR is fine.

G-Ambatte:
By experimenting with the Marked.js demo, I think the code should look something like this:

```
const lexerTokens = Markdown.marked.lexer(text); // <= Need to use appropriate parser for renderer
return (lexerTokens.findLast((lexerToken)=>lexerToken.type === 'heading' && lexerToken.depth === headerSize)?.tokens.map((token)=>return token.text).join('') || 'PART 1 | SECTION NAME');
```

G-Ambatte:
I've changed the approach to use the previously mentioned Markdown Lexer tokens to detect the appropriate header. So far, testing indicates that the new code detects H1 and H2 headers created using `{header}\n=` and `{heading}\n-`, as well as `# {heading}` and `## {heading}` (and `### {heading}` through `###### {heading}`).

calculuschild:
Merging! Thanks @G-Ambatte 💯 🎉 ⭐

Issue #1804, opened 2021-11-05
Create default Brew structure


G-Ambatte:
As discussed in #1801, the processing and checking that a Brew contains all of the required properties - default values where not specified by the loaded brew. Theoretically, this default brew could be created in the `homebrew.jsx` Routes file, merged (assuming `_.merge` or `_.assign`, whichever is appropriate), then passed to all Routes.

Issue #2427, opened 2022-10-08
Saving directly from Home Page does not use V3 renderer


calculuschild:
When a user clicks the blue "Save Now" button from the Home page, which now uses V3, the resulting saved document should also use V3. It appears this is not always the case, resulting in the users getting confused why their beautiful work is getting "scrambled" (turned into Legacy) when they hit that button.

https://www.reddit.com/r/homebrewery/comments/xy9en4/every_time_i_save_it_changes_my_homebrew_around/?utm_source=share&utm_medium=android_app&utm_name=androidcss&utm_term=1&utm_content=share_button

G-Ambatte:
Got it - `homePage.jsx`, the `save` function was ONLY sending the brew's text to be saved.

![image](https://user-images.githubusercontent.com/19826659/194730465-de7ccfda-94e4-4090-8359-e42c87cddd63.png)

This is resolved by sending the current state of the entire brew object to be saved.

![image](https://user-images.githubusercontent.com/19826659/194730491-1087ae75-2253-46de-9eb5-42998c143924.png)

PR incoming shortly.

PR #2429, opened 2022-10-09
Add default brew on load and before save

G-Ambatte:
While investigating #2427, it became apparent that the default brew settings are not being applied in the save function, nor are they being applied at the time that the brew is loaded.

This Draft PR uses the Lodash `defaults` function to apply the properties of a default brew object to the brew object before the data is saved, and also as a final step when the brew is loaded. This function ONLY copies the property from the default when the property is missing or undefined on the source object.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -29,7 +29,6 @@ const splitTextStyleAndMetadata = (brew)=>{
 		brew.style = brew.text.slice(7, index - 1);
 		brew.text = brew.text.slice(index + 5);
 	}
-	_.defaults(brew, { 'renderer': 'legacy', 'theme': '5ePHB' });
 };
 
 const sanitizeBrew = (brew, accessType)=>{
--- a/server/homebrew.model.js
+++ b/server/homebrew.model.js
@@ -46,8 +46,6 @@ HomebrewSchema.statics.get = function(query, fields=null){
 				unzipped = zlib.inflateRawSync(brews[0].textBin);
 				brews[0].text = unzipped.toString();
 			}
-			if(!brews[0].renderer)
-				brews[0].renderer = 'legacy';
 			return resolve(brews[0]);
 		});
 	});
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -9,6 +9,18 @@ const yaml = require('js-yaml');
 const asyncHandler = require('express-async-handler');
 const { nanoid } = require('nanoid');
 
+const DEFAULT_BREW = {
+	title       : 'Untitled Brew',
+	description : '',
+	renderer    : 'V3',
+	tags        : [],
+	systems     : [],
+	thumbnail   : '',
+	published   : false,
+	pageCount   : 1,
+	theme       : '5ePHB'
+};
+
 // const getTopBrews = (cb) => {
 // 	HomebrewModel.find().sort({ views: -1 }).limit(5).exec(function(err, brews) {
 // 		cb(brews);
@@ -66,6 +78,8 @@ const getBrew = (accessType)=>{
 		if(typeof stub.tags === 'string') {
 			stub.tags = [];
 		}
+
+		_.defaults(stub, DEFAULT_BREW);
 		req.brew = stub;
 
 		next();
@@ -130,6 +144,8 @@ const beforeNewSave = (account, brew)=>{
 
 	brew.authors = (account) ? [account.username] : [];
 	brew.text = mergeBrewText(brew);
+
+	_.defaults(brew, DEFAULT_BREW);
 };
 
 const newGoogleBrew = async (account, brew, res)=>{


G-Ambatte:
This would resolve #1804.

jeddai:
I know this is still a draft but it generally looks good to me, haven't run it yet though

G-Ambatte:
I was thinking that the default brew might be better as part of a configuration file, rather than embedded in the source code - it makes sense (to me, at least) that all of the default values of the brew properties are deliberately exposed to be user configurable.

jeddai:
Yeah ultimately I think it would be great with two additions:
* As we explore user settings, users should be able to set their defaults
* Having it be configurable at runtime would be pretty nice because then we can make changes without code changes. This will introduce an additional layer of risk, however.

G-Ambatte:
> * As we explore user settings, users should be able to set their defaults

Agreed - once we have the UI Page merged, a "user settings" type page - probably just `/settings/` can be developed. I think with the latest change @calculuschild requested, the UI Page is content-agnostic enough that it could support such a form.
At this point, this is a system default, rather than a user specific default. However I believe that Lodash's `defaults` function supports additional objects, applied left to right, so `_.defaults(brew, DEFAULT_BREW);` could just become `_.defaults(brew, USER_DEFAULT_BREW, DEFAULT_BREW);`, once a USER_DEFAULT_BREW is actually implemented.

> * Having it be configurable at runtime would be pretty nice because then we can make changes without code changes. This will introduce an additional layer of risk, however.

There is the potential that users that have old brews without any value in the `renderer` property will have their brews rendered in `v3` instead of `legacy` (as recently seen when users saved from the Home page, only `brew.text` was sent to the save function). We could potentially render based on `brew.createdAt` data (e.g. `if(!brew.renderer && brew.createdAt <= SOME_TIMESTAMP) { brew.renderer = 'legacy'; }`).

calculuschild:
Let's leave this default brew in the source code for now just for simplicity....

server/homebrew.model.js
```diff
@@ -46,8 +46,6 @@ HomebrewSchema.statics.get = function(query, fields=null){
 				unzipped = zlib.inflateRawSync(brews[0].textBin);
 				brews[0].text = unzipped.toString();
 			}
-			if(!brews[0].renderer)
-				brews[0].renderer = 'legacy';
```
calculuschild, line 50: Is removing this going to cause old legacy brews to mistakenly render using V3? (although in general I agree with moving all of the "brew cleanup/defaulting" to a central location in homebrew.api.js)

calculuschild (commented):
Ok, so now, can we start to remove/clean up some of the `getDefaultProps` brews in the various pages? If we can keep all of our "default" values kind of centralized here at the point of brew retrieval, we can probably eliminate some potential points of conflict and a lot of code bloat. We have like 8 different versions of "default brew" scattered across the various components of the edit page. 

- [ ] `editPage.jsx`
- [ ] `newPage.jsx`
- [ ] `printPage.jsx`
- [ ] `sharePage.jsx`
- [ ] `brewItem.jsx`
- [ ] `editor.jsx`
- [ ] `brewRenderer.jsx`
- [ ] `metaDataEditor.jsx`, 

I imagine if we are smart about it we should be able to guarantee that at each of these stages, we always have a valid brew object being passed in.

server/homebrew.model.js
G-Ambatte: Short answer is "yes". Anything that was created and not updated since before the `renderer` value was added will use the default value; as this was done before v3 was available, by definition they must be written in legacy and may/likely will break when rendered in v3.
We can potentially differentiate on the `createdAt` or `updatedAt` values, as anything earlier than v3 going live should default to legacy. Alternatively, we could leave the default renderer value on loading as `legacy`, as all recent brews should include the renderer tag with an explicit value.

calculuschild: Right, given that the only case of a missing value should be before V3, let's leave the default as "legacy".

calculuschild: In that sense we use the term "default" in the sense of "values were missing, so we default back to what we guess they should be" rather than "this is what a blank brew should look like"

G-Ambatte: I'm fairly sure that if we alter the "on load" default to something like `_.defaults(brew, loadDefaults, DEFAULT_BREW);` then any properties specified in `loadDefaults` will be - by definition - unable to be absent at the point that the `DEFAULT_BREW` properties are applied.
So `loadDefaults` only needs to be a diff of brew properties for loading vs saving.

I'll try to experiment with it later today.

G-Ambatte:
The only I'm immediately suspicious of is `newPage.jsx`, as all of the others will be passed a brew object from somewhere else. At the moment this PR only targets saving and loading, but `newPage.jsx` will be creating a brew effectively from nothing... I suspect that we could make it call for a default brew.

I was also thinking that there may be some properties that SHOULD be excluded from save defaults, like `views` - as this is only changed by a separate function, it shouldn't be saved by any other DB activity. Same with `createdAt` - it should never be changed.

I'll percolate on it for a while.

server/homebrew.model.js
G-Ambatte: I've experimented with it a bit today, and `_.defaults` only works with properties that are `undefined` - but sometimes `renderer` is returning an empty string instead.
While we can simply use `assignWith` and craft a customizer object that replaces any falsy value with the default... but that would not necessarily work properly with actual Boolean values. I've pushed a commit with a custom `assign` function that is most of the way there, I think.

calculuschild:
@jeddai, @AlexeySachkov Do either of you know how we might automate testing the retrieval of some dummy brews with different missing values to verify this fills in the default values correctly? I can test manually but we've been making pretty frequent updates to the API; I want to have some future-proofing.

jeddai:
💯 yes. I'll update #2548 to branch off of this PR to add tests there if that's okay.

calculuschild:
Perfect.

server/app.js
```diff
@@ -406,6 +407,7 @@ if(isLocalEnvironment){
 //Render the page
 const templateFn = require('./../client/template.js');
 app.use(asyncHandler(async (req, res, next)=>{
+	const brew = _.defaults(req.brew, DEFAULT_BREW);
```
calculuschild, line 410: @G-Ambatte Do we still need this here? Is this just for the case we haven't `get`'d a brew yet? I.e., for new brews?

calculuschild:
I think this is ready... just need to test:

- [ ] When loading brew with `renderer = ''`, properly fills in `legacy`
- [ ] When loading brew with `tags = ''`, properly fill in `[]`
- [ ] When saving from `/edit`, the above item changes are recorded
- [ ] When saving from `/new`, correctly keeps renderer
- [ ] When starting a `/new` brew, defaults to V3, empty description, empty title, "example" style text, empty brew text, unpublished
- [ ] After saving from `/new`, title becomes "Untitled Brew", everything else stays as-is if untouched
- [ ] When visiting `/print`, `/share`, `/edit` `/`, all brews load properly

PR #2440, opened 2022-10-18
Display an Image Preview on URL Hover

Gazook89:
This adds a simple preview of an image when hovering over a URL in the editor.  It only displays for certain URLs, ending in:

- .jpg
- .jpeg
- .png
- .gif
- .webp
- .avif
- .apng
- .svg

In my opinion, this is handy when you start to get a lot of images and you aren't sure which links correspond to which images in the preview window.  Currently the image preview position is just offset from the cursor position-- maybe it's better to have it consistently located in a corner of the editor?  Easy change if that is preferred.

### preview

image

*This doesn't resolve any open issues and there hasn't been any discussion on it, so I'm interested to see if this is helpful or not...if not, i'm not married to it and we can close it out.*


Gazook89:
I'm going to close this PR since it seems there isn't much desire for it, and if that changes it is easy to bring back.

PR #2442, opened 2022-10-19
Add Ability to Remove Brews from Recent Items

### LINKED ISSUE #1918: Deleting a brew should remove it from Recent Items navbar list
Issue raised on [reddit](https://old.reddit.com/r/homebrewery/comments/rllw7s/issue_when_i_delete_a_brew_it_still_shows_up_in/).   After having deleted a brew, the brew link still exists in localStorage and pulls up in the Recent Items list from the navbar and navigates to a blank page.

If fixing this, might also make the small CSS change needed to fix #1881

Gazook89:
This PR fixes #1918 by adding the ability to remove brews from the Recent Brews lists.  There is a 'remove' button that displays on hover over an item.  It pulls up the recent items stored in localStorage, removes the desired brew based on the `<a href=...>` and sends the new list of recent items back into localStorage.    I think it's pretty straightforward without adding anything extraneous. 

image


calculuschild (changes requested):
I like it. What do you think about tweaking the style just a bit?

- Move the button off to the side (or make transparent like on the `brewItems`) so you can still see the item you are clicking on?
- Have some "on-hover" style (again, we can replicate the style used on brewItems)

One last concern; is it obvious that this is just clearing the recent items list and not deleting the file? It works for me but it might confuse some. @ericscheid ?

Gazook89:
I think if it’s moved to the side it will look weird unless also scaled up to fit better vertically…meaning, if the right margin is reduced, the top and bottom margin have to be reduced to match. 

And setting it transparent means the symbol will be more difficult to separate from the text behind it. And since we are already not showing full titles (except in tooltips), covering it while hovered is not really a concern (in my opinion). I didn’t want to make the button too small, either, to avoid the existing text. 

Regarding clear vs delete: because we use the trash symbol elsewhere and the “x” symbol here I think it’s on the right track. But also it would benefit from a title attribute for a tooltip. 

I’m not home until the evening of the 31st though so can’t make any changes myself until then.

calculuschild:
> I think if it’s moved to the side it will look weird unless also scaled up to fit better

What I meant was to have the button extend out the side.

Gazook89:
> (or make transparent like brewItems)

I'm not sure I understood this correctly-- first one is 50% and second is 70% opacity, and third is "on hover" at 100%:
image image image

Here is something similar (70%), but full height:
image

I don't think having the button protrude out the side is appealing.  I started working on it but am going to head to bed now.  But I just can't imagine having a single protrusion sticking out and jumping around as you move your cursor over different items.  For many titles the dropdown is too narrow already, so losing a little bit more only on hover isn't a problem (for me), and obstructing the timestamp also isn't much of a loss, especially since the font color of the timestamp is nigh impossible to read on the :hover background color of the recentItem anyway.

calculuschild:
> Here is something similar (70%), but full height:

image

This one looks good; having the full bar darken also matches the style in the User page, which is what I was referring to: 
image

Gazook89:
Okay I have this updated.  I didn't spot any issues in my local env so it's ready to merge as long as no one else bumps into issues.

Gazook89:
Side note:  I don't know if I understand the "Changes Requested" aspect of this Github PR (in terms using GH correctly).   Normally I see comments on specific lines of code which opens sort of another 'thread', which I can later 'resolve' which closes them up.  But here, there didn't seem to be comments on any particular lines of code.  At the bottom, where it says "1 change requested", I can expand for details and it takes me to the "Files Changed" tab with some lines highlighted --- but it's not clear to me if you made changes here, or just 'marked' them somehow?  Ultimately, is it possible to 'officially' indicate that the requested changes were made and 'resolve' it, like I do with those threads?     Or do i just say "okay all done" and you need to change the status of your review?

calculuschild:
>I don't know if I understand the "Changes Requested" aspect of this Github PR

When you review a PR, you can add comments to any single part of the code, but also Github asks you for an "overall" review status, which can be "comment only", or "accept", or "request changes". Basically, github blocks merging a PR if anyone reviewed with the "request changes" option, or you can set a minimum number of "accept" reviewers before merging is allowed. It's a safety measure to prevent accidentally merging in case something needs to be fixed.

It's something the reviewer has to do to make it green. This is what I see on my screen :

image

I'm not sure what happened here that the highlighted lines didn't give you a "resolve" button, but that's a separate thing from the "1 change requested" status. As long as you fixed the issue and let me know I can change that status once I agree it's fixed.

calculuschild:
That all said, I will test run this PR and then merge. I think it looks good :+1:

ericscheid:
> Regarding clear vs delete: because we use the trash symbol elsewhere and the `×` symbol here I think it’s on the right track. But also it would benefit from a title attribute for a tooltip.

Concur.

calculuschild (approved):

PR #2453, opened 2022-10-20
Change `<meta>` tags based on Current Page

### LINKED ISSUE #2323: Improve Open Graph metadata on home page (and other non-brew pages)
Links to our home page don't make a good impression.

image
image

Gazook89:
This fixes #2323 by adding a function to template.js that determines the current page URL and sets the meta tags per below (pulled from the related Issue):

| data | `homebrewery....com` & fallback | `/share/` | `/edit/` | `/new/` | `/user/` | 
|:---|:---|:---|:---|:---|:---|
| **Site Name** | "The Homebrewery - Make your Homebrew look legit!" | 'The Homebrewery - Make your Homebrew content look legit!' | 'The Homebrewery - Make your Homebrew content look legit!' | 'The Homebrewery - Make your Homebrew content look legit!' | 'The Homebrewery - Make your Homebrew content look legit!' |
| **Title** |  "The Homebrewery" | *brew title* or 'Untitled Brew' | *brew title* or 'Untitled Brew' | "New" | "*username*'s  Collection" |
| **Description** | "Homepage" | *brew description* or `''` | *brew description* or `''` | "Start crafting your homebrew on the Homebrewery!" | 'View my collection of homebrew on the Homebrewery.' |
| **Thumbnail** | d20 logo | *brew thumbnail* or `''` | *brew thumbnail* or `''` | ~ | *user thumbnail* or `''` |
| **Type** | website | article | article | website | website (or profile) |

| data | `/source/` & `/download/` & `/print/` | `/changelog/` | `/migrate/` | 
|:---|:---|:---|:---|
| **Site Name** | ~ | 'The Homebrewery - Make your Homebrew content look legit!' | 'The Homebrewery - Make your Homebrew content look legit!' | 
| **Title** |  ~ | "Changelog" | "Migration Guide " | 
| **Description** | ~ | "Development changelog." | "Take the first steps migrating your brew from the Legacy renderer to the v3 renderer" | 
| **Thumbnail** | ~ | ~ | ~ | 
| **Type** | ~ | website | website | 

\**user description could be a future user-specific field set in Account settings to give a brief description of the user, such as contact info or marketing/promotion/bio info.  Similar deal with user thumbnail*

For reference, here is a Discord embed.  From the top, it is `site name`, `title`, `description`, `image`.  We could also affect the `theme` which would set the color of the tiny left-hand border which is currently dark gray, if we wanted.  Also, @jeddai changed the image size with an earlier PR, but maybe that hasn't gone live yet (i just screenshot this from a discord channel a minute ago).  

image


G-Ambatte:
I haven't tested it yet, but I suspect that these changes might restore functionality to the seldom-used Admin page.

client/template.js
```diff
@@ -1,28 +1,67 @@
...
+		});
+	}
+
+	const ogTags = [];
+	Object.entries(ogMeta).forEach(([key, value])=>{
+		if(!value) return;
+		const tag = `<meta property="og:${key}" content="${value}">`;
+		ogTags.push(tag);
+	});
+	const ogMetaTags = ogTags.join('\n');
```
calculuschild, line 44: I think the idea behind this PR is good, but I'm worried about adding yet another code location where each URL has different behavior. Could we possibly move the different metadata objects into `app.js`, and then in `template.js` the only change is this step you added that goes through a metadata object being passed in via `props`?

calculuschild (changes requested):

client/template.js
Gazook89: Yes I can do that, that sounds much better. Even as I added this as-is I thought “this seems like a lot to put into template.js “

Gazook89: Just to be sure I'm on the right track, is the `home page v3` comment on app.js wrong/a typo?  It should be `home page legacy`, right?   

```js
//Home page
app.get('/', (req, res, next)=>{
	req.brew = {
		text     : welcomeText,
		renderer : 'V3'
	};
	splitTextStyleAndMetadata(req.brew);
	return next();
});

//Home page v3
app.get('/legacy', (req, res, next)=>{
	req.brew = {
		text     : welcomeTextLegacy,
		renderer : 'legacy'
	};
	splitTextStyleAndMetadata(req.brew);
	return next();
});
```

Gazook89: Just pushed a commit on this.  It's not complete, but I'm going to bed and thought I'd push this up in case someone can say "looks good so far" or "wait hold on a minute".  

If it looks okay, I do have a question-- how could I put metatags in the `/source/` page with this method?  It seems /source/ doesn't use the template, and has an empty `<head>` tag.

calculuschild: I don't think source needs a meta tag, since it literally just sends the brew file contents. That is, if you visit the source link the *only* thing you should get is brew contents, free of any external tags or scripts.

Gazook89: You don't think there is value in displaying an embed 'card' when posting a link to a /source/ page?  To me it just seems like any other page on the site.

But leaving the `head` empty is fine with me, too-- less work.

calculuschild: > To me it just seems like any other page on the site.

In this case, it *isn't* a normal web page. It's meant to be just plain text, like opening a .txt file. That should be all you get.

Gazook89: Okay, I just pushed a commit on this that I think is *close*, if not complete.  I've updated the original comment in this PR to reflect the current setup of the meta tags on each page, and there have been some changes from what I originally had.  

I guess there are two parts to review at this point:  the basic structure of my code, and the decisions on what each page should display.

Gazook89:
I have no idea why this fails the `/v3_preview` test......

calculuschild (approved):

PR #2457, opened 2022-10-23
Set Language Attribute of the `.pages` Element

### LINKED ISSUE #1343: Feature request: Allow setting document language for hyphenation
It would be nice if Homebrewery allowed setting document language other than English. All it would do would be placing a `lang` attribute on the appropriate HTML element.

The main use case is hyphenation, which requires the correct HTML `lang` attribute to be set. Specifically: I would like to typeset documents in English and Russian with hyphenation. While I can set `hyphens: auto` in CSS, currently neither language works because there is no `lang` attribute in the document, so the browser doesn't know which language to hyphenate for.

Gazook89:
This pr resolves #1343 by adding a "Language" property in the properties editor.   The input value is then sent to the BrewRenderer component and inserted into the `.pages` element.

The property input field is a datalist that shows the language code + the name of the language in that language.  Adding languages to this list is easy, just add the code and the "name" is automatically added.  There is a very simple pattern validation, and an error message that displays under the input field....this isn't my favorite, since the text is so small and it's on a gray background (so it is difficult to get high contrast), but this can be changed to match whatever is decided on PR #2374 so i didn't spend a lot of time on it.   Here is a picture of it now:

image

The pattern itself could be made significantly more robust, or we could use a new library to actually check to see if the language tag is valid (mentioned in attached Issue).

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -28,7 +28,7 @@ const MetadataEditor = createClass({
 				systems     : [],
 				renderer    : 'legacy',
 				theme       : '5ePHB',
-				lang        : ''
+				lang        : 'en'
 			},
 			onChange : ()=>{}
 		};
@@ -56,8 +56,8 @@ const MetadataEditor = createClass({
 			...this.props.metadata,
 			[name] : e.target.value
 		});
-		console.log(this.props.metadata[name]);
 	},
+
 	handleSystem : function(system, e){
 		if(e.target.checked){
 			this.props.metadata.systems.push(system);
@@ -101,6 +101,32 @@ const MetadataEditor = createClass({
 			.end(function(err, res){
 				window.location.href = '/';
 			});
+	},
+
+	constructLanguageRegExp : function(){
+		const grandfathered = '(' +
+		/* irregular */ '(en-GB-oed|i-ami|i-bnn|i-default|i-enochian|i-hak|i-klingon|i-lux|i-mingo|i-navajo|i-pwn|i-tao|i-tay|i-tsu|sgn-BE-FR|sgn-BE-NL|sgn-CH-DE)' +
+		'|' +
+		/* regular */ '(art-lojban|cel-gaulish|no-bok|no-nyn|zh-guoyu|zh-hakka|zh-min|zh-min-nan|zh-xiang)' +
+		')';
+		const langtag = '(' +
+			'(' + (
+			'([A-Za-z]{2,3}(-' +
+				'([A-Za-z]{3}(-[A-Za-z]{3}){0,2})' +
+				')?)|[A-Za-z]{4}|[A-Za-z]{5,8})'
+		) +
+			'(-' + '([A-Za-z]{4})' + ')?' +
+			'(-' + '([A-Za-z]{2}|[0-9]{3})' + ')?' +
+			'(-' + '([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3})' + ')*' +
+			'(-' + '(' + (
+			/* singleton */ '[0-9A-WY-Za-wy-z]' +
+				'(-[A-Za-z0-9]{2,8})+)'
+		) +
+			')*' +
+			'(-' + '(x(-[A-Za-z0-9]{1,8})+)' + ')?' +
+			')';
+		return  '^(' + grandfathered + '|' + langtag + '|' + '(x(-[A-Za-z0-9]{1,8})+)' + ')$';
+
 	},
 
 	renderSystems : function(){
@@ -225,6 +251,32 @@ const MetadataEditor = createClass({
 		</div>;
 	},
 
+	renderLanguageDropdown : function(){
+		const langCodes = ['en', 'de', 'fr', 'ja', 'es', 'it'];
+		const listLanguages = ()=>{
+			return _.map(langCodes, (code)=>{
+				const languageNames = new Intl.DisplayNames([code], { type: 'language' });
+				return <option value={`${code}`}>{`${languageNames.of(code)}`}</option>;
+			});
+		};
+
+
+
+		return <div className='field language'>
+			<label>language</label>
+			<input type='text' className='value'
+				value={this.props.metadata.lang}
+				onChange={(e)=>this.handleFieldChange('lang', e)}
+				list='languageList'
+				pattern='[a-zA-Z]{2,3}(-.*)?' />
+			<datalist id='languageList'>
+				{listLanguages()}
+			</datalist>
+			<span class='validity'>Must be 2-3 letters, optionally followed by '-...'</span>
+		</div>;
+	},
+
+
 	render : function(){
 		return <div className='metadataEditor'>
 			<div className='field title'>
@@ -269,12 +321,7 @@ const MetadataEditor = createClass({
 				</div>
 			</div>
 
-			<div className='field language'>
-				<label>language</label>
-				<input type='text' className='value'
-					value={this.props.metadata.lang}
-					onChange={(e)=>this.handleFieldChange('lang', e)} />
-			</div>
+			{this.renderLanguageDropdown()}
 
 			{this.renderThemeDropdown()}
 
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -56,6 +56,7 @@ const MetadataEditor = createClass({
 			...this.props.metadata,
 			[name] : e.target.value
 		});
+		console.log(this.props.metadata[name]);
 	},
 	handleSystem : function(system, e){
 		if(e.target.checked){
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -27,7 +27,8 @@ const MetadataEditor = createClass({
 				authors     : [],
 				systems     : [],
 				renderer    : 'legacy',
-				theme       : '5ePHB'
+				theme       : '5ePHB',
+				lang        : 'en'
 			},
 			onChange : ()=>{}
 		};
@@ -261,6 +262,13 @@ const MetadataEditor = createClass({
 				<div className='value'>
 					{this.renderSystems()}
 				</div>
+			</div>
+
+			<div className='field language'>
+				<label>language</label>
+				<input type='text' className='value'
+					value={this.props.metadata.lang}
+					onChange={(e)=>this.handleFieldChange('lang', e)} />
 			</div>
 
 			{this.renderThemeDropdown()}
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -28,7 +28,7 @@ const MetadataEditor = createClass({
 				systems     : [],
 				renderer    : 'legacy',
 				theme       : '5ePHB',
-				lang        : 'en'
+				lang        : ''
 			},
 			onChange : ()=>{}
 		};
--- a/server/homebrew.model.js
+++ b/server/homebrew.model.js
@@ -15,7 +15,7 @@ const HomebrewSchema = mongoose.Schema({
 	description : { type: String, default: '' },
 	tags        : [String],
 	systems     : [String],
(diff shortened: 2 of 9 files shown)


ericscheid:
Perhaps add a link to a resource that explains (in friendly terms) what language codes exist?

Perhaps laid out like 
`Language:   [_________________] _Language codes?_`

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -225,6 +251,32 @@ const MetadataEditor = createClass({
...
+		};
+
+
+
+		return <div className='field language'>
+			<label>language</label>
+			<input type='text' className='value'
+				value={this.props.metadata.lang}
+				onChange={(e)=>this.handleFieldChange('lang', e)}
+				list='languageList'
```
ericscheid, line 270: Anytime an `input` has a `list` attribute, it is usually a good idea to turn off `autocomplete` .. otherwise the UI for autocomplete can also appear, and the input now has two drop-down lists of suggestions. It still works, but it looks very silly.

Gazook89: committed and pushed

Gazook89:
One issue I can foresee is numerous versions of the question "why doesn't changing the language in properties translate my document to that language?".   I'm not sure how to avoid that...maybe changing the label to "Document Language"?   Or maybe we just tackle it if/when it becomes a problem.

Gazook89:
Things left to do:

- [ ] currently the validation pattern just shows an error, but still accepts the value (and saves it to the document).  Likely it should only accept valid patterns.
- [ ] correctly set the default to 'en'
- [ ] confirm choices for the datalist.

server/homebrew.model.js
```diff
@@ -15,6 +15,7 @@ const HomebrewSchema = mongoose.Schema({
 	description : { type: String, default: '' },
 	tags        : [String],
 	systems     : [String],
+	lang        : { type: String, default: '' },
```
calculuschild, line 18: should the default be "en"?

Gazook89: updated.

Gazook89:
This is ready for review I think.

Right now:
1. When creating 'New' document, it defaults the Language input and the `.pages` element to `en`.
2. If you change the input, it updates the `.pages` element and also saves correctly to the `METAKEY` in localStorage.
3. So if you reload the page (or leave and return to the properties editor), it keeps the input value.
4. If you delete the value to an empty string, the `.pages` element will revert the lang attribute to `en`, and the next time you reopen properties editor it will show `en`.
5. Everything on the Edit page works pretty much the same way.

Before merging, might want to confirm the following:
1. What goes in the 'suggested' languages datalist?
2. Are we satisfied with the current minimal pattern?  Basically just requires 2-3 letters, and optionally followed by `-anything`.     Or, I just put together this regex which I think covers the ['common pattern' given on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#language_tag_syntax):

```js
/^[a-zA-Z]{2,3}(-[A-Z]{1}[a-z]{3})?(-(?>[0-9]{3}|[A-Z]{2}))?$/gm
```
[*regex101.com*](https://regex101.com/r/5uqXQ2/1)

3. Finally, I'm just not sure about where to put a helpful link to the BCP 47 / language code reference list.  It could go in small text below the input....?  There could be little 'info-circle' buttons that open tooltips....?  I'm sort of the opinion that having a datalist of the most common languages is going to be enough until proven otherwise.

Gazook89:
i updated the regex to the slightly more strict version I posted above (with a minor change, which I edited that previous comment to reflect).  And I added a couple more language tags suggested by Eric.  Finally, I fixed the autocomplete problem.

calculuschild:
With `en` as the default, the dropdown only lists English and Swedish as options:

image

Similarly, when any other option is selected, the dropdown lists only the current selection:

image

Looks like it's doing more of a "match" (svenska and english both have "en") rather than a selection box. Was very confusing to me, thinking: "why are we only providing English and Swedish as options?"


This has another curious interaction, where typing the first letter of a language triggers both the "autocomplete" match *and* the validator:

image


Last, the dropdown is in a different position when the field is blank. Wherever it ends up, I prefer the dropdown in a consistent location:

image

Gazook89:
> Looks like it's doing more of an "autocomplete match" 

Yeah, it doesn't look like there is method to display all suggested options in datalist.  Datalist is conceptually a part of the autocomplete tag.  I did a bit of searching for how to display all options while still using datalist and didn't see anything, but there is probably a way to do it with js.  Or will just have to go with a js combobox. 

>  I prefer the dropdown in a consistent location:

Again, [after some searching](https://stackoverflow.com/questions/13693482/is-there-a-way-to-apply-a-css-style-on-html5-datalist-options), it doesn't appear that datalists can be given any css styling, so I can't pick a standard position.  There are libraries out there that provide custom inputs, or could just make our own.

> This has another curious interaction, where typing the first letter of a language triggers both the "autocomplete" match and the validator:

Would this bother you if they didn't overlap?  If so, making our own combobox and setting it to appear to the right would be one remedy.   Otherwise, would need to figure out debounce on the field validation.

I have another PR I just started in which I want a combobox, so maybe that's just what needs to be done.

ericscheid:
1.) The html `autocomplete` thing is a different thing from the `datalist` thing. You generally don't want both active on an input, mainly because of the clashing/competing UIs. (I think I might have mentioned this when playing with datalist for tags).

- `autocomplete` is a mechanism for automatically filling a specific value from either frequent usage or from a user profile, possibly offering a choice of same
- `datalist` is a mechanism for providing an extensive list of possible values for an input

One is saying ″here are recommendations specific to you″ and the other is saying "here is the full menu of options". Doing both simultaneously is incoherent.

(BTW, is `autoComplete` a React thing? The html attribute name is `autocomplete`.)

2.) For the validation thing you might want try delaying validation via a debounce so it only kicks in when the user stops typing. Alternatively, allow `[a-z]` while the input has focus, but use the fuller regex on blur.

3.) As to defaulting `en` .. maybe the field should be empty but with a placeholder of simply `en` (i.e. not a narrative description, a literal example value). We can leave the field blank when saving, as the renderer will insert `en` if blank. The datalist itself demonstrates a range of values so we don't need an extended list of examples in the placeholder.

4.) If at all possible, could the input element be shrunk to be closer in width to what the normal range of values will likely be. I'm not sure what CSS would be needed since there's `flex` in play here. Maybe `.metadataEditor .field.language>.value { flex: unset; width: 150px; }`. Doing this means the datalist will be physically constrained as to where it could appear.

calculuschild:
>  Or will just have to go with a js combobox.

I would probably prefer this anyway, since we already have custom dropdown menus implemented in multiple places.

>  I can't pick a standard position. There are libraries out there that provide custom inputs, or could just make our own.

Let's just make use of the dropdowns we already have (see the Themes selector). Making it text-editable should be trivial, and then we can have a more consistent style as well rather than suddenly using datalists in one place.

> Would this bother you if they didn't overlap?

That would help, but I would prefer they just don't appear at the same time if possible. If we are being shown an auto-complete suggestion, we already know the input isn't valid. We don't need both at once.

>Otherwise, would need to figure out debounce on the field validation.

We should probably do this anyway (different PR maybe?). Having an error show up immediately when I'm still typing is a little bit aggressive.

calculuschild:
> (BTW, is `autoComplete` a React thing? The html attribute name is `autocomplete`.)

Since React is using "JSX" rather than HTML, a few properties have slightly different spellings/capitalizations to not clash with existing Javascript names during executing (such as `className` vs `class` for CSS styling) . But when rendered to the DOM it does get converted to the standard HTML attribute.

Gazook89:
> Edit: Ehhh.... maybe the datalist is fine. I just don't like that it hides options when you click on it and there is something in the box already. The autosuggest behavior of the datalist can stay (but turn autocomplete=off as ericsheid says, since that's a different thing). But I sure would like On Click to show the full list like a normal select box instead of limited to matching the current value.

I would imagine that onClick would just clear the current input value, such that the full list then displays.  That's the easiest method I can think of.

Would you consider a library like https://github.com/Codebrahma/React-combobox?

Gazook89:
Here is a gif of a custom react component, a combobox:
combobox

The dropdown list is basically the same as the current Themes dropdown, meaning it is just a div with a collection of div options.  As such, it can be styled with css and the behavior can be changed so that it avoids overlapping the validation warning but I didn't go that far yet.

I made it because I was having trouble at first with just adding an input to the options in a satisfactory way.  Also, I have another idea in mind which would benefit from a combobox, and I thought i would make a generic component to use with that and this.  It's a little mucky right now, like a blind man dancing in a china shop.

ericscheid:
What happens if you type "ch" into the box, hoping to find the language code for Chinese?

Gazook89:
> What happens if you type "ch" into the box, hoping to find the language code for Chinese?

It does nothing.  At least not as setup, but could be added as an option since I don't think that would be tremendously important for this language list.  But if wanting to use for another field like tags as you mentioned, then yes, that would be nifty to replicate the datalist functionality.   

If just looking to keep the datalist, but always show "all" options when initially clicking into the text input rather than the filtered list, then it's probably best to reset the input to an empty string `''` whenever you click into it.   I personally don't see any other way while still using `<datalist>`.

-----
Side note, I'm not sure someone who wants their document set to 'chinese' would type 'chinese'....I would assume they would type in chinese script?  I have no idea.

ericscheid:
The point of the “ch” example is that the language code for Chinese is “zh”.

Gazook89:
Okay so now my custom combobox component can:

1. display a list of items in a dropdown.  The items are just divs defined like the current Languages options in the parent component and passed into this component.  
2. enable or disable auto-suggestions.  This just filters options using an `.includes()` method, which means an input of `e` returns `de, de-ch, es, en, ...`.   It could easily be changed to match only `es, en, ...` or even have that as a sub-option when enabling autoSuggest.
3. enable or disable resetting the auto-suggestions on initial click into input box.    If enabled, when you click into the input box it will show the entire option list in the dropdown.  Thereafter the dropdown will switch to autoSuggest with any text input (but not if selecting an option from the dropdown).

combobox

I have this in a branch from this current PR.  Probably it could have been spun-off from *master* but I wanted to use the Language field as a test.  I'm still screwing around with it before I push it to my remote.

calculuschild:
I like this behavior. :+1: And with this working we should be able to apply any custom styling unlike the native datalist.

ericscheid:
Looks quite good.

Does the auto filter suggest `fr` is the user types `s` ?

(not that anyone would type “s” for French, but it’s a simple example that demonstrates a particular filtering behaviour)

Gazook89:
> Does the auto filter suggest `fr` is the user types `s` ?

I just made a couple changes, so there is an option to do this.  I just published my new branch off of this one, and starting to write my PR (to merge into this, and then hopefully this into upstream/master).

Gazook89:
My new PR is https://github.com/Gazook89/homebrewery/pull/1

It is a PR to merge into this branch, so I think I have to be the one to merge it but let me know if you are able to look at it and let me know.  I just didn't want to get too deep into it before merging into what is already in *this* branch.

calculuschild:
@Gazook89 I looked over your other PR. Looks stable enough to have you go ahead and merge it in here. We can try to push this through before it gets too much further behind Master.

Gazook89:
I got my other PR merged into here with the combo-box, and I caught this up to *master*.

It's been a long time since I really looked at this, so i'm not sure where I left off-- if it was at the 'end' or somewhere in the 'middle'.  Though, it appears to work in my browser with a tiny amount of testing.

Gazook89:
One issue i'm noticing is that an invalid value is still kept, even though it flashes the error.  I think likely the best route is turn invalid values to `en` as a default.

Gazook89:
> As @ericscheid says, we should probably debounce the validation (different PR maybe?). Having an error show up immediately when I'm still typing is a little bit aggressive.

I spent some time trying to figure out how to do this, and couldn't get it.  I was using `_.debounce` and adding it to the `handleFieldChange` method, but it creates a new debounce method on every *input*/*keystroke* so it doesn't actually do anything worthwhile.  I also tried in a few other places, but never quite the result we want.  Some stackoverflow answers suggest it has to do with the shadow DOM and suggested react hooks or other options, but I'm just not familiar enough with the concepts that I can just put them into this PR. I think this would have to be another PR, and possibly at some point rework our input elements to work better with validation.

calculuschild:
Yay! Thanks @Gazook89 💯 🎉 ⭐

PR #2469, opened 2022-10-28
update twitter:card meta tag to have "summary" as the content

jeddai:
Updates the `twitter:card` meta tag's content to be `summary`, enabling the thumbnail image to be inline with the embed.

image


calculuschild:
Looks good! 🎉

PR #2470, opened 2022-10-28
update getBrew usages to not fetch google brew during updates

jeddai:
This updates `getBrew` to have an optional boolean parameter that defaults as true. When set to false, it will prevent the method from fetching the google half of a brew. This is only used during update calls right now, but we could consider additional methods as they cause issues.


calculuschild:
Does updating a brew need to fetch the Google at all? Don't we just overwrite all of the relevant properties anyway with that initial "_.assign" step?  `	let brew = _.assign(req.brew, excludePropsFromUpdate(req.body));`

The only properties we end up sending through the update in `googleActions.js` are items that we already have locally, I think:

https://github.com/naturalcrit/homebrewery/blob/7d6a3da2e2cd473777d0b0435397fa7fa7b880b3/server/googleActions.js#L134-L155

jeddai:
Hmm we could do that by replacing the call to `save` on the brew with an `updateOne` with a query for the brew id and the user as an author? That would complicate the case of a stub not existing yet, however. If you think the benefits are worth the tradeoffs I'm happy to test that change.

calculuschild:
Hm, that could save us another transaction I suppose.... But you are right it seems it would introduce more edge cases we need to consider, like nonexistent stubs or no-authors.

How about we just stick with the current "findOne" for now.

jeddai:
Sounds good to me. I'll handle it in the authorship PR.

server/homebrew.api.js
```diff
@@ -234,30 +237,31 @@ const updateBrew = async (req, res)=>{
...
-
-	if(!brew.markModified) {
-		// If it wasn't in the database, create a new db brew
-		brew = new HomebrewModel(brew);
+	// we need the tag type change in both getBrew and here to handle the case where we don't have a stub on which to perform the modification
+	if(typeof brew.tags === 'string') {
+		brew.tags = [];
 	}
 
-	brew.markModified('authors');
```
jeddai, line 246: I didn't see a reason to keep markModified, it functions without it. If there is one please feel free to share!

server/homebrew.api.js
```diff
@@ -43,9 +43,12 @@ const getBrew = (accessType)=>{
 				}
 			});
 		stub = stub?.toObject();
+		if(stub?.authors && !stub?.authors.includes(req.account.username)) {
+			throw 'Current logged in user does not have access to this brew.';
+		}
```
jeddai, line 48: Oh I went ahead and wrapped the initial authorship piece into this PR. This will prevent any unwanted edits.

jeddai: Actually I'm backing this out and putting in its own PR so we can release it separately if we want.

server/homebrew.api.js
calculuschild: I think you are right. markModified is only needed if Mongoose can't naturally detect that a field was updated (dates or untyped arrays), but "authors" and "systems" are both typed to arrays of strings in the Mongoose Schema so that should be fine.....

server/homebrew.api.js
```diff
@@ -234,30 +234,31 @@ const updateBrew = async (req, res)=>{
...
 		brew.authors = _.uniq(_.concat(brew.authors, req.account.username));
 	}
-
-	// Fetch the brew from the database again (if it existed there to begin with), and assign the existing brew to it
-	brew = _.assign(await HomebrewModel.findOne({ _id: brew._id }), brew);
-
-	if(!brew.markModified) {
-		// If it wasn't in the database, create a new db brew
-		brew = new HomebrewModel(brew);
+	// we need the tag type change in both getBrew and here to handle the case where we don't have a stub on which to perform the modification
```
calculuschild, line 237: When would this ever occur? The only case I can think of is opening a Google Brew that was never stubbed. In that case, we have never saved tags to Google Brews (stubs came before tags we implemented). I propose instead we modify the `getGoogleBrew()` function in `googleActions.js` so it just.... doesn't try to retrieve the tags field. Then we don't have to worry about doing this conversion here.

jeddai: Ah that's a good point! Yeah I'll make that change, good call.

server/homebrew.api.js
```diff
@@ -27,7 +27,7 @@ const getId = (req)=>{
 	return { id, googleId };
 };
 
-const getBrew = (accessType)=>{
+const getBrew = (accessType, fetchGoogle = true)=>{
```
calculuschild, line 30: Can we change the phrasing here from `fetchGoogle` to `stubOnly` or `metadataOnly` or something? Somehow that seems more accurate to what we want to do here (i.e., are we retrieving a complete brew for display, or are we just looking up a few brew properties so we can update/etc.).

Then, in a later PR we can even further reduce our bandwidth usage by making `stubOnly` mode not retrieve brew text contents, which is the majority of a brew's file size. That could save us a lot of errors for users with big brews that end up timing out after repeated autosaves takes up all the traffic.

calculuschild (changes requested):
I think just changing `fetchGoogle` to a more generic "whole brew or partial brew" name is the last thing I see. I can also do it myself but want to see if it sounds reasonable to you first.

jeddai:
Yep! I'll get that change made today!

jeddai:
Done and ready for a final review

server/homebrew.api.js
```diff
@@ -59,14 +59,14 @@ const getBrew = (accessType)=>{
 		}
 
 		// If after all of that we still don't have a brew, throw an exception
-		if(!stub) {
+		if(!stub && !stubOnly) {
 			throw 'Brew not found in Homebrewery database or Google Drive';
 		}
```
calculuschild, line 64: Can you help me understand this change?  If we can't find any record of the brew (no stub and nothing from Google Drive), then shouldn't we always throw here whether `stubOnly` is true or not?

server/homebrew.api.js
```diff
@@ -235,29 +235,26 @@ const updateBrew = async (req, res)=>{
...
-			console.error(err);
-			res.status(err.status || 500).send(err.message || 'Unable to save brew to Homebrewery database');
-		});
 	if(!saved) return;
 	// Call and wait for afterSave to complete
 	const after = await afterSave();
 	if(!after) return;
 
-	res.status(200).send(saved);
+	res.status(200).send(brew);
```
calculuschild, line 257: Theoretically, `brew` and `saved` should be the same, but I think we still want to send `saved` because `editPage.jsx` uses that result for a couple things.

jeddai: So saved is either a brew, or an update result, using brew makes sure it's always the brew. updateOne doesn't return the brew (idk there might be a way to change that, lemme Google)

server/homebrew.api.js
jeddai: If there's no stub and it's only getting the stub that handles Google drive pre-stub updates

server/homebrew.api.js
calculuschild: Ah, so looking at `editPage.jsx`, at least one reason we need `saved` is if we are saving an unstubbed Google brew, `saved` will contain the newly-minted `editID` and `shareID`, which will be needed for subsequent saves.

We may need to use `findOneAndUpdate()` to get a return value , using the `options.returnDocument='after'` option to get the resulting document: https://mongoosejs.com/docs/api.html#model_Model-findOneAndUpdate

jeddai: So from what I can tell findOneAndUpdate returns the doc before it was modified https://mongoosejs.com/docs/tutorials/findoneandupdate.html#getting-started

calculuschild: Yes but if you look down a little further you can set an option to return it after modifying.

jeddai: Oh cool! Okay yeah I'll do that. Sorry I'm on my phone right now 😅

server/homebrew.api.js
calculuschild: Hm.... OK, that makes sense. After this PR I might go in and rearrange some logic further up in this function because something in the back of my mind thinks there is something weird going on but I can't put my finger on it.

jeddai: There might be 🤷 any extra eyes would be greatly appreciated. This is why I want to unit test it 😅

calculuschild: Yes we definitely need better testing on this project....

PR #2474, opened 2022-10-28
[WIP] User Info

### LINKED ISSUE #1908: User settings
## Editor Settings

We've added a number of editor-related changes recently, and I think it would be beneficial for users to be able to modify certain editor settings and it be persistent in the application. A super easy solution there would be to have them stored in local storage and loaded when the editor is loaded.

These could be in an `object` type `editorSettings` value on the user schema. This would allow for a more gentle-handed approach to its schema as new settings are added.

As some examples of things that we might want to include in the editor settings, we could include the following:

- enable/disable active line highlighting
- enable/disable code folding
- enable/disable line numbers?
- change tab size
- option to indent with tabs vs spaces
- enable/disable auto-closing html/curly brace tags
- enable/disable showing trailing spaces
- editor theme?

Any other options/ideas people have are welcome!

G-Ambatte:
This draft PR is to demonstrate the progress of adding a new UserInfo collection to the current NaturalCrit database used by Homebrewery.

This PR will resolve #1908 when complete.

Current goals:
- [x] add last user activity to UserInfo
  - [x] create user with default properties if not already present
- [x] add ability to look up UserInfo to Admin page
  - [x] fix Admin page
- [ ] add user options to global object
- [ ] move Recent Brews data storage to UserInfo
- [ ] user awards?
- [ ] ???


dbolack-ab:
Looking at this, I am curious about the unchecked boxes. Can we expand those to show what information is going in there?

dbolack-ab:
I think a "Display Name" would be nice as well.

5e-Cleric:
This is 2840 commits behind master, with last change being over a year ago, this PR has been sitting so long it would need to start from scratch anyway.

Record of this is tracked in issue #1908

PR #2475, opened 2022-10-28
Update thumbnail display

lucastucious:
There is a proposition to use thumbnail in a better way

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
@@ -6,12 +6,13 @@
 	box-sizing                  : border-box;
 	box-sizing                  : border-box;
 	overflow                    : hidden;
-	width                       : 48%;
+	width                       : 310px;
+    min-width                   : 200px;
 	min-height                  : 105px;
 	margin-right                : 15px;
+	margin-left                 : 15px;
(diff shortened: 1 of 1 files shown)


calculuschild:
Thanks for the PR!

Personally, I'm not sure this is the route we want to go since it makes each brew item so much taller so only a few items fit on the page. In the future, we may add some menu to customize the appearance of your user page, but until then I think the current style is easier to sort through to find what you need.

Thoughts @Gazook89 @G-Ambatte @ericscheid @jeddai ?

lucastucious:
My only issue with the current display is that almost all my brews have a thumbnail and a "long name so the brew name is difficult to read

G-Ambatte:
I believe there has been previous discussion about potentially adding a similar kind of UserPage styling behind a switch, with the value stored in local storage, probably cycling through with a changing FontAwesome icon:

image
image

I don't anticipate that it would be difficult functionality to implement, especially as UserPage is already retrieving data from local storage when loaded. It _should_ be relatively simple to add an additional variable and a function to switch the applied style.

Of course, once UserInfo is complete, the variable can be added to the list of things to be moved from local storage to the centralized collection.

G-Ambatte:
I've punched out a functional version now and put it in Draft PR #2476.

calculuschild:
Closing in favor of #2476

Issue #2405, opened 2022-09-28
[UserPage] Make UserPage theme agnostic


G-Ambatte:
### Your idea:

As Homebrewery introduces more themes than just straight 5E PHB, having other parts of the website - like the UserPage - use the 5E PHB style becomes inappropriate. I suggest - once Themes are live - we alter the UserPage style to a theme-agnostic style, possibly mimicking the new UI Page style.

### Please confirm:

- [X] I have searched the Issues tracker for any duplicate requests and found none.

dbolack-ab:
Are you thinking of pushing it down to Blank or a distinct new theme?

dbolack-ab:
@G-Ambatte Would the Vault/Userpage merge solve for this or is that another step needed after?

G-Ambatte:
> Are you thinking of pushing it down to Blank or a distinct new theme?

I believe the intention was to use a style similar to the Error Page, so that it is quite clearly different from the 5EPHB Theme.

Vault has it's own styling that already makes it quite different from the User Page, so this change might not significantly affect Vault.

5e-Cleric:
Related, i made small changes to the Properties Tab, does anybody like it?

<img width="534" height="593" alt="Image" src="https://github.com/user-attachments/assets/13fecd14-17d8-46b5-969f-ac0603f28853" />

ericscheid:
> Related, i made small changes to the Properties Tab, does anybody like it?

1. you made the section headers clearer 👍 
2. you lightened the background 👎   
.. so the visual of the tab being attached to the Properties btn as a tab is gone. <img width="208" height="76" alt="Image" src="https://github.com/user-attachments/assets/9a4d478b-336e-4e3b-96ee-510f306515dd" />  
(not a biggie, the other editors don't have that "tabbed" UI look)
3. anything else?

Overall: 👍

Gazook89:
I do like the lighter color, but guaranteed there will be grumblings from users who have set everything they can to "dark".  

Can the properties editor inherit the color scheme of the codemirror editors?  Without looking, i'm not sure if those editor color values are reachable by the properties editor.  Before it comes up, I would prefer matching to the other editors over any notion of Properties editor styling being chained to a broader site-wide "dark mode/light mode" toggle (which I otherwise support).

5e-Cleric:
> I do like the lighter color, but guaranteed there will be grumblings from users who have set everything they can to "dark".  
> 
[...]

PR #2476, opened 2022-10-29
[UserPage] Update thumbnail display - with switch function

G-Ambatte:
This Draft PR builds from the styling of @lucastucious' PR #2475 and incorporates a switching function on the User Page.

- [x] Implement the new style as an additional class in LESS
- [x] Add ListPage functionality to change CSS class on BrewItems
- [x] Save and load state to/from local storage


lucastucious:
Maybe we could improve the userpage. No need to be a "page" (215*279) and could be a responsive one. So large screen can display more BrewItem

Gazook89:
IMHO, there should be a 'list' view that is more 'utilitarian' in purpose...shows more brews in a smaller space in a uniform way.  It displays Title, created date, most recent update, views, published status, save location, pages, etc all at once.  This would be the default for viewing your own brew collection.

Then the other view is our 'cards' view, but with an eye towards 'showcasing'.  Meaning that information like page views, page count, and even Description are mostly hidden until revealed.   And things like Title, Thumbnail, and Tags are made the focus.  This card view would be the default for when viewing another users account, though the option to switch to List would be available.

Both options should be paginated, only displaying 50(?) brews at once.

I don't think there is a compelling reason to why we separate Published from Unpublished brews, either.  Maybe before we had the better filters/sort options it made sense, but ultimately when I sort my brews alphabetically I don't want to have to look through two lists if I'm not sure on it's published status.   Published status should just be added as another filter option and otherwise just throw it all together.   This would simplify the design considerations of the brew collection page.

Perhaps it would be useful to incorporate the "Recent" brews into the brew collection as well, following the above formats but in it's own spot (similar to how we have Published and Unpublished separated now).  Though I'm not entirely sold on this.

Gazook89:
Because i'm tired of vscode, I mocked up my vision for the "card" format of the brew collection, below.  These would be over the blue background, *without* any sort of `.page` element.  They would just be arranged either with Flex and `wrap` to responsively adjust to the screen real estate, or setup as some sort of carousel.

image

It shows some functionality we don't have, such as bookmarking and favoriting.

lucastucious:
I really like your mockup, but what about description length? Do we make a description character limit or trim the display?

ericscheid:
> Maybe we could improve the userpage. No need to be a "page" (215*279) and could be a responsive one. So large screen can display more BrewItem, with more columns

Just a quick note: the user "page" has a width, but is not limited in length.

ericscheid:
> I don't think there is a compelling reason to why we separate Published from Unpublished brews, either. Maybe before we had the better filters/sort options it made sense, but ultimately when I sort my brews alphabetically I don't want to have to look through two lists if I'm not sure on it's published status. Published status should just be added as another filter option and otherwise just throw it all together. This would simplify the design considerations of the brew collection page.

The compelling case is that it clearly shows what the non-owning user will see when they visit your userpage, and as the owning user you don't want to be making any mistakes in that regard. Think of it as a WYSIWYG approach.

My list of brews is quite long, and I've long stopped manually scrolling/scanning for the brew I want. I just use the filter option (and before that the browser's in-page-search).

Gazook89:
I'm not convinced that published status needs to be separated out for that reason.  The 'pro' is as you say-- a method of checking what visitors will see when they come to your page.  The 'con' is that every time you visit your own collection you are sorting through two different lists of what is essentially the same thing.  I personally almost *never* visit another users profile, but visit my own possibly every time I use the Homebrewery (*and everyone knows that personal anecdotes make for the most compelling evidence...*).

Also, as in your case with the large amounts of brews, you aren't even using the separated collections anyway, preferring to just use the Filter.   So for a user with many brews a doubled collection likely isn't much help since they are just filtering.  And a user with fewer brews has less to assess.    As long as Published Status is still visible either with an icon similar to how we display Google Drive status or alternative visual signal (colored border perhaps?), I don't think much is lost compared to what is gained.  

And in the end, adding a "Published Status" filter or sort option would be the final 'nail in the coffin'.

ericscheid:
True, not using published/not for searching. 

But am using the display as reinforcement/reminder of what I have published/not (without having to parse out some buried detail from each brew, having scanned the entire list of brews).

If I was manually scrolling/scanning for a brew, I'd probably remember if the brew I wanted is completed and finished, vs being in some draft state — which for me pretty much informs whether the brew is published or not. Thus, if I'm wanting to add some notes to a draft/in-progress brew then I know I should be looking in the Unpublished pile. If I have a correction or update to a brew I've finished then it is probably in my Published Brews.

Pretty much no-one visits another user's profile at the moment, but mainly because we don't particularly facilitate or promote that activity. And for a while doing so would be disappointing because any published gdrive brews wouldn't even show up (even now, at least until stubs get created by authors of the brews).

Masonry layouts can be sorted, but reading the sorted order can be a bit of an effort.

Gazook89:
So I'm hearing "if I browsed all my brews by scrolling through them, I would want to do so by published status.  But i don't scroll through all my brews".   This could easily be done by adding Published status as a filter, right?

image

And "Published" != "Finished".....If I have brews that I consider complete but just don't really want public, it would complicate that mental processing to think "Well, I need to update this brew that I think is finished, but i'm not sure if it was published or not?".   But again, just being able to filter by published status in the search bar would take care of it.

Gazook89:
Here are the two different views as I imagine them...first with the card view, unchanged from above besides proportions a little bit and using the same "hover" setup etc:

image

The second is the List view.  See more brews, but without the thumbnail.  The "Info" button may either expand the row to show the description, or just show the description in a tooltip on hover/click (my vote is for the latter).   I wouldn't be surprised if there was criticism about 'crowded tags' here....Possibly the allotted space for them could be expanded if they started right after the title?  Or, same width but just make it only a single row and show fewer tags.  As with the Cards, tags that are not displayed can be revealed by clicking the '...' tag.  On hover, each row background lightens a smidgen, and displays the toggle buttons for 'starred, bookmarked, published'.

image

G-Ambatte:
I've pushed an initial pass at a new view type - List

image

When all groups are collapsed, it reverts to the original styling, so it's still WIP right now.

G-Ambatte:
This is all just styling right now, so the brew groups are unchanged.

Gazook89:
I can also help with the css if you’d like, too— I could create a branch off yours or just contribute it via comment to this branch. I don’t know if you enjoy doing css stuff or not.

G-Ambatte:
> I can also help with the css if you’d like, too— I could create a branch off yours or just contribute it via comment to this branch. I don’t know if you enjoy doing css stuff or not.

I'm no CSS wizard, I can make the code add classes when the button cycles, but much of my CSS is where I've played with it in the Inspector until I think it looks good, then copy-paste to the LESS. So any assistance with styling would be appreciated.

I am thinking that for the List view, the underlying data needs to change - I've hidden things like `description` with `display: none`, but I can't make a HB icon appear for locally stored brews when there's no element to attach to (because the icon only appears for Google-stored brews).
I don't think that it would necessarily be a bad thing to have two versions of the icons (true/false), as the same amount of visual space is occupied by each icon and thus the layout will always be consistent. One of the challenges right now is that the icons in `info` are currently in a `flexbox` to distribute them evenly, but this does not give rise to consistent columns, as there may be as few as three and as many as six elements.

5e-Cleric:
This could be the future UI for searching brews, letting the user change from list view to table view.

5e-Cleric:
This PR is pretty much forgotten, isn't it? i'll take a look, see if i can improve the UI and bring an update later today.

G-Ambatte:
> This PR is pretty much forgotten, isn't it?

Not forgotten, just simmering in the background. This PR is part of the reason I added a BrewItem icon for brews in the MongoDB database.

5e-Cleric:
Going to try implementing a simmilar UI in #3263

dbolack-ab:
is anything in particular blocking this?

5e-Cleric:
> is anything in particular blocking this?

pretty much forgotten, i could take care of this after i close #2586 if no one beats me to it

Gazook89:
I think there was light murmuring about what really distinguishes a User Page from the Vault with a certain author filter.

ericscheid:
- user page will have meta-navigation by folder
- vault page has meta-navigation by pagination

- user page has sort-by and filter tools
- vault (currently) does not (which don't work too well with paginated collections)

dbolack-ab:
While I understand that Vault and the User Page currently do not act like this, can anyone explain why the User Page should not simply be an implementation of the Vault page with a forced user filter ( current user ) and extra icons for brew operations?

calculuschild:
The Vault page is actually an inplementation if the user page with added searching and pagination.

5e-Cleric:
However #4354 goes, this is not that PR, this is about userpage styles being changed into a more modern approach. Unless we can reach an agreement over the next week, we should close this.

dbolack-ab:
Been a week...

ericscheid:
> The above image is just a very quick mockup--- i think there would actually need to be something that visually distinguishes 'sorts' from 'filters', and 'published status' would be a filter. Other filters would be tags, GDrive status, etc. And maybe those filters are just applied by inserting them with some syntax into the 'filter/search' input....similar to how Github does it.

Expecting authors to figure out a syntax for the filter field might not work well.

As a possible solution to how to separate sorts from filters:

`Sort by: Sort by : Title • Created Date • Updated Date • Views • 🔍 [filter title/description/tags] X Y Z`

.. where X, Y, and Z are toggle buttons for filter by storage-location (greyed → d20 → gdrive → greyed), filter by published, filter by favourited, filter by bookmarked, etc.

icons for filter options, positioned after the filter field

5e-Cleric:
This PR is stale and needs a restart, can be logged as part of https://github.com/naturalcrit/homebrewery/issues/2405 or be logged into its own issue. Closing

PR #2501, opened 2022-11-14
Reg exp crash in curly syntax

### LINKED ISSUE #2393: Mustache Div regex has exponential time for each CSS tag
This line slows to a crawl when trying to type a space and then another character when there are 6+ separate CSS properties in a mustache div:

https://github.com/naturalcrit/homebrewery/blob/6ade9925d766ce592514d7ee1f3bebe521a417b2/shared/naturalcrit/markdown.js#L94

image

Something with this regex seems to be causing exponential slowdown with every added CSS property (five is instantaneous, six takes 1s, seven takes 5s, 8 takes 20+ seconds)

https://github.com/naturalcrit/homebrewery/blob/6ade9925d766ce592514d7ee1f3bebe521a417b2/shared/naturalcrit/markdown.js#L85

Gazook89:
Fixes #2393 by applying the recommended change to the regexp patterns for curly blocks, spans, and injection.  

It can be merged as-is (assuming no issues when reviewing) or there are some other changes that could be made:

1. Move the common 'guts' of the pattern out into a constant (*string*) at the top of Markdown.js and insert it when needed between whatever bracket setup is needed for each type of curly token.
2.  Adjust the pattern to allow for css properties that utilize `spaces` in their values such as `box-shadow:1px 2px 0 black`, maybe by allowing `'` in the values.
3. Add styling in the Editor for Injection syntax.


calculuschild:
I'll give it a test run but it looks good at first glance.

> 1. Move the common 'guts' of the pattern out into a constant (*string*) at the top of Markdown.js and insert it when needed between whatever bracket setup is needed for each type of curly token.

Yes, this is possible, by putting some keyword in the regex like `/(.*someregex)KEYWORD.*/`. I *think* you have to convert the RegEx to a string first (regex.source or something), then you can do a string.replace on the keyword with whatever regex you want to inject, and finally convert back into regex (`new RegExp(string)`).

*but*, make it a separate PR. I'm really liking keeping the PRs small and minimal.

2.  Adjust the pattern to allow for css properties that utilize `spaces` in their values such as `box-shadow:1px 2px 0 black`, maybe by allowing `'` in the values.

We already allow for this using quotes: `box-shadow:"1px 2px 0 black"`

3. Add styling in the Editor for Injection syntax.

Let's leave for a separate PR

Gazook89:
> We already allow for this using quotes: box-shadow:"1px 2px 0 black"

Oh, i'm just so used to single quotes.  Noted!

> Add styling in the Editor for Injection syntax.

I just got this working so i'll open a new PR.

calculuschild:
I made a small change to the RegExs because they would crash if the curly braces were empty. Seems to be all working now :+1:

PR #2503, opened 2022-11-14
Add syntax highlighting for injections

### LINKED ISSUE #1670: [v3]  Additional markdown highlighting for injection syntax
I don't have strongly formed thoughts about what it should look like, but I think we need to add some markdown highlighting for injection brackets, *and/or* for image syntax?   Here is an example of something that just doesn't stand out from the text:

image

I don't have strong feelings about this, so feel free to dismiss and close.   But it bothered me just enough to raise it as a possible issue.

Gazook89:
*This is a branch of PR #2501 (which I'm now realizing it likely didn't need to be...it should have branched directly from master)*

It fixes #1670 by giving the curly injection syntax a green color in the Editor.   I thought Orange would be too close to the red used for Spans, and I don't think we are using green yet?  Maybe there should be some plan for colors but if CodeMirror themes are ever added it'll be a moot point.

I do have concerns that I have unnecessarily doubled up on code here where I might not have had to.

image

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -139,7 +139,7 @@ const Editor = createClass({
 
 						
 						if(line.includes('{') && line.includes('}')){
-							const regex = /{(?:(?=(:(?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"':{}\s]*))\1)*}/g;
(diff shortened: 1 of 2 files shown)


calculuschild:
@Gazook89 I had to make some tweaks to the regex on #2501 . Can you update this to use the same syntax? Should be a matter of just merging with master and then copying over the updated regex for the injector highlighting.

Gazook89:
Oh duh, I could remove the bracket count piece…..

PR #2507, opened 2022-11-16
Move common meta tags to default object

### LINKED ISSUE #2012: Document Social Embedding
When document share urls are linked from various social platforms, an embed should be able to be obtained from the Homebrewery site. Embeds are typically handled by document head attributes, primarily opengraph values.

My proposal includes a few things:
- Fix https://github.com/stolksdorf/vitreum/blob/master/headtags.js
    - if you have multiple Meta objects it will duplicate them, which will break the embedding
- A default Homebrewery embed for the main page, v3 preview, and migrate pages
- Dynamic Homebrewery embeds for document share pages
- imageUrl document attribute that will display as the image for embeds

Example:
image

### LINKED ISSUE #2497: OG Metatags have duplicate entries for each route
Possible opportunity for code cleanup. Following #2453 each route has its own OG Meta tags, but many of them are duplicated. May be worth creating a "default" object that each route can modify with its one or two unique values.

Gazook89:
This PR is meant to clean up some duplicate lines introduced by the first change to OG Meta Tags, and resolves #2497.  

Also, it adds metatags to the new Account page.

This + original PR for metatags  resolves #2012 as well.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -376,10 +376,9 @@ app.get('/account', asyncHandler(async (req, res, next)=>{
 	req.brew = data;
 
 	req.ogMeta = { ...defaultMetaTags,
-		title       : `${req.params.username}'s Account`,
+		title       : `Account Page`,
 		description : null,
 		image       : null,
-		// type        :  could be 'profile'?
 	};
 
 	return next();
--- a/server/app.js
+++ b/server/app.js
@@ -149,7 +149,7 @@ app.get('/changelog', async (req, res, next)=>{
 	req.ogMeta = { ...defaultMetaTags,
 		title       : 'Changelog',
 		description : 'Development changelog.',
-		thumbnail   : null,
+		thumbnail   : null
 	};
 
 	splitTextStyleAndMetadata(req.brew);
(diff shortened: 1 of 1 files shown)


G-Ambatte:
The `/account` page doesn't show a specific user's account information, it shows the account information of the currently logged in user.

server/app.js
```diff
@@ -378,6 +374,14 @@ app.get('/account', asyncHandler(async (req, res, next)=>{
 	}
 
 	req.brew = data;
+
+	req.ogMeta = { ...defaultMetaTags,
+		title       : `${req.params.username}'s Account`,
```
G-Ambatte, line 379: This isn't account information for a specific user, but rather the account information of the currently logged in user, so I think this title needs to be updated.

Gazook89:
This + original PR for metatags would resolve #2012 as well.

server/app.js
Gazook89: Oh good catch. I just changed it to something generic, "Account Page".

calculuschild:
Looks good! Merging! :tada: :100: 🤌

PR #2509, opened 2022-11-17
Disable changes from non authors

jeddai:
After fetching a brew stub, if the authors list doesn't contain the logged in user we throw an error.

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -43,6 +43,9 @@ const getBrew = (accessType, fetchGoogle = true)=>{
 				}
 			});
 		stub = stub?.toObject();
+		if(stub?.authors && !stub?.authors.includes(req.account.username)) {
+			throw 'Current logged in user does not have access to this brew.';
+		}
 
 		// If there is a google id, try to find the google brew
 		if(fetchGoogle && (googleId || stub?.googleId)) {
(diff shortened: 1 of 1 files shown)


server/homebrew.api.js
```diff
@@ -43,9 +43,12 @@ const getBrew = (accessType)=>{
 				}
 			});
 		stub = stub?.toObject();
+		if(stub?.authors && !stub?.authors.includes(req.account.username)) {
+			throw 'Current logged in user does not have access to this brew.';
+		}
```
calculuschild, line 48: `getBrew` is also used to retrieve brews for sharing. This will block non-authors from viewing a share file.

jeddai: Just added an addition to this if to make sure the accessType is 'edit'!

ericscheid: What happens if an anonymous user requests an edit URL? Does `[req.account.username]`produce an empty string or is it an invalid reference?

jeddai: Oop lemme check

jeddai: Yep that was an issue, fixed in https://github.com/naturalcrit/homebrewery/pull/2509/commits/6b8db74a2b51a17cbbac378ca99c0429f05f7aa3

PR #2523, opened 2022-11-28
Add link to instructions to refresh Google creds

### LINKED ISSUE #2520: Account page show current Google auth status
### Your idea:

The Account page only shows if the user has a Google ID linked to their account. It may be handy to also display if the connection is currently working. We get a lot of users who have used Google Drive in the past and the credentials expired, etc., so if we tell them to check their account page this could be handy.

- [ ] Display YES/NO if the auth check worked.
- [ ] If NO, display a link to the logout/renew credentials steps.

G-Ambatte:
This PR resolves #2520.

This PR adds a link to the instructions for a user to refresh their Google credentials to the Account Page in the event that the list of Google files are unable to be received.

### DIFF
--- a/client/homebrew/pages/accountPage/accountPage.jsx
+++ b/client/homebrew/pages/accountPage/accountPage.jsx
@@ -42,7 +42,6 @@ const AccountPage = createClass({
 	},
 
 	renderUiItems : function() {
-		// console.log(this.props.uiItems);
 		return 	<>
 			<div className='dataGroup'>
 				<h1>Account Information  <i className='fas fa-user'></i></h1>
@@ -56,7 +55,11 @@ const AccountPage = createClass({
 			<div className='dataGroup'>
 				<h3>Google Information <i className='fab fa-google-drive'></i></h3>
 				<p><strong>Linked to Google: </strong> {this.props.uiItems.googleId ? 'YES' : 'NO'}</p>
-				{this.props.uiItems.googleId ? <p><strong>Brews on Google Drive: </strong> {this.props.uiItems.fileCount || '-'}</p> : '' }
+				{this.props.uiItems.googleId &&
+					<p>
+						<strong>Brews on Google Drive: </strong> {this.props.uiItems.fileCount != '-' ? this.props.uiItems.fileCount : <>Unable to retrieve files - <a href='https://github.com/naturalcrit/homebrewery/discussions/1580'>follow these steps to renew your Google credentials.</a></>}
+					</p>
+				}
 			</div>
 		</>;
 	},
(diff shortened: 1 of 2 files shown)


G-Ambatte:
Here's how it looks on my local install:

image

Gazook89:
Should the instructions just be hosted on HB directly, rather than sending to github?

calculuschild:
Probably eventually. But I think for this PR it's ok.

client/homebrew/pages/accountPage/accountPage.jsx
```diff
@@ -56,7 +55,11 @@ const AccountPage = createClass({
 			<div className='dataGroup'>
 				<h3>Google Information <i className='fab fa-google-drive'></i></h3>
 				<p><strong>Linked to Google: </strong> {this.props.uiItems.googleId ? 'YES' : 'NO'}</p>
-				{this.props.uiItems.googleId ? <p><strong>Brews on Google Drive: </strong> {this.props.uiItems.fileCount || '-'}</p> : '' }
+				{this.props.uiItems.googleId &&
+					<p>
+						<strong>Brews on Google Drive: </strong> {this.props.uiItems.fileCount != '-' ? this.props.uiItems.fileCount : <>Unable to retrieve files - <a href='https://github.com/naturalcrit/homebrewery/discussions/1580'>follow these steps to renew your Google credentials.</a></>}
```
calculuschild, line 60: Minor change: Instead of passing in `-` from `app.js` when there are 0 google brews, can we pass in 0? Since the `-` doesn't seem like it will be used here anymore, and it might be a source of confusion why we are using a hyphen in this one spot and nowhere else.

calculuschild:
@G-Ambatte do you want to finish this one up and I can get it merged?

G-Ambatte:
Sorry, I saw the request for changes while I was at work, thought "I'll deal with that tonight" and by the time I got home, it had completely slipped my mind.
I'll do my best to knock these out quickly today.

client/homebrew/pages/accountPage/accountPage.jsx
G-Ambatte: That's a good question: 0 should be a valid result, whereas a failure to retrieve any results should be a separate issue. I'll think on this today.

G-Ambatte: I made sure that the MongoDB query could actually return 0, but it was still failing... I figured out that because 0 as a number is falsy, while `app.js` was passing a zero, the subsequent checks on `accountPage.jsx` were reporting the same as if they'd failed.
But "0" as a string is truthy, so converting the passed data from a number to string has allowed 0 to be a valid result.

Now, I just need to replicate this for Google Drive brews...

G-Ambatte: Tested with Google brews, they're now working correctly. I've just pushed a commit that works with either 0 Mongo or Google brews, and still differentiates out the error to the login instructions when listing the Google Drive folder is not permitted.

G-Ambatte:
Google linked - no permissions:
image

Google linked - full permissions, 0 Google files:
image

MongoDB - 0 brews:
image

PR #2528, opened 2022-11-29
Change error message about googleActions on local startup.

### LINKED ISSUE #2424: Local - Google Auth / Service calls
### Renderer

- [ ] Legacy
- [X] v3

### Browser

Edge

### Operating System

Windows

### What happened?

Hallo,

I followed the Quick Start guide trying to get a local copy of Homebrewery running.  This results in an error on startup, since I've not configured my localhost to work with Google Docs.

```
> node server.js

Error: Must pass in a JSON object containing the Google auth settings.
    at AuthPlus.fromJSON (C:\homebrewery\node_modules\google-auth-library\build\src\auth\googleauth.js:298:19)
    at Object.<anonymous> (C:\homebrewery\server\googleActions.js:13:28)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:94:18)
    at Object.<anonymous> (C:\homebrewery\server\homebrew.api.js:6:23)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
Please make sure that a Google Service Account is set up properly in your config files.
server on port: 8000
```

Additionally, when I try to use the local API to PUT changes to a Brew, I get Gaxios errors:

```
GaxiosError: Daily Limit for Unauthenticated Use Exceeded. Continued use requires signup.
    at Gaxios._request (C:\homebrewery\node_modules\gaxios\build\src\gaxios.js:130:23)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

Any chance there can be a disabler wrapper around the various Google service calls? 

### Code

_No response_

Gazook89:
This PR simply checks to see if `keys` exist in googleActions.js and if not, throws a less ~~scary~~ verbose error message.

Fixes #2424

### DIFF
--- a/server/googleActions.js
+++ b/server/googleActions.js
@@ -7,7 +7,7 @@ const config = require('./config.js');
 
 let serviceAuth;
 if(!config.get('service_account')){
-	console.log('No Google Service Account in config files - Google Drive functionality will not function.');
+	console.log('No Google Service Account in config files - Google Drive integration will not be available.');
 } else  {
 	const keys = typeof(config.get('service_account')) == 'string' ?
 		JSON.parse(config.get('service_account')) :
@@ -15,12 +15,10 @@ if(!config.get('service_account')){
 
 	try {
 		serviceAuth = google.auth.fromJSON(keys);
-		serviceAuth.scopes = [
-			'https://www.googleapis.com/auth/drive'
-		];
+		serviceAuth.scopes = ['https://www.googleapis.com/auth/drive'];
 	} catch (err) {
 		console.warn(err);
-		console.log('Please make sure that a Google Service Account is set up properly in your config files.');
+		console.log('Please make sure the Google Service Account is set up properly in your config files.');
 	}
 }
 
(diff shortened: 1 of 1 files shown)


calculuschild (changes requested):
Rather than put this inside the try/catch block, let's move this out to right below where we define `keys`. Then if keys doesn't exist, we can immediately show your message, and don't have to enter the try/catch block in the first place. I.e., we shouldn't have to wait for service auth generation to try and fail if keys don't even exist.

If keys *does* exist, then we move ahead with the try/catch block, and the error in there can be rephrased to indicate that there is a service account found, but it might not be properly set up.

Gazook89:
So I think I went one step further with last commit- it doesn't try to create keys if 'service_account' isn't in the config, and displays the new message.

PR #2527, opened 2022-11-29
Fix og:image and og:site_name

Gazook89:
This fixes some typo's I made in the OG meta tags.  

`og:siteName` => `og:site_name`
`og:thumbnail` => `og:image`

Before merging, let me know if you prefer `edit` and `share` links to display the d20 logo by default if the user hasn't set their own thumbnail.  Currently it is set to not display anything.

PR #2530, opened 2022-11-30
Fix crash opening Download page with non-latin title

Gazook89:
This PR fixes an [issue reported on reddit](https://old.reddit.com/r/homebrewery/comments/z8arm9/homebrews_with_nonlatin_characters_in_the_title/) where the user has a brew title with non-latin characters, and choosing to "Download" it lead to an error page.  

Content Disposition must be utf-8 encoded.  So this PR [follows the MDN docs for doing this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).

Example title:  Cześć or καλή μέρα

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -192,12 +192,21 @@ app.get('/download/:id', asyncHandler(getBrew('share')), (req, res)=>{
 	sanitizeBrew(brew, 'share');
 	const prefix = 'HB - ';
 
+	const encodeRFC5987ValueChars = (str)=>{
+		return (
+			encodeURIComponent(str)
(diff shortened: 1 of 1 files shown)


calculuschild (commented):
Looks good, but includes several changes from #2527 as well (which I have a small comment for). Can you make sure this isn't going to overwrite those tweaks I asked for? I.e., when you make changes to #2527, make the same changes here, or redo this PR so it doesn't overlap with #2527

Gazook89:
Oh dang, didn’t notice that. I had a little hiccup when doing the branch but thought I had it resolved. 

I’ll likely just delete this branch and start fresh since it was such a simple change.

PR #2537, opened 2022-12-01
change injector regex to work with safari

### LINKED ISSUE #2531: styling {injector} is broken in Safari
### Renderer

v3

### Browser

Safari

### Operating System

MacOS

### What happened?

Loading any page causes Safari to fail incomplete.

An error in the console leads to this:
image

The source is https://github.com/naturalcrit/homebrewery/blob/ee05392d27ce76999b95bfb8f990d17a1e6d2b1a/client/homebrew/editor/editor.jsx#L142

The regex for styling `{injectors}` begins with a negative look-behind, but that is not supported in Safari per https://caniuse.com/js-regexp-lookbehind

Suggestion: this `(?<!{)` be changed to a non-capturing group `(?:[^{]?)` — that is, the `{injector}` pattern should be preceded with zero-or-one character which is not a `{`.

### Code

```gfm
# heading
{injector}
```

Gazook89:
This PR fixes #2531 by updating the injector syntax to work with Safari.


client/homebrew/editor/editor.jsx
```diff
@@ -139,10 +139,10 @@ const Editor = createClass({
 
 						// Highlight injectors {style}
 						if(line.includes('{') && line.includes('}')){
-							const regex = /(?<!{){(?=((?::(?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"':{}\s]*)*))\1}/g;
+							const regex = /(?:^|[^{|\n])({(?=((?::(?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"':{}\s]*)*))\2})/gm;
```
calculuschild, line 142: Is there a reason we want to exclude injectors occurring after pipes `|`  with `[^{|\n]` ?

Gazook89: No.  When i wrote it I was thinking of it as "OR" but clearly that isn't needed in the brackets. Fixed.

Another question:  Do we want the highlight/injection to happen if there is a `\s` / `space` before the brackets?  

a. `## My text is here {class}`   
b. `## My text is here{class}`

Should both be injected, or just *b*?   

In my opinion, both should work for readability sake.  However, if *A* is allowed then it wouldn't be possible to write "The alien came around the corner {and started blasting}."    Obviously, parentheses would be the logical choice, but it may preclude stylistic choices.

We don't have an escape character for these things, right?

Gazook89: For now, I just removed that pipe from the pattern.  So if that works for you, merge it.

Gazook89: For now I just pushed the fix to the pipe.  I tested it in Safari.

calculuschild: Yeah I'm pretty sure we inject with any number of beginning spaces too.  A backslash will escape the injector

Gazook89: okay then i think this is good to go?

PR #2547, opened 2022-12-06
Fix tags validation issues

jeddai:
Add callIfExists method, which will call a chosen method only if it exists

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -59,8 +59,6 @@ const MetadataEditor = createClass({
 	},
 
 	handleFieldChange : function(name, e){
-		callIfExists(e, 'persist');
-
 		// load validation rules, and check input value against them
 		const inputRules = validations[name] ?? [];
 		const validationErr = inputRules.map((rule)=>rule(e.target.value)).filter(Boolean);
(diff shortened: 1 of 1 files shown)


calculuschild:
Is the issue occurring for more than just `.persist()` then?

jeddai:
Yep, so I was going to try and pass in an event that I create from the document node but since it's a special react event and has special methods I couldn't do that. Since tags have a more proactive approach to validation I'm not concerned about this validation not working for them (in the short term), but this will make it so the tags at least function.

calculuschild:
To follow on, do we even need persist at all? I haven't really looked into why we need it if Tags can't even access it anyway.

If we do use this solution, would it be better to define the function outside of the handler so we aren't re-defining it with every validation check or whatever (every keypress)?

jeddai:
Yeah from what I can tell until React v17 `persist` is necessary because any calls to the synthetic event methods when it _hasn't_ been persisted are considered illegal invocations.

Personally, my long-term recommendation would be to move away from the synthetic event-based validation error popups and use an inline or block-based error message.

But yes you're right, I should probably move it to outside the method. Maybe just a static in the JS file.

calculuschild:
I did just update us to React 17. Does that mean we get rid of Persist then?

jeddai:
Oooh that's a good question, I'll check.

calculuschild:
Did you get a chance to check? Otherwise I can just merge it now so our users can have tags working again.

jeddai:
Yep, I just removed the persist call but we do still need the new function for the other event methods we call.

PR #2548, opened 2022-12-06
Add api tests

jeddai:
Here's the beginnings of what can become our API tests!


jeddai:
The actual code changes in the API file are things I have open on other PRs plus some shifting around of the code into a single object (so it can be mocked properly).

jeddai:
In the current state, if you run `npm test -- --coverage` you'll see that we have about 38% coverage on the main api as compared to 15% on the main branch.

server/homebrew.api.spec.js
```diff
@@ -0,0 +1,377 @@
...
+				body : {
+					googleId : '12345'
+				}
+			});
+
+			expect(id).toEqual('abcdefgh');
+			expect(googleId).toEqual('12345');
+		});
+
+		it('should return id and google id params', ()=>{
```
AlexeySachkov, line 77:
```suggestion
		it('should return id and google id from params', ()=>{
```

server/homebrew.api.spec.js
```diff
@@ -0,0 +1,377 @@
...
+			const testBrew = { title: 'test brew', authors: [], tags: 'tag' };
+			api.getId = jest.fn(()=>({ id: '1', googleId: undefined }));
+			model.get = jest.fn(()=>toBrewPromise(testBrew));
+
+			const fn = api.getBrew('share', true);
+			const req = { brew: {} };
+			const next = jest.fn();
+			await fn(req, null, next);
+
+			expect(req.brew.tags).toEqual([]);
```
AlexeySachkov, line 145: Just out of my own curiosity, could you please explain why we see an empty array as `tags` here? What exactly are we testing here? Do I understand it correctly that we may have string representation for tags in our database and we have some code which drops that field and replaces it with an empty array? If so, why do we need that special handling, can't we ensure that DB already contains data in a correct format?

server/homebrew.api.spec.js
```diff
@@ -0,0 +1,377 @@
...
+		it('does not throw if no authors', async ()=>{
+			api.getId = jest.fn(()=>({ id: '1', googleId: undefined }));
+			model.get = jest.fn(()=>toBrewPromise({ title: 'test brew', authors: [] }));
+
+			const fn = api.getBrew('edit', true);
+			const req = { brew: {} };
+			const next = jest.fn();
+			await fn(req, null, next);
+
+			expect(next).toHaveBeenCalled();
```
AlexeySachkov, line 175: Do we expect `{ title: 'test brew', authors: [] }` to be returned or we expect to see an empty brew? I would suggest that we test `req.brew` result as well: for purposes of documenting API behavior and testing it.

server/homebrew.api.spec.js
```diff
@@ -0,0 +1,377 @@
...
+			const brew = {
+				...hbBrew,
+				title : undefined
+			};
+			api.beforeNewSave({}, brew);
+
+			expect(brew.title).toEqual('brew text');
+		});
+
+		it('does not set the title if present', ()=>{
```
AlexeySachkov, line 330:
```suggestion
		it('does not override the title if present', ()=>{
```

AlexeySachkov (commented):
Yay! Glad to see more testing, thanks for that! A few minor comments and some questions

server/homebrew.api.spec.js
ericscheid: You are correct that there are string representations for tags in the database.

Doing a mass-conversion to an empty array would be a high-impact database operation, possibly resulting in some downtime. We don't have formal procedures for "downtime for maintenance", so such an operation has been punted. For now.

Additionally, the code does need to handle brews with invalid values .. which might be retrieved from google drive storage. Somewhere there's a bit of work with implementing a Postel Robustness filter on the retrieve/save functions. IIRC this work hasn't been refactored in.

calculuschild: To clarify:

Old brews have strings for tags, but *they are all empty strings*. We switched to arrays before the Tags UI was implemented, so there was no way to actually create a tag when it was still strings. That is, all actual tags are correctly implemented in arrays, new brews will have an empty array, and it is only old brews that may have an empty string.

jeddai: Yep! This test is merely testing the logic that exists in the getBrew function that performs the replacement from string to empty array.

jeddai:
I've updated based on feedback, and this now includes tests for the delete method. I'm working on the update tests, but any comments on what I have now are welcome!

server/homebrew.api.js
```diff
@@ -17,344 +17,329 @@ const { DEFAULT_BREW, DEFAULT_BREW_LOAD } = require('./brewDefaults.js');
...
+router.put('/api/:id', asyncHandler(api.getBrew('edit', true)), asyncHandler(api.updateBrew));
+router.put('/api/update/:id', asyncHandler(api.getBrew('edit', true)), asyncHandler(api.updateBrew));
+router.delete('/api/:id', asyncHandler(api.deleteBrew));
+router.get('/api/remove/:id', asyncHandler(api.deleteBrew));
 
-module.exports = {
-	homebrewApi : router,
-	getBrew
-};
+module.exports = api;
```
calculuschild, line 345: By wrapping the whole API here, do we need to change the locations where `getbrew` was being used to now be API.getbrew? (I.e., in app.js)

jeddai: Nope! That's actually the beauty of it. Since the `module.exports` is set to `api`, nothing outside the file needs to change. As for `router`, it's set onto `api` by reference, so the routes still work when referencing it by `router` from within the file.

calculuschild: Ah, you are right. It's just another object as before. Carry on :)

jeddai:
Test coverage threshold piece works! 🎉 https://app.circleci.com/pipelines/github/naturalcrit/homebrewery/3077/workflows/4b683b4b-665a-4803-89e4-87f17087ca7d/jobs/4041

PR #2554, opened 2022-12-08
[WIP] [BrewRenderer] Add brew scaling

G-Ambatte:
This PR adds some initial functionality to zoom the content of the Brew Renderer frame.

This is still WIP so a styling pass has not yet been applied.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -42,6 +42,7 @@ const BrewRenderer = createClass({
 			viewablePageNumber : 0,
 			height             : 0,
 			isMounted          : false,
+			scale              : 1,
 
 			pages          : pages,
 			usePPR         : pages.length >= PPR_THRESHOLD,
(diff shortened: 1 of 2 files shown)


G-Ambatte:
I had a lightbulb moment earlier today: I suspect that this will work better if, rather than using `scale`, we manipulate the page in 3D space using `perspective` and `translateZ`.

Sample styling code:
```
transform: perspective(10px) translateZ(0px);
transition-duration: 0.5s;
z-index: 99;
```

calculuschild:
@G-Ambatte What's the status on this one? Did it make it out of WIP status?

5e-Cleric:
Question: why is this feature not using the CSS property [Zoom](https://developer.mozilla.org/en-US/docs/Web/CSS/zoom)

I know it does not have the best support, but like `:has`, the only failing browser is FF, which is not a problem for such a simple feature.
As far as i know, reccomendation is to use Scale rather than Zoom, but zoom does not produce the same problems as scale does.

G-Ambatte:
> @G-Ambatte What's the status on this one? Did it make it out of WIP status?

Not exactly; regardless of the method used to cause the zoom (`zoom`, `scale`, `perspective`) there needed to be a function to return to regular zoom levels, and this needed to trigger before use of other functions (like "match view to source" or "match source to view").
It also caused strangeness with the current page number in the page info panel, but that might be resolved with the new Page component.

This can probably be closed; any future work is likely better starting from a clean slate.

calculuschild:
Alright. Gonna close for now then.

PR #2556, opened 2022-12-09
[LOCAL ONLY] Get Google Drive folder name from config

### LINKED ISSUE #2555: [LOCAL Only] Change Google Drive folder name
### Your idea:

As local installs can use the same Google account is the live website, this will result in two Homebrewery folders in Google Drive, with no easy way to differentiate the two. Adding the ability to modify the folder name, ideally through a `config/local.conf` or similar file, will make it much easier for those of us running Google-linked local installs to locate our files.

G-Ambatte:
This PR resolves #2555.

This PR gets the Google Drive folder name from the configuration files under the name `google_folder_name`. If no such config element exists, the name defaults to `Homebrewery`.


calculuschild:
I don't think we need this; the user can just go into Google Drive and change the folder name there?

calculuschild:
Closing this. Easily solved by just going into Google Drive and renaming your folder.

PR #2561, opened 2022-12-14
Add ability to invite authors + version comparison check

jeddai:
This PR completes the following

- [x] add ability to invite authors
- [x] allows removal/edit of _invited_ authors
- [ ] allows removal of other authors
    - I'd like to handle this in a separate PR
- [x] moves version from google -> stub
- [x] checks/updates version on save
- [x] adds a new error message for version mismatch
- [x] groups metadata fields

image
image

Partially completes #1987


client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -296,6 +296,23 @@ const MetadataEditor = createClass({
...
+
+			<h1 className='sectionHead'>Authors</h1>
+
+			{this.renderAuthors()}
+
+			<StringArrayEditor label='invited authors' valuePatterns={[/.+/]}
+				validators={[(v)=>!this.props.metadata.authors.includes(v)]}
+				placeholder='invite author' unique={true}
+				values={this.props.metadata.invitedAuthors}
+				notes={['Invited authors are case sensitive.', 'After adding an invited author, send them the edit link. There, they can choose to accept or decline the invitation.']}
```
jeddai, line 309: > There, they can choose to accept or decline the invitation.

That's the intent, but I'm going to handle that in a separate PR. For now, the users can secure their invitation by saving.

calculuschild:
Just a little bit of weirdness on the style:
button misaligned
image
image

I know this was already there with tags but just pointing it out if we want to tweak a bit.

calculuschild:
Invited authors can invite other authors before accepting their own invite.

jeddai:
https://github.com/naturalcrit/homebrewery/pull/2561#issuecomment-1352178305

I can't seem to be able to recreate this. Can you elaborate on how to get this to show up?

calculuschild:
> I can't seem to be able to recreate this. Can you elaborate on how to get this to show up?

I'm not doing anything special, this is just what it looks like on my screen when editing a tag or author. It's fine when not editing the tag, but it seems the border on the text box is interfering with the alignment somehow. What does it look like for you when editing a tag?

jeddai:
image

calculuschild:
Yup, that's it. The tag being edited has it's check button out of alignment. (Or the text box is out of alignment with the check)

jeddai:
🤦 I thought you were talking about the more rounded border radius on the right side of the input in your image. Yeah I'll try to fix that haha

server/homebrew.api.js
```diff
@@ -187,7 +189,13 @@ const newBrew = async (req, res)=>{
 
 const updateBrew = async (req, res)=>{
 	// Initialize brew from request and body, destructure query params, set a constant for the google id, and set the initial value for the after-save method
-	let brew = _.assign(req.brew, excludePropsFromUpdate(req.body));
+	const updateBrew = excludePropsFromUpdate(req.body);
```
calculuschild, line 192: Let's be careful about using the same variable name as the function here. I think earlier we had an issue of "X is not a function" because of variable name conflicts.

jeddai: OOP I did not noticed I'd done that. Changing that now

calculuschild (commented):
Only other thing I can think of is just spicing up the "you do not have access to this brew" error. Something like:

> The current logged in user does not have editor access to this brew. If you believe you should have access to this brew, ask the file owner to invite you as an author by opening the brew, viewing the **PROPERTIES ( i )** tab, and adding your username to the "invited authors" list. You can then try to access this document again.

We can make it look "pretty" in a different PR with an actual formatted error page.

calculuschild:
Wait.... oddly, I am able to access the edit page of brews via the old "long" google ID link whether I am an other or not. If you have Google Drive set up on your local install you can test accessing my brew with this: /edit/1Er3bUFJdSFemDEAwPHRULG4_LG7idR0mEmw7HQyD6_Bx

jeddai:
Ah I might need to move the if statement

server/homebrew.api.js
```diff
@@ -43,7 +43,9 @@ const getBrew = (accessType, stubOnly = false)=>{
 				}
 			});
 		stub = stub?.toObject();
-		if(accessType === 'edit' && stub?.authors?.length > 0 && !stub?.authors.includes(req.account?.username)) {
+		const authorsExistAndIsNotAuthor = stub?.authors?.length > 0 && !stub?.authors.includes(req.account?.username);
+		const isNotInvited = stub?.invitedAuthors?.length > 0 && !stub?.invitedAuthors.includes(req.account?.username);
```
calculuschild, line 47: If `stub` doesn't exist, or `invitedAuthors` doesn't exist or is size 0, `isNotInvited` will be FALSE and the user will be allowed access.

The negative variable name is also making the logic tricky to follow (at least for me). Can we make this `isInvited`?

.... actually.... can we change all of these names?

Change `authorsExistAndIsNotAuthor` into `authorsExist` and `isAuthor`, and `isNotInvited` into `isInvited`?

Then the access blocking logic becomes `if(accessType === 'edit' && (authorsExist && !isAuthor || i!sInvited))` or whatever the parentheses need to be.

server/homebrew.api.js
```diff
@@ -62,6 +57,14 @@ const getBrew = (accessType, stubOnly = false)=>{
 			// Combine the Homebrewery stub with the google brew, or if the stub doesn't exist just use the google brew
 			stub = stub ? _.assign({ ...excludeStubProps(stub), stubbed: true }, excludeGoogleProps(googleBrew)) : googleBrew;
 		}
+		const authorsExist = stub?.authors?.length > 0;
+		const isAuthor = (stub?.authors || [])?.includes(req.account?.username);
```
calculuschild, line 61: Do we need the `|| []` ? If its an empty array it can't include the current username.
```
const isAuthor = stub?.authors?.includes(req.account?.username);

or

const isAuthor = authorsExist && stub.authors.includes(req.account?.username);
```

server/homebrew.api.js
```diff
@@ -62,6 +57,14 @@ const getBrew = (accessType, stubOnly = false)=>{
...
+		const isInvited = (stub?.invitedAuthors || []).includes(req.account?.username);
```
calculuschild, line 62: Same comment here.

server/homebrew.api.js
jeddai: Yeah for some reason I thought that js would barf at the method call on undefined but I just double checked and it seems fine

calculuschild: I think the Elvis stops evaluating as soon as it sees a property doesn't exist so it doesn't even get far enough to call the method.

server/homebrew.api.js
jeddai: Both are resolved now!

jeddai:
@calculuschild if it's okay I'll try to fix the tag alignment issue in a separate PR

calculuschild:
> @calculuschild if it's okay I'll try to fix the tag alignment issue in a separate PR

That's fine. I'll test and merge this tonight if possible.

server/homebrew.api.js
```diff
@@ -60,6 +57,14 @@ const getBrew = (accessType, stubOnly = false)=>{
 			// Combine the Homebrewery stub with the google brew, or if the stub doesn't exist just use the google brew
 			stub = stub ? _.assign({ ...excludeStubProps(stub), stubbed: true }, excludeGoogleProps(googleBrew)) : googleBrew;
 		}
+		const authorsExist = stub?.authors?.length > 0;
+		const isAuthor = stub?.authors?.includes(req.account?.username);
+		const isInvited = stub?.invitedAuthors?.includes(req.account?.username);
+		if(accessType === 'edit' && (authorsExist && (!isAuthor || !isInvited))) {
```
calculuschild, line 63: This is causing another issue; now nobody can access any brews because if you are an author, you are not invited, and if you are invited, you are not an author. That is, `(!isAuthor || !isInvited)` is always true.

```suggestion
		if(accessType === 'edit' && (authorsExist && !(isAuthor || isInvited))) {
```

jeddai: 🤦 thank you

calculuschild:
All seems to work! Thanks! 💯 🎉  ⭐

G-Ambatte:
As per the dicussion in Gitter chat today, it would be useful if, when the Edit Page is loaded and the logged in user is in the Invited Authors but not the Authors list, that they receive a prompt (possibly `confirm()`) requesting to trigger a save to accept the invitation and be added to the Authors list.

PR #2562, opened 2022-12-15
[WIP] Update SharePage when brew is edited

G-Ambatte:
I was experimenting with the BroadcastChannel functionality and managed to create the following:

- When SharePage is loaded, subscribe to a BroadcastChannel with name `shareId`
- When EditPage is saved, post a new message in the `shareId` BroadcastChannel
- When SharePage receives an update message, confirm `shareId` and `origin` match; if so, perform a `window.location.reload()`, refreshing the brew and allowing effectively real time updates

Future improvements:
- a window reload loses the place in the brew, returning to the top; this can probably be resolved by instead sending the brew content and updating the brew state on the SharePage, causing only a re-render rather than a whole refresh.

Primarily, I see this as being useful for DMs who wish to present to their parties on a second screen - pull up EditPage on the hidden screen, and SharePage on the second monitor. Alter EditPage - for example, toggle an item's visibility from hidden to visible - and suddenly it will appear on the SharePage.

### DIFF
--- a/client/homebrew/pages/sharePage/sharePage.jsx
+++ b/client/homebrew/pages/sharePage/sharePage.jsx
@@ -12,6 +12,7 @@ const Account = require('../../navbar/account.navitem.jsx');
 
 const BrewRenderer = require('../../brewRenderer/brewRenderer.jsx');
 
+let broadcastChannel;
 
 const SharePage = createClass({
 	displayName     : 'SharePage',
@@ -32,10 +33,16 @@ const SharePage = createClass({
 
 	componentDidMount : function() {
 		document.addEventListener('keydown', this.handleControlKeys);
+
+		const channelName = this.props.brew.shareId;
+		broadcastChannel = new window.BroadcastChannel(channelName);
+		broadcastChannel.onmessage = (event)=>{this.messageReceived(event);};
 	},
 
 	componentWillUnmount : function() {
 		document.removeEventListener('keydown', this.handleControlKeys);
+
+		broadcastChannel.close();
 	},
 
 	handleControlKeys : function(e){
@@ -45,6 +52,13 @@ const SharePage = createClass({
 			window.open(`/print/${this.processShareId()}?dialog=true`, '_blank').focus();
 			e.stopPropagation();
 			e.preventDefault();
+		}
+	},
+
+	messageReceived : function(e){
+		console.log(`Message RX: ${new Date}`);
+		if(e.origin === window.location.origin && e.data === `brewUpdate:${this.props.brew.shareId}`) {
+			window.location.reload();
 		}
 	},
 
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -26,6 +26,8 @@ const googleDriveInactive = require('../../googleDriveMono.png');
 
 const SAVE_TIMEOUT = 3000;
 
+let broadcastChannel;
+
 const EditPage = createClass({
 	displayName     : 'EditPage',
 	getDefaultProps : function() {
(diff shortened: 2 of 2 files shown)


client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -48,6 +55,13 @@ const SharePage = createClass({
 		}
 	},
 
+	messageReceived : function(e){
+		console.log(`Message RX: ${new Date}`);
+		if(e.origin === window.location.origin && e.data === `brewUpdate:${this.props.brew.shareId}`) {
```
ericscheid, line 60: Theoretically, the same origin check shouldn't be necessary since by definition only same origin tabs (etc) can communicate. Did you find otherwise?

https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel

> The BroadcastChannel interface represents a named channel that any [browsing context](https://developer.mozilla.org/en-US/docs/Glossary/Browsing_context) of a given [origin](https://developer.mozilla.org/en-US/docs/Glossary/Origin) can subscribe to. It allows communication between different documents (in different windows, tabs, frames or iframes) of the same origin.

ericscheid:
This is very cool, btw.

client/homebrew/pages/sharePage/sharePage.jsx
G-Ambatte: The origin check was left over from previous experimentation with [window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage); after some time and a distinct lack of success, I realized that BroadcastChannel was likely the better candidate for what I was trying to achieve.

I should also remove that unnecessary `console.log` call, it was only to show that messages were being received.

G-Ambatte: I've removed the console log, and all of the received message checks, instead using JSON.stringify and JSON.parse to pass the entire `brew` from EditPage directly to the SharePage, which is then applied as a state update, causing a re-render, rather than a reload.

As predicted, this eliminates the loss of place in the brew on the SharePage when an update comes through from EditPage.

G-Ambatte:
I think this is pretty close to complete, actually. What else might we want from this functionality? What else is even possible?

Gazook89:
I’m not advocating for it really, but maybe make it so that any update from the previous version is briefly highlighted before fading out?  I don’t really know how this would work for anything besides just regular text changes, like changes to divs etc.

G-Ambatte:
It's not so much collaboration as it is a local subscription. I was thinking about CM6's atomic change model, which would as I understand it is essentially just sending the diff, which would be basically perfect for this kind of system.

One thing I did come up overnight is that if EditPage emitted a brewUpdate event on changing editors as well as on save, then EditPage would only ever need to send the content of the currentEditor (on save) or the editor just left (on Editor change). This would reduce the amount of data being transmitted via the BroadcastChannel.
Once that's in place, a diff mechanism could be added to further reduce transmitted data.

dbolack-ab:
> I think this is pretty close to complete, actually. What else might we want from this functionality? What else is even possible?

Maybe a notify box?

5e-Cleric:
Perhaps subtly highlight the changed page, like other software does with the text

dbolack-ab:
> Perhaps subtly highlight the changed page, like other software does with the text

The CM 'page' ( between two `\page` ?)  where the change occurred?

G-Ambatte:
I will note here that using the `event-stream` source created in PR #4034, instead of a BroadcastChannel, would allow this functionality to occur across browsers and even devices.

5e-Cleric:
This would preferably work with the patch functionality, if it is fixed that is.

PR #2589, opened 2023-01-07
adjust frontend error handling, add client/server mismatch middleware

jeddai:
Testing this is rather simple! 
1. Navigate to the home, new, or edit page.
2. Open the js console, and set `version` to anything that isn't the current version.
3. Trigger a save.
4. The error should appear, explaining what's wrong.

The logic is also quite simple. 
* Whenever an http call is made to the API, a superagent middleware pulls the version out of `global`, and places it on a header called `Homebrewery-Version`.
* The API now has a middleware that checks the version in that header against the server's version. If they don't match, it returns a 412 error.
  * I checked all the http response codes and couldn't find one that matched exactly, but 412 seemed reasonable based off its definition.

~At this time this is not complete, I still need to finish the brews page and metadata editor's deletion error handling.~


jeddai:
@calculuschild would you mind doing a quick check of this and sharing your thoughts before I wrap it up?

server/middleware.js
```diff
@@ -0,0 +1,17 @@
+const middleware = {
```
calculuschild, line 1: Can we rename to something more meaningful? `checkClientVersionMiddleware` or something?

client/template.js
```diff
@@ -22,6 +22,9 @@ const template = async function(name, title='', props = {}){
 		<body>
 			<main id="reactRoot">${require(`../build/${name}/ssr.js`)(props)}</main>
 			<script src=${`/${name}/bundle.js`}></script>
+			<script>
+				sessionStorage.setItem('version', '${props.version}');
+			</script>
```
calculuschild, line 27: We already have access to the version via `global.version` set in Homebrew.jsx (the root React container). This is how we access it to display the version number on the nav bar.

We also set a handful of other globals there. That may or may not be the best place to put them, but at least for Version let's have it in only one location to keep our data? If you want to have it in Session Storage, let's change the nav bar to also use session storage. Otherwise jus use `global.version` and remove this from the template.

(also, I think I prefer to not handle this inside the template; I've kind of intended it to just load and boot the application rather than setting variables and stuff.)

calculuschild (changes requested):
This seems to work pretty well. You can go ahead and finish this up. I just had some comments on file structure and formatting mostly. The functionality is good.

client/template.js
jeddai: > We already have access to the version via `global.version` set in Homebrew.jsx (the root React container). This is how we access it to display the version number on the nav bar.

Ah I didn't realize that, yeah I'll get rid of the session storage stuff

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -266,75 +262,8 @@ const EditPage = createClass({
...
-				<div className='errorContainer'>
-					Looks like there was a problem saving. <br />
-					Report the issue <a target='_blank' rel='noopener noreferrer'
-						href={`https://github.com/naturalcrit/homebrewery/issues/new?template=save_issue.yml&error-code=${encodeURIComponent(errMsg)}`}>
-						here
-					</a>.
-				</div>
-			</Nav.item>;
+		if(this.state.error){
+			return require('../../utils/render-error-nav-item.js')(this, this.state.error);
```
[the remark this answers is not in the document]
jeddai: Hm I don't know how doable that'll be considering this is sort of built into other nav items

client/homebrew/utils/render-error-nav-item.js
```diff
@@ -0,0 +1,71 @@
+const React = require('react');
```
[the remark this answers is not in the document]
jeddai: Oop that's a good point, thanks

jeddai:
@calculuschild this is ready for a full review!

jeddai:
My own coverage requirements got me. I think it's just because I added another line. I'll increase the threshold tomorrow.

client/homebrew/pages/editPage/editPage.jsx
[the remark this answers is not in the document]
jeddai: Did it! Realized I can replace the component where it overlays but add the component where there isn't a save button.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -205,8 +200,9 @@ const EditPage = createClass({
 			.send(brew)
 			.catch((err)=>{
 				console.log('Error Updating Local Brew');
-				this.setState({ errors: err });
+				this.setState({ error: err.response });
```
calculuschild, line 203: Not sure if this line is the cause, but the error messages generated to send to GitHub is this on the live site:

image

But seems to be missing a lot of info with this PR:

image

client/homebrew/navbar/error-navitem.jsx
```diff
@@ -0,0 +1,84 @@
...
+			this.props.parent.setState(state);
+		};
+
+		const error = this.props.error;
+		const status = error.status;
+		const message = error.body?.message;
+		let errMsg = '';
+		try {
+			errMsg += `${error.toString()}\n\n`;
+			errMsg += `\`\`\`\n${error.stack}\n`;
```
calculuschild, line 30: Yeah, using only the `.response` part of the Error object, we seem to be losing the `.toString()` and `.stack` properties.

client/homebrew/pages/editPage/editPage.jsx
jeddai: image
After my adjustments we get the stack trace, while it's not pretty I feel like it's more complete than it was before

PR #2604, opened 2023-01-13
Try @Layers to not need `.page` in user style

calculuschild:
Probably needs tuning, but thanks to a suggestion from @5e-Cleric, it seems CSS Layers might eliminate the need for users to add `.page` to every custom style in the Style tab.

So far this uses 5 layers; later overrides previous, and non-layered styles win everything, except default browser styles:

0) *Not layered* : Built-in Browser styles
1) **bundle** (base Homebrewery CSS: bundle.css)
2) **V3_Blank** (blank theme)
3) **V3_PHB** (PHB theme)
4) **V3_DMG** (DMG theme, build on top of PHB)
5) **styleTab** (user styles)
6) *Not layered* inline CSS

There will need to be separate layer names for each other theme, which I haven't gotten to yet.

##### Todo
- [x] Test this doesn't break any of our built-in themes
- [ ] Test this doesn't break existing brews that currently use `.page`
- [x] Test this doesn't break the core Homebrewery site (limit to the BrewRenderer iFrame)
- [x] Wrap the rest of the themes in layers
- [x] Make this work on the `/print` page.


5e-Cleric:
I don't know how CSS works in the HB, but be noted that Layers take order when first defined, in order to avoid this behaviour a good option would be to define them beforehand, like so:

`@layer bundle, V3_Blank, V3_DMG, styleTab;`

And mention that layers can be used in `@import` rules and may be declared before them.

5e-Cleric:
All themes' layers could potentially be nested inside another layer, i believe it is worth considering, in order to keep all together.

calculuschild:
> I don't know how CSS works in the HB, but be noted that Layers take order when first defined, in order to avoid this behaviour a good option would be to define them beforehand, like so:

Rather than pre-define a list, I am simply loading each layer sequentially as each successive step of the brew rendering process occurs.

So the layer order is being determined by load order in this case, and can work for any custom future theme without having to manage a list here.

jeddai:
So with this layering piece, this effectively guarantees anything in the style tab takes precedence over anything set by defaults or themes?

5e-Cleric:
@jeddai 
> So with this layering piece, this effectively guarantees anything in the style tab takes precedence over anything set by defaults or themes?

From the CSS Documentation:
> Cascade layers (like declarations) are sorted by order of appearance, see [§ 6.4.3 Layer Ordering](https://w3c.github.io/csswg-drafts/css-cascade-5/#layer-ordering). When comparing declarations that belong to different layers, then for [normal](https://w3c.github.io/csswg-drafts/css-cascade-5/#normal) rules the declaration whose [cascade layer](https://w3c.github.io/csswg-drafts/css-cascade-5/#cascade-layers) is latest in the layer order wins, and for [important](https://w3c.github.io/csswg-drafts/css-cascade-5/#important) rules the declaration whose cascade layer is earliest wins.

Do notice that !important declarations in an EARLIER defined layer take precedence over !important declarations declared later (that's just how !important works), as writing !important declarations is a bad practice this is not a problem but i dare say an advantage.

jeddai:
Awesome, yeah this seems like it's going to be a great change. I'll code review and test in just a bit

ericscheid:
When testing, please check for any catastrophic failures with older browsers (e.g. Safari 12 on older iOS hardware) or non-Chrome browsers.

Hmm .. what is "older"? According to caniuse.com that would be "more than 11 months" — yikes!

(I don't anticipate any catastrophic fails with the current database of brews/styles. If I understand @layers correctly, only brews skipping the `.page` rule specificity are likely to rely on @layers, right?)

5e-Cleric:
> When testing, please check for any catastrophic failures with older browsers (e.g. Safari 12 on older iOS hardware) or non-Chrome browsers.
> 
> Hmm .. what is "older"? According to caniuse.com that would be "more than 11 months" — yikes!
> 
> (I don't anticipate any catastrophic fails with the current database of brews/styles. If I understand @layers correctly, only brews skipping the `.page` rule specificity are likely to rely on @layers, right?)

Afaik, that is correct.

5e-Cleric:
This feature also changes the need for .phb and .page in all snippets that inject code, they don't need to be updated, but they should be.

Issue #1899, opened 2021-12-17
Brew Themes


calculuschild:
Swappable brew themes will allow users to change documents from PHB style to MM, DMG, non-WOTC styles, and eventually custom and shared styles from other users.

### Built-in Themes

We already have the capability to live-swap style sheets (we do this to change between Legacy and V3), and we could get away with simply adding more styles to that list and swap them in a similar way. However, as the list of Theme grow, there are more special cases such as unique snippets, fonts, image resources, etc. that we may also need to consider part of the Theme. Hence, I propose that Themes consist of a set of multiple files linked together:

```
Theme 1 (i.e. DMG)/
 ├─ Settings.json	 : V3 or Legacy, Creator credit, Base Theme
 │                         (i.e. DMG is built on PHB, so default to Base Theme behavior where not defined)
 ├─ Styles.less		 : Relevant CSS
 ├─ Snippets.json        : List of snippets specific to this Theme. May overwrite snippets of Base Style
 └─ Widgets.json         : List of editor widgets, as proposed in #1870 
                           (i.e., "if this line has '{{monster', show checkbox to add/remove 'frame' class")
Theme 2/

Theme 3/

Fonts/
 ├─ Font1.woff2
 └─ Font2.woff2

Assets/
 ├─ Inkstain1.woff2
 └─ Inkstain2.woff2
```

The `widgets.json` might have a structure similar to the following. Clicking on a line that matches the `prefix` would signal to CodeMirror to pop up some widget (a color picker, checkboxes, number entry boxes, etc.):

```
{
  name         : 'monster',
  prefix       : '{{monster',
  widgets :  [
    {
      name : 'frame',
      type : checkBox,
      text : 'frame'
    },
    {
      name : backgroundColor,
      type : colorPicker,
      text : background-color:
    },
    {
      name : positionX,
      type : slider,
      text : left:
    }
  ]
}
```

However, getting Widgets doesn't necessarily need to come in the first release of this. That's a bonus feature that will be filled out afterward via #1870. It's just important to see how it will fit into the file structure.

Themes would be selected via a dropdown menu in the Metadata tab of the **Current Brew** that the user is currently editing, preferably with some sort of image preview while hovering over options, including perhaps a "blank" theme with no CSS applied at all except for page sizes and basic column behavior for those who want to fully build a theme from scratch in their Style tab. Any styles a user has in their "style" tab will be applied over the top of the selected theme as normal.

### Custom Themes

[...]

PR #2614, opened 2023-01-18
Theme's Previews

5e-Cleric:
Completes #1899 

There are four commits because i didn't know local.json went into config folder and therefore .gitignore, my bad.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -59,6 +59,12 @@ For a full record of development, visit our [Github Page](https://github.com/nat
 
 ### v3.6.0
 {{taskList
+##### calculuschild
+
+	* [x] Fix Google Drive brews sometimes duplicating
+
+Fixes issues [#2603](https://github.com/naturalcrit/homebrewery/issues/2603)
+
 ##### Jeddai
 
 * [x] Add unit tests with full coverage for the Homebrewery API
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -109,7 +109,7 @@ If you believe you should have access to this brew, ask the file owner to invite
 	excludePropsFromUpdate : (brew)=>{
 		// Remove undesired properties
 		const modified = _.clone(brew);
-		const propsToExclude = ['_id', 'views', 'lastViewed', 'editId', 'shareId', 'googleId'];
+		const propsToExclude = ['_id', 'views', 'lastViewed'];
 		for (const prop of propsToExclude) {
 			delete modified[prop];
 		}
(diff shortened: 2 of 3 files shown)


scripts/buildHomebrew.js
```diff
@@ -72,6 +72,7 @@ fs.emptyDirSync('./build');
 		themeData.path = dir;
 		themes.V3[dir] = (themeData);
 		fs.copy(`./themes/V3/${dir}/dropdownTexture.png`, `./build/themes/V3/${dir}/dropdownTexture.png`);
+		fs.copy(`./themes/V3/${dir}/dropdownPreview.jpg`, `./build/themes/V3/${dir}/dropdownPreview.jpg`);
```
G-Ambatte, line 75: Would there be benefit in creating a theme-specific asset folder (e.g. `./themes/V3/${dir}/assets/`) and moving that to the build directory, rather than listing each item to be copied individually?
I suspect ultimately it doesn't matter one way or the other right now, but it may simplify adding images to additional themes in the future.

5e-Cleric: That would make sense, there are only 4 themes now but it may be expanded, and there is no reason against it.

calculuschild: Yeah, eventually that can be done. Where there are some overlaps between assets I just had them all in a root "assets" folder. How about we re-evaluate the folder structure once we have a few more themes on hand. We might see some patterns of common files.

calculuschild (changes requested):
This looks pretty good! Thanks!

Just a few comments on the preview images:
- I think we should choose a generic title in the H1 title, rather than "Dungeon Master's Guide Theme", just to give ourselves a bit of distance from WOTC copyright.
- Let's also replace the cat image with our Homebrewery mug from the home page.
- Remove the "THIS IS A CAT WARRIOR" header. Instead, somewhere add an H3 header that shows the underline and some text under.
- Can we move the watercolor stains away from the first paragraph so it doesn't blur with the Drop caps?
- Change the watercolor stain colors slightly depending on the background to look prettier. Remove it from the "blank" theme entirely.
- Shorten the monster block if room is needed to fit more text to space items out.
- Can we get a shot that doesn't show the weird box artifact around the drop caps? I think this tends to occur more in PDFs than in the actual site.

5e-Cleric:
I closed this by error while trying to fix a few problems with my fork, will reopen as soon as i can remake the previews following your guidelines. 
Will get to it tomorrow, at 8 or 9am CST, if you can do it before, even if it's only some of them, feel free to do so.

5e-Cleric:
> Can we get a shot that doesn't show the weird box artifact around the drop caps? I think this tends to occur more in PDFs than in the actual site.

That is correct, that happens because i downloaded the brews as pdfs and then converted them into jpgs, because i figured that would be easier than trying to crop multiple images to the pixel.

calculuschild:
I gotcha. In that case I think I'd still prefer a "glitch-free" picture that is a few pixels off. By the time it's shrunk down I don't think it makes much difference.

G-Ambatte:
Can always do a `Capture node screenshot` from the Inspector tool from the PrintPage, that's probably the fastest way and skips the whole PDF>JPG step.

From [this sample brew](https://homebrewery.naturalcrit.com/share/u0qzpes4Pig9) at 80% browser zoom so it would all fit on the screen at once, I captured these results, changing only the Theme between each capture:

PHB:
homebrewery naturalcrit com_print_u0qzpes4Pig9_dialog=true

DMG:
homebrewery naturalcrit com_print_u0qzpes4Pig9_dialog=true (1)

Journal:
homebrewery naturalcrit com_print_u0qzpes4Pig9_dialog=true (2)

Blank:
homebrewery naturalcrit com_print_u0qzpes4Pig9_dialog=true (3)

5e-Cleric:
That's great, did you add it to the code or you want me to? it'll take like a minute

PR #2623, opened 2023-01-22
Cloning views fix

5e-Cleric:
Right now view count gets cloned with the brew contents, this takes care of that, artifically setting views to 0 when a cloned brew is created.


calculuschild:
This is good, although it includes a lot of extraneous changes from your other PRs as well. Can you remake this making sure to use the current Master branch as the base?

5e-Cleric:
> This is good, although it includes a lot of extraneous changes from your other PRs as well. Can you remake this making sure to use the current Master branch as the base?

wow, must've pushed changes in the wrong branch, let me fix that

5e-Cleric:
i erased my other commits and merged master into it, that's it, right? @calculuschild

PR #2625, opened 2023-01-23
Windows install script using Chocolatey

G-Ambatte:
This PR adds a Windows Powershell script that downloads and installs the Homebrewery perquisites and the project itself.

- [x] install.ps1
- [x] README.WINDOWS.md


calculuschild:
I'm just going to trust that this works since it doesn't look like it impacts the live site in any way. @G-Ambatte you get to be the point-of-contact if someone has issues getting this to work.

One thought (can be a separate PR). Would it be possible to create a shortcut to a batch file or something that the user can double-click next time they want to run, instead of trying to remember to type `npm start`?

G-Ambatte:
> I'm just going to trust that this works since it doesn't look like it impacts the live site in any way. @G-Ambatte you get to be the point-of-contact if someone has issues getting this to work.
> 
> One thought (can be a separate PR). Would it be possible to create a shortcut to a batch file or something that the user can double-click next time they want to run, instead of trying to remember to type `npm start`?

Created a new issue to track this.

PR #2649, opened 2023-01-27
Cover snippet

5e-Cleric:
This is the initial draft for the snippet, here a preview:

image

Updated to the last commit, svg images, and NaturalCrit's Logo


themes/V3/5ePHB/style.less
```diff
@@ -677,6 +677,135 @@ body {
...
 }
+//*****************************
+// *      COVER PAGE
+// *****************************/
+
+.page {
+	&:has(.coverPage):after {
+		display: none;
+	}
+	.coverPage {
```
Gazook89, line 688: I wonder if it'd be a good idea to include the class name "default", and set the css to target `.page .default.coverPage`.  This way someone could still use the class name `coverPage` themselves and completely ignore the default styling without having to unset/override a bunch of existing properties.....and then if they want to have the default styling they just add `default`.  

I don't know if that makes much sense, though.  Especially with `@layers`, maybe that makes this a moot point.

themes/assets/assets.less
```diff
@@ -10,6 +10,9 @@
 @monsterBorderImage       : url('/assets/monsterBorderFancy.png');
 @codeBorderImage          : url('/assets/codeBorder.png');
 @classTableDecoration     : url('/assets/classTableDecoration.png');
+@coverPageLogo            : url('/assets/coverPageLogo.png');
+@coverPageBanner          : url('/assets/coverPageBanner.png');
+@coverPageHorizontalRule  : url('/assets/coverPageHorizontalRule.png');
```
Gazook89, line 15: I think this should be edited to remove the existing drop shadow.  Since the other two graphic elements have no drop-shadow baked into them, this one should match that expectation.  Then, as we do with the others, apply the drop-shadow with css.

(or the other way around-- bake the drop-shadow into these files and remove the css).

calculuschild: Agreed. Let's keep the graphics clean where possible and apply effects via css

themes/V3/5ePHB/style.less
5e-Cleric: That could be a thing, but it's the same as users using `hardCover` or `myCover` instead of `coverPage`, which they will, because they won't know they can use `coverpage` without importing styles.

themes/assets/assets.less
5e-Cleric: so, removing the drop shadow from the horizontal rule? that'd be no problem

5e-Cleric: @calculuschild @Gazook89 should i make them vectorial? that would make them use less space, right?

calculuschild: Yes, if they appear the same and don't have any weird behavior with positioning and sizing compared to pngs.

5e-Cleric: they appear fine and behave as a png would.

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
```
calculuschild: Instead of wrapping the whole coverpage in one big DIV, can we make more use of `:has`? As in, just put `{{coverPage}}` at the top of the page, and that when that class is detected, the whole page will be put into "coverPage" mode.

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
+		> p > img {
+			position: absolute;
+			inset: 0;
+			height: 100%;
+			min-width: 100%;
+		}
+		&:before {
+			background-image: @homebreweryLogo;
+			background-position: center;
+			background-size: 75px;
```
calculuschild: In general, let's try to set absolute sizes using cm instead of px. We haven't been perfect about this, but the basic rule we've been following is:

- **cm** for sizes related to the page (width of an image, font size, padding, etc.)
- **px** for sizes related to CSS filters and specific pixel cutoffs (borders, shadows)

In the same vein, we try to avoid `em` for setting font sizes, and instead use `cm`. EM is ok for line heights (i.e., when a line is 1.2 times the height of the font, use `em 1.2`), but we've run into a lot of issues in the past where we change one font size and the whole document breaks.

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
+			--shadow-x3: var(--shadow-x2), var(--shadow-x2), var(--shadow-x2);
+			text-shadow: var(--shadow-x3), var(--shadow-x3), var(--shadow-x3);
+			text-transform: uppercase;
+			max-width: 18ch;
+			margin-top: 3px;
+			color: white;
+			font-family: NodestoCapsCondensed;
+			font-size: 6.6em;
+			font-weight: normal;
+			line-height: 72px;
```
calculuschild: Use `cm` for the font size and then `em` for the line height.

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
+			filter: drop-shadow(0 0 5px black);
+		}
+		h1 {
+			--shadow-x0: #000 0px 0px 3.5px;
+			--shadow-x1: var(--shadow-x0), var(--shadow-x0), var(--shadow-x0);
+			--shadow-x2: var(--shadow-x1), var(--shadow-x1), var(--shadow-x1);
+			--shadow-x3: var(--shadow-x2), var(--shadow-x2), var(--shadow-x2);
+			text-shadow: var(--shadow-x3), var(--shadow-x3), var(--shadow-x3);
+			text-transform: uppercase;
+			max-width: 18ch;
```
calculuschild: Use `cm` instead of ch.

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
+			margin-top: 3px;
+			color: white;
+			font-family: NodestoCapsCondensed;
+			font-size: 6.6em;
```
calculuschild: `cm`

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
+			font-weight: normal;
+			line-height: 72px;
+			word-break: break-word;
+			transform: scaleX(0.94);
+		}
+		hr {
+			position: relative;
+			top: -16px;
```
calculuschild: cm

themes/V3/5ePHB/style.less
```diff
@@ -677,6 +678,136 @@ body {
...
+			--shadow-x3: var(--shadow-x2), var(--shadow-x2), var(--shadow-x2);
+			text-shadow: var(--shadow-x3), var(--shadow-x3), var(--shadow-x3);
+			font-family: NodestoCapsCondensed;
+			font-weight: normal;
+			font-size: 2.5em;
+			letter-spacing: 0.5px;
+			max-width: 46ch;
+			color: white;
+		}
+		.banner {
```
calculuschild: Can the banner be made to work as a span, like `{{banner 5e Homebrew}}`?

themes/V3/5ePHB/snippets/coverpage.gen.js
```diff
@@ -1,124 +1,123 @@
...
+## ${_.sample(subtitles)}
+___ 
+
+{{banner
+5e HOMEBREW
+}}
+
+{{footnote
+Everything a Homebrew needs to create a pretty document <br> for the world's greatest roleplaying game
+}}
```
calculuschild, line 118:
```suggestion
{{footnote
Everything a Homebrew needs to create a pretty document\
for the world's greatest roleplaying game
}}
```

themes/V3/5ePHB/snippets/coverpage.gen.js
```diff
@@ -1,124 +1,123 @@
...
-  .page:nth-child(2n+1) .pageNumber { right: inherit !important; left: 2px !important; }
-  .page:nth-child(2n)::after { transform: scaleX(1); }
-  .page:nth-child(2n+1)::after { transform: scaleX(-1); }
-  .page:nth-child(2n) .footnote { left: inherit; text-align: right; }
-  .page:nth-child(2n+1) .footnote { left: 80px; text-align: left; }
-</style>
+{{coverPage
 
-{{margin-top:225px}}
+![image](https://i.imgur.com/Mqx8Vf7.png) {left:0px}
```
calculuschild, line 105: Can we make the `{left:0px}` similar to the `imageMasks` snippet? That is, make the image at `bottom:0; left:0;` in the CSS by default, then put `height: 100%` here? Just for consistency between snippets.

calculuschild (changes requested):
Very nice work. I suggest just some CSS cleanup and editing the font files if you are able to do that.

themes/V3/5ePHB/snippets/coverpage.gen.js
```diff
@@ -1,124 +1,123 @@
...
-  .page#p1{ text-align:center; counter-increment: none; }
-  .page#p1:after{ display:none; }
-  .page:nth-child(2n) .pageNumber { left: inherit !important; right: 2px !important; }
-  .page:nth-child(2n+1) .pageNumber { right: inherit !important; left: 2px !important; }
-  .page:nth-child(2n)::after { transform: scaleX(1); }
-  .page:nth-child(2n+1)::after { transform: scaleX(-1); }
-  .page:nth-child(2n) .footnote { left: inherit; text-align: right; }
-  .page:nth-child(2n+1) .footnote { left: 80px; text-align: left; }
-</style>
+{{coverPage
```
calculuschild, line 103: remove the extra newlines here. Just start with ``return `{{coverPage``

Actually, you can use

```
return dedent`
   XYZ
   ABC
   123`
```

and it will let you use the indentation in the source code without looking ugly in the resulting snippet. See the `imageMask.gen.js` file.

themes/V3/5ePHB/style.less
5e-Cleric: that is a good idea, and then all props can be set as `.page:has(.coverPage) hr` and so on (when processed)

to be hones with you, i don't really see the difference, i would argue that having the entire cover be an element the user understands easier what components make the cover.

calculuschild: In short, using `:has` ensures that the *whole page* is the cover. Making it a Div means it's easy for a user to accidently put components outside of the Div.

themes/V3/5ePHB/snippets/coverpage.gen.js
5e-Cleric: this was already done, `inset: 0;` is the shorthand for bottom, left, right and top, and height was set to 100%.

themes/V3/5ePHB/snippets/coverpage.gen.js
5e-Cleric: i'm making changes to this text, and will just let it break naturally.

5e-Cleric:
> Very nice work. I suggest just some CSS cleanup and editing the font files if you are able to do that.

what changes in the fonts?

Issue #2650, opened 2023-01-28
Cloning copies published state and last edited metadata


5e-Cleric:
### Renderer, browser and OS
v3, Chrome, Windows

### Description
Related to #2623, cloning also copies the last edited date, created date, and the published state, i believe this is a bug.

You can recreate this bug at any time by cloning any brew.

Desired behavior:
Cloning a brew should make it its own brew, with new created and updated brew, without it's view count or published state, only the content and therefore its page count should remain.

PR #2651, opened 2023-01-28
fix created, updated published last viewed

5e-Cleric:
Fixing #2650


calculuschild:
Thanks @5e-Cleric for noticing the bug and providing a fix!

Issue #2280, opened 2022-08-17
The eñe character in ScalySansSmallCapsRemake (V3) is borked


ericscheid:
ScalySansSmallCapsRemake is used in V3 for H5 headings.

In Legacy, the font-face is ScalySansSmallCaps, and the glyph is correct.

<img width="970" alt="image" src="https://user-images.githubusercontent.com/2605467/185114940-c035f7be-09ea-4d7b-920f-3d8629dec9a2.png">

Put this into a V3 brew..
````
##### Gravedad del Daño in V3

##### Gravedad del Daño in Legacy
{font-family:ScalySansSmallCaps}

<style>
/* copied from 5ePhbLegacy.style.css */
@font-face {
  font-family: ScalySansSmallCaps;
  src: url('../fonts/5e legacy/Scaly Sans Caps.woff2');
  font-weight: normal;
  font-style: normal
}
</style>

````

5e-Cleric:
Releasing a PR to fix this right now.

calculuschild:
We definitely do not want to use the Legacy font in V3. The V3 font fixes several other issues with that existed in the Legacy font. Just fixing one character is not a good enough tradeoff.

What we actually want to do is go into the V3 font and just add the missing character.

PR #2661, opened 2023-02-04
U+00D1 and U+00F1 fix

5e-Cleric:
Fixes https://github.com/naturalcrit/homebrewery/issues/2280 by using the legacy font file in v3 too.


calculuschild:
We definitely do not want to use the Legacy font in V3. The V3 font fixes several other issues with that existed in the Legacy font. Just fixing one character is not a good enough tradeoff.

What we actually want to do is go into the V3 font and just add the missing character.

5e-Cleric:
>  The V3 font fixes several other issues with that existed in the Legacy font. 

did not know that! well then i'll add the missing characters.

What were the other issues? so i can avoid making them.

calculuschild:
Issues included:

- Badly-formed glyphs (self-intersecting paths, etc)
- Inconsistent character size
- Inconsistent character spacing
- Uneven vertical alignment
- Other missing characters
- Invalid metadata

As long as you just add the new characters and don't touch the others, it should be fine. Probably easiest is to just duplicate the normal 'n' so it's already sized and spaced, then put the tilde on top.

calculuschild:
Oh also, check the new font wife by wife with the original v3 font. I've noticed some font editors re-encode things and it makes the characters thicker than before.

5e-Cleric:
For the last commit i took the v3 font and manually created the following glyphs: Ñ, ñ, ç, Ç, Ý, ý, #, ^, ¿, ', ", ¡, ·

The characters do not seem to have gotten thicker.

calculuschild:
Perhaps if you can take a screenshot of the two fonts side by side to make it easier to compare and make sure nothing got broken.

5e-Cleric:
Wait, there is an error in the font metadata, bc of the editor

5e-Cleric:
image
In the left(Ñ and first A) the new version, in the right, the V3 one. they are exactly the same

calculuschild:
Can you screenshot the whole alphabet in use on the website? To compare font letter spacing and sizing?

5e-Cleric:
The next step would be to add those characters in italics, bold and bold italics, but i don't think that is as prioritary, as per the dnd style that font is never used in caps or in italics.

5e-Cleric:
> Can you screenshot the whole alphabet in use on the website side by side with the original? To compare font letter spacing and sizing?

You mean a document with lorem ipsum? sure

ericscheid:
> > Can you screenshot the whole alphabet in use on the website side by side with the original? To compare font letter spacing and sizing?
> 
> You mean a document with lorem ipsum? sure. But with special characters?

Don't need filler text, do need letters for comparison:

````
{{wide
ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz (V3)
ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz (new font)

ÅÑÜØetc (V3)
ÅÑÜØetc (new font)
}}
````

5e-Cleric:
image

calculuschild (changes requested):
There appears to be some strange spacing issues on the new characters. For instance I would expect Ñ and N to have the same spacing between adjacent characters:

image

This test file is here in case it's helpful to copy it for testing: https://homebrewery.naturalcrit.com/share/2bpr4wLt4Knu Note that not all characters need to match necessarily (\* and ^ for example), but at least for the alphabetic characters there should be more consistency.

5e-Cleric:
fixed it, should have the exact same width now, most of them.

Note: italics, bold and bold italics should also be changed, but unless someone wants to pick it up from here, i am done with editing fonts.

calculuschild:
Something is still off. It looks like the characters are not centered properly.

image

**Edit:** I had some time and went ahead and fixed it. Looks like the Left and Right Side Bearings were not quite right.

image

calculuschild:
Thanks @5e-Cleric ! 💯 :tada: ⭐

PR #2681, opened 2023-02-20
[EDITOR] Fix div links #2680

### LINKED ISSUE #2680: [EDITOR] Table of Contents internal links don't work in <div>
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

Generating a table of contents from the snippet in Legacy results in a section of code that does not function correctly if the brew is later transitioned to V3.

Digging deeper for a minimal reproduction, this bug affects any link inside a `<div>`. Crucially, this bug does NOT appear if there is a line between the opening `<div>` line and the line the link is on.

Investigation has shown that the V3 Markdown `html` block processing code returns the internal text of the block after being parsed by Marked - but crucially, this parsed data is rendered by the default renderer, NOT the custom renderer.

It appears that the solution is to change `/shared/naturalcrit/markdown.js:L14` from:

    return `${openTag} ${Marked.parse(html)} </div>`;

to:

    return `${openTag} ${Marked.parse(html, { renderer: renderer })} </div>`;

---

Discovered while investigating this issue on Reddit: https://redd.it/1158cwd

---

Minimal issue reproduction code is attached.

### Code

```gfm
<div>
[Has no _self Attribute](#p1)
</div>
```

G-Ambatte:
This PR fixes #2680.

This PR adds the renderer option to the HTML block function in the V3 Markdown processor, which causes internal links in `<div>` blocks to be processed correctly - specifically, the output has the `target=_self` attribute.


G-Ambatte:
I realized that it would be better to set the renderer in the Marked options, so any future iteration that implements a `Marked.parse` call will automatically use the correct renderer.

Issue #2687, opened 2023-02-23
mug image on home page is `http`


ericscheid:
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

The mug image in the example brew of the home page content is specified with an `http` url, not `https`. 

<img width="861" alt="image" src="https://user-images.githubusercontent.com/2605467/220793948-1e6a0cd9-72fc-426b-9bc9-ce1a036c9c77.png">

A warning appears in the console:

<img width="844" alt="image" src="https://user-images.githubusercontent.com/2605467/220793767-0bd9e2b9-5965-42d4-8089-68cc7cb338e6.png">

## Solution: 

edit the url to use `https`

### Code

```gfm
![homebrew mug](https://i.imgur.com/hMna6G0.png) {position:absolute,bottom:20px,left:130px,width:220px}
```

G-Ambatte:
`/client/homebrew/pages/homePage/welcome_msg.md` line 39:

![image](https://user-images.githubusercontent.com/19826659/220794308-2870a32f-cc48-4a09-b7cb-8f77187449c8.png)

Solution: change `http:` to `https:`

PR #2694, opened 2023-02-25
Use secure protocol for Homebrewery mug image

MichielDeMey:
Fixes https://github.com/naturalcrit/homebrewery/issues/2687

### DIFF
--- a/client/homebrew/pages/homePage/welcome_msg.md
+++ b/client/homebrew/pages/homePage/welcome_msg.md
@@ -36,7 +36,7 @@ After clicking the "Print" item in the navbar a new page will open and a print d
 If you want to save ink or have a monochrome printer, add the **PRINT → {{fas,fa-tint}} Ink Friendly** snippet to your brew!
 }}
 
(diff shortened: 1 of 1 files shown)


MichielDeMey:
Looks like some flakey tests. 👀

calculuschild:
Thanks for this. Yes, there is one test that sometimes times out before the test environment is fully spun up. If you have circleci rerun the test it should work.

Also if you have any ideas besides just extending the timeout limit.

MichielDeMey:
Looks like I don't have the required permissions to re-run the test suite.

In addition, I created a trace of `require('app.js')`.
This should give us some insight into what it's loading and where most time is spent: 
https://chromedevtools.github.io/timeline-viewer/?loadTimelineFromURL=https://gist.githubusercontent.com/MichielDeMey/a7b6621d2089c2b91cb90a394f00d32e/raw/0b565eb1c9480bc725e42f3b419455449f7204d8/app.js.trace

MichielDeMey:
I've also gone ahead and created a draft PR to address the app startup time which might help with the test flakiness: https://github.com/naturalcrit/homebrewery/pull/2696

PR #2696, opened 2023-02-26
[Draft] Trimming Google APIs

MichielDeMey:
**NOTE**: Unfortunately, I'm unable to test the Google Drive integration due to the external naturalcrit login page. Anyone able/willing to help me out?

---

This PR is an attempt at trimming the Google APIs package used since we only use the Google Drive API.

Taking into account that the Google APIs account for about half the `require('app.js')` time, this is a significant speedup.

This reduces the require time of the Google API package by a factor of 10.

Before (**649ms**):
Screenshot 2023-02-26 at 14 31 25

After (**64ms**):
Screenshot 2023-02-26 at 14 31 55

Ultimately, this might also help with test flakiness.


server/googleActions.js
```diff
@@ -136,7 +134,7 @@ const GoogleActions = {
 	},
 
 	updateGoogleBrew : async (brew)=>{
```
MichielDeMey, line 136: Is there a specific reason we're not passing the OAuth2 client in the params as `auth` as we're doing with the other functions? e.g. `newGoogleBrew(auth, brew)`, `listGoogleBrews (auth)`

calculuschild: Yes, there is a reason.  At the top of this file we already set the `serviceAuth` which makes it the default for all of the api requests (line 25 above which you deleted). The only cases where we pass in a separate `auth` is when we want to require the specific user to be logged in.

For example, anybody with editor access to the file can update it (uses `serviceAuth`, no need to pass in `auth`). However only the document owner themselves is allowed to delete the file (we validate their user credentials and pass in that OAuth2 token instead).

So this change: 

```
const drive = drive({ version: 'v3', auth: serviceAuth || config.get('google_api_key') });
```

Should be undone, since the default was already being set with

```
google.options({ auth: serviceAuth || config.get('google_api_key') });
```

calculuschild:
> **NOTE**: Unfortunately, I'm unable to test the Google Drive integration due to the external naturalcrit login page. Anyone able/willing to help me out?

Just undo the auth changes and I can test this. Thanks for looking for a soution!

calculuschild:
Hi @MichielDeMey, I'd like to get this one merged. Would you be able to apply the requested changes, or should I take care of it for you?

server/googleActions.js
MichielDeMey: Well, the global `google.options` is no longer available when using the Google Drive specific package. So I had to remove that line.

I believe the only way around this is by passing the auth to the `drive()` constructor as an option, which I've done in the auth commit.

MichielDeMey:
@calculuschild Apologies, I got sidetracked. 😅

server/googleActions.js
calculuschild: I see. In that case, let's just put an object up top to hold our default settings, and then pass in that object into the `drive()` constructor where needed. Just so all of our default options are handled in one place.

Something like:

```
defaultAuth = {serviceAuth || config.get('google_api_key')};


//
//
//

const drive = drive({ version: 'v3', auth: ...defaultAuth });

MichielDeMey: @calculuschild Implemented your suggestion 👌

calculuschild:
Had to tweak things slightly but I think its working now. Thanks! Seems to run the tests a bit faster but we'll have to see if this is enough.

Thanks! 💯 🎉 ⭐

PR #2710, opened 2023-03-01
Updated Google Drive icon

MichielDeMey:
Google updated its product icons in 2020, this PR reflects the new icon used.

I've also taken the liberty to replace the PNG file with a smaller SVG
I've also gone ahead and used a CSS filter instead of a separate image for the inactive state

Active
localhost_8000_edit_AfR6BrakRlgo

Inactive
localhost_8000_edit_AfR6BrakRlgo (1)


Gazook89:
I like it.  One question, I really don't know the answer:  Is it better to have an inline `filter` style or to set the class to `inactive` and then add css styling to the site's stylesheet for `.... .inactive`?   I don't have a preference either way, just wondering if anyone else does.

5e-Cleric:
> I like it.  One question, I really don't know the answer:  Is it better to have an inline `filter` style or to set the class to `inactive` and then add css styling to the site's stylesheet for `.... .inactive`?   I don't have a preference either way, just wondering if anyone else does.

Best practice would be separation of code, it would be better to have the css, or less, separated from the markup, and avoid inline.

In regards to size, optimisation and speed, just as fast, though it may be faster to have in-line, depending on how are the markup and styling compiled.

MichielDeMey:
I've gone ahead and moved the styling to the separate stylesheet. 👍

calculuschild:
Thanks @MichielDeMey ! 🎉 💯 ⭐

PR #2711, opened 2023-03-01
Back cover snippet

5e-Cleric:
preview:

image


5e-Cleric:
I believe this should be done now, this is the best i can do without being able to use local HB for now

calculuschild:
Thanks @5e-Cleric ! 🎉 💯 :star2:

PR #2714, opened 2023-03-02
Inside Cover Snippet

5e-Cleric:
Preview vs PHB:
image
image
</div>


themes/assets/assets.less
```diff
@@ -13,6 +13,7 @@
 @naturalCritLogo          : url('/assets/naturalCritLogo.svg');
 @coverPageBanner          : url('/assets/coverPageBanner.svg');
 @horizontalRule           : url('/assets/horizontalRule.svg');
+@insideCoverMask          : url('/assets/insideCoverMask.png');
```
calculuschild, line 16: We can add some more masks for this. Currently we have "edges" and "corners", but we can add a "centers" mask set and then use that here.

5e-Cleric: thought about it, flamableconcrete made some

calculuschild: Yes. He gave me all of his original files. I can try to put smaller PR together that just adds a few of those.

Gazook89: I thought the corner masks we had were already just circles, pushed to the corners?  Meaning the corners and 'center' masks could just be the same file.

5e-Cleric: I believe they are corners but mirrored for use in all 4 corners, but they look awful in the center

calculuschild: @5e-Cleric is correct. They are the corner images just mirrored four ways so they can slide into any corner. If somebody wants to edit them up so they blend more seamlessly feel free to do so. But they are also not full-page masks either.

For this inside cover we need a larger mask anyway. Unfortunately the "circle" ones from flamableconcrete are pretty small and look like they might be some early attempts because they all use a single brush which doesn't look great:

(I call this one "the sawblade")
image

Anyone want to make a set of like 10 or more "page-width blob" masks for me and I'll throw them into a snippet?

5e-Cleric: > Anyone want to make a set of like 10 or more "page-width blob" masks for me and I'll throw them into a snippet?

i have the appropiate programs and brush set ready on my computer, I will try to do some tomorrow and edit this comment whether I succeed or not.

calculuschild:
Blocked by #2741

calculuschild:
Ok, I think we're good here! Thanks @5e-Cleric ! 🎉 ⭐ 💯

PR #2723, opened 2023-03-09
Fix Injection Blocks Crashing to White Screen

### LINKED ISSUE #2712: Brew crashes due to multiple text formatting tags
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

### Code

```gfm
##### [Grievous Wounds](#p27)
{text-align:center}
```

Gazook89:
This PR fixes #2712. 

I'm not going to give a big description yet, just putting it up here so someone can look at it.

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -137,14 +137,13 @@ const mustacheInjectInline = {
 			if(!lastToken)
 				return false;
 
-			const tags = ` ${processStyleTags(match[1])}`;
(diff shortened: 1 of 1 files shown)


Gazook89:
I have a better solution (is *actually* the way it is supposed to be done).  I want to fix a markdown highlighting error, though, and kid just woke up from nap.  so later tonight i'll push that up probably in new PR.

Gazook89:
closing this in favor of a better PR.

Issue #2505, opened 2022-11-14
Fix a few Markdown tests


Gazook89:
I was just looking at some tests in `tests/markdown/mustache-span.test.js` and there are a handful with `FIXME` tags that seem like they are fixed.  "Testing" isn't something I've really looked at until now, so before I submit a PR to just clean these up I thought I'd check my work.  If I remember right, @AlexeySachkov is the person to talk about tests with.......?

```js
test('Renders a mustache span with text and class', function() {
	const source = '{{my-class text}}';
	const rendered = Markdown.render(source);
	// FIXME: why do we have those two extra spaces after closing "?
	expect(rendered).toBe('<span class="inline-block my-class"  >text</span>');
});
```

There are 4 or 5 similar tests that have the same or similar "FIXME" tags about extra spaces, and it looks like the expected result in the test has just been changed to match the undesired behavior rather than fixing whatever is causing the spaces.  But on the live site those extra spaces *aren't* present....so not sure how it passes this test since it is expecting the spaces to be there?

Gazook89:
Hold the phone....I guess maybe the browser dev tools is just automatically removing those spaces so I think they aren't there.  But when I "fix" the test, it fails.  

Okay.

AlexeySachkov:
Hi @Gazook89,

> I guess maybe the browser dev tools is just automatically removing those spaces

Yes, most likely this is what happening.

> There are 4 or 5 similar tests that have the same or similar "FIXME" tags about extra spaces, and it looks like the expected result in the test has just been changed to match the undesired behavior rather than fixing whatever is causing the spaces. 

My aim back than was to simply at least enable some testing and I decided that a couple of extra spaces which are not user-visible (i.e. they are in between tag arguments, not inside its "body") not require immediate bug fix. We might even decide that it doesn't worth fixing at all

PR #2724, opened 2023-03-10
Fix Injections from Crashing, and Unit Tests

### LINKED ISSUE #2712: Brew crashes due to multiple text formatting tags

Gazook89:
This fixes #2712.   It basically reduces the differences between the mustacheInjectInline and mustacheInjectBlock methods that we've had for awhile.  No more crashing to white screen. 

I still need to add unit tests as suggested by @calculuschild , and also I would like to fix the injectBlock syntax highlighting-- currently every injection is highlighted if it's at the start of a new line.  Ideally, it should not highlight if the preceding line is just another injection by itself.  I think this can be done with a negative lookbehind, but my understanding is that is not supported by Safari yet (though may be soon, next 2 months?).  I also thought about setting a variable to true/false that is updated with each pass of the highlighter on each injection, so that successive injections are ignored-- but that would require a good bit more lines and I couldn't use the same regex for inline and block injections (i think).


ericscheid:
Another syntax highlighting error:
image

Gazook89:
> Another syntax highlighting error:  (highlighting inside code fences)

I just pushed a fix for this-- no highlighting inside code fences.  However, it adds a layer of nesting for this logic, and exceeds the nesting limit we have set for the repo.  I'm not sure how to avoid this in a way that is less messy.  Here is basically what I added:

```js
let codeFence = false;

.......

if(line.match(/^```$/)){
		codeFence = !codeFence;
};

if(codeFence == false){
	...all the existing highlighting Logic....
}
```

ericscheid:
> if(line.match(/^```$/)){

code fences can start with 4 backticks, and can also have a code-language string.

Gazook89:
@ericscheid  --- what about `/^`{3,4}[A-Za-z]{0,5}/m`

Gazook89:
## Original Crashing Problem Fixed
Just an update on this...i think the original problem is fixed.  Successive injections do not crash the app to a white screen, and are left as plain text in the brew.  This works with inline injections as well (which in fact doesn't have an issue to begin with).
```
{{ text}}
{color:red}
{color:blue}
```
comes out to: 
image

## Syntax Highlighting
This PR also includes a little detour where I removed syntax highlighting inside code fences, described in the previous comment.  

I *have not* fixed the highlighting of successive block-injections.  I think I need to replace the `.forEach()` loop with a normal `for` loop that wraps all of the line-by-line highlighting, so that each line can be seen in the context of the rest of the lines as we `markText()`.  **I'd like to do this in a different PR, though, and likely redo the code fence highlighting**.
image

## Unit Tests
Unit tests have been added for Injection syntax.  I still have some more to write, particularly related to Block injections.  At the same time, I have grouped the mustache-syntax tests into groups (using `describe([name], ()=>{[tests]}`), so that Inline, Injection, and (when added) Block syntaxes can be tested separately.  I also added a package.json script to run those individual groups.  I just got tired of seeing irrelevant tests.

I'll add Block tests, too.

ericscheid:
> ```
> {{
> text
> }}
> {color:red}
> {color:blue}
> ```

Quick question: what is the intended effect of that brew code (assuming the crashing is averted)?

Is the block `text` meant to have both injections applied (which in this case means the element might well render with `style="color:red; color: blue;"` and thence styled by the browser as simply `color:blue`). If the second injection were to instead be `{font-weight:bold}`, would or should `text` be red and bolded?

ericscheid:
> @ericscheid --- what about
> 
> ```js
> /^`{3,4}[A-Za-z]{0,5}/m
> ```

or simpler

```js
/^`{3,4}[A-Z]*/im
```

Actually, no .. the `<code>` render does accept other characters for the language token, such as the underscore `_`. It will generally ignore anything after a space ` ` (on the fence line). I suppose one could delve into the `<code>` rendering code for the actual criteria, or more simply just have:

```js
/^`{3,4}.*/m
```

or maybe even 

```js
/^`{3,4}/m
```

calculuschild:
> Quick question: what is the intended effect of that brew code (assuming the crashing is averted)?

Our current interpretation is that an injector can't be chained onto another injector. So only the first gets applied, and the second appears as plain text in the document since it would be considered invalid syntax.

That may not be what happens live right now, but this PR should make it consistent and adds unit tests clarifying the desired behavior.

Gazook89:
Highlighting issues i'm just going to leave for another PR-- hopefully it's okay to leave what I have for code fences in this one, understanding it could be improved later (maybe it's naive to think this is a niche token).

Another issue:
```md
- Talk to Groupie
- Astonishing Chant of Chocolate
- Alchemical Evocation of the Goths
- Extra-Planar Spell of Irritation
- Occult Globe of Salad Dressing
{color:red}
```
The above injects the style into the last `<li>` element, which I didn't expect.  I would have expected that `li` would be an inline token, and *the style would be applied to the `<ul>` block element.
```md
- Talk to Groupie
- Astonishing Chant of Chocolate
- Alchemical Evocation of the Goths
- Extra-Planar Spell of Irritation
- Occult Globe of Salad Dressing{color:red}
```
This tokenizes the injection but doesn't apply the style to any element.  It essentially just disappears.
```md
- Talk to Groupie
- Astonishing Chant of Chocolate
- Alchemical Evocation of the Goths
- Extra-Planar Spell of Irritation
- Occult Globe of Salad Dressing

{color:red}
```
The same happens here-- the injection disappears without being applied to anything.

This is all true on the live site right now.  Is this worth examining?  I think it would be tricky because `ul` is handled by the default marked.js stuff, so this wouldn't be a simple "fix" like I'm doing with our custom spans and divs.

Gazook89:
Also fixes #2505

I also see this failed some circleCI stuff. I’ll come back to it

dbolack-ab:
Still a going concern?

Gazook89:
I'm going to close this PR.  The original issue, the crashing, is fixed.  Also in this PR I tried to address an issue with code highlighting within `pre` blocks, but that is also no longer an issue on the live site.  And it looks like along the way additional tests were added via other PRs-- I'm not sure what overlap there is with the tests in this PR and the current tests, but I think this PR is sufficiently stale that I'm not going to re-animate it (trying to merge `master` into it gave me 6 conflicted files-- blech).

PR #2730, opened 2023-03-13
Run Tests sequentially

calculuschild:
One possible solution to tests timing out on CI

https://jestjs.io/docs/troubleshooting#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server

Also raising the timeout limit just 'cuz.

I also noticed the `test:coverage` option re-runs all the tests a second time. Perhaps it's worth just having circleci do `run:test --coverage --runInBand` instead of breaking them all out into separate steps so we can test coverage at the same time as everything else in one pass. @jeddai ?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
       - image: cimg/node:16.11.0
 
     working_directory: ~/homebrewery
-    parallelism: 4
+    parallelism: 1
 
     steps:
       - attach_workspace:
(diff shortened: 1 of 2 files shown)


Gazook89:
Will this fix my circle CI issues on my pr?

calculuschild:
> Will this fix my circle CI issues on my pr?

I was already able to make it pass by just re-running the test. But hopefully this will make manually re-running unnecessary in the future.

PR #2733, opened 2023-03-14
Add Metadata Nav Item to SharePage

### LINKED ISSUE #1679: Add ability to view brew metadata on Share Page // Add "brew version" metadata field
The metadata for a brew is not viewable on a Share Page, but may contain important information for a reader.  Such information as Author(s), Systems, Description would be helpful.   

A new metadata field for "Version" could be helpful, too, such that an author could make it clear if there are revisions (though not going as far as adding a real versioning or diff system).  Just a flexible text field that allows an author to type "version 4.5"  or even "updated Jan 14, 2022".  Arguably that could be done in the Description or Title fields, but a separate field would be cleaner and consistent across all brews.  (if this should be broken into a separate issue, i can do that).

G-Ambatte:
This PR resolves #1679.

This PR adds a new Nav Item to the Share Page titled "METADATA" which on hover, reveals a window showing the brew's authors, tags, systems, and description.

To do list:
- [x] Add METADATA nav item
- [x] Show window on hover
  - [x] Description
  - [x] Tags
  - [x] Systems
  - [x] Authors
- [x] Apply styling to match mock up

---

Styling checklist:

- [x] The box is slightly overlapping the nav bar itself, but behind the "title" button. This is more noticeable on a narrow screen where the nav bar becomes two lines tall.
- [x] Can we style the "tags" with a round pill shape as we do everywhere else?
- [x] Could we perhaps unify the style with the "do you want to send this to Google" popup box (from now on, "the Popup")? Add a bit of a border, etc.? I also propose alternating background color of the rows instead of the divider lines (mockup examples below).
- [x] Some of the lines will be mostly empty. Could we narrow the max width (~440px?)
- [x] Can the dropdown be centered under the title button instead of centered on the page? This would also simplify this without needing a separate "window wrapper" class. See the Popup for how to achieve this.
- [x] Can we add a transition animation similar to the Popup?

---

- [x] Confirm lines with missing metadata return the expected no data message - "No systems", "No tags", etc.
- [x] Make authors link to the relevant UserPage

---

FIX:

- [x] UserPage dropdown menu bug


G-Ambatte:
Some adjustments from this morning's code shuffling:

image

G-Ambatte:
I've said it before, I'll say it again: `display: flex` is black magic.

image

Gazook89:
I wonder if the labels should be right-aligned?  What are they in the metadata editor?

Gazook89:
Though that would look bad when very skinny

G-Ambatte:
Current state of the PR:

image

The transparent full width element that allows the window to be centered was sitting over the drop down menus, causing them to immediately close when the metadata window was open. I had to modify the `z-index` of the drop down menus to have it sit over the new invisible window.

calculuschild (changes requested):
This is great! Just some comments:

- The box is slightly overlapping the nav bar itself, but behind the "title" button. This is more noticeable on a narrow screen where the nav bar becomes two lines tall.
- Can we style the "tags" with a round pill shape as we do everywhere else?
- Could we perhaps unify the style with the "do you want to send this to Google" popup box (from now on, "the Popup")? Add a bit of a border, etc.? I also propose alternating background color of the rows instead of the divider lines (mockup examples below).
- Some of the lines will be mostly empty. Could we narrow the max width (~440px?) 
- Can the dropdown be centered under the title button instead of centered on the page? This would also simplify this without needing a separate "window wrapper" class. See the Popup for how to achieve this.
- Can we add a transition animation similar to the Popup?

Strange overlapping
image

Proposed style:
image

ericscheid:
Also, change `Last Updated` to just `Updated`

Is there a way we can have the `(i)` icon in a subdued hue which goes full brightness when the title area is hovered?

Gazook89:
Should it only display fields that have values?

calculuschild:
> Is there a way we can have the `(i)` icon in a subdued hue which goes full brightness when the title area is hovered?

Whatever the behavior, it should be identical to all the other nav bar buttons.

ericscheid:
> > Is there a way we can have the `(i)` icon in a subdued hue which goes full brightness when the title area is hovered?
> 
> Whatever the behavior, it should be identical to all the other nav bar buttons.

Hmm .. whole nav segment changes bgcolor and any associated drop-down auto appears?

OK, with that going on don't bother with any special visual treatment of the `(i)` icon.

5e-Cleric:
> Should it only display fields that have values?

I think this might be a bad idea, instead we could have a `There is no X` text.

G-Ambatte:
Converting this to a checklist for actions:

- [ ] The box is slightly overlapping the nav bar itself, but behind the "title" button. This is more noticeable on a narrow screen where the nav bar becomes two lines tall.
- [ ] Can we style the "tags" with a round pill shape as we do everywhere else?
- [ ] Could we perhaps unify the style with the "do you want to send this to Google" popup box (from now on, "the Popup")? Add a bit of a border, etc.? I also propose alternating background color of the rows instead of the divider lines (mockup examples below).
- [ ] Some of the lines will be mostly empty. Could we narrow the max width (~440px?)
- [ ] Can the dropdown be centered under the title button instead of centered on the page? This would also simplify this without needing a separate "window wrapper" class. See the Popup for how to achieve this.
- [ ] Can we add a transition animation similar to the Popup?

G-Ambatte:
> I think this might be a bad idea, instead we could have a `There is no X` text.

The error checking was MEANT to display "No authors" or "No systems" or "No tags". Clearly that wasn't working correctly; I've added that check to the To Do list.

G-Ambatte:
Right, fixed the "no data" error messages:

image

G-Ambatte:
This is most of the checklist completed:

image

5e-Cleric:
Another point, to help users learn about this new feature, i believe this menu should be open by default, at least a few weeks

G-Ambatte:
I've completed most of the checklist now, but I'm breaking for the night.
The last thing I've added at this point is the min and max widths; the only issue as a result of this is that the window is now no longer centered. I don't expect that this will be particularly difficult to resolve once I'm able to return to the issue.

G-Ambatte:
Full screen width (hovering over title):
image

Skinny (not hovering over title):
image

Brew with no tags, no systems, and no description:
image

Mouse over the author "test":
image
image
image

---

Unless there's something else that needs to be added, corrected, or otherwise changed, I think this one is done!

G-Ambatte:
Dammit, there's a bug with the UserPage as a result of this; I assume it's probably due to z-index changes in some of the stuff I was experimenting with earlier. I'll dig into it and resolve prior to closing this out.

G-Ambatte:
Ok, I'm not aware of any further bugs induced by this PR; it should be ready for review/merging.

calculuschild (approved):

PR #2741, opened 2023-03-16
Water color image masks part 2 - Center Masks

5e-Cleric:
Updating #2373 with 10 center masks


calculuschild (changes requested):
Just needs the masks to be larger so the entire cutout fits without cutting off edges, or make some of the cutouts smaller so they fit without edges.

Everything else should be working correctly now.

calculuschild:
Looks good! Thanks @5e-Cleric ! :tada: ⭐ :100:

PR #2743, opened 2023-03-19
Add direct printing

### LINKED ISSUE #2742: Print directly from pages, rather than via PrintPage
### Your idea:

I was investigating how one might call the print function from within an iFrame, and after some experimentation, I discovered that the following command works to open a print dialogue:

    window.frames['BrewRenderer'].contentWindow.print()

However, the print preview shows as result quite unlike the desired output.  
In order to print the full Brew, the following CSS needs to be added:

```
@media print {
	.brewRenderer {
		height: 100% !important;
		overflow-y: unset;
	}
	.brewRenderer .pages {
		margin: 0px;
	}
	.brewRenderer .pages>.page {
		box-shadow: unset;
	}
}
```

As this makes printing from the New/Edit/Share pages possible, there is potentially no longer a requirement for a separate PrintPage.

Implementation Tasks:
- [ ] Edit `navbar/print.navitem.jsx` to call the print function directly instead of linking to `/print`
  - [ ] As a React Function Component, rather than Class Component (optional?)
- [ ] Add `@media print` CSS to BrewRenderer styling
- [ ] Update NewPage to use the standard PrintNavItem rather than a custom LocalPrintNavItem
- [ ] Test Print button functionality on all pages
  - [ ] EditPage
  - [ ] SharePage
  - [ ] NewPage
- [ ] Remove Print (possibly optional?)
  - [ ] Remove PrintPage files
  - [ ] Remove `/print` route
    - [ ] homebrew.js routes
    - [ ] app.js funtionality

G-Ambatte:
This PR resolves #2742.

The removal of the PrintPage and `/print` route have been left for a future PR, if their removal is deemed necessary.

Implementation Tasks:
- [x] Edit `navbar/print.navitem.jsx` to call the print function directly instead of linking to `/print`
  - [x] As a React Function Component, rather than Class Component (optional?)
- [x] Add `@media print` CSS to BrewRenderer styling
- [x] Update NewPage to use the standard PrintNavItem rather than a custom LocalPrintNavItem
- [x] Test Print button functionality on all pages
  - [x] EditPage
  - [x] SharePage
  - [x] NewPage


G-Ambatte:
After going back and forth and re-writing this from scratch, I eventually realized that the PrintNavItem (previous PrintLink) was actually _already_ written as a Function Component.

G-Ambatte:
Possibly fatal issue:

When Partial Page Rendering is active, the content of the unrendered pages literally does not exist and thus cannot be passed to the printer.

5e-Cleric:
As pointed out in gitter  a possible solution  to the Partial Page Rendering(PPR) problem would be to create a temporary iframe that just loads the existing /page to then render the whole document for printing.

It could possible even be hidden visually from the user.

calculuschild (commented):
I'm not sure I see the problem this is trying to solve. Can you elaborate?
`/print` is being used for other links. For instance it's what Google is using to track several of our reported issues on Chrome.

5e-Cleric:
> Possibly fatal issue:
> 
> When Partial Page Rendering is active, the content of the unrendered pages literally does not exist and thus cannot be passed to the printer.

image
this is the result, zoomed to see the icon in the top left of the page
all pages are printed, but those not loaded have the loading icon

calculuschild:
I think I'll close this issue for now. I'm not sure the issue this is trying to solve actually needs to be solved. Further discussion can happen on that open issue.

PR #2752, opened 2023-03-24
Prevent Injection if preceded by another Injection token

### LINKED ISSUE #2712: Brew crashes due to multiple text formatting tags

Gazook89:
This is a *short and sweet* fix for #2712 (fixes #2712).  It just adds an escape hatch in case of two consecutive inject blocks.

Note:  This is sort of a *half measure*, and should be improved in a future PR.  Specifically, you will see that consecutive mustacheInjectBlock tokens are rendered as plain text after the first, but mustacheInjectInline does not render the subsequent tokens as text (they are absorbed into the previous token).   I think the standard behavior that is desired is to follow the Block process, so that anything that isn't being tokenized is just coming out as plain text.  I don't want to bog this PR down with any further delays, so I'm going ahead without worrying about this for now.

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -134,7 +134,7 @@ const mustacheInjectInline = {
 		const match = inlineRegex.exec(src);
 		if(match) {
 			const lastToken = tokens[tokens.length - 1];
-			if(!lastToken)
+			if(!lastToken || lastToken.type == 'mustacheInjectInline')
 				return false;
 
 			const tags = ` ${processStyleTags(match[1])}`;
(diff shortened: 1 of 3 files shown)


calculuschild (changes requested):
Can you just add the unit test showing that you can do this now without crashing? Even though inline and block treat the extra injectors differently I'd like a test there just as kind of a marker that *this type of failure* shouldnt ever happen again.

Gazook89:
Okay added a unit test for double injections for each Inline and Block (Block is what caused the crash, though).  

there are `FIXME`'s in there, too, because the injection code has got bigger problems.

PR #2755, opened 2023-03-26
Index Snippet - PHB

5e-Cleric:
Preview vs PHB:
image
image
<div>

With greater font size and line height, for better readabilty.


Gazook89:
I think the text should match the PHB as much as possible, even if the readability suffers.  The goal should always be to match the source, and then let users make desired changes.  This makes it easy to simply say "well this is what matches the book" for any complaints.  Otherwise we will get complaints about why it doesn't match, which will be harder to explain ("we thought ours looked better than the official...").

5e-Cleric:
> I think the text should match the PHB as much as possible, even if the readability suffers.

Ok then, will try to measure text size in a physical copy to get the correct size.

calculuschild (approved):

PR #2768, opened 2023-04-02
Add editor themes

### LINKED ISSUE #362: Request: Editor Dark Theme
It looks like you're using CodeMirror for syntax highlighting in the editor. Would it be possible to allow us to select other themes available to CodeMirror? Some of the dark ones they have are much easier on my eyes when using them for long periods.

G-Ambatte:
This PR closes #362.

This PR adds the CodeMirror themes to the build folder, adds a CSS stylesheet link to the Editor object, and enables the transmission of that value downstream to the CodeMirror Editor object(s). This values is set via a new icon in the SnippetBar (currently raw text entry, needs a better selector).

To do:
 - [x] better theme selector (dropdown?)
 - [x] save selection in local storage
 - [x] load selection from local storage


calculuschild:
@Gazook89 look at https://github.com/naturalcrit/homebrewery/pull/2768 which supercedes this.

Gazook89:
This is 2768

calculuschild:
D'oh.

What I was hoping to say is the selector for this is no longer in the account page. Its in the edit page.

Gazook89:
D'oh.

I see that now, so much for skimming it between doing other things.  Okay, well all is resolved.

Except for our custom highlighting--- any idea how that plays out?  Or how we can improve that to flexibly adapt?

calculuschild:
Not sure about custom highlighting, but I think I'd be ok leaving it for a future PR. I suspect codemirror has documentation somewhere.

5e-Cleric:
As we have some custom styling for some elements, those do not change when theme changes, this reduces contrast and therefore visibility of some elements

### EXAMPLE:
`\page` element:
image

`\column` element:
image

Curly brackets
image


Theme: Vibrant Ink

G-Ambatte:
I resolved the merge conflicts, but now it looks like something isn't working quite right... I'll have a look when I get free from work in a few hours.

G-Ambatte:
It works in Chrome, Firefox looks to be having issues with the dropdown for some reason.

themes/codeMirror/editorThemes.json
```diff
@@ -0,0 +1,68 @@
...
+"tomorrow-night-eighties",
+"ttcn",
+"twilight",
+"vibrant-ink",
+"xq-dark",
+"xq-light",
+"yeti",
+"yonce",
+"zenburn",
+"default"
```
calculuschild, line 67: Is it possible to put 'default' back at the top of the list so people can easily find it if they want to go back?

themes/codeMirror/customEditorStyles.less
```diff
@@ -0,0 +1,68 @@
...
+    &.cm-s-seti,
+    &.cm-s-shadowfox,
+    &.cm-s-the-matrix,
+    &.cm-s-tomorrow-night-bright,
+    &.cm-s-tomorrow-night-eighties,
+    &.cm-s-twilight,
+    &.cm-s-vibrant-ink,
+    &.cm-s-xq-dark,
+    &.cm-s-yonce,
+    &.cm-s-zenburn
```
calculuschild, line 50: Ah, nice way to recolor our custom items.

In a future PR, would it be possible to move the default coloring for "light" styles in here as well? Just to keep similar items in the same place?

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -159,6 +164,7 @@ const CodeEditor = createClass({
 			autoCloseTags     : true,
 			styleActiveLine   : true,
 			showTrailingSpace : false,
+			theme             : this.props.editorTheme
```
calculuschild, line 167: Here's another place where we are using a different name.  Can we make sure to consistently name any of these related variables "editorTheme" so as not to confuse with "brew themes"? Should make tracing the variable down the props chain a little easier too.

server/app.js
```diff
@@ -417,9 +419,10 @@ const templateFn = require('./../client/template.js');
 const renderPage = async (req, res)=>{
 	// Create configuration object
 	const configuration = {
-		local       : isLocalEnvironment,
-		publicUrl   : config.get('publicUrl') ?? '',
-		environment : nodeEnv
+		local            : isLocalEnvironment,
+		publicUrl        : config.get('publicUrl') ?? '',
+		environment      : nodeEnv,
+		codeMirrorThemes : codeMirrorThemes
```
calculuschild, line 425: Can we try to consistently name these related variables some variation of "editorThemes" instead of "codeMirrorThemes", "themes", etc.

server/app.js
```diff
@@ -17,6 +17,8 @@ const asyncHandler = require('express-async-handler');
 
 const { DEFAULT_BREW } = require('./brewDefaults.js');
 
+const codeMirrorThemes = require('../themes/codeMirror/editorThemes.json');
```
calculuschild, line 20: Can we move the loading of this file into the `snippetbar.jsx` perhaps? I don't see it getting used anywhere else, so I'm not sure it needs to be in the global config.

calculuschild (changes requested):
This works really well! Added some small comments but I think it's 99% ready.

themes/codeMirror/editorThemes.json
G-Ambatte: I've modified the build script that creates this file to always put "default" as the first item in the JSON array.

themes/codeMirror/customEditorStyles.less
G-Ambatte: I've added a section to the LESS for the CM themes with light backgrounds - it's a much shorter list. As the Linter will complain about an empty style, the styling currently sets the CSS variable `--dummyVar` to a value of `'currently unused'`.

shared/naturalcrit/codeEditor/codeEditor.jsx
G-Ambatte: Pretty sure that particular instance is required to set the CodeMirror internal theme parameter. The rest of `codeEditor.jsx` uses `editorTheme`. I'll check through the other files for consistency.

server/app.js
G-Ambatte: `EditorThemes.json` is now copied to the build folder and is loaded in `SnippetBar.jsx` rather than `App.js`.

calculuschild (approved):

calculuschild:
Sorry this one took so long! Thanks @G-Ambatte!

💯 🎉 🥇

PR #2774, opened 2023-04-04
Mustache syntax unit tests

### LINKED ISSUE #2505: Fix a few Markdown tests

Gazook89:
This PR adds a new set of unit tests for the mustache syntax.  Currently the repo only has tests for mustache *spans* or "inline".  This PR will add tests for *divs* or "blocks" and also for injections.  

The PR:
1. Adds a jest library that allows a custom message in the jest test results for each test.  In this PR, it means the test result displays the input syntax along with the expected output.
2. Changes the mustache-span test file to be inclusive of all mustache syntax tests (spans, blocks, injections).  Each type of syntax is separated into groups using `describe()` so they can be run individually.
3. Modifies package.json to include scripts for running each `describe` group separately (ie `npm run test:mustache-syntax:block`)
4. Removes the stack trace from the test results to reduce unneeded clutter.
5. Adds a bunch of tests.  

Many of the tests are *failing* at this point.  Some of them like the injection tests are because injections are not actually working properly (they replace attributes rather than appending).  Some of them are because we return extra whitespace characters that the browser then strips out, but don't cause issues otherwise.  

I am going to mark those tests as `.failing()` with a comment on why-- this will allow them to "pass".   Then when they are fixed they will actually "fail" their tests and we have to remove that method.

Example:

```
it.failing('Renders a div with a single ID', function() {
		const source = dedent`{{#cat,#dog
		Sample text.
		}}`;
		const rendered = Markdown.render(source).trimReturns();
		expect(rendered, `Input:\n${source}`, { showPrefix: false }).toBe(`<div class="block" id="cat"><p>Sample text.</p></div>`);
	});
```

fixes #2505


Gazook89:
Something like 3 of the tests naturally pass, without the `.failing` method added on.  For Inline and Block, I added a specific note for each (though they generally are due to the same issue).  For Injection, I just left one comment for the batch since that needs a bigger change I think, and the issue actually impacts what is rendered in the browser unlike the others.

I think this is ready for a review.  Probably some more tests can be written, but I think this is a pretty good start and I'm all tested out for now.

PR #2796, opened 2023-04-23
Update in the DMG theme

5e-Cleric:
Adding correct drop cap letters colors, and correct table colors when inside a note.


calculuschild:
Thanks @5e-Cleric ! 💯 👍 😎

PR #2797, opened 2023-04-23
Quote Snippet

5e-Cleric:
Preview vs PHB:
image
image
<div>

The first line of the text is small caps in the PHB, but not in the DMG


Gazook89:
My quick thoughts:
- I think the snippet should utilize single asterisks to create the italic text.  Just because that is what we advise users to do in other snippets and on the homepage. I understand that single underscores *can* be used, but our current convention has been to not use them.
- The `::first-line` pseudo-selector is being applied to all descendant `p` tags inside `.quote`.  I believe you only want it on the first paragraph:  `.page .quote p:first-child::first-line` will fix that, or you could use `:first-of-type` in case there was some weird non-paragraph element at the beginning.
- Using an unordered list item for the attribution seems a little odd.  This is definitely not something I think I'm *right* about, but it might be worth considering just using a named span instead, as given below.  I think this would lead to less unexpected behavior, and feels like less of an abuse of HTML styling.  However, for a typical user, just using a hyphen *is* pretty straightforward.

```
{{quote
The thief crept through the shadows, his eyes scanning the room for any sign of danger. He knew that one false move could mean the difference between success and failure, and he was determined to come out on top.

Laboris amet quis eiusmod nostrud ut nostrud eu consectetur. Excepteur excepteur duis labore amet non et irure ad adipisicing.
				
{{attribution Isadora Fireheart, *Moonlit Magic*}}
}}
```

And the additional styling:
```css
.page .quote .attribution {
	font-style: normal;
	display: block;
	text-align: right;
}

.page .quote .attribution:before {
	content: '---';
	margin-right: .2em;
}
```

ericscheid:
I believe the use of an unordered list item for the attribution is due to the plain-text convention of writing an attribution with a leading dash or pair of dashes (emulating an em-dash). Which markdown then converts to an unordered list (usually with a bullet point instead of a dash, just to further confound things).

So, it's a curious case of being _markdown in spirit_, while also working around how markdown typically works.

Ideally, though overweening, I'd like to see a line that starts with `-- ` be parsed by markdown as needing to be rendered as a `<cite />` element. (Note too that `--` gets rendered as `&mdash;`)

Is there CSS syntax for selecting elements that contain particular text?

----

BTW, I would modify your `.page .quote .attribution:before` rule to have `content: '&mdash;'`

Gazook89:
> Is there CSS syntax for selecting elements that contain particular text?

No, it can only select and select-based-on elements (and pseudo-elements), not the text node inside elements.  

And I know about em-dash, but also I noticed a difference between `--` and `---` when putting this together.  I liked the longer dash, but I'm not married to it.  I'm not sure if there is a style convention we want to follow (besides what is in the source books).

>  I'd like to see a line that starts with -- be parsed by markdown as needing to be rendered as a <cite /> element. 

This would need to be add to the custom markdown rules in the renderer.  Not impossible, but it does require summoning the regex demons.  Maybe I'll take a whack at it in the near-ish future (though my in-laws are coming for a week so not *super* soon).

ericscheid:
> I'm not sure that <cite> is the best tag

Yeh, I saw that w3c specs commentary too. The w3c specs don't seem to have considered the use case of an speaker being quoted (their examples are inline references as content).

We do have `<blockquote>` relevant markedown. Marked will still render text with lines starting with `> ` as inside a blockquote .. but the V3 phb-style.css stripped away all styling.

---

> I noticed a difference between `--` and `---` when putting this together

It would appear that the former gets rendered as the unicode character for an en-dash, and the latter as an em-dash. 

(They don't get rendered as html-entities, mea culpa).

---

It would be nice if this..
```markdown
> lorem ipsum fancy words
>
>--- Qwotty the Quotidian (5th Sermon of the Infinities)
```

got rendered as ..
```html
<blockquote>
<p>lorem ipsum fancy words</p>
<cite>&mdash; Qwotty the Quotidian (5th Sermon of the Infinities)</cite>
</blockquote>
```

ericscheid:
Just remembered .. there's prior art with fancy quote/note blocks where special classes get assigned in the outer curly block which drive whether particular special styling is applied. We also do this with `{{monster,framed,wide`. The cite/signature would always be the last paragraph, and thus:

```markdown
{{quote,attribution
The thief crept through the shadows, his eyes scanning the room for any sign of danger. He knew that one false move could mean the difference between success and failure, and he was determined to come out on top.

Laboris amet quis eiusmod nostrud ut nostrud eu consectetur. Excepteur excepteur duis labore amet non et irure ad adipisicing.
				
--- Isadora Fireheart, *Moonlit Magic*
}}
```

with styles of 
```css
.page .quote.attribution p:last-of-type {
	font-style: normal;
	display: block;
	text-align: right;
}
```

Gazook89:
So further reading, [from the html spec](https://www.w3.org/TR/html5-author/the-blockquote-element.html#the-blockquote-element):

> Attribution for the quotation, if any, must be placed outside the [blockquote](http://www.w3.org/TR/2012/WD-html5-20121025/single-page.html#the-blockquote-element) element.

So I'm leaning towards either ignoring the spec and doing the attribution inside the blockquote, or the `<figure>` option:

```html
<figure>
    <blockquote>Here is my quoted text</blockquote>
    <figcaption>Here is my attribution</figcaption>
</figure>
```

And this would still just use the normal `blockquote` markdown syntax, and only be changed into the above if there was an attribution token at the bottom:

```md
> Here is my quoted text
>
> ---Here is my attribution
```

One question with the above approach is if a markdown blockquote without an attribution should also be placed inside a `<figure>` element for consistency, or just left out by itself.

5e-Cleric:
```
{{quote
> “Twas beauty killed the beast,‘ said the Dean, who liked to say things like that.
>
> ‘No it wasn‘t,‘ said the Chair. ‘It was it splatting into the ground like that.” 

{{attribution ― Terry Pratchett, Moving Pictures}}
}}
```
This syntax seems overly complex and unnecessary to me, but i do get that blockquotes and spans are made for this.

Applying Gazook's first thoughts, except the italic text, users can add it if they like, if we add the asterisks, the first line will also be italics, instead, we can drop the italics and let users add the two simple asterisks.

Changing atttribution to a span, as i believe it is best until we have markdown syntax to generate cite elements.

I think having both blockquote and a block syntax will be confusing for users, but if Calculus agrees with you, the changes will be made, requesting review as soon as i have updated this PR.

calculuschild:
After reading through all this, my vote now is this approach:

```
{{quote
The thief crept through the shadows, his eyes scanning the room for any sign of danger.
He knew that one false move could mean the difference between success and failure,
and he was determined to come out on top.

Laboris amet quis eiusmod nostrud ut nostrud eu consectetur. Excepteur excepteur duis
labore amet non et irure ad adipisicing.
				
{{attribution -- Isadora Fireheart, *Moonlit Magic*}}
}}
```

Although this does get me thinking about perhaps a more generalized Markdown addon, instead of limiting the focus to `<cite>` elements: text alignment. Similar to how Markdown tables make use of `----:` to right-align a column, I wonder if some similar syntax could be useful:

```
--: This paragraph is right-aligned.
-:- This paragraph is centered.
:-- This paragraph is left-aligned (the default)
:-: This paragraph is justified
```

And then some related syntax for adding horizontal spaces:
```
:::::
This paragraph is shifted down 5 lines.

:>>>>> This paragraph starts with 5 &nbsp; and has 3 more :>>> in the middle

calculuschild (approved):

calculuschild:
Merging this now. Thanks @5e-Cleric ! 🎉 💯 🔥

PR #2804, opened 2023-04-25
[NEW THEME]: Tasha's Cauldron for Everything (TCE)

5e-Cleric:
### ToDo

- [x] Snippets
    - [x] Comment note
    - [x] Blue header
    - [ ] Page Art Border (need more stains)
    - [ ] image description font:
    
image


dbolack-ab:
Would it be desirable to use this as a test case for the "naturacrit user" user themes?

5e-Cleric:
> Would it be desirable to use this as a test case for the "naturacrit user" user themes?

Actually, with your brew themes PR, we'd want to start this from scratch using https://homebrewery.naturalcrit.com/share/10_ZaGXAzLWNcn_ybM9K1uiMZ_o31Lt4f5CxliV-7LFuV as a start.

Don't remember if we needed special snippets, hope not.

5e-Cleric:
> I'm going to look into what's needed to move this into a Brew theme

If i remember correctly, the only thing left for this PR were assets, specifically, header page art:

image

The discussion ended with needing more of those (at least 10) in order to move forward, because we could not use the official art. And we don't really have an artist in the house.

dbolack-ab:
> The discussion ended with needing more of those (at least 10) in order to move forward, because we could not use the official art. And we don't really have an artist in the house.

So 10 flowery watercolor headers?

5e-Cleric:
> > The discussion ended with needing more of those (at least 10) in order to move forward, because we could not use the official art. And we don't really have an artist in the house.
> 
> So 10 flowery watercolor headers?

If i remember correctly, yes

Gazook89:
Do you want to take the time to get Master merged into this?  It doesn't seem like it even has Themes selector at this point?

Gazook89:
image

yea or nay?  i can make several these, but takes a little bit of time.

Gazook89:
Just occurring to me that kaiburr has a template for this already: https://homebrewery.naturalcrit.com/share/5CxliV-7LFuV

Should i just stop and this PR get closed out?

5e-Cleric:
His art is exctracted from tashas, not sure it can be used even in dmsguild, but I am no going to be making the decals anyway, this was here waiting in case some artist showed up or something. close if you want.

dbolack-ab:
Perhaps then, we should close this?

Issue #1556, opened 2021-08-11
auto-indent in editor?


ericscheid:
When entering brew text into the editor, in some circumstances (hard to pin down) the editor will auto-insert a <kbd>TAB</kbd> indent on entering a line break.

### Before hitting <kbd>RETURN</kbd>
![two lines need separating..](https://user-images.githubusercontent.com/2605467/129061613-05c871d8-268b-4863-9e14-15607bbbe6ac.png)

### After hitting <kbd>RETURN</kbd>
![second line indented with a tab](https://user-images.githubusercontent.com/2605467/129061453-a1cda02c-9b73-4387-afcc-d8700fcfdf1b.png)

This behaviour is undesirable.

----

One use case for auto-indent would be when continuing an indenting of text, like adding a new line here could be done by simply pressing <kbd>RETURN</kbd>
![a bunch of lines, all indented, ready for more typing](https://user-images.githubusercontent.com/2605467/129061233-5a09a088-1683-4c43-a4f0-8836a3d3cc9c.png)

And also continuing indented sub-lists.

----

Has an obvious impact with v3 brews because code blocks are now rendered as code blocks.

ericscheid:
Doesn't happen when entering lines of plain markdown text. Does happen inside `<blockquote/>` and other html codes.

Seems to be a CodeMirror thing .. auto-indenting when entering code, specifically html code here.
https://codemirror.net/doc/manual.html#option_smartIndent
Default is true.

We could maybe add a line for `smartIndent: false,` somewhere in here: https://github.com/naturalcrit/homebrewery/blob/a48c74b2e714e769c0fe714c189ea785990458f8/shared/naturalcrit/codeEditor/codeEditor.jsx#L41-L65

There might be other CodeMirror specific bits involved though, so requires experimenting.

ericscheid:
See also #1202

Gazook89:
I tested your suggestion, just added `smartIndent : false,`  and it seems to work well.  It stopped the indenting inside `<div> </div>`.  

I thought it would take away smart indenting for things like lists (and nested lists), but it did not.  And interestingly, it keeps the appropriate indent level if you add the indent yourself (with 2 spaces rather than a tab).  

So in the below case, I entered the second div with two leading spaces, and then when I press enter the next line has those two spaces.  I thought we would lose that for sure.  So I think this is a slam dunk, at least until someone can show me why I am dead wrong.

```
{{div1
  {{div2

  }}

}}
```

[...]

PR #2839, opened 2023-05-20
Custom function to use spaces for indent

### LINKED ISSUE #2092: Auto-indent in html mode is problematic
If a brew has some html container (e.g. `<div> ... </div>`) and a new line is entered (e.g. breaking a paragraph into two) then the a <kbd>TAB</kbd> character is added at the front of the line.

Unfortunately, this clashes with markdown processing of that content, with either the line now being rendered as within `<code>...</code>`, or unexpected indenting of bullet point lists.

This is fixable in editing of course — just delete the unnecessary <kbd>TAB</kbd>. Annoying and tedious, also not always obvious for the case of bullet lists.

Recommendation: turn off the auto-indent option for html mode.

image

calculuschild:
Fixes #2092, #1556

Apparently Codemirror doesn't have a built-in way to just use spaces instead of tabs for indentation. This PR uses a custom script to change that behavior, based on this discussion on the CodeMirror github:

https://github.com/codemirror/codemirror5/issues/988#issuecomment-549644684

- Tab: indent using spaces at the current location (will use 1 or 2 spaces, to line up with the next tab boundary)
- Tab while text is selected: indent whole line
- Shift+Tab : unindent whole line

Also removes the "smartIndent" feature which auto-indents inside HTML blocks when pressing `return`. Manual indenting and then pressing `return` will maintain the current indent.

You can test the behavior at https://homebrewery-pr-2839.herokuapp.com/

### DIFF
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -97,11 +97,13 @@ const CodeEditor = createClass({
 		this.codeMirror = CodeMirror(this.refs.editor, {
 			lineNumbers       : true,
 			lineWrapping      : this.props.wrap,
-			indentWithTabs    : true,
+			indentWithTabs    : false,
 			tabSize           : 2,
 			historyEventDelay : 250,
 			scrollPastEnd     : true,
 			extraKeys         : {
+				'Tab'              : this.indent,
+				'Shift-Tab'        : this.dedent,
 				'Ctrl-B'           : this.makeBold,
 				'Cmd-B'            : this.makeBold,
 				'Ctrl-I'           : this.makeItalic,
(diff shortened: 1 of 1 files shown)


shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -171,6 +174,19 @@ const CodeEditor = createClass({
 		this.updateSize();
 	},
 
+	indent : function () {
+		const cm = this.codeMirror;
+			if (cm.somethingSelected()) {
+				cm.execCommand('indentMore');
+			} else {
+				cm.execCommand('insertSoftTab');
+			}
```
ericscheid, line 183: ironic indent?

ericscheid (approved):
All looks good. Tested on staging, looks good. 

Pasting a leading TAB character in leaves a TAB character there which will behave differently, so we can expect the occasional bit of confusion from reddit. Applying `indentMore` or `indentLess` will convert those TABs to spaces. TABs which are internal to line content (e.g. in tables, used to keep columns tidy) will remain untouched (good, though these would need pasting to get in there).

shared/naturalcrit/codeEditor/codeEditor.jsx
calculuschild: I'm not sure what you mean?

ericscheid: Hmm .. sorry, my snippet wasn't 100% clear.

Line 178 is indented 2 tab stops, but the sibling statement immediately following (lines 179:183) is indented 3 tab stops.

Gazook89:
Looks good to me, both the result and the code (besides that extra layer of indent on the `if` statement that Eric pointed out). 

I do think that this is going to irk *someone*, but in my opinion it's an improvement.  However, there are likely lots of old brews that get copied (in part or in full) into new brews, so there is going to be this weird sort of hybrid situation that might trip people up.   I don't know if there is a smart way to convert tabs into spaces when pasting into a new brew?  Even if it requires a dialog prompt that asks about it?  

In either case, i think it's good to go.

PR #2843, opened 2023-05-22
Fix StringArrayEditor unique `key` error

Gazook89:
This is a very simple fix for a console error that happens in the Properties editor when the StringArrayEditor component is rendered.  Specifically, for the Invited Authors bit.  That bit includes a "notes" prop that gets sent to a `.map()` method, but the output `<p>` elements did not utilize an index to create a unique key.

Super simple fix to remove a very long error message.

EDIT:  I'll add that while struggling with this for a bit I came across this SO post: https://stackoverflow.com/a/43892905

If i'm reading it correctly, it is suggesting that using `index` for  the `key` prop in changing arrays may not be a good idea.  I didn't really want to go into much further after I got this working, but mentioning it in case someone thinks we need to worry about it.

### DIFF
--- a/client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
+++ b/client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
@@ -140,7 +140,7 @@ const StringArrayEditor = createClass({
 					</div>
 				</div>
 
-				{this.props.notes ? this.props.notes.map((n)=><p><small>{n}</small></p>) : null}
+				{this.props.notes ? this.props.notes.map((n, index)=><p key={index}><small>{n}</small></p>) : null}
 			</div>
 		</div>;
 	}


jeddai (approved):

calculuschild:
Looks good! I think the point of that stackoverflow post is that `key` needs to be a persistent identifier for each item. When React need to re-render the page, it uses the `key` to quickly locate the right object in the DOM to modify (since the HTML contents of the div itself might change). If the array has a potential of objects being rearranged or changing their position in the array, it's possible React will get confused and re-render the wrong item.

i.e.:

- `[key:1 name:Author1, key:2 name:Author2, key:3, name:Author3]`
- -> User renames Author3 to Candace
- `[key:1 name:Author1, key:2 name:Author2, key:3, name:Candace]` React can quickly see that key3 needs to be redrawn
- -> User deletes Author 2
- `[key:1 name:Author1, key:2 Candace]` Candace becomes key2, which confuses React because object keys are not supposed to change. It's not sure if key2 has changed name, or if Candace has changed key.


In practice I'm not sure how much difference it makes; I think it might appear as a temporary blink showing the wrong text. It would require coming up with some kind of alternative key that is unique and predictable (not randomly-generated), but doesn't depend on the position in the array.

However, I think this PR is good enough for now to at least  clear the error logs. Things seemed to render fine without any keys at all, so at a minimum I don't think this makes anything worse.

PR #2847, opened 2023-05-26
Fix stylelint

calculuschild:
- Fixes the "single declarations inline" custom rule being too aggressive
- Tweaks some rules based on feedback in #2846 
- Adds a default stylelint config which includes most "avoid errors" rules


Gazook89:
Running this now, and seems things work as expected.  However, one thing:
image

Any idea why this "no double quotes" thing can't do a *quick fix*?  Seems like it should be straightforward to replace double quotes with single quotes.  The stylistic docs say "The [fix option](https://stylelint.io/user-guide/options#fix) can automatically fix most of the problems reported by this rule".  Just weird.

Gazook89:
As discussed, move the single quotes rule below the font-family-name-quotes rule.  Refresh IDE after doing this so the VS Code extension updates.  Then it should all work as expected.

I've got no other comment on this for right now.  Looks good to me.

PR #2880, opened 2023-06-10
Widgets are upon us.

### LINKED ISSUE #1870: Snippet Options Interface

jeddai:
This will fix #1870! Here's my [previous PR](https://github.com/jeddai/homebrewery/pull/1) for the same thing, this one is just branched off main more recently and uses CodeMirror's CSS spec for style value autocomplete.

**This PR can be previewed at [https://homebrewery-pr-2880.herokuapp.com/](https://homebrewery-pr-2880.herokuapp.com/).**

Commits:
- add the concept of widgets and widget fields

---

TODO:
- [x] Move `themeWidgets` into explicit theme files
- [ ] Finish adding widgets for various snippets
- [ ] Set up widget select box and image preview for text+number = image snippets

---

https://github.com/naturalcrit/homebrewery/assets/4909307/674f61ef-c0f2-4745-a872-35b659b5eada


Gazook89:
I'm assuming it is too difficult to narrow the options in the select menu to just valid values for each property?  The dataset for all those values/properties would need to be organized that way, I assume, and it isn't already?

calculuschild:
> ...narrow the options in the select menu to just valid values for each property?

I agree, if possible we should only auto-complete with valid values.

calculuschild:
Alright, so as I see it now, you have implemented 3-and-a-half widget types:

- A sort of generic **CSS value Autocomplete** widget. If we can get it to autocomplete only valid values for the current field, I can see this helping non-technical people make tweaks to the CSS, as a fallback for less specific widgets.
- A **Related classes/fields** widget that detects "monster" and suggests "frame" and "wide" as relevant options, but also provides "top" as a possible field for the "artist" snippet.
- An **Image** widget that is similar to the **Related fields** widget, but uses the "injection brackets"
- A **numbers** widget sub-mode of the **CSS value Autocomplete** that applies when the text contains numbers, and can be incremented/decremented with arrow keys.

Am I correct in understanding that this CSS Autocomplete is intended to act as a fallback for less specific widgets? I.e., I'm imagining that eventually our "widgets.json" file, it would be possible to mark "number-based" properties (top, left, etc.) explicitly so they don't just rely on using RegEx to detect a number in the text field, but instead appear formatted with increment/decrement arrows by default. Not sure how much of that you intend for this PR and how much to push off until later).

---

As for other feedback, I'm not going to dig too much into the code yet until the behavior is where you want it, but I have some UI/UX thoughts:

- I kinda liked the single-row layout you had before. I liked being able to still see the contents of the block being edited. Not sure how well everything would fit, but it looked a bit more streamlined to me. The current behavior seems to have a mix of both:

image

- I would like to be able to close out of the widgets just by clicking away.

- Could you clarify what other possible "Field types" we might see? It seems currently you just have "STYLE" which appears to be just searching a pattern like "property:value".

jeddai:
> If we can get it to autocomplete only valid values for the current field

Done

> Am I correct in understanding that this CSS Autocomplete is intended to act as a fallback for any property that doesn't have a specified widget?

Yes, they are present as a fallback, but if you declare something specific that takes precedent. So we can definitely add a number-only option. I think I'd prefer to do that in a separate PR

> If we make it inline, what do you think about resizing each field to only be as large as its contents, and maybe color different types of widgets differently to keep widgets visibly separated?

I do like that, this is where having a more strongly defined UI framework would be super helpful.

> I would like to be able to close out of the widgets just by clicking away.

That is potentially tricky but doable, yeah.

> Could you clarify what other possible "Field types" we might see?

I'm honestly not sure. I'm mostly keeping it configurable for configuration's sake

jeddai:
Okay we have a MAJOR update.
* Added `Hints` component
  * Manages the keypress/click events and rendering for the autocomplete
  * Lots of logic in here that goes back and forth between fields, codeEditor, and hints
* I'm realizing only now exactly how much code y'all have to review for this and I am so sorry.
* I'm going to write tests covering all of this

jeddai:
Short demo of the updated autocomplete

https://github.com/naturalcrit/homebrewery/assets/4909307/4c5b282b-3888-49cf-ad72-001ebf6bb68a

Gazook89:
image

The HTTPS option is editable.  And that's *fine*, but it doesn't seem to work correctly.  You can type into the field, but the text isn't displayed in the field, only inserted into the markdown....and can't be deleted.

jeddai:
Yeahhhhhhh I'm working on that still 😬

5e-Cleric:
When a widget is open and an input chosen, scrolling pulls the autocomplete dropdown, as it is a fixed component:

image

5e-Cleric:
For the Color, and gap between the pills, i reccomend a 5px gap, and bringing the general menu UI color in:

image

With darker editor themes, a simple invert css filter should suffice, if needed.

Another option would be to have different color for widgets depending on what kind of snippet they edit (tables green, PHB golden, text editor blue, and images red).

jeddai:
Closing this PR in favor of #2905. All feedback here will be implemented there.

Issue #802, opened 2018-11-11
[Large brews][Feature] Add command line tool for rendering large brews into a pdf


AlexeySachkov:
A lot of users have problems with rendering large brews: Chrome hangs, crashes, etc.

I propose some kind of a solution/workaround for this problem. The main idea is to distribute some console application which perform conversion from brew source code into a pdf without using browser.

Here is example of usage:
```
render-brew source_code.txt -o my_brew.pdf # converts brew code saved in source_code.txt into a pdf file named my_brew
render-brew http://homebrewery.naturalcrit.com/share/HkWWfylvum -o my-brew-2.pfd # converts brew by using share link into a pdf file named my-brew-2
```

I expect that such solution will require less resources (RAM) to render large brews

tcyrus:
I would definitely appreciate something like this. From looking at the code, it seems like something like this could be accomplished by using the code in [`printPage.jsx`](https://github.com/naturalcrit/homebrewery/blob/master/client/homebrew/pages/printPage/printPage.jsx) and combining it with [Puppeteer](https://pptr.dev/)

tcyrus:
@AlexeySachkov Your second command can be done using something like this

```
chrome --headless --print-to-pdf="my-brew-2.pdf" https://homebrewery.naturalcrit.com/print/HkWWfylvum
```

tcyrus:
@AlexeySachkov I also built something which might be able to help you with your first command

https://github.com/tcyrus/homebrewery-standalone

dbolack-ab:
I think this should be closed.

Issue #2824, opened 2023-05-10
Possibility of Homebrewery as `npm` package


T3sT3ro:
### Your idea:

Hi, I have been searching through open and closed issues, but I didn't find any about supporting adding homebrewery as a package dependency.

Reading README I saw that it requires a local Mongo installation to make it work. I was wondering why is it needed, and is it possible to extract/separate different parts of the app into separate modules and packages. Things like creating HTML previews, masks, the markdown parser, styles, table generations and similar. For example If I wanted to have a standalone app with notes, I would want to be able to pass some markdown files to a standalone homebrewery preview. Is it possible, am I missing something?

dbolack-ab:
We could possibly convert `/shared/naturalcrit/markdown.js` into an NPM.

That's largely what Dungeons and Markdown does - which might work for you for this purpose, however, D&M hasn't been updated in ages and is a little stale as a result.

I guess I need to update my fork.

calculuschild:
A first step would probably be pulling the individual Markdown extensions out into their own packages. We already did this for Emojis, Extended Tables, and Smartypants Lite.

dbolack-ab:
That sounds like V4 territory

dbolack-ab:
This is proceeding at speed.

Issue #1926, opened 2021-12-26
[LOCAL ONLY] Flat File Structure


G-Ambatte:
Rather than saving documents to a MongoDB collection, some users in the past have requested that they would prefer to save to a local text file. Some have suggested placing these files in their own separate Git repository/sub-module to allow for versioning of their files, e.g. #346, #808.

I would suggest that we create a save system that will create files locally, but contain this behind a `process.env.NODE_ENV == 'local'` check so that it only applies to users who are running local installs.

calculuschild:
I think setting up a local text file save would be fine, especially now that we have metadata embedded into the brews themselves. However we still want local installs that can work with MongoDB for development. So perhaps we could rename the current `local` environment variable to `dev` or something, and then `local` would be just a self-contained to use the local file system?

G-Ambatte:
One option would be to change the Google icon from a toggle (Mongo/Google) to a dropdown (Mongo/Google/Local).
- Google would be greyed out and unselectable unless the user has signed in with Google Drive.
- Local would only be available where `process.env.NODE_ENV == 'local'`.

This dropdown option could also be extended as additional storage options are incorporated. This could be added to the UserInfo framework as an accountwide default setting, once that PR is completed.

dbolack-ab:
There has been no discussion on this inthree years. Close?

Issue #1101, opened 2020-11-16
[Suggestion] Github sync, like Drive sync is working


ericscheid:
> I know it's probably pretty low on the wish-list, but I'd imagine there are a number of people who would love to sync their brews with their own github repo.

https://www.reddit.com/r/homebrewery/comments/jtrq9z/suggestion_github_sync_like_drive_sync_is_working/

dbolack-ab:
For this to finish implementation we'd need a place to store the user's github token  and ensure it could be seen by homebrewer.

If this is not desirable I think we should close this as a Won't Implement.

5e-Cleric:
right, so modify the user db?

Gazook89:
Is the implication here that if we are storing a github token on behalf of users, that we are opening ourselves up to greater security concerns?   I don't know enough about these things so it's a genuine question.

With the Google token, my understanding is that even we have read/write access it is narrowed to a very small scope (in a specific folder), right?  And with Github, it is similar, it only grants access to a specific repo (the user would create a repo for the Homebrewery stuff).

I'm just trying to read into the tone of `If this is not desirable...` as if you are raising an eyebrow or are skeptical that we would want to do this.  Correct me if i'm wrong.

dbolack-ab:
> right, so modify the user db?

We'd have to if you want sharing.

dbolack-ab:
> Is the implication here that if we are storing a github token on behalf of users, that we are opening ourselves up to greater security concerns? I don't know enough about these things so it's a genuine question.

No. We would store a user-provided access token configured to allow as little as possible to create and read gists. Gists can be marked as Secret on creation - which I think we should use - but this behaves EXACTLY like an unpublished brew regarding discoverability.

If we switch from a gist model ( remember, a gist is a git repo wrapper and simplification ) to a repository model we can lock that down further but I feel that is overkill.

There are two big advantages from adding this feature:

 * Significantly document history for the brew. Each save is a git commit.
 * Github rate limiting is much better behaved. :)

 
> I'm just trying to read into the tone of `If this is not desirable...` as if you are raising an eyebrow or are skeptical that we would want to do this. Correct me if i'm wrong.

If this feature is not desirable, this issue should be closed.

5e-Cleric:
This feature is desirable and should be added in my opinion.

calculuschild:
Fortunately the Naturalcrit site already uses a generic authentication package for the Google Login stuff. Passport.js.

[...]

PR #2886, opened 2023-06-14
[WIP - Discussion] Command Line HTML generation

### LINKED ISSUE #1035: [request] command-line interface without need starting Chrome
I suggest making an standalone app that does not depend on Chrome thing, Something similar to Latex. It can run from command line without starting any server. 

like this: 
```
homebrewery input.md output.pdf 
```

Because I can write stuff on file system (markdown file with text editor) and integrate with .git easily without need to copy and paste stuff from file to Chrome every time I build.

G-Ambatte:
This PR resolves #1035.

This PR adds a `/cli` folder which contains `process.js`, a script that takes several parameters:

* input : the filename of the downloaded source of a Homebrewery brew
* output : the filename of the desired output file
* renderer : the renderer to use (valid options are "legacy" and "v3")
* overwrite : (OPTIONAL) overwrite the output file if it already exists

Defaults for these parameters can be stored in the `/cli/config.json` file.

The output relies on a build of the Homebrewery project having completed, as it acquires the compiled `bundle.css` from the `/build` folder.

To be perfectly clear and explicit: the output of this script is still HTML, and needs to be loaded into Chrome to be printed to PDF. Currently there is a large delay before the page loads, but I believe that this is likely due to missing asset files, rather than any rendering issue (as the render was completed offline, and the HTML is static).

---

The current purpose of this Draft PR is to create discussion about what the users and/or the developers might want from such a CLI tool, and what the expected output might look like. Similar Issues include #1035 and #802; I'm sure that there are probably more.


G-Ambatte:
Tagging potentially related open issues:

#661 
#802 
#956 
#1035
#1101
#2824 

This also has the potential to be precursory work for the following issue:

#1926 

If you agree/disagree with the inclusion of these issues, please let me know in the comments!

calculuschild:
> The current purpose of this Draft PR is to create discussion about what the users and/or the developers might want from such a CLI tool

I personally have no use for such a tool. Besides #802 and #1035, the issues I see linked here don't seem to have any relation to a CLI at all, but I may be mistaken.

I'm all for more features, but from reading those issues, it seems people are rather looking for an "export to embeddable HTML" button on the website itself.

5e-Cleric:
As Calculuschild, i don't see the need for such a tool, if i understand it correctly, this would be a CLI(Command Line Interface, for those too lazy to google it) to transform markdown files into html, following the syntax and all of the homebrewery.

But isn't the homebrewery itself already doing that better? by allowing real time rendering?

Furthermore, even if the tool itself went down, the repository will stay up, and the instructions to install locally are very simple.

What is the motivation behind such a tool?

If as the boss sugests, this would be to create an embeddable html, i would suggest building a system that would allow a request for a file, and deliver its html if the file has some sort of metadata option ticked, such as published.
Actually that is not so bad of an idea.

srfraser:
> What is the motivation behind such a tool?

I came looking for just this feature. When I write documents, I store them in version control, so that I can try out changes, revert to previous versions, and other useful features. So while I do use homebrewery interactively and the live rendering is immensely useful for getting things looking good, my goal is to generate the finished document from the version committed to version control. Part of that is being able to run a CI pipeline to then generate the files required, once a commit has been made - That way I know the documents are based on what was committed, not what I was last working on.

Gazook89:
@srfraser  check out this project that takes the rendering parts of HB and puts it in a VS Code extension: https://github.com/Spjak/DungeonsAndMarkdown

this allows you to use HB locally, with your own md files that can be stored anywhere, even committing them all to a git system along with other benefits.

srfraser:
Nice, thank you for the recommendation, I should be able to cannibalise that for my needs

poolpog:
Hello! I am looking for a feature just like this (*or* a POSTable Rest API that accepts a markdown file as payload and returns rendered HTML or a PDF)

Similar @srfraser I much prefer to put documentation into version control, and process the document source as needed. In addition, this allows me to do things such as programmatically generate Markdown from data, such as to create roll tables for wandering monsters or whatever.

Using a web UI is great for immediate feedback, but terrible if I want to automate document production.

poolpog:
In case anyone else wants to use this, here's a quick writeup on how I got this to work without installing Node or NPM. On a Macbook, MacOS Monterey

**NOTE**: Source dir for my local markdown file in this example is: `$HOME/Projects/poolpog/wandering-monster-tables/`

Add Git Fork to my local repo or clone git fork as listed in that draft PR

    git remote add G-Ambatte git@github.com:G-Ambatte/homebrewery.git

Checkout branch that has the CLI code

    git checkout remotes/G-Ambatte/experimentalCommandLineBrewProcess

(Don't worry about the "Detached HEAD" warnings -- we aren't making any code changes)

Build local docker image with new code checkout

    docker-compose build

Run node command in docker; mount local dirs for output. You'll have to specify full paths everywhere.

    docker run -it --rm \
        -v $HOME/Projects/poolpog/wandering-monster-tables/:/usr/src/app/workdir homebrewery \
        node cli/process.js --input /usr/src/app/workdir/my-brew.md  \
        --output /usr/src/app/workdir/my-brew.html \
        --renderer v3 --overwrite

dbolack-ab:
Suggest closing. Once the markdown.js can become a module this is better suited as a demonstration tool for said module.

dbolack-ab:
Above condition has occurred.

5e-Cleric:
Agreed, Sorry @G-Ambatte but this is better suited as its own possibly desktop tool, now that the markdown processing has been split into its own npm package.

PR #2889, opened 2023-06-17
Remove whitespace from brew titles on UserPage

### LINKED ISSUE #2775: Whitespaces in brew title aren't trimmed
Brew titles can be saved with as many spaces users want, but spaces before the word influence sorting by title.

This should be easily solved by simply trimming those spaces when saving the brew.

G-Ambatte:
This PR resolves #2775.

This PR performs a `trim()` on all brew titles before they are passed to the base ListPage.


calculuschild:
Would it make more sense to do this in the `listPage.jsx` where we already do other similar "cleanup" of brew titles to aid alphabetical sorting (deburr, etc.)?

https://github.com/naturalcrit/homebrewery/blob/1e004977bea73f112197f959d8dd2a58a4d11ac5/client/homebrew/pages/basePages/listPage/listPage.jsx#L92

G-Ambatte:
> 1. Would it make more sense to do this in the `listPage.jsx` where we already do other similar "cleanup" of brew titles to aid alphabetical sorting (deburr, etc.)?

I'm pretty sure that the whitespace collapses gives the appearance of trimming anyway. I've added it to the ListPage sort as well, as well as the UserPage brew collection generation.

>     2. This PR also doesn't address the suggestion of "just fix the title before it gets saved". Do we want to _actually change_ brew titles? Or do we only care about making them alphabetically sort in the userpage? I'm not sure it's necessary (maybe some people _want_ random untrimmed spaces.... but I'm not sure.  Thoughts?

I've added a couple of lines to trim the title before the initial save, and to perform a trim before saving updates.

At this stage, I have also expanded this slightly to include trimming the brew's description.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -28,6 +28,7 @@ const UserPage = createClass({
 		const usernameWithS = this.props.username + (this.props.username.endsWith('s') ? `’` : `’s`);
 
 		const brews = _.groupBy(this.props.brews, (brew)=>{
+			brew.title = brew.title.trim();
```
calculuschild, line 31: My only comment is I think we should move this "cleanup" step into the homebrew.api.js page when we first retrieve the brew list, so all the information coming into this user page is already "clean".

Similar to how we fill in missing values with defaults when getting old brews. Let's keep all the data cleanup at the point of retrieval where possible.

calculuschild (changes requested):

client/homebrew/pages/userPage/userPage.jsx
G-Ambatte: I've just pushed a commit where the brew data cleanup is done in `app.js` instead; I haven't been able to test it yet but I expect that it will pass without issue.

G-Ambatte: I've been able to test this on my local install now and it all appears to work as expected,

calculuschild (approved):

PR #2892, opened 2023-06-19
Basic ErrorPage functionality

### LINKED ISSUE #1924: Improve the "Can't get that" error page for missing brews
Hitting a blank white page with only "can't get that" as the error is frustrating, ugly, and confusing for users.  We should improve that error page.

- [ ] Look pretty
- [ ] Emit (and display) a relevant error code instead of just 500
- [ ] Provide instructions to direct the user to the next step
- [ ] A button to optionally remove the brew from "Recent brews"

G-Ambatte:
This PR resolves #1924.

This PR improves on the current error handling functionality, redirecting to an ErrorPage when an error occurs.

Currently the existing (ancient) ErrorPage has been used; styling and functionality requirements will need to be finalized and implemented before this PR can be considered ready for merging.

**Error examples:**
- Google file not found:
image
- Author does not have editor access to the brew:
image


5e-Cleric:
I would suggest the following text to replace the GD 404 brew text:

___

## Your brew is not here

This error tells us your file was saved on your Google Drive, but the link you tried to open doesn't work anymore. Since the Homebrewery cannot delete your your personal files from Google Drive without you clicking the delete button, there are three most likely possibilities:

 - You have accidentally deleted the Google Drive files. Look on your Google Drive and make sure the Homebrewery folder is still there, and that it holds your brews as text files.

 - You have accidentally changed the sharing settings for your files. If the files are still on Google Drive, change all of them to be shared "with everyone who has the link" so the Homebrewery can access them.

 - Rarely, it might mean that your Google Account is full or has been closed by Google (for inactivity, violating some policy of theirs). Make sure you can still access your Google Drive normally and upload/download files to it.

If you can't find it, Google Drive usually puts your file in your Trash folder for 30 days. Assuming you didn't empty the trash right after, it might be worth checking. You can also look on the right side of the page while logged into Google Drive and look at the Activity tab. This can help you pin down the exact date the brew was deleted and by whom.

If you still can't find it, some people have had success asking Google to recover accidentally deleted files. You can visit this link https://support.google.com/drive/answer/1716222?hl=en&ref_topic=7000946

At the bottom of the page there is a button that says "Send yourself an Email" and you will receive instructions on how to request the files be restored.

Also note, at no point are you "forced" to save your files on Google Drive. You can always set any brew to be stored on your own Google Drive, or on the Homebrewery servers, by clicking the Google drive icon by the brew title and choosing "transfer my brew to/from Google Drive".

- The Devs

___

Please suggest any change needed, this is the answer we've been giving to people on reddit and here.

G-Ambatte:
> I would suggest the following text to replace the GD 404 brew text:

At this point in time, I have created an `errorIndex.json` file, which is just a JSON object using the reported error code as the key; I have used the suggested text on the "404" error:

image

Possibly this approach would work better with Homebrewery-specific internal error numbers, as the possibility exists that some other function will return a 404 and gets the Google Drive file missing error message.

---

Attempting to access a brew's Edit Page while not logged in as an author (not in `errorIndex.json`) results in this:

image

G-Ambatte:
One downside of using a cookie to store the error information and expiring the cookie as soon as the error page has been loaded, is that refreshing the error page causes the error information to be lost.

An alternative that does not suffer this issue is to store the error information in query parameters instead - the error information will survive a refresh.

G-Ambatte:
Changing the existing ErrorPage to use the base UIPage:

Google file missing:
image

Not an Author:
image

server/homebrew.api.js
```diff
@@ -308,7 +311,7 @@ If you believe you should have access to this brew, ask the file owner to invite
 				await HomebrewModel.deleteOne({ _id: brew._id })
 					.catch((err)=>{
 						console.error(err);
-						throw { status: 500, message: 'Error while removing' };
+						throw { ...new Error, status: 500, message: 'Error while removing' };
```
calculuschild, line 314: I *think* this is equivalent to:

```suggestion
						throw { Error('Error while removing', {status: 500}) };
```

By default the "Error" object uses its first field as the `message:`. This reads more clearly to me without needing the spread operator, etc.

calculuschild:
I made some changes so we don't have to redirect at all (user gets to keep the current URL in case it was a typo causing the error),  and hence we don't need to do the cookie thing.

> Possibly this approach would work better with Homebrewery-specific internal error numbers

Yes. We should add a separate HBErrorCode or something to each error. The 404 status as you point out will occur for more than just the "google drive file missing" case, and each should have its own error text.

server/homebrew.api.js
G-Ambatte: At this point, I've shifted to using `throw { [...] }`, which allows us to add custom fields. We can revisit this if necessary in the current commit.

G-Ambatte:
Current state of the PR:

Google Drive file has been deleted:
image

Brew does not exist in HB or Google Drive:
image

User is not an author:
image

server/homebrew.api.js
```diff
@@ -57,22 +57,23 @@ const api = {
...
 				stub = stub ? _.assign({ ...api.excludeStubProps(stub), stubbed: true }, api.excludeGoogleProps(googleBrew)) : googleBrew;
 			}
 			const authorsExist = stub?.authors?.length > 0;
 			const isAuthor = stub?.authors?.includes(req.account?.username);
 			const isInvited = stub?.invitedAuthors?.includes(req.account?.username);
 			if(accessType === 'edit' && (authorsExist && !(isAuthor || isInvited))) {
-				throw `The current logged in user does not have editor access to this brew.
-
-If you believe you should have access to this brew, ask the file owner to invite you as an author by opening the brew, viewing the Properties tab, and adding your username to the "invited authors" list. You can then try to access this document again.`;
+				throw { name: 'Access Error', message: 'User is not an Author', status: 401, HBErrorCode: '03', authors: stub.authors };
```
calculuschild, line 71: Can we make two different errors?

- The user is logged in, but not in the current authors.
- The user is *not* logged in at all (so the error doesn't say "The current logged in user....")

along the lines of: 
```suggestion
				if(req.account) {
					throw { name: 'Access Error', message: 'User is not an Author', status: 401, HBErrorCode: '03', authors: stub.authors };
				}
				else {
					throw { name: 'Access Error', message: 'User is not logged in', status: 401, HBErrorCode: '04', authors: stub.authors };
				}
```

server/homebrew.api.js
```diff
@@ -57,22 +57,23 @@ const api = {
 						googleError = err;
 					});
 				// Throw any error caught while attempting to retrieve Google brew.
-				if(googleError) throw googleError;
+				if(googleError) {
+					const reason = googleError.errors[0].reason;
+					throw { ...googleError, HBErrorCode: reason == 'notFound' ? '02' : '01' };
```
calculuschild, line 62: Similar to below comment, can we make split 02 into two to help users troubleshoot these different scenarios?

- The user is logged in, and this brew is under their account (you deleted your google drive files by accident!)
- The user is not logged in, or this brew is not under their account (this brew is owned by a different account, which explains why you don't see a record of it in your google drive. Are you logged into the right account?)

server/homebrew.api.js
G-Ambatte: Implemented in commit [e28b6e7](https://github.com/naturalcrit/homebrewery/pull/2892/commits/e28b6e7a1926e88041ab8b4175b2128878f9a43e)

server/homebrew.api.js
G-Ambatte: At this stage, I don't think we can - I haven't been able to spot any way to differentiate "you don't have access" from "this doesn't exist". I'll leave this unresolved for now, in case we have any flashes of genius later on.

calculuschild: The error is the same (brew doesn't exist), so the response from Google would be the same.

What I'm asking is to add a little more info to differentiate:

- This brew doesn't exist, *and you own it* as first author
- This brew doesn't exist, *and you don't own it*


to help explain the cases where users made the brew on an old Google Drive account and then switched to a different login and have continued editing as an invited author. I.e., to track down which account deleted the brew.

G-Ambatte: This should be in the latest commit.

calculuschild:
Ok, I've made one more round of tweaks to the error texts. I'm sure there's plenty more improvement that can be made, but we can always reword things later. I put screenshots of the main errors here for ease of review, and adding the others as reviewers for final comments. Thoughts on the error text or style? Anything major we can put in a separate PR, but small things we can change here and then send live.

| **Error** | **Screenshot** |
|:-----:|:-----:|
| Google Brew not found |   |
| Google Brew not found 2 |   |
| Google Brew not found 3 |   |
| Not signed in to edit |   |
| Not in author list |   |
| Brew not found |   |

server/homebrew.api.js
```diff
@@ -77,7 +77,7 @@ const api = {
 				if(req.account){
 					throw { ...accessError, message: 'User is not an Author', HBErrorCode: '03', authors: stub.authors, brewTitle: stub.title };
 				}
-				throw { ...accessError, message: 'User is not logged in', HBErrorCode: '04' };
+				throw { ...accessError, message: 'User is not logged in', HBErrorCode: '04', authors: stub.authors, brewTitle: stub.title };
```
G-Ambatte, line 80: Adding these additional properties has broken in `invalid author` test; we'll need to update the expected test result in the API spec document to get the test to pass again.

calculuschild: Whoops. I'll fix that real quick.

G-Ambatte:
I think it's looking pretty good! That one failing test should be a pretty minor update to correct.

G-Ambatte:
I've just tried pushing a commit from the work laptop - I had to edit the file in Notepad++ but I don't think I did anything to anger the Linter.

G-Ambatte:
Confirmed - test is fixed now.

server/homebrew.api.js
G-Ambatte: Done - its resolved in [this commit](https://github.com/naturalcrit/homebrewery/pull/2892/commits/a7040e554ac2d3a7b144f0983b31c604697d2f16).

G-Ambatte:
@calculuschild : would you agree that this PR will resolve #1924?

calculuschild:
Yes.

G-Ambatte:
> Yes.

I've updated the top comment to link the original Issue to this PR.

PR #2902, opened 2023-07-03
[NEW THEME]: LEO

Gazook89:
Here is a rough draft of a theme called Low Earth Orbit, or LEO.  I was motivated by the fact we have this Themes capability but nothing really beyond DND, Blank, and the Old Journal one I created a long time ago.  The Journal one is pretty niche, and has a some slightly complex CSS stuff, and I think it's sort of hard for a user to "customize" since it's so wrapped up in it's background page texture.

With LEO, I wanted to put a non-DND theme together that didn't rely on static/hosted images for background textures or page elements and was instead designed using CSS for as much of the total styling as possible.  And, ideally, stay away from very complex CSS, though I'm not sure I've accomplished that.  Hopefully I could create a decent looking theme that was simpler to customize with basically just colors and different border styles.  

Low Earth Orbit as name comes from my original inspiration, pulp sci-fi book covers.  Lots of that yellow and red lettering, and basic solid or dotted lines.  It's a fairly tough source material because those books *only* had illustrated covers, and of course the actual pages are basically just text.  I've tried looking for similar "retro" pulpy sci-fi magazines but it's not super easy to find the *insides* of those.  But I've only really done web searches...I may stop at a couple stores near me that should have some physical copies in stock.

Ultimately, I may shift gears slightly towards a non-sci-fi pulp fiction theme...it's not a big jump, but should be more recognizable/generic.  Perhaps two themes, one built on the other?

Anyway, enough talk, let me know if you think there is anything *here* we want to move forward with, or changes, or if you have some source material that you think we should try to incorporate.  At this point, this is far from complete (Tables? Alternating pages?  h4 & h5?  Cover page?) so it's sort of a preview before I put more effort into it.

Screenshots:
image
image
image


dbolack-ab:
I like this.

Gazook89:
going to close this PR.  Might open a discussion about theme suggestions (the graphic part of that).

PR #2905, opened 2023-07-03
Editor widgets for snippets

### LINKED ISSUE #1870: Snippet Options Interface

jeddai:
## Description
This is a move of #2880. Because of how complicated this is, we've moved the branch into the repo. Branches can be made off this branch, and pull requests can be opened targeting this branch.

Code explainer video: https://youtu.be/pd42hBXhXa0

## From #2880:

This will fix #1870! Here's my [previous PR](https://github.com/jeddai/homebrewery/pull/1) for the same thing, this one is just branched off main more recently and uses CodeMirror's CSS spec for style value autocomplete.

**This PR can be previewed at [https://homebrewery-pr-2905.herokuapp.com/](https://homebrewery-pr-2905.herokuapp.com/).**

Commits:
- add the concept of widgets and widget fields

---

TODO:
- [x] Move `themeWidgets` into explicit theme files
- [x] Finish adding widgets for various snippets
- [x] Set up widget select box and image preview for text+number = image snippets

---

https://github.com/naturalcrit/homebrewery/assets/4909307/674f61ef-c0f2-4745-a872-35b659b5eada


shared/naturalcrit/codeEditor/helpers/widgets.js
```diff
@@ -0,0 +1,93 @@
...
+				coverGutter : false,
+				noHScroll   : true,
+				className   : `snippet-options-widget ${widgetOption.name}-widget ${widgetOption.name}-widget-${n}`
+			});
+		}
+	};
+
+	return {
+		removeLineWidgets : (widget)=>{
+			cm.removeLineWidget(widget);
```
calculuschild, line 64: I know `cm.removeLineWidget` is in the example online, but I don't see it in the main documentation?

However `widget.clear()` seems to do the same thing. May be better for documentation sake?
https://codemirror.net/5/doc/manual.html#widget_clear

shared/naturalcrit/codeEditor/helpers/widgets.js
```diff
@@ -0,0 +1,93 @@
...
+	}));
+
+	const updateLineWidgets = (n, remove)=>{
+		const { text, widgets } =  cm.lineInfo(n);
+		const widgetOption = widgetOptions.find((option)=>!!text.match(option.pattern));
+		if(!widgetOption) return;
+		if(!!widgets) {
+			for (const widget of widgets) {
+				widgetOption.createWidget(n, widget.node);
+			}
```
calculuschild, line 51: I may need help understanding exactly what happens here. `widget` is a CodeMirror object, right? And `widget.node` is the HTML element representing that widget? So `createWidget` is just replacing the HTML on the existing widget, right? Perhaps we can rename `createWidget` to something like `updateWidgetHTML` or `updateWidgetElement`?

themes/V3/5ePHB/widgets.js
```diff
@@ -0,0 +1,24 @@
+const { WIDGET_TYPE, FIELD_TYPE } = require('../../../shared/naturalcrit/codeEditor/helpers/widget-elements/constants');
+
+module.exports = [{
+	name    : 'monster',
+	type    : WIDGET_TYPE.SNIPPET,
+	classes : ['frame', 'wide']
```
calculuschild, line 6: Alright, so after going through the video, I have some concern about the extensibility of this approach dividing the widgets up to have a `classes` property (that makes checkboxes) and a separate `fields` property (that makes textboxes). I foresee eventually running into confusion when we add other types of entry methods (is a colorpicker a "field"? is a watercolor image selector a "class"? Are there cases where we want checkboxes for fields and not just classes?

I propose reorganizing these properties into a single `widgets` array where each component is identified by its button type (checkbox/textbox), rather than its CSS type (class/field):
```suggestion
	name    : 'monster',
	type    : WIDGET_TYPE.SNIPPET,
	widgets : [
		{
			name : frame,
			type : checkbox
		},
		{
			name : wide,
			type : checkbox
		},
		{
			name : opacity,
			type : textbox
		},
	]
```

I think this could greatly simplify the `createWidgets()` step

shared/naturalcrit/codeEditor/helpers/widget-elements/constants.js
```diff
@@ -0,0 +1,30 @@
+export const UNITS = ['cm', 'mm', 'in', 'px', 'pt', 'pc', 'em', 'ex', 'ch', 'rem', 'vw', 'vh', 'vmin', 'vmax', '%'];
+
+export const HINT_TYPE = {
+	VALUE         : 0,
+	NUMBER_SUFFIX : 1
+};
+
+export const WIDGET_TYPE = {
```
calculuschild, line 8: I think we may need to step back and make sure our definitions are all the same.

To me a `widget` is one of these: 
image

But this code indicates that the widget is this:
image

themes/V3/5ePHB/widgets.js
calculuschild: So in `index.js` where you define how to render the `cClass` element and the `field` element we would rename those to `checkbox` and `textbox`, and then later add `colorselector` and `imageselector`, etc.

shared/naturalcrit/codeEditor/helpers/widget-elements/index.js
```diff
@@ -0,0 +1,117 @@
...
+			const id = `${_.kebabCase(prefix.replace('{{', ''))}-${_.kebabCase(cClass)}-${n}`;
+			const frameChange = (e)=>{
+				if(!!e.target && e.target.checked)
+					cm.replaceRange(`,${cClass}`, CodeMirror.Pos(n, prefix.length), CodeMirror.Pos(n, prefix.length), '+insert');
+				else {
+					const start = text.indexOf(`,${cClass}`);
+					if(start > -1)
+						cm.replaceRange('', CodeMirror.Pos(n, start), CodeMirror.Pos(n, start + cClass.length + 1), '-delete');
+					else
+						e.target.checked = true;
```
calculuschild, line 87: Is this what we want to do in this state? By the time we get here, we know that `cClass` was not found, so shouldn't we leave the checkbox unchecked to show that it's missing?

shared/naturalcrit/codeEditor/helpers/widget-elements/index.js
```diff
@@ -0,0 +1,117 @@
...
+				}
+			};
+			return <React.Fragment key={`${_.kebabCase(prefix)}-${cClass}-${n}`}>
+				<input type='checkbox' id={id} onChange={frameChange} checked={_.includes(text, `,${cClass}`)}/>
+				<label htmlFor={id}>{_.startCase(cClass)}</label>
+			</React.Fragment>;
+		},
+		field : function(cm, n, field) {
```
calculuschild, line 95: I don't know what you think about this, but what if we refactored a bit to move *all* the logic for each widget element out into the `widget-elements` folder. So then all these processing steps/replacement logic lives inside the `Field.jsx` file instead of being another part that has to be passed around.

`cClass` could get the same treatment; we move it into its own file in `widget-elements`, so all widget-elements follow the same file structure., and all of their logic is self-contained. `index.js` just serves to list them all.

themes/V3/5ePHB/widgets.js
jeddai: Yep that all makes sense. My one comment on it all is that I'd prefer we stick to the actual HTML input types -- so `checkbox` is fine, but I'd suggest `text` instead of `textbox`. And then for `imageselector` we'd use `select` and have appropriate additional fields on the object for options with name/value/image/etc...

shared/naturalcrit/codeEditor/helpers/widget-elements/constants.js
jeddai: Fair, that should be `SNIPPET_TYPE`

themes/V3/5ePHB/widgets.js
calculuschild: My only comment to that is we might run into some custom input types (like maybe "color selector" is one of those) that doesn't have a standard HTML Element. But deciding on a name for that can be handled when we get to it.

shared/naturalcrit/codeEditor/helpers/widget-elements/constants.js
calculuschild: Ok, in that case could we rename these types to: BLOCK, INLINE, and INJECTOR? Since the single { we use for images can be applied to other elements to.

shared/naturalcrit/codeEditor/helpers/widgets.js
jeddai: Fair, I'll update it to `render`

shared/naturalcrit/codeEditor/helpers/widget-elements/index.js
```diff
@@ -1,119 +1,7 @@
...
-				}
-				cm.replaceRange(value, CodeMirror.Pos(n, index), CodeMirror.Pos(n, index + current.length), '+insert');
-			};
-			return <React.Fragment key={`${field.name}-${n}`}>
-				<Field field={field} value={value} n={n} onChange={inputChange} setHints={(f, h)=>setHints(h, f)} getStyleHints={getStyleHints}/>
-			</React.Fragment>;
-		}
-	};
+module.exports = {
+	Field    : Field,
```
calculuschild, line 5: Can we call this one "Text" now? To match `FIELD_TYPE.TEXT`? Seems we are using the word Field to mean both this textbox widget specifically, and all widget types in general.

shared/naturalcrit/codeEditor/helpers/widget-elements/constants.js
```diff
@@ -5,26 +5,27 @@ export const HINT_TYPE = {
 	NUMBER_SUFFIX : 1
 };
 
-export const WIDGET_TYPE = {
-	SNIPPET        : 0,
+export const SNIPPET_TYPE = {
+	DEFAULT        : 0,
 	INLINE_SNIPPET : 1,
 	IMAGE          : 2,
```
calculuschild: Can we name these "BLOCK", "INLINE", AND "INJECTOR"?

shared/naturalcrit/codeEditor/helpers/widgets.js
```diff
@@ -1,38 +1,111 @@
...
+	'indeterminate'    : 1, 'invalid'          : 1, 'lang'             : 1, 'last-child'       : 1, 'last-of-type'     : 1,
+	'link'             : 1, 'not'              : 1, 'nth-child'        : 1, 'nth-last-child'   : 1, 'nth-last-of-type' : 1,
+	'nth-of-type'      : 1, 'only-of-type'     : 1, 'only-child'       : 1, 'optional'         : 1, 'out-of-range'     : 1,
+	'placeholder'      : 1, 'read-only'        : 1, 'read-write'       : 1, 'required'         : 1, 'root'             : 1,
+	'selection'        : 1, 'target'           : 1, 'valid'            : 1, 'visited'          : 1
+};
+
+const genKey = (...args)=>args.join('-');
 
 module.exports = function(CodeMirror, widgets, cm, setHints) {
```
calculuschild, line 20: cm is the codemirror instance, and CodeMirror is the whole library, right? Do we need to have CodeMirror as a parameter here? Could it be `require`'d as a dependency in this file instead?

jeddai: Y'know what yes, _but_ I need to pull its declaration out into its own JS file

jeddai: Done!

shared/naturalcrit/codeEditor/helpers/widget-elements/constants.js
jeddai: Done!

shared/naturalcrit/codeEditor/helpers/widget-elements/index.js
jeddai: Done!

jeddai:
@Gazook89 hyphens, quotes, function values, and links are all taken care of!

@5e-Cleric I updated the coloring to match the snippet bar, let me know what you think

jeddai:
I think the last thing we really need for an mvp is probably to add as many widget options for the base theme as possible without requiring the select box widget or other types of widgets.

calculuschild:
> I think the last thing we really need for an mvp is probably to add as many widget options for the base theme as possible without requiring the select box widget or other types of widgets.

Do you mean adding to the `PHB5E/widgets.js` file?

jeddai:
Yep! And then testing to make sure it all works how we expect. 

I'm just not sure what additional snippets would benefit from this.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -169,9 +146,61 @@ const CodeEditor = createClass({
 		});
 		closeTag.autoCloseCurlyBraces(CodeMirror, this.codeMirror);
 
+		this.setState({
```
calculuschild, line 149: I'm not sure `widgetUtils` needs to be in the component state, since it looks like it's only set once here when the component is first mounted. State is meant to hold variables that should trigger a re-render of the component when they change, but I don't think widgetUtils ever changes after this first `require`.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -169,9 +146,61 @@ const CodeEditor = createClass({
...
+				const widget = this.state.widgetUtils.updateLineWidgets(line);
+				if(widget) {
+					this.setState({
+						widgets : {
+							[line] : widget
+						}
+					});
+				}
+			}
+		});
```
calculuschild, line 182: Do we need to use keys to track a list of widgets for multiple lines at a time, or can we simply make `this.state.widgets` hold the one currently-selected line?

Seems like the intent is only one line of widgets is open at a time anyway; clicking on a line closes the widgets on all other lines.

I.e., simplify this to:

```
		this.codeMirror.on('cursorActivity', (cm)=>{
			const { line } = cm.getCursor();
			const { widgets } = this.codeMirror.lineInfo(line);

			if(!widgets) {
				this.state.widgets.clear(); //clears the previously-opened line of widgets
				this.setState({
					widgets : this.state.widgetUtils.updateLineWidgets(line) || []
				});
			}
		});
```

jeddai: As things currently stand if you click on the gutter button you can open multiple widgets at a time

shared/naturalcrit/codeEditor/codeEditor.jsx
jeddai: I did that so the widget utils are defined per-component instance. I know technically I can throw them on the class as a member var but I felt putting it in the state made more sense because I've noticed some odd things with component member vars

jeddai:
Added a field type for image selection (watercolor, image center mask snippets). It includes a modal and a list of images within it!

jeddai:
> @5e-Cleric: Another option would be to have different color for widgets depending on what kind of snippet they edit (tables green, PHB golden, text editor blue, and images red).

I like this, we could base it off of what color the snippet itself is? I do think this needs to come in a later PR though

> @5e-Cleric: When a widget is open and an input chosen, scrolling pulls the autocomplete dropdown, as it is a fixed component

Yeah I'm not sure what's up with that, I'll compare somethings between my code and CodeMirror's and see if I can fix that

calculuschild:
I can't seem to delete the last letter in the text widgets:

image

calculuschild:
> @5e-Cleric: Another option would be to have different color for widgets depending on what kind of snippet they edit (tables green, PHB golden, text editor blue, and images red).

Along those lines, I think it would be helpful to indicate where the different widgets are coming from using some color scheme. For example, any fields coming from the `monster` snippet could be grouped/colored similarly, and snippets that aren't present yet but are suggested might be faded out or something.

image

jeddai:
> I can't seem to delete the last letter in the text widgets:

Fixed

> the hints seems to be back to suggesting absolutely everything

I think this might've been fixed by proxy for color at least — added a color-selector input

> would be helpful to indicate where the different widgets are coming from using some color scheme

Done based off the scheme you provided

5e-Cleric:
As per right now, the js is picking non-existent properties:

image


doing so, you can add a coma in the right place inside the left input, and it will automatically separate them into two correct inputs, i am unsure if this is a desired behaviour.


On a separate matter:

this needs more work, as they are edge masks and are not really visible in this selector
image

calculuschild:
> this needs more work, as they are edge masks and are not really visible in this selector

Perhaps we need to make a separate set of thumbnail images for the masks that shows a nice closeup of the edge.

jeddai:
> doing so, you can add a coma in the right place inside the left input, and it will automatically separate them into two correct inputs, i am unsure if this is a desired behaviour.

Considering this is caused by the user incorrectly defining this section of their brew, I would argue this is acceptable. If you're noting that the CSS property doesn't exist, frankly that's either going to have to be on the user or we're going to need to define CSS linting rules for the hb doc type

calculuschild:
I agree that's on the user if they type invalid properties. It keeps it flexible in case it's some custom property/browser feature that a linter would miss.

5e-Cleric:
From the conversation in Gitter chat, a new widget is suggested for the Table of Contents snippet, with the following features:

- Option to be wide or not

- Option to choose between 1 to 3 columns

- Option to choose how many levels of headers are wanted: from 2 all the way to 6, either by checkboxes or a slider or select.

This still needs some work on the TOC generation, as headers within most per default blocks should be ignored, such as monster or note blocks. Even a full rework of the TOC generation may be the best idea,, as stated in the chat:

> it might be worth updating the TOC generator to use the Markdown Lexer to get a tree of tokens instead of just searching for `###` in the text. That would let us see if the headers are nested inside other objects or not which would make it easier to handle headers inside monster blocks, etc.

Gazook89:
> > this needs more work, as they are edge masks and are not really visible in this selector
> 
> Perhaps we need to make a separate set of thumbnail images for the masks that shows a nice closeup of the edge.

I think it might get tricky to have a separate thumbnail, because if you try to focus on the "edge" at too close of a scale, you lose the overall look?  But of course, I suppose the position of the edge is already selected once you are looking at these thumbnails, so not a big deal.  

Anyway, here is a quick take on making the modal a little easier to look to at, below.  I think that it's important to see the overall page shape, so gave it that gradient as a stand-in for the page/image being masked.  I also *feel like* the overall style of HB is square corners rather than rounded on ui elements, so removed the border radius on the modal but left it on the image mask choices.

image

(I just did this in the dev tools, not with VS Code, but I can replicate it again as a branch of this PR if we want to continue down this road).

5e-Cleric:
> (I just did this in the dev tools, not with VS Code, but I can replicate it again as a branch of this PR if we want to continue down this road).

Looks good to me, although it is very difficult to see the difference between them
i guess corner and center are easier to diferentiate

calculuschild:
> Option to choose between 1 to 3 columns

I would just leave it open to any number.

> Option to choose how many levels of headers are wanted: from 2 all the way to 6, either by checkboxes or a slider or select.

I'm not sure how this would translate to CSS for the element.

Gazook89:
TOC CSS Widget options:  Column-count, leader lines, visible footer/bar

calculuschild:
> I think it might get tricky to have a separate thumbnail, because if you try to focus on the "edge" at too close of a scale, you lose the overall look? But of course, I suppose the position of the edge is already selected once you are looking at these thumbnails, so not a big deal.

For the edges at least, the overall shape is just a line; the texture up close is the only real difference. Plus, the edge masks are deliberately made quite a bit larger than the actual page size to allow rotating around in any orientation and still cover the entire page (i.e., diagonally), so the majority of the image in the thumbnail is usually going to be cut off the page anyway.

Mask element highlighted:

image

Actual page size compared to thumbnail (green):

image

And given that most of the vertical portion of the image is solid black or white, I think we could just focus in like this:

image

5e-Cleric:
Orphaned for 3 years, closing.

PR #2906, opened 2023-07-03
[WIDGETS] Simplify click-outside close widget logic

calculuschild:
use `Element.matches()` which lets us use a CSS selector to see if the target is a child of the widget.

### DIFF
--- a/shared/naturalcrit/codeEditor/codeEditor.jsx
+++ b/shared/naturalcrit/codeEditor/codeEditor.jsx
@@ -453,16 +453,8 @@ const CodeEditor = createClass({
 		};
 	},
 	handleMouseDown : function(e) {
-		let target = e.target;
-		let found = false;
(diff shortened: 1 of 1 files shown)


shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -453,16 +453,8 @@ const CodeEditor = createClass({
...
-		while (target.parentElement) {
-			target = target.parentElement;
-			if(target.classList.contains('CodeMirror-linewidget')) {
-				found = true;
-				break;
-			}
-		}
-		if(!found) {
+		// Close open widgets if click outside of a widget
+		if(!e.target.matches('.CodeMirror-linewidget *')) {
```
jeddai, line 457: Oh this is dope, didn't know about this method!

PR #2944, opened 2023-07-19
Add save location setting to local storage

### LINKED ISSUE #2943: Add option to set default save location (Homebrewery/Google Drive)
### Your idea:

It would be useful if users could set a default save location - that is, if a user has linked their account to Google but has decided that they do want to store their brews on the internal Homebrewery database rather than in Google Drive.
This would allow those who want to use Google as an authentication system only to do so.

This can currently be achieved by saving the new brew, then clicking the Google Drive button to transfer the brew back to Homebrewery. While this is only a few additional steps, it is an additional load on the user to carry out on every new brew - and one that we could eliminate by resolving this issue.

G-Ambatte:
This PR resolves #2943.

This PR adds the ability to set the default save location to Homebrewery's internal database when the user has linked their HB account to Google (i.e. the user no longer wishes to use Google Drive for storage).


client/homebrew/pages/accountPage/accountPage.jsx
```diff
@@ -29,6 +31,32 @@ const AccountPage = createClass({
...
+	componentDidMount : function(){
+		if(!this.state.saveLocation && this.props.uiItems.username) {
+			SAVEKEY = `HOMEBREWERY-DEFAULT-SAVE-LOCATION-${this.props.uiItems.username}`;
+			let saveLocation =  window.localStorage.getItem(SAVEKEY) || '';
+			if(saveLocation == '') {
+				saveLocation = 'HOMEBREWERY';
+				if(this.state.uiItems.googleId) {
+					saveLocation = 'GOOGLE-DRIVE';
+				}
+			};
```
calculuschild, line 43:
```suggestion
			let saveLocation =  window.localStorage.getItem(SAVEKEY);
			saveLocation = saveLocation ?? this.state.uiItems.googleId ? 'GOOGLE-DRIVE' : 'HOMEBREWERY';
```

client/homebrew/pages/accountPage/accountPage.jsx
```diff
@@ -29,6 +31,32 @@ const AccountPage = createClass({
...
+
+	makeActive : function(newSelection){
+		if(this.state.saveLocation == newSelection) return;
+		window.localStorage.setItem(SAVEKEY, newSelection);
+		this.setState({
+			saveLocation : newSelection
+		});
+	},
+
+	renderButton : function(name, key, condition=true){
```
calculuschild, line 56: Can we rename "condition" to something more descriptive? Like "shouldRender" or something?

calculuschild (changes requested):
Just a couple small cleanup items.

client/homebrew/pages/accountPage/accountPage.jsx
G-Ambatte: Applied in commit [7feaa51](https://github.com/naturalcrit/homebrewery/pull/2944/commits/7feaa51de081957dfa6730f11dafe1811cd78dbd)

PR #2945, opened 2023-07-19
exclude covers from TOC generation

### LINKED ISSUE #2920: TOC should ignore covers
# Introduction to the issue

The table of contents snippet tracks down headers to be created, this is the desired effect, but not every header should be tracked.

As in the PHB, the Cover page, the inside cover, and the back cover should not figure in the table of contents.

Furthermore, from the Part covers, only the number of the part should be copied:

image

As you can see, only Part 1 figures, but not Creating a character, which its the name part.

This Table of contents will have to be modified again if other snippets require so.

5e-Cleric:
Fixes #2920 

there is still the matter that h3 should be ignored to keep the theme true to the book standard, but i feel it would be detrimental to the user experience overall.

### DIFF
--- a/themes/V3/5ePHB/snippets/tableOfContents.gen.js
+++ b/themes/V3/5ePHB/snippets/tableOfContents.gen.js
@@ -29,21 +29,23 @@ const getTOC = (pages)=>{
 
 	const res = [];
 	_.each(pages, (page, pageNum)=>{
-		const lines = page.split('\n');
-		_.each(lines, (line)=>{
-			if(_.startsWith(line, '# ')){
-				const title = line.replace('# ', '');
-				add1(title, pageNum);
-			}
-			if(_.startsWith(line, '## ')){
-				const title = line.replace('## ', '');
-				add2(title, pageNum);
-			}
-			if(_.startsWith(line, '### ')){
-				const title = line.replace('### ', '');
-				add3(title, pageNum);
-			}
-		});
+		if(!page.includes("{{frontCover}}") && !page.includes("{{insideCover}}") && !page.includes("{{partCover}}") && !page.includes("{{backCover}}")) {
+			const lines = page.split('\n');
+			_.each(lines, (line)=>{
+				if(_.startsWith(line, '# ')){
+					const title = line.replace('# ', '');
+					add1(title, pageNum);
+				}
+				if(_.startsWith(line, '## ')){
+					const title = line.replace('## ', '');
+					add2(title, pageNum);
+				}
+				if(_.startsWith(line, '### ')){
+					const title = line.replace('### ', '');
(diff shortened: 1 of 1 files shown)


Gazook89:
I'll point out that this also doesn't fix a third point in your Issue:  that the *Part* pages are having both their headers included in the generated TOC.  Since the generator is just grabbing every h1, h2, and h3 it is always going to grab the h2 *part* name.   Programmatically cutting out that one particular h2 header might be a pain.

The simpler solution, which requires us claiming "BETA!", is to reduce the h2 header into base text that is styled to look like that h2 header.  Semantically, the Part name shouldn't be a header anyway, because it has no content itself before the next header.

### Current:
```
{{partCover}}

# PART X
## Lost Caverns

{{imageMaskEdge8,--offset:10cm,--rotation:180
  ![Background image](https://i.imgur.com/9TU96xY.jpg){position:absolute,bottom:0,left:0,height:100%}
}}
```

### Revised:
```
{{partCover}}

# PART X
Lost Caverns

{{imageMaskEdge8,--offset:10cm,--rotation:180
  ![Background image](https://i.imgur.com/9TU96xY.jpg){position:absolute,bottom:0,left:0,height:100%}
}}
```

#### Revised CSS
```css
.page:has(.PartCover) h1 + p:first-letter {
	all: unset;
}

.page:has(.partCover) h1 + p {
	font-family: Overpass;
	font-size: .45cm;
	position: relative;
	margin-top: -0.7em;
	font-weight: 800;
	line-height: 1.1em;
	margin-left: auto;
	margin-right: auto;
	display: block;
	text-align: center;
}
```

Screenshot of the result, replacing the h2 with regular text:
image

5e-Cleric:
> I'll point out that this also doesn't fix a third point in your Issue: that the Part pages are having both their headers included in the generated TOC.

Yes it does, since it ignores any heading inside a page with the part cover snippet, perhaps i wrote the class wrong?

ericscheid:
A stray thought .. perhaps we could set it up to use `querySelectorAll()` to find the headings for the TOC, and let the brew author edit the css selectors (for extensibility).

Gazook89:
> > I'll point out that this also doesn't fix a third point in your Issue: that the Part pages are having both their headers included in the generated TOC.
> 
> Yes it does, since it ignores any heading inside a page with the part cover snippet, perhaps i wrote the class wrong?

So it wouldn’t grab “Part 1” either?

5e-Cleric:
> So it wouldn’t grab “Part 1” either?

It shouldn't

calculuschild:
Yep, this looks like it's functioning properly. All the different coverpage types are totally ignored so any headers inside them are not added to the TOC. :+1:

I'll merge this. Thanks @5e-Cleric !

PR #2955, opened 2023-07-27
Add User auth token to getBrew

### LINKED ISSUE #2954: Brews removed from Google Drive cannot be deleted from "My Brews" page.
See https://www.reddit.com/r/homebrewery/comments/158k1z5/brew_with_missing_google_drive_file_cannot_be/?utm_source=share&utm_medium=android_app&utm_name=androidcss&utm_term=1&utm_content=2

Logs seem to show we hit an error trying to retrieve the brew since it doesn't exist, and so we fail out without proceeding to just remove the stub.

G-Ambatte:
This PR resolves #2954.

This PR adds the ability to pass an authentication token to the `getBrew` function, defaulting to the `defaultAuth` (Service Account) when no auth token is provided. This functionality is then used by `deleteBrew`, so any brew that can be seen by the user (and thus appear on their User Page) can be deleted.


G-Ambatte:
It looks like these changes have caused quite a lot of tests to break. I'll try to look into it in the next few days.

server/homebrew.api.js
```diff
@@ -287,15 +287,21 @@ const api = {
 
 		res.status(200).send(saved);
 	},
+	getGoogleAuth : function(account, res){
+		const auth = GoogleActions.authCheck(account, res);
+		return auth;
+	},
 	deleteGoogleBrew : async (account, id, editId, res)=>{
-		const auth = await GoogleActions.authCheck(account, res);
+		const auth = api.getGoogleAuth(account, res);
```
calculuschild, line 295: I think we want to make sure we keep the `await` here, since it's possible getting the Auth needs to get refreshed credentials from Google; we don't want to get `undefined` as a result if that call to Google has any delay. In general any call to the `googleActions` stuff should be awaited.

server/homebrew.api.js
```diff
@@ -287,15 +287,21 @@ const api = {
...
 		await GoogleActions.deleteGoogleBrew(auth, id, editId);
 		return true;
 	},
 	deleteBrew : async (req, res, next)=>{
 		// Delete an orphaned stub if its Google brew doesn't exist
 		try {
-			await api.getBrew('edit')(req, res, ()=>{});
+			const { googleId } = api.getId(req);
+			const auth = googleId ? api.getGoogleAuth(req.account, res) : undefined;
+			await api.getBrew('edit', false, auth)(req, res, ()=>{});
```
calculuschild, line 304: Can you confirm this allows the "restricted sharing" google brew to be deleted?

Is this going to cause issues with co-authors who want to delete the brew from their own account in normal cases? i.e., if User A makes the Google Brew, invites User B as an author, then if User B tries to delete, my guess is it will fail because the Service Account has delete permissions but User B's account doesn't.

calculuschild (changes requested):
Small changes and some questions.

server/homebrew.api.js
G-Ambatte: 1. Yes, it does - the document is trashed from the Google account to the Bin.
2. If I understand the scenario, Author B shouldn't be deleting the brew, but rather only removing themselves from the list of authors - which from memory is held in the MongoDB stub. If the owner of the Google Drive file attempts to delete it but there are other authors, the brew is moved to MongoDB and the original owner removed from the authors list.

server/homebrew.api.js
```diff
@@ -287,15 +287,21 @@ const api = {
 
 		res.status(200).send(saved);
 	},
+	getGoogleAuth : function(account, res){
+		const auth = GoogleActions.authCheck(account, res);
+		return auth;
+	},
```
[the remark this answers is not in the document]
G-Ambatte: I originally put that in thinking it would see more use; I'll revert the change now that development has shown that not to be the case.

server/homebrew.api.js
G-Ambatte: Yes - brain fade moment. I'll revert that.

server/homebrew.api.js
calculuschild: 2) Yes, I mean removing themselves as an author. But with Author B no longer using ServiceAccount, isn't `getBrew()` going to fail for them and exit `deleteBrew()` prematurely before they are removed as an author later on in this function?

calculuschild: @G-Ambatte Can you confirm that for a brew with two co-authors, if I delete a brew as Author B (non-owner), it just removes from my account and does not delete the original (or cause an error because it can't retrieve the brew without serviceauth)?

server/homebrew.api.js
calculuschild: If you reverted this can you push up the change?

server/homebrew.api.js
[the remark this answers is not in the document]
calculuschild: If you reverted this can you push up the change?

calculuschild:
@G-Ambatte I think this one is really close but for the two open comments I have up there, and the breaking tests. Do you have any insight on those?

server/homebrew.api.js
G-Ambatte: > I think deleteBrew() will not work as expected for Author B.

I've been going over this again, and I think you're exactly right - if two authors both are linked to Google, the delete request will fail because of incorrect permissions, which Google normally returns as a 404 - there is no externally identifiable difference between "blocked because it doesn't exist" and "blocked because disallowed". This will trigger the `err.HBErrorCode == '02'` block and delete the whole stub.

I also realize now that this is because that is the intended purpose of this section of code; I was modifying the wrong section of code for my intended purpose.

5e-Cleric:
Okay, i have been informed that there is a blocking error in this PR, related to google auth(as is the whole PR).

I know nothing about this, is anyone in the mood to try to look into it?

calculuschild:
Closing in favor of #3943

PR #2961, opened 2023-07-31
Simple adjustments for mobile viewing

Gazook89:
This PR takes a few simple steps to improve the mobile aspect of the app without going to far into the weeds.

It adds a `viewport` meta tag to set the initial size of the app to the device width and height (edit: I totally missed that we already have this meta tag, even when i *looked* for it).  This isn't perfect, because the mobile browsers (at least iOS browsers) cover a portion of the viewport with their address bars or other items like that, but it's better.  Further improvement could likely be done by utilizing `window.screen.availHeight` or some such, but I didn't want to get into the JS of it all.

It also changes the `onmousedown` and similar events on the splitView component, and uses `onpointerdown` etc.  Now the divider bar can be moved with a touch.  One improvement that I'm a little stuck on is that the pointer event seems to require a 'hard press' or long press on the divider before it "picks it up" and allows you to move it.  I wasn't sure if this was a bad thing, though.

This PR doesn't touch any real CSS, doesn't move buttons around, so everything is basically the same. It still works on desktop.

I used my iphone with Safari and Firefox (webkit) to test it.

### DIFF
--- a/client/template.js
+++ b/client/template.js
@@ -11,8 +11,7 @@ const template = async function(name, title='', props = {}){
 	return `<!DOCTYPE html>
 	<html>
 		<head>
-			<meta name="viewport" content="width=device-width, initial-scale=1" />
-
+			<meta name="viewport" content="width=device-width, initial-scale=1, height=device-height, interactive-widget=resizes-visual" />
 			<link href="//use.fontawesome.com/releases/v5.15.1/css/all.css" rel="stylesheet" />
 			<link href="//fonts.googleapis.com/css?family=Open+Sans:400,300,600,700" rel="stylesheet" type="text/css" />
 			<link href=${`/${name}/bundle.css`} rel='stylesheet' />
--- a/client/template.js
+++ b/client/template.js
@@ -11,6 +11,8 @@ const template = async function(name, title='', props = {}){
 	return `<!DOCTYPE html>
 	<html>
 		<head>
+			<meta name="viewport" content="width=device-width, initial-scale=1" />
+
 			<link href="//use.fontawesome.com/releases/v5.15.1/css/all.css" rel="stylesheet" />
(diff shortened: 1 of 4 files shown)


Gazook89:
https://javascript.info/pointer-events

Just started reading this article and I think I can make improvements to this PR before merge

Gazook89:
I have another thing to possibly add to his PR:  setting the font-size in the Editor to 16px if viewing on mobile (rather than the default `100%`.  This is because iOS forces a zoom on inputs if the font-size is smaller than that, which is very annoying *for me*, but I suppose is an accessibility thing.  Likely the best option is to actually set a *minimum* size of 16px, in case someone has user styles that makes it larger.

Question though: is it enough to do a media query for `@media screen and (pointer : coarse) {...}` for all touch screen devices, or should I set a max-width on it to to focus only on smaller screens?

Sometimes even with a larger touchscreen device it's nice to have larger text than normal because you can make selections/insert text easier.

calculuschild (changes requested):
Ok, this seems to make things quite a bit less buggy on mobile.

Just remove the console logs and I'll merge.

calculuschild:
>This is because iOS forces a zoom on inputs if the font-size is smaller than that, which is very annoying for me, but I suppose is an accessibility thing. 

Yeah there must be something similar on Android because I notice the same thing. This PR seems to stop that. I personally thing 16px looks absolutely *huge* on mobile, but the again a phone isn't a great device for mass text editing anyway, so I'd rather have big font than random zooming.

Gazook89:
Random zooming is particularly bad on HB because there isn’t much room to pinch-zoom out back to fill size (for the UI). Most of the screen space is text input or an iFrame. 

I’ll remove console logs maybe this afternoon

Gazook89:
I guess I already had logs removed, just didn't push that commit until now.  Should be good to go?

PR #2969, opened 2023-08-07
Add Google file link to UserPage BrewItem

G-Ambatte:
This PR follows up on a side issue raised in #2954.

This PR adds a link to the file on Google Drive via the Drive Icon on the User Page. In the event that a user cannot find their brew document but it does still exist in their Drive somewhere, they can use this link to locate the file.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
@@ -94,7 +94,9 @@ const BrewItem = createClass({
 		if(!this.props.brew.googleId) return;
 
 		return <span>
-			<img className='googleDriveIcon' src={googleDriveIcon} alt='googleDriveIcon' />
+			<a href={this.props.brew.webViewLink} target='_blank'>
+				<img className='googleDriveIcon' src={googleDriveIcon} alt='googleDriveIcon' />
+			</a>
 		</span>;
 	},
 
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
@@ -92,14 +92,16 @@ const BrewItem = createClass({
 	},
 
 	renderStorageIcon : function(){
-		if(!this.props.brew.googleId) return <span title='Internal Homebrewery Database'>
-			<img className='homebreweryIcon' src={homebreweryIcon} alt='homebreweryIcon' />
-		</span>;
+		if(this.props.brew.googleId) {
+			return <span title={this.props.brew.webViewLink ? 'Your Google Drive Storage': 'Another User\'s Google Drive Storage'}>
+				<a href={this.props.brew.webViewLink} target='_blank'>
+					<img className='googleDriveIcon' src={googleDriveIcon} alt='googleDriveIcon' />
+				</a>
+			</span>;
+		}
 
-		return <span title='Google Drive Storage'>
-			<a href={this.props.brew.webViewLink} target='_blank'>
-				<img className='googleDriveIcon' src={googleDriveIcon} alt='googleDriveIcon' />
-			</a>
+		return <span title='Homebrewery Storage'>
+			<img className='homebreweryIcon' src={homebreweryIcon} alt='homebreweryIcon' />
 		</span>;
 	},
 
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
@@ -7,6 +7,7 @@ const moment = require('moment');
 const request = require('../../../../utils/request-middleware.js');
 
 const googleDriveIcon = require('../../../../googleDrive.svg');
+const homebreweryIcon = require('../../../../thumbnail.png');
 const dedent = require('dedent-tabs').default;
 
 const BrewItem = createClass({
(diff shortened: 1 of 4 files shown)


G-Ambatte:
The second part of this PR adds a small icon for brews stored on the local MongoDB database, to differentiate them from the brews stored on Google Drive.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
```diff
@@ -90,11 +91,15 @@ const BrewItem = createClass({
...
-	renderGoogleDriveIcon : function(){
-		if(!this.props.brew.googleId) return;
+	renderStorageIcon : function(){
+		if(!this.props.brew.googleId) return <span title='Internal Homebrewery Database'>
+			<img className='homebreweryIcon' src={homebreweryIcon} alt='homebreweryIcon' />
+		</span>;
 
-		return <span>
-			<img className='googleDriveIcon' src={googleDriveIcon} alt='googleDriveIcon' />
+		return <span title='Google Drive Storage'>
```
Gazook89, line 99: Maybe the Title should be “Saved on [User]’s Google Drive”?

ericscheid:
Does webViewLink work if the user looking at the Users Page is not the owner of that User Page?

For example, if anyone but me looks at https://homebrewery.naturalcrit.com/user/erics, will the Google Drive link for the one brew I have there on gdrive work for them?

G-Ambatte:
> will the Google Drive link for the one brew I have there on gdrive work for them?

A quick perusal of the code shows that when `ownAccount` is false, `GoogleActions.listGoogleBrews()` is not performed, so `webViewLink` cannot be retrieved from Google Drive. The information about the listed brews that are stored on Google Drive comes from the internal Brew stub.
`webViewLink` has not been added to the `fields` array, so even if it does get stored in the Brew stub, it will not be retrieved.

When `webViewLink` is undefined, the resulting `<a>` element has no `href`, and looks like this:

image

calculuschild (commented):
Ok, so as pointed out, this only works *if you are logged in to the owner's google drive account.* If not, the webViewLink is undefined since we are just reading from the stub without actually performing `listGoogleBrews()`; Often the reason a user can't find the file is because they aren't logged in to the right account. So what should we display to the user in that case?

**Edit:** It looks like the webViewLink can by derived by just using the Google ID we have, so we could apply this to all brews on the page, even those from a Stub:

- **googleID** : 1eynUmZHfmTc6fDQupTvxgxlhxGyrwoPw
- **webViewLink** : https://drive.google.com/file/d/1eynUmZHfmTc6fDQupTvxgxlhxGyrwoPw/view?usp=drivesdk

Then the link would work for everyone, even if logged out (or in the wrong account). Displaying whether that file exists is another question, but visiting the link would at least show an error from google in that case:

image

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
```diff
@@ -90,11 +91,15 @@ const BrewItem = createClass({
 		</a>;
 	},
 
-	renderGoogleDriveIcon : function(){
-		if(!this.props.brew.googleId) return;
+	renderStorageIcon : function(){
+		if(!this.props.brew.googleId) return <span title='Internal Homebrewery Database'>
```
calculuschild, line 95: Instead of the negative, `(!this.props.brew.googleId)`, can we reorder the logic so we can use the positive case (`if(this.props.brew.googleId)`?

calculuschild, line 95: Change the title to just `Homebrewery Storage`.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
calculuschild: This could be done with the `owners` field from the Google API. Unfortunately it would only work for brews owned by the current logged-in user where we call `listGoogleBrews()`

Maybe in that case we could display `Another user's Google Drive Storage` to at least make clear that you aren't logged in to the owning account.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
Gazook89: What about NaturalCrit Storage?

calculuschild: Either way. But I think elsewhere (the Account page) we refer to it as Homebrewery Storage already.

G-Ambatte: Absolutely - as you can see, I was just throwing in some output for local storage, rather than the previous nothing.

G-Ambatte:
> Often the reason a user can't find the file is because they aren't logged in to the right account. So what should we display to the user in that case?
> 
> **Edit:** It looks like the webViewLink can by derived by just using the Google ID which we have, so we could apply this to all brews on the page, even those from a Stub:
> 
>     * **googleID** : 1eynUmZHfmTc6fDQupTvxgxlhxGyrwoPw
> 
>     * **webViewLink** : https://drive.google.com/file/d/1eynUmZHfmTc6fDQupTvxgxlhxGyrwoPw/view?usp=drivesdk
> 
> Then the link would work for everyone, even if logged out (or in the wrong account).

Given that every Google brew needs to be set as 'Anyone with the link can Edit' in order for things to work, exposing the link is problematic, from a data security perspective.
My personal tendency would be towards providing the link on the owner's account, "Another user's Google Drive" for non-owners of Google Drive docs, and "Homebrewery Storage" for internal brews.

calculuschild:
> Given that every Google brew needs to be set as 'Anyone with the link can Edit' in order for things to work, exposing the link is problematic, from a data security perspective.

Hmm, good point. I was thinking only about users who are looking at their own accounts, meaning they are already also authors and should have edit privileges anyway, but this would expose the link to any random visitor to your user page too.

What about showing the link only if you are looking at your own brews page? I think there may still be some benefit to being able to link to files you are an author on, even if you don't own it. For the same reasons as before (user has multiple Google accounts and has "lost" the file, etc.)?

G-Ambatte:
> What about showing the link only if you are looking at your own brews page?

That seems reasonable, I think - if you're an author, you can edit the brew, so being able to see the source file poses a minimal additional security exposure. The only negative case I can think of immediately is a removed author might be able to retain access via the Google Drive link in their history. The upside to that is that Google Drive version history allows for the relatively painless restoration of the file in the event that a malicious ex-author vandalizes a file.

G-Ambatte:
Also, the presence of a `this.props.webViewLink` implies the current user's account is the owner, so it might be a better property to test than `authors[0]`.

calculuschild:
Ok, I think this is good enough as is. I'll go ahead and merge. Thanks @G-Ambatte !

PR #2981, opened 2023-08-11
Cleaning CSS for uniformity

5e-Cleric:


themes/V3/5ePHB/style.less
```diff
@@ -684,7 +684,7 @@ h5 + table{
 	columns    : 1;
 	text-align : center;
 	&:after {
-		all: unset;
+		display: none;
```
calculuschild, line 687: Any particular reason for changing this? ` all:unset` completely removes the footer. `display:none` just hides it.

themes/V3/5ePHB/snippets/coverpage.gen.js
```diff
@@ -149,8 +149,6 @@ module.exports = {
 			![](/assets/naturalCritLogoWhite.svg)
 
 			Homebrewery.Naturalcrit.com
-			}}
-
-			\page`;
+			}}`;
```
calculuschild, line 152: The "coverPage" snippets are designed to create a whole new page, with the `\page` break to make sure it is separated cleanly from the following content. I would prefer to keep the page break if possible.

themes/V3/5ePHB/style.less
```diff
@@ -780,16 +780,16 @@ h5 + table{
...
 	}
 	h1 {
-		font-family		 : NodestoCapsCondensed;
-		font-weight		 : normal;
-		font-size		   : 2.1cm;
-		margin-top     : 1.2cm;
+		font-family	   : NodestoCapsCondensed;
+		font-weight	   : normal;
+		font-size	   : 2.1cm;
+		margin-top	   : 1.2cm;
```
calculuschild, line 789: All of the spacing issues should be fixable automatically if you run stylelint on this page. That will probably fix a bunch of other small things too.

themes/V3/5ePHB/style.less
5e-Cleric: Precisely, found users wanting to have a footer in pages that don't have one normally, this change makes easier to bring it back instead of having to redeclare every property. We use display none in the TOC page, so this helps with consistency too.

themes/V3/5ePHB/snippets/coverpage.gen.js
5e-Cleric: This is only for the back cover, do users need a page after that?

themes/V3/5ePHB/style.less
calculuschild: Ok, that makes sense. Thanks!

themes/V3/5ePHB/snippets/coverpage.gen.js
calculuschild: Ok if it's only the back cover that's fine.

calculuschild (commented):
Before merging this I would just want to see an example brew that uses these changed measurements in different combinations so we can show side-by-side how everything lines up. We don't want to accidentally throw off anybody's columns of text because something is one pixel taller or shorter now.

themes/V3/5ePHB/style.less
```diff
@@ -901,8 +901,8 @@ h5 + table{
 			color: #fff;
 			font-family: NodestoCapsWide;
 			font-size: .4cm;
-			letter-spacing: 0.08em;
-			line-height: 1em;
+			letter-spacing: 0.036cm;
+			line-height: .4cm;
```
Gazook89, line 905: These two are all good as well.

image

themes/V3/5ePHB/style.less
```diff
@@ -942,8 +942,8 @@ h5 + table{
 			font-family: Overpass;
 			font-size: 0.45cm;
 			position: relative;
-			margin-top: -0.7em;
-			line-height: 1.1em;
+			margin-top: -0.315cm;
+			line-height: 0.495cm;
```
Gazook89, line 946: And finally, this one is all good

image

Gazook89 (changes requested):
Overall I think this is all really good.  There is one conversion that could be improved and a dropped a magic number into the review of that line.  

Some users may experience some shifting if they have used custom font properties since the change is from a relative measure (`em`) to an absolute measure (`cm`), but overall I think it's best to go ahead with these changes now.

The github deployment wouldn't work for me, so I cloned the branch onto my local to look at it.  I think that before this can merge, the `upstream/master` needs to be merged in, and there are a few non-serious conflicts.  At the same time, StyleLint can be run to clean up the last few lint problems-- it might be best to save those Styelint changes to another PR (after some other PR's are closed) though in order to isolate *actual style changes* from *linter changes*.

themes/V3/5ePHB/style.less
```diff
@@ -876,8 +876,8 @@ h5 + table{
 	}
 	p {
 		font-family: Overpass;
-		line-height: 1.5em;
 		font-size: 0.332cm;
+		line-height: 0.5cm;
```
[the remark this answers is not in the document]
5e-Cleric: I don't see any difference in line height in the first two paragraphs, but i do see a difference in letter-spacing. it is always a pain in the butt to figure which is and by how much.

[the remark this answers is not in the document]
Gazook89: Sorry, this is the example with that corrected .35. I should have included the current .5 for comparison. I can do that later today.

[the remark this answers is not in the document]
Gazook89: Here is the current PR (0.5 line-height) on top of the production site:

image

Again, not huge, but in this type of block text, which can have many lines, the difference compounds.

calculuschild:
@Gazook89 @5e-Cleric What's left to fix here? Are there any final tweaks to apply based on Gazook89's testing?

5e-Cleric:
> @Gazook89 @5e-Cleric What's left to fix here? Are there any final tweaks to apply based on Gazook89's testing?

his first review comment, i will try to have this done tomorrow, if i don't feel free to poke me with a stick.

5e-Cleric:
All done, except linting, check gitter chat for the problem left.

calculuschild:
Ok. I'll merge this now.

Just note your linting step linted the *whole* project, so suddenly this PR had hundreds of changes across 15 files. I was able to undo that but just remember after linting to remove any unrelated files from your commits before pushing up changes.

PR #2984, opened 2023-08-12
Add Mozilla column flow fix

### LINKED ISSUE #2982: Add snippet to change column flow
### Your idea:

The following snippet of code is often used to change the way elements flow within columns in user's brews - commonly due to their browser not supporting column flow in the same way or extent that Chrome does (e.g. Firefox).

Adding this snippet would allow us to redirect users having these issues to an "trusted and official" Style snippet, rather than "copy and paste this code into your document" which should only carry all the trust one would normally associate with talking to a stranger over the Internet.

```
.page .columnWrapper {
  column-fill: auto;
 }
```

G-Ambatte:
This PR resolves #2982.

This PR adds a Mozilla-specific section of styling that changes the Brew column-fill property from `balance` to `auto`, which causes the rendering in Firefox to more accurately match that of Chrome.

**Current**
Chrome:
image

Firefox:
image

**PR**
Chrome:
image

Firefox:
image


calculuschild (commented):
As discussed on the gitter page, I would prefer to just change this for all browsers if it looks like the behavior is going to be the same. As long as test wraps around wide elements properly (and the wide elements themselves also wrap the same in both browsers).

We were using a hack for Chrome, but if its not necessary anymore we can avoid adding special rules for Firefox.

G-Ambatte:
In Chrome, we can set the `column-fill` property of `.page` to `auto` with no negative effects as long as the `break-after` property of `.columnSplit` remains at `always`.
In Firefox, we need to set `column-fill` property of `.columnWrapper` to `inherit` in order for the brew to display correctly.

I'm going to close this PR and create a new one with only the universal changes.

PR #3024, opened 2023-09-04
Shift manual save to use debounced save function

### LINKED ISSUE #2641: Version Error
### Error Code

```shell
"This brew has been modified on another device."
```

### Your description of what happened:

Manually initiating a save sometimes results in the error "This brew has been modified on another device." Repeating the manual save works correctly and the error disappears.

I suspect this is due to the auto-save being in progress when the manual save is initiated, resulting in two calls to the save function with the same internal version number. The error does not seem to occur when triggering two manual saves in close proximity.

image

G-Ambatte:
This PR resolves #2641.

This PR shifts the manual save process to use the same debounced save function as the auto-save functionality. This will eliminate the race condition that leads to the version mismatch error ("This brew has been changed on another device").

As every change triggers the debounced autosave function, this change causes the pending function to be fired immediately using Lodash's built-in `flush` function. This also has the happy side effect of clearing the pending autosave when the manual save is triggered early, reducing the total number of save (and thus DB writes).

### DIFF
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -91,7 +91,7 @@ const EditPage = createClass({
 		if(!(e.ctrlKey || e.metaKey)) return;
 		const S_KEY = 83;
 		const P_KEY = 80;
-		if(e.keyCode == S_KEY) this.save();
+		if(e.keyCode == S_KEY) this.trySave(true);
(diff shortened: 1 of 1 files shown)


calculuschild:
Nice. Seems like an easy fix. Hopefully this was the main source of this bug; we have had a lot of people reporting this on Reddit lately.

Thanks @G-Ambatte ! 🎉 ⭐ 💯

PR #3059, opened 2023-09-20
Radix UI

Gazook89:
_Woops, pushed the "create draft" thinking I'd still have a chance to type stuff.  Anywhere, here is an edit._

# +3203, -1322

This PR is just a real doozy of a PR and I'm pretty much creating it so that the effort I put into the branch isn't lost forever in case I lose interest again for many months (though winter is coming, and I expect I'll be in the basement a lot more, again).

This PR is my effort to rework the UI into something more at-home on mobile devices but also to revise the desktop UI as well.  I've sort of mentioned this effort in other Discussions, Issues, and in Gitter occasionally and I don't want to go into all the specifics right now, but a couple of points I've kept in mind:
1. At the very least, the Share page should work better on mobile devices for those just perusing published brews.
2. The current navbar is too crowded already, as is the snippet bar.  Rearranging some pieces into logical groupings and adding some more button real estate will be helpful now and for future feature additions
3. Avoid having to tell people "you need to add CSS in the Style Editor *(which is the Paint brush icon to the right of the snippet menus)*"  (example today: https://discord.com/channels/113801089362558983/950919905266507806/1155663047868825720)
4. Start breaking some common components into their own components (which I haven't gone *too* deep into just yet but I've been making a list)
5. Dark theme

There are some other points that have come up while doing this that I haven't really added to the above points yet, but they include:
1. Why are the New, Share, and Edit pages not just one page or component?  Why is the Save functions repeated on those pages?
2. Separate the editors into components that can just be plugged into the editor container (which in this PR is a Radix Tabs component), so that additional tabs/editors can be plugged in when needed.

I started this a long time ago, and it's extremely piece-meal and amateurish as I just learn more about Radix (and in fact took a long time to settle between Radix and Shoelace, and I'm not even sure I picked correctly) and JS in general.  I sort of expect to keep working on this, but in the end it likely needs to be rebooted as a fresh branch and implemented in a more straightforward manner/line of commits, if it's used *at all*.  I know that in the meantime a lot of PRs have been merged that fix some of the same issues or take things in a different direction.  It is what it is.

I wanted to get this where someone else could see it, and I *think* that if it's deployed as a test, it *should* work well enough to get the gist of it.  It *definitely* does not have everything working:
a. on iOS there is some issue with clicking on the editor tabs....iOS does something weird with touch events and I spent a lot of time this spring trying to figure it out but then I paused work on it and have since forgotten where I left off....Right now, I'm not enthused about that.
b. You should be able to go to the home page, new page, share page, and edit page.  User page doesn't work.  I assume Source and Print work but maybe not.
c. Some things have just been commented out if they weren't my focus and were breaking things.  Usually navbar related stuff.
d. keep in mind, *major* work in progress...some things that might seem like a poor design decision may just be a temporary casualty (though it could be poor design too)

Anyway, that's a lot of typing for now.  ciao.


calculuschild:
> Why are the New, Share, and Edit pages not just one page or component? Why is the Save functions repeated on those pages?

See #1423,  #1799 , #1800, #1801

This has been in the works for literal years, but always dies out before it gets completed. Feel free to tackle that one if you want.

calculuschild:
Tried to deploy this just now but it need to have the conflicts resolved. The version of NPM and Node have a conflict.

Gazook89:
I'm going to close this PR as it's too much all at once to be practical, and if it's not going to happen as-is, then there is no reason to keep it open.  I still am interested in getting some things overhauled but I'm going to restart and take smaller bites with a better understanding of how things work from the outset.

PR #3066, opened 2023-09-27
Single Quote in in-line styles

5e-Cleric:
This PR introduces the possibility of using single quotation marks inside inline styles to be able to use strings as values.

It does so by changing the regex pattern for mustache spans, blocks and injects.

Note: in order to use them, you still need double quotation marks:
`--variable:"'string'"` is a correct and expected syntax.


tests/markdown/mustache-syntax.test.js
```diff
@@ -291,13 +317,13 @@ describe('Injection: When an injection tag follows an element', ()=>{
 			expect(rendered, `Input:\n${source}`, { showPrefix: false }).toBe('<div class="block" style="color:red;"><p>text</p></div>');
 		});
 
-		it.failing('renders a div "text" with two injected styles', function() {
+		it.failing('renders a div "text" with two injected styles,included quotes', function() {
```
calculuschild, line 320: Rather than overwriting this test, which was meant to test two normal CSS properties, can you please create a separate test for your string variable?

calculuschild:
@5e-Cleric Looks good! Sorry this took so long! 🎉 ⭐ 🎆

PR #3068, opened 2023-09-28
Check for author in author list before increasing views.

### LINKED ISSUE #3037: Authors view should not count for view-count
As i said in the title, i don't think an author's checking of the share page should be counted in the `views` metadata of a brew.

Gazook89:
This PR fixes #3037 by doing a check whether the current user is included in the brew's author list before increasing the view count.

**I have not tested this on brews with multiple authors**, but I think that shouldn't be an issue?

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -327,14 +327,17 @@ app.get('/share/:id', asyncHandler(getBrew('share')), asyncHandler(async (req, r
 		type        : 'article'
 	};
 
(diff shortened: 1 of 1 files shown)


calculuschild:
There was a bug causing /share to crash when not logged in. `req.account` doesn't exist so it would crash checking for `req.account.username`. Small logic tweak and now seems to work as expected.

Thanks @Gazook89 !

PR #3071, opened 2023-09-28
Add basic localStorage mechanism to check user status at save

### LINKED ISSUE #2975: Saving an owned brew while not logged in causes perpetual spinning save icon
If you somehow reach the edit page of a brew without being logged in properly as one of the authors (you signed out while the tab was already open, you are removed as an author while the document is open, etc.), making an edit and then trying to save will give you a perpetual spinning save icon, instead of any warning that you need to be logged in.

This is not clear to the user that they need to refresh their login.

For reference, the server logs this error:

image

And the client browser simply gets this:

image

Coming from this line:
image

It shouldn't be possible for this to happen under most normal use, but it can lead to lost work if the user isn't getting a proper warning that the save is failing.

Gazook89:
This is a draft, and likely a weak one, to fix #2975.  It doesn't pop any error right now besides a little message in the console.  It also only works on **LOCAL** because I didn't want to dig into the NaturalCrit system (I thought I might have to) and because I'm sort of doubtful this is the best method anyway.  

### Method
This is basically creating a localStorage key `username` anytime you do a local login.  Then, if you are logged in with multiple tabs, if you log out from any of those tabs the localStorage key is deleted which updates in all tabs.  When you try to save in any of those tabs it checks for that `username` key and only continues with the save if the key exists.

### Possible alternatives and concerns
I wanted to try checking not just that the user was signed in, but was the correct user (in case they sign out and switch accounts in one tab, and try to save in another), so I'm dropping the username into localStorage.  If we didn't care about that, we could just check for the existence of the nc_session cookie.  

I'm not super sure on security of this?  I'm nervous that someone could just change the localStorage key value to whatever they want....but not sure what that would accomplish?  

I'm not very sure about how the log-in system works on the production site, and I don't think i'm ready to learn(?), so maybe this method won't translate well.

It'd be neat to immediately show the effect of logging out across all open tabs, right in the navbar:  when you log out, it could automatically update the `<Account />` menu component to remove the username.  As I was reading, I thought I saw a method of doing this with React Router but I didn't dig into it.  Possibly it's easier in newer React with useContext()?

### DIFF
--- a/client/homebrew/navbar/account.navitem.jsx
+++ b/client/homebrew/navbar/account.navitem.jsx
@@ -32,6 +32,7 @@ const Account = createClass({
 				}
 				domain = domainArray.join('.');
 			}
+			window.localStorage.removeItem('username');
 			document.cookie = `nc_session=;expires=Thu, 01 Jan 1970 00:00:01 GMT;path=/;samesite=lax;${domain ? `domain=${domain}` : ''}`;
 			window.location = '/';
 		}
@@ -53,7 +54,7 @@ const Account = createClass({
 					console.warn(err);
 				});
 		if(!token) return;
-
+		window.localStorage.setItem('username', username);
 		document.cookie = `nc_session=${token};expires=${expiry};path=/;samesite=lax;${window.domain ? `domain=${window.domain}` : ''}`;
 		window.location.reload(true);
 	},
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -183,6 +183,13 @@ const EditPage = createClass({
 	save : async function(){
 		if(this.debounceSave && this.debounceSave.cancel) this.debounceSave.cancel();
 
+		if(this.state.brew.authors.includes(window.localStorage.getItem('username') || '') == false){
+			console.log('not logged in');
+			return;
+		};
+
+
+
 		this.setState((prevState)=>({
 			isSaving   : true,
 			error      : null,


calculuschild:
I don't think any of this is necessary, as the user status is already tracked in a cookie. Once a user signs out of one tab, the cookie is updated for all tabs.

Gazook89:
Okay, I don't think I understand how to surface the error that I get on the VS Code side to the browser side.  I see this in VS Code:
```
{
  name: 'Access Error',
  status: 401,
  message: 'User is not logged in',
  HBErrorCode: '04',
  authors: [ 'Gazook89' ],
  brewTitle: 'Here is my test brew.'
}
```
which i would like to pass to `this.state.error` to use the error notification component, but I'm not even sure at what point this is getting logged to the vs code console. 

I would think that the issue is that `res.body` returns null, and then fails when trying to replaceState since there isn't a `editId` property on it.  And I can (and have) set up a check to see if `res.body` exists and exit the save function if it doesn't.  But I don't know how to trigger that error notification component / how to setState of `this.state.error` with an appropriate value.
```js
		const res = await request
			.put(`/api/update/${brew.editId}${params}`)
			.send(brew)
			.catch((err)=>{
				console.log('Error Updating Local Brew');
				this.setState({ error: err });
			});
		if(!res) return;

		this.savedBrew = res.body;
		history.replaceState(null, null, `/edit/${this.savedBrew.editId}`);
```

I'm likely over(under)thinking about what needs to be done.

calculuschild:
Ok, so if you look in `homebrew.api.js`, you should see several places where we generate errors. You can search for '04' to see where this error is generated, inside the `getBrew()` function.

https://github.com/naturalcrit/homebrewery/blob/52dcc3b53cbd197b92ebe0f9d314f3d9d08d9eba/server/homebrew.api.js#L79-L84

The error gets `throw`n, which eventually lands in `app.js` which will `console.log` it and try to send an error page. 

https://github.com/naturalcrit/homebrewery/blob/52dcc3b53cbd197b92ebe0f9d314f3d9d08d9eba/server/app.js#L471-L492

Ideally, that bit of code should be able to differentiate between sending an error page and sending an API error (possibly by checking if `req.originalUrl` is a page like `/edit/` or an API path like `/update/`), but we didn't get that far. So instead, when you want to respond to the `editPage` save request, we have separately added `res.send()` functions all around `homebrew.api.js`. I don't think we put one for this error. We just log to the console and then throw the error.

If you add a `res.status.send` around there, it should get received into the `state.error` object, and then in `homebrew/navbar/error-navitem.jsx` you can specify what the error popup will say based on the contents of the error.

The trick is making sure sending the response there still allows the the error page to work when trying to access a brew you aren't an author for.

dbolack-ab:
Would this not be better handled with IP ?

calculuschild:
> Would this not be better handled with IP ?

This PR adds a new way to check the user's login status as a step toward solving issue #2975.

However, *we already have a way to check login status*, which makes this PR obsolete. (I will probably just close this).

Issue #2975 is just that we aren't correctly displaying a warning to the user if they get logged out while they are editing.

PR #3074, opened 2023-10-01
Add basic Homebrewery credits snippet

### LINKED ISSUE #283: Crediting you
I really like this product. It has so freed up my time for manual making. Sooo,  I'd like you to include a preset markdown (if that's what you call it) that I can click that automatically puts your credit in the format you'd like it to be. Is the bottom of the last page okay? Maybe you even have a logo you'd like to see displayed. 

You deserve it.

G-Ambatte:
This PR resolves #283.

This PR adds a snippet and styling to help users quickly and easily add a short credit for Homebrewery.


themes/V3/5ePHB/snippets.js
```diff
@@ -212,6 +212,21 @@ module.exports = [
...
+				name : 'Homebrewery Credit',
+				icon : 'fas fa-dice-d20',
+				gen  : function(){
+					return dedent`
+						{{homebreweryCredits
+						Made With
+						
+						{{homebreweryIcon,gold}}
+						
+						Homebrewery.Naturalcrit.com
```
Gazook89, line 226: Two things:  1) Should this be a link to the homebrewery?  2)  NaturalCrit should have the C capitalized as well.

Gazook89 (commented):
Do you think PHB is the best menu for it?  I think this snippet is more universal, and could likely fit better in the Text Editor menu.

themes/V3/5ePHB/snippets.js
G-Ambatte: Dealing with 2) first, I have just pushed a commit that capitalizes the C.
On 1), this is an interesting question. Should this be a link (and thus https://homebrewery.naturalcrit.com) or just the name of the project ("The Homebrewery"), or both ([The Homebrewery](https://homebrewery.naturalcrit.com))??? As previously mentioned, this is currently a duplicate of the bottom of the back cover, where it exists as the text of the link, but not actually a link.

G-Ambatte:
> Do you think PHB is the best menu for it? I think this snippet is more universal, and could likely fit better in the Text Editor menu.

I thought that the PHB was the best place for right now, given that it is used as a base for other themes. However, arguably the Blank theme might be even better, given that no matter what theme is, the Homebrewery editor is being used and thus the credit snippet might be desired.
However I figured that getting a functional version out in a PR was the most useful thing - it can be transplanted to a different theme if required.

themes/V3/5ePHB/snippets.js
Gazook89: Actually, having the URL (homebrewery.NaturalCrit.com) is helpful for print documents.  But it should also be a link.

Gazook89:
> given that it is used as a base for other themes. 

It should go in the base theme, since it should be available in *all* themes.  Different themes can style it differently, but the overall structure should likely remain the same.

themes/V3/5ePHB/snippets.js
G-Ambatte: I've updated it now, to this:

image

The website address is a link, it responds to mouse over with an underline:

image

G-Ambatte:
I've moved it over to the Blank theme now.

5e-Cleric:
I love the colors, but contrast-wise, not a good fit imo:
image

Can't add a shadow with a mask-image in there. if there was an easier way of adding an svg without all the html, that would be better.

Additionally, i thought the idea for the logo was more in the path of a Cauldron? i get that we don't want to spend an awful amount of time discussing branding, but we should be consistent, if we decide to stick to the naturalcrit logo, that is fine, but we should consider linking it to naturalcrit.com.

Gazook89 (changes requested):
It might be silly to suggest, but it might be good to consider leaving the heavy styling change to another PR, and this PR just be about getting a simple long-awaited snippet out.  

Changes to any brand image (the one image we use that could qualify as such) is likely a bigger longer discussion, and this little snippet PR shouldn't need to shoulder that. 

Here's kind of my desired options, most desirable first:
1. remove the gold styling, including the masking and background color
2. remove the gold styling, but continue to use masking/background color so that the gold styling can be added later after further consideration
3. keep the styling, but only in the CSS.....adjust the snippet to not include `,gold`, leaving it as a 'secret' option that users can use if they want.  Perhaps with codemirror widgets, the `gold` class option will be made apparent as an option.

5e-Cleric:
Those are all valid points, and I agree, pushing this forward is more important than specifics.

themes/V3/Blank/snippets.js
```diff
@@ -111,6 +111,22 @@ module.exports = [
...
 			},
+			{
+				name : 'Homebrewery Credit',
+				icon : 'fas fa-dice-d20',
+				gen  : function(){
+					return dedent`
+						{{homebreweryCredits
+						Made With
+						
+						{{homebreweryIcon,gold}}
```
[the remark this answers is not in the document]
G-Ambatte: I always planned to leave the styling in the LESS, but I was of two minds about leaving the `,gold` in the default snippet. I'm not opposed to taking it out of the snippet text.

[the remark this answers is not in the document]
5e-Cleric: I agree with Gazook, we might want to offer color alternatives in the future, but for now, i would keep that in the credits snippet itself.

calculuschild (changes requested):
Looks good. My only comment is along with the gold coloring; we don't use gold anywhere else in the our branding and its hard to see against the parchment texture. I would just make it default black with maybe "red" as the extra class instead of "gold", since that matches the only "special" color we have used.

5e-Cleric:
Gazook made a comment regarding that, we don't really want to get into Branding in this issue, and rather focus on getting the snippet out, so let's just make it black, and get on with it.

G-Ambatte:
- [x] Icon now defaults to black
- [x] Red added as additional available class to change color

calculuschild (approved):

calculuschild:
Cool! I'll merge this now. Thanks @G-Ambatte !

💯 🎉 ⭐

PR #3079, opened 2023-10-03
fix indent in monster blocks

### LINKED ISSUE #527: Headers After Lists Are Improperly Indented
Any header characterized by`> ***text***` following a bulleted list is improperly indented one tab too far in.

### Additional Details

**Share Link** : http://homebrewery.naturalcrit.com/share/B1W4EvcOp-

See any class features that come after a list. For example, "Natural Explorer", "Two-Weapon Fighting", "Primeval Awareness", and "Horde Breaker" all exhibit this bug.

5e-Cleric:
Should fix #527


Gazook89 (approved):
I think commit does fix #527, with a nice single line.  Only notes I could add, and certainly not needed, is to update the comment on that css declaration (and I don't even know exactly what I would write that wasn't too word).

And as noted in that issue, i offered a minor side issue related to notes:  I think the bullet points are too large compared ot the source material. The regular list bullet could be changed to use the unicode "bullet operator", `&#8729;`.  Of course there is a good chance that character doesn't exist in the fonts.

The screenshot from HB is overlaid above a screenshot of a 5e book (tashas).  You can see the HB bullet point is very large.

image

This might just another PR thing, since the proper way is likely to fix the font file itself (which would be easy, to scale down a circle to the right size).

calculuschild (changes requested):
This has an issue where multi-paragraph entries are now not indented properly:

MM:
image

HB now:
image


This PR:
image

5e-Cleric (commented):
fixed in the last commit

calculuschild:
Thanks @5e-Cleric ! 💯 ⭐ 🎉

PR #3089, opened 2023-10-09
\page as Marked token

G-Ambatte:
This PR is to demonstrate the use of \page as a Marked token, allowing us to render the entire document in a single `Marked.parse` call, rather than breaking up the document into individual pages as a preprocessing step.

This should immediately create unique heading IDs throughout the document (or at least, the ones created by Marked will be unique), and allow other future enhancements to affect the entire document, rather than just one page.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -33,20 +33,24 @@ const BrewRenderer = createClass({
 	},
 	getInitialState : function() {
 		let pages;
+		let pageCount;
 		if(this.props.renderer == 'legacy') {
 			pages = this.props.text.split('\\page');
+			pageCount = pages.length;
 		} else {
 			// pages = this.props.text.split(/^\\page$/gm);
 			pages = [this.props.text];
+			pageCount = this.props.text.split(/^\\page$/gm).length;
 		}
 
 		return {
-			viewablePageNumber : 0,
+			viewablePageNumber : 1,
 			height             : 0,
 			isMounted          : false,
 
 			pages          : pages,
-			usePPR         : pages.length >= PPR_THRESHOLD,
+			pageCount      : pageCount,
+			usePPR         : pageCount >= PPR_THRESHOLD,
 			visibility     : 'hidden',
 			initialContent : `<!DOCTYPE html><html><head>
 												<link href="//use.fontawesome.com/releases/v5.15.1/css/all.css" rel="stylesheet" />
@@ -66,14 +70,18 @@ const BrewRenderer = createClass({
 	componentDidUpdate : function(prevProps) {
 		if(prevProps.text !== this.props.text) {
 			let pages;
+			let pageCount;
 			if(this.props.renderer == 'legacy') {
 				pages = this.props.text.split('\\page');
+				pageCount = pages.length;
 			} else {
 				pages = [this.props.text];
+				pageCount = this.props.text.split(/^\\page$/gm).length;
 			}
 			this.setState({
-				pages  : pages,
-				usePPR : pages.length >= PPR_THRESHOLD
+				pages     : pages,
+				pageCount : pageCount,
+				usePPR    : pageCount >= PPR_THRESHOLD
 			});
 		}
 	},
@@ -85,9 +93,21 @@ const BrewRenderer = createClass({
 	},
 
 	handleScroll : function(e){
-		const target = e.target;
+		let currentPage = 1;
+		if(window) {
+			const pageCollection = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('page');
+			const brewRendererHeight = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0).getBoundingClientRect().height;
+
+			for (const page of pageCollection) {
+				if(page.getBoundingClientRect().bottom > (brewRendererHeight / 2)) {
+					currentPage = parseInt(page.id.slice(1)) || 1;
+					break;
+				}
+			}
+		}
+
 		this.setState((prevState)=>({
-			viewablePageNumber : Math.floor(target.scrollTop / target.scrollHeight * prevState.pages.length)
+			viewablePageNumber : currentPage
 		}));
 	},
 
@@ -121,7 +141,7 @@ const BrewRenderer = createClass({
 				{this.props.renderer}
 			</div>
 			<div>
-				{this.state.viewablePageNumber + 1} / {this.state.pages.length}
+				{this.state.viewablePageNumber} / {this.state.pageCount}
 			</div>
 		</div>;
 	},
@@ -154,7 +174,7 @@ const BrewRenderer = createClass({
 		else {
 			cleanPageText += `\n\n&nbsp;\n\\column\n&nbsp;`; //Artificial column break at page end to emulate column-fill:auto (until `wide` is used, when column-fill:balance will reappear)
 			return (
-				<div className='pages' ref='pages' lang={`${this.props.lang || 'en'}`} dangerouslySetInnerHTML={{ __html: Markdown.render(cleanPageText || ' ') }} />
+				<div className='pages' ref='pages' lang={`${this.props.lang || 'en'}`} dangerouslySetInnerHTML={{ __html: Markdown.render(cleanPageText || ' ', this.state.usePPR ? this.state.viewablePageNumber : -1, 2) }} />
 			);
 		}
 	},
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -36,7 +36,8 @@ const BrewRenderer = createClass({
 		if(this.props.renderer == 'legacy') {
 			pages = this.props.text.split('\\page');
 		} else {
-			pages = this.props.text.split(/^\\page$/gm);
+			// pages = this.props.text.split(/^\\page$/gm);
+			pages = [this.props.text];
 		}
 
 		return {
@@ -68,7 +69,7 @@ const BrewRenderer = createClass({
 			if(this.props.renderer == 'legacy') {
 				pages = this.props.text.split('\\page');
 			} else {
-				pages = this.props.text.split(/^\\page$/gm);
+				pages = [this.props.text];
 			}
 			this.setState({
 				pages  : pages,
@@ -149,13 +150,11 @@ const BrewRenderer = createClass({
 	renderPage : function(pageText, index){
 		let cleanPageText = this.sanitizeScriptTags(pageText);
 		if(this.props.renderer == 'legacy')
-			return <div className='phb page' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(cleanPageText) }} key={index} />;
+			return <div className='pages' ref='pages' lang={`${this.props.lang || 'en'}`}><div className='phb page' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(cleanPageText) }} key={index} /></div>;
 		else {
 			cleanPageText += `\n\n&nbsp;\n\\column\n&nbsp;`; //Artificial column break at page end to emulate column-fill:auto (until `wide` is used, when column-fill:balance will reappear)
 			return (
-				<div className='page' id={`p${index + 1}`} key={index} >
-					<div className='columnWrapper' dangerouslySetInnerHTML={{ __html: Markdown.render(cleanPageText) }} />
-				</div>
+				<div className='pages' ref='pages' lang={`${this.props.lang || 'en'}`} dangerouslySetInnerHTML={{ __html: Markdown.render(cleanPageText || ' ') }} />
 			);
 		}
 	},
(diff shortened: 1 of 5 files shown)


shared/naturalcrit/markdown.js
```diff
@@ -238,7 +268,38 @@ const definitionLists = {
...
-Marked.use({ extensions: [mustacheSpans, mustacheDivs, mustacheInjectInline, definitionLists] });
+let pageBlockTopLevel = true;
+let pageBlockNumber = 0;
+const pageBlocks = {
+	name  : 'pageBlock',
+	level : 'block',
+	start(src) { return pageBlockTopLevel; },
+	tokenizer(src, tokens) {
+		if(!pageBlockTopLevel) return false;
+		const pageArray = src.split(/^\\page$/gm);
```
calculuschild, line 279: Since each token just needs to get the next page between `\page`, we can put a limit on how much splitting occurs to save time:

```suggestion
		const pageArray = src.split(/^\\page$/gm,1);
```

calculuschild: May need to be 2 depending how `split` works.

shared/naturalcrit/markdown.js
```diff
@@ -238,7 +268,38 @@ const definitionLists = {
 	}
 };
 
-Marked.use({ extensions: [mustacheSpans, mustacheDivs, mustacheInjectInline, definitionLists] });
+let pageBlockTopLevel = true;
+let pageBlockNumber = 0;
+const pageBlocks = {
+	name  : 'pageBlock',
+	level : 'block',
+	start(src) { return pageBlockTopLevel; },
```
calculuschild, line 276: `start:` isn't supposed to return a boolean. It returns the index of where the token starts.

Since Marked scans for tokens in the order the extensions are loaded, it's possible for it to see one token and jump ahead in the string, leaving later tokens unprocessed. `start` places a marker at any likely places so it knows to stop and check before jumping ahead.

For example, if we had:

```
{{curlystuff

DL :: STUFF

}}
```

And we had loaded the DL extension first, Marked would jump past the curlies and not process them because it was looking for DL first.

With `start` filled in for both (using a simplified regex that just looks for `{{`), it knows to make sure to stop at the curlies and check the other extensions before moving on.

shared/naturalcrit/markdown.js
```diff
@@ -238,7 +268,38 @@ const definitionLists = {
...
+
+		const token = {
+			type       : 'pageBlock',
+			raw        : `${pageArray[0]}\\page`,
+			text       : pageArray[0],
+			pageNumber : pageBlockNumber,
+			tokens     : []
+		};
+
+		pageBlockTopLevel = false;
```
calculuschild, line 291: I'm not sure what `pageBlockTopLevel` is contributing to this extension?

shared/naturalcrit/markdown.js
calculuschild: In our case, page blocks should always be assumed to start at the very beginning of the current string, so we should have `start: 0` or just no `start` at all.

Having `start: true` gets converted to 1, so it's adding a marker 1 char after the start. Extensions are always checked at the beginning of the string too, so the first time the page extension comes around it already eats up the whole page, and that marker will get discarded anyway.

shared/naturalcrit/markdown.js
```diff
@@ -238,7 +268,38 @@ const definitionLists = {
...
+	start(src) { return pageBlockTopLevel; },
+	tokenizer(src, tokens) {
+		if(!pageBlockTopLevel) return false;
+		const pageArray = src.split(/^\\page$/gm);
+		if(this.lexer.tokens.length == 0 && this.lexer.state.top) pageBlockNumber = 0;
+		pageBlockNumber++;
+
+		const token = {
+			type       : 'pageBlock',
+			raw        : `${pageArray[0]}\\page`,
```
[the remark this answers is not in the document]
calculuschild: This can probably be checked by looking if the first split is an empty string "".  If so, we know the current string starts with `\page` (in which case we grab the second split).

shared/naturalcrit/markdown.js
G-Ambatte: Given that any string - even one that does not contain the `\page` token - should be wrapped in a pageBlock, the processing must then trigger on any given string, including strings that contain other block tokens. But without limitation, this will result in an infinite sequence of pageBlocks.
Given the limitation that no pageBlock can contain another pageBlock, I created the `pageBlockTopLevel` boolean to determine if the current iteration of the tokenizer is running inside a pageBlock token and immediately cancel creating further pageBlock tokens. Perhaps a better name would have been `hasPageBlockParent`, with a default value to `false`.

It's possible that changing the `start(src)` function can be manipulated to produce a similar result.

calculuschild: Ok I see. I didn't see the check at the beginning of the Tokenizer.

Yeah I don't think the Start function is doing *anything* now. You can probably just remove it and it should work the same.

shared/naturalcrit/markdown.js
G-Ambatte: Applied.

shared/naturalcrit/markdown.js
G-Ambatte: As per other comments, `start(src)` has been removed as it was no longer doing anything.

5e-Cleric:
as  noted today on gitter, we might want to transform link references into actual links when rendering, otherwise the reference is not found and the links become invalid

G-Ambatte:
> as  noted today on gitter, we might want to transform link references into actual links when rendering, otherwise the reference is not found and the links become invalid

This might not be an issue, I believe that ref link definition values are generated at tokenization, if so being modified prior to rendering will not affect the ref link calls. They might not even leave a token to be modified.

shared/naturalcrit/markdown.js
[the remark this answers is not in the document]
G-Ambatte: The latest commit should have improved this.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -84,14 +93,27 @@ const BrewRenderer = createClass({
...
+			const pageCollection = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('page');
+			const brewRendererHeight = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0).getBoundingClientRect().height;
+
+			for (const page of pageCollection) {
+				if(page.getBoundingClientRect().bottom > (brewRendererHeight / 2)) {
+					currentPage = parseInt(page.id.slice(1)) || 1;
+					break;
+				}
+			}
+		}
```
calculuschild, line 107: This seems a lot more complicated to do essentially the same thing as before. What is the difference here? Is this just to handle different-sized pages?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -149,13 +171,12 @@ const BrewRenderer = createClass({
 	renderPage : function(pageText, index){
 		let cleanPageText = this.sanitizeScriptTags(pageText);
 		if(this.props.renderer == 'legacy')
-			return <div className='phb page' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(cleanPageText) }} key={index} />;
+			return <div className='pages' ref='pages' lang={`${this.props.lang || 'en'}`}><div className='phb page' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(cleanPageText) }} key={index} /></div>;
```
calculuschild, line 174: This now renders a separate `pages` div around each individual page, rather than one large `pages` object that contains all pages. Can you help me understand the reasoning for this?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -149,13 +171,12 @@ const BrewRenderer = createClass({
...
 		else {
-			cleanPageText += `\n\n&nbsp;\n\\column\n&nbsp;`; //Artificial column break at page end to emulate column-fill:auto (until `wide` is used, when column-fill:balance will reappear)
+			cleanPageText = cleanPageText.split(/^\\page$/gm).join(`\n\n&nbsp;\n\\column\n&nbsp;\n\\page\n`); //Artificial column break at page end to emulate column-fill:auto (until `wide` is used, when column-fill:balance will reappear)
+
 			return (
-				<div className='page' id={`p${index + 1}`} key={index} >
-					<div className='columnWrapper' dangerouslySetInnerHTML={{ __html: Markdown.render(cleanPageText) }} />
-				</div>
+				<div className='pages' ref='pages' lang={`${this.props.lang || 'en'}`} dangerouslySetInnerHTML={{ __html: Markdown.render(cleanPageText || ' ', this.state.usePPR ? this.state.viewablePageNumber : -1, 3) }} />
```
calculuschild, line 179: This logic seems to be duplicating the steps that `renderPages()` already does, of deciding which pages to render based on PPR being active.

server/homebrew.api.js
```diff
@@ -116,7 +116,10 @@ const api = {
 		return text;
 	},
 	getGoodBrewTitle : (text)=>{
-		const tokens = Markdown.marked.lexer(text);
+		let tokens = Markdown.marked.lexer(text);
+		while (tokens[0].type == 'pageBlock'){
```
calculuschild, line 120: Why not an `if` here? Is it possible to nest pages that we need to loop through more than once?

client/homebrew/brewRenderer/brewRenderer.jsx
G-Ambatte: That is a mistake; I was so focused on V3 that I missed that change to Legacy. I'll dig into it and correct.

client/homebrew/brewRenderer/brewRenderer.jsx
G-Ambatte: This is a copy/paste of the page detection from the Brew movement function; I needed a more robust system once I realized that the new rendering method broke the page info panel.
I can absolutely swap out to `target` to simplify the logic, or see if this can be reworked to be simpler.

client/homebrew/brewRenderer/brewRenderer.jsx
G-Ambatte: Effectively, yes - as V3 now renders everything as a single page, the V3 renderer needs to be passed the PPR information so it can prune the non-rendered pages.
So Legacy performs the PPR in `brewRenderer.jsx` while V3 performs PPR in `markdown.js`.

shared/naturalcrit/markdown.js
```diff
@@ -7,14 +7,44 @@ const { gfmHeadingId: MarkedGFMHeadingId } = require('marked-gfm-heading-id');
...
+
+			this.lexer.blockTokens(html, item.tokens);
+
+			return item;
+		}
+		return false;
+	},
+	renderer(token) {
+		return `${token.openTag} ${this.parser.parse(token.tokens)} </div>`;
 	}
```
calculuschild, line 47: This does not work. Try testing this text. Both should have functional markdown:

```
<div>

Will this *render?*

</div>

<div>
Will this *render?*
</div>
```

The problem is most of the logic in the original code here is applied to the Renderer, but this change has put it inside the Tokenizer, so the tokens being generated aren't correct.

I agree this bit of code should be made into correct "Markdown Extension" format. However, I think it's outside the scope of this PR and will need more work that complicates the `\page` token stuff. Can you remove it from here and put into a separate PR we can better test in isolation?

server/homebrew.api.js
G-Ambatte: It _shouldn't_ be possible for the first token to be anything other than `pageBlock`. I seem to recall using `if` and having it not work, but I don't recall any specific details. I'll switch to `if` and report if problems arise.

shared/naturalcrit/markdown.js
G-Ambatte: Oh, I remember this now! Because this was using a separate instance of the renderer, it was generating a `pageBlock` to contain the HTML div; that's why I ended up including this change here.

G-Ambatte: Something is definitely screwy, though:

image

calculuschild: > this was using a separate instance of the renderer

I don't see where a separate instance is being used? We still have `Marked.use({ renderer: renderer, mangle: false });` where the original HTML handling code is being applied to the same instance.

G-Ambatte: I think we cut this out, make it it's own PR, get it merged
Then come back for the larger `\page` token PR

G-Ambatte: > I don't see where a separate instance is being used?

I think it was the `Marked.parse()` call; although I could be wrong about that.

calculuschild:
Do you happen to have a nice long, heavy V3 document to test the performance of this?

shared/naturalcrit/markdown.js
G-Ambatte: With this change reverted, I'm not seeing the behaviour that I was seeing earlier - I probably blamed this particular chunk of code for something unrelated.

calculuschild:
Going to close this in favor of #3156

However thank you for your work here, @G-Ambatte . I think this revealed the way forward and lessons learned will help get variables working.

PR #3099, opened 2023-10-14
Add Share link to Unauthorized Edit Access error message

### LINKED ISSUE #3086: Add Share link to "Not an Author" error page
### Your idea:

It is not uncommon for users to post their Edit page details when they intend to post their Share page information instead - this happens often when newer users are requesting support, and don't necessarily understand the difference between the two. Currently, we block Edit links on the subreddit, harking back to the days when anyone with the Edit link could access and change the brew content, but now that Authors have been enforced, it is no longer an issue.

What I suggest is that we maintain the existing error message, but incorporate a method in which the Share link for a brew can be accessed or derived from that page, primarily for support reasons.
Alternatively, we could develop some other method - perhaps using the oft overlooked Admin pages, only accessible to elevated users - to retrieve a brew's Share ID from the Edit ID to assist in troubleshooting.

G-Ambatte:
This PR resolves #3086.

When a new user asks for help in the public forums, we often ask for a link to their document. While we specify that we need the `/share/` link, quite often the inexperienced user will instead post the `/edit/` links. This has been a problem since... well, forever.

This PR adds a link to the brew's `/share/` page to the error message that is received when a user attempts to access an brew that they are not an author of.

### DIFF
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -22,18 +22,18 @@ const errorIndex = (props)=>{
 			## We can't find this brew in Google Drive!
 			
 			This file was saved on Google Drive, but this link doesn't work anymore.
-			${ props.brew.authors?.length > 0
-				? `Note that this brew belongs to the Homebrewery account **${ props.brew.authors[0] }**,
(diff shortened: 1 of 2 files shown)


calculuschild:
Looks good. Thanks @G-Ambatte!

PR #3116, opened 2023-10-27
Fix Firefox printing

### LINKED ISSUE #3115: [Firefox] Brew prints alternate blank pages
### Your idea:

When printing from Firefox, every odd page is blank.

After some experimentation, it appears that this is caused by the `page-break-before: always` and `page-break-after: always` properties being set on each `page` element. Removing these properties had no noticeable effect on printing in Chrome.

G-Ambatte:
This PR resolves #3115.

This PR removes the `page-break-before: always` and `page-break-after: always` properties from the `page` elements in both the Blank and 5ePHB theme styling. This corrects the alternate blank page issue seen when printing from Firefox (and presumably other browsers) while having no determinable effect on printing from Chrome.


calculuschild:
Ok. I agree I don't see any negative effect to doing this. Merging!

Thanks @G-Ambatte 💯 🎉 ⭐

Issue #2510, opened 2022-11-17
Improve error popup for User Google Drive storage out of space


Brink81:
### Error Code

```shell
Error: Forbidden


Error: Forbidden
    at Request.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:75356:19)
    at Emitter.emit (https://homebrewery.naturalcrit.com/homebrew/bundle.js:1966:20)
    at xhr.onreadystatechange (https://homebrewery.naturalcrit.com/homebrew/bundle.js:75665:10)
{
  "status": 403,
  "method": "PUT",
  "url": "/api/update/g1LAcYR05zRg"
}
```
```


### Your description of what happened:

"Oops" error while Saving or Auto-saving on Google Chrome, Mozilla Firefox, and even Microsoft Edge.

The code "<div class="navDropdownContainer save-menu"><div class="navItem save saved">auto-saved.</div>"
seems to be malfunctioning.

**Comment by calculuschild** [comment]
According to our logs, it was trying to save to your Google Drive, but your drive is full.

**Comment by calculuschild** [comment]
This is an opportunity to improve the error messaging for this particular google response. Example response from Google here (personal info removed):

```

 GaxiosError: The user's Drive storage quota has been exceeded.
     at Gaxios._request (/app/node_modules/gaxios/build/src/gaxios.js:130:23)
     at runMicrotasks (<anonymous>)
     at processTicksAndRejections (node:internal/process/task_queues:96:5)
     at async JWT.requestAsync (/app/node_modules/google-auth-library/build/src/auth/oauth2client.js:368:18)
     at async Object.updateGoogleBrew (/app/server/googleActions.js:137:3)
     at async updateBrew (/app/server/homebrew.api.js:215:19) {
   response: {
     config: {
       url: 'https://www.googleapis.com/upload/drive/v3/files/xxxxxxxxxxxxxxxxxxxx?uploadType=multipart',
       method: 'PATCH',
       userAgentDirectives: [Array],
       paramsSerializer: [Function (anonymous)],
       data: [PassThrough],
       headers: [Object],
       params: [Object],
       validateStatus: [Function (anonymous)],
       retry: true,
       body: [PassThrough],
       responseType: 'json',
       retryConfig: [Object]
     },
     data: { error: [Object] },
     headers: {
[...]

PR #3118, opened 2023-10-28
Identify API errors and send error response

### LINKED ISSUE #2975: Saving an owned brew while not logged in causes perpetual spinning save icon

G-Ambatte:
This PR does not fully resolve #2975, #2510, but does unblock them.

This PR adds an identification method to the error handling middleware function and identifies errors arising from API calls (via `req.originalUrl.startsWith('/api/')`), which causes an error response to be sent (`res.status().send()`) rather than `res.send(page)`.


server/app.js
```diff
@@ -472,9 +472,18 @@ const getPureError = (error)=>{
 };
 
 app.use(async (err, req, res, next)=>{
-	const status = err.status || err.code || 500;
+	err.originalUrl = req.originalUrl;
 	console.error(err);
 
+	if(err.originalUrl?.startsWith('/api/')) {
+		// console.log('API error');
+		res.status(err?.status || err?.response?.status || 500).send(err.message || err);
```
Gazook89, line 480: Do you need the `?` operator on `err?.status` and `err?.response`?  Presumably there *is* an error if you've entered this `if...then` statement.  

If anything, they could be replaced by throwing the `?` in the `if(err?.originalURL....` condition?

Gazook89 (approved):
The server side stuff is usually beyond me, though this looks a simple PR.  Looks good to me (though I left one comment).

server/app.js
G-Ambatte: I think that entire line came directly from `homebrew.api.js`, you are correct - there is no way that we can enter that branch of the `if` with a non-existent `err`.

calculuschild:
Ok, this looks good! It does unstick #2975 by allowing the failed request to actually return an error instead of nothing, but to fully resolve the issues mentioned above, they need to have a custom error message displayed in `error-navitem.jsx` instead of the generic "oops! problem saving".

Thanks @G-Ambatte! 💯 ⭐ 🪄

Issue #380, opened 2017-04-13
Wrapping text around an image


badgermaniac:
First, thanks for the awesome tool. 

I have been inserting images into my docs and positioning them in various ways. However, I haven't figured out how to wrap text around an image.  

I have cut out my images so they have transparent backgrounds, leaving the border irregular as oppose to rectangular, but now I want the text to "fill in the gaps" (see pg. 125 in PHB). 

Any code help would be appreciated. Thanks.

### Additional Details

**Share Link** :

or

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

PASTE BREW CODE HERE

</pre></code></details>

BaronMurph:
Haven't actually figured out how to add multiple spaces (or, like, a horizontal in-line spacer) myself, so I'll admit to leaving a comment here in part for selfish reasons in case anyone does. 

What I _can_ tell you, however, is that you can add line breaks without spaces in front (along the lines of https://i0.wp.com/www.critical-hits.com/wp-content/uploads/2014/08/Bulette.jpg), just by adding a double space at the end of a line.
`So just like this with two spaces over there -->  `
`And then this show up without the customary spaces in front. Note that this is on the next line, not two lines down - I only pressed the Enter key once, so to speak, not twice.`

badgermaniac:
Thanks. This helps for aligning images on one side of text, so it is s start.

BaronMurph:
Hell yeah, found it.
&nbsp; adds a single space, and you can just keep adding 'em. It'll look ugly on code ('cause you need to keep adding it, so two spaces would be &nbsp;&nbsp; and three &nbsp;&nbsp;&nbsp;) but it works.
>&nbsp;&nbsp;&nbsp; like this

Biotronic:
The solution you really want for this is CSS's [shape-outside](https://css-tricks.com/almanac/properties/s/shape-outside/). Here's an example of how to do it:

http://homebrewery.naturalcrit.com/share/Hkgyq2eJ-

Biotronic:
This report could be turned into an enhancement with a bit of javascript and a CSS class:

```
$(".wrap-image").each(function() {
    if ($(this).attr("src")) {
        $(this).css("shape-outside", "url("+$(this).attr("src")+")"); 
    }
});
```

This will set the 'shape image' of any image to its image, and thus automatically do what I did in the previous post.

calculuschild:
[...]

PR #3126, opened 2023-11-05
Create Solution for simplified shaped image wrap.

dbolack-ab:
This adds a pattern for creating macro substitutions in mustacheInjectInline to create complex style properties that depend on information in the mustache-impacted token.

Creates $twl and $twr macros for images to use shape-outside.

$twl is changed to shape-outside: url(image_url), float: left
$twr is changed to shape-outside: url(image_url), float: right

$twl-_numberunit_ is changed to shape-outside: url(image_url), float: left;shape-margin:_numberunit_
$twr-_numberunit_ is changed to shape-outside: url(image_url), float: right;shape-margin:_numberunit_

where _numberunit_ is a number and CSS valid unit.

$twl-_number_ is changed to shape-outside: url(image_url), float: left;shape-image-threshold:_number_
$twr-_number_ is changed to shape-outside: url(image_url), float: right;shape-image-threshold:_number_

where _number_ is a floating point number.

In all cases, image_url is the href value from lastToken.

### Testing
 - [x] Verify adding $twl floats target left and uses the default shape-margin distance
 - [x] Verify adding $twr floats target right and uses the default shape-margin distance
 - [x] Verify adding $twl-10px floats target left and sets the shape-margin distance to 10px
 - [x] Verify adding $twr-10px floats target right and sets the shape-margin distance to 10px
 - [x] Verify adding $twl-0.5 floats target left and sets the shape-image-threshold to 0.5
 - [x] Verify adding $twr-0.5 floats target right and sets the shape-image-threshold to 0.5

Note that the actual numerical values in the tests are not important so long as they are enough to show a visible change.

The FAQ is also updated with how to use this macro or perform the tasks manually.

Solves #380


5e-Cleric:
Everything seems to be working fine, i like the macro, we could update the snippet list with a floating image using this macro. So that users can see it in use. Maybe update the cat image with a background-less version of it.

Gazook89:
I also generally like this.  I think that if snippets are created, or a FAQ given, it should include `position:relative;` and then a left/right adjustment as well.  And, it should be included in the Homepage guide rather than FAQ (or both).

```
![homebrew mug](http://i.imgur.com/hMna6G0.png) {$swl,width:280px,left:-100px,position:relative}
```
image

I also think that maybe `swr` and `swl` are not descriptive enough-- I'm assuming it stands for `shape wrap left/right`.  I think it could be `twl` `twr`  (text wrap left/right).  Or even full words.  Or `sol` `sor`.   Shape Wrap just seems like a combination of the two that doesn't really have any meaning to anyone who doesn't have experience with `shape-outside`.

dbolack-ab:
> shape-margin: 5px

The text's I found matched URL up with shape-image-threshold which is what I set up.

Do we have a preference if both are an option?

dbolack-ab:
This has been broken by recent  merges and I find myself wondering is this a hammer is search of a nail when looking at fixing it.

5e-Cleric:
Is this feature working right now?

dbolack-ab:
> Is this feature working right now?

Was last I tested it.

dbolack-ab:
I found a basic fail in the macro symbol choices. Reworking.

5e-Cleric (approved):
Seems to be working alright, and the changes are minimal! perhaps we could improve the syntax to be actually user friendly, but for now it is good!

5e-Cleric:
> I found a basic fail in the macro symbol choices. Reworking.

Apparently i was also testing wrong, so

dbolack-ab:
> Seems to be working alright, and the changes are minimal! perhaps we could improve the syntax to be actually user friendly, but for now it is good!

Open to suggestions. I couldn't come up with a shortened name that was understandable so I went with initials.

5e-Cleric:
> > Seems to be working alright, and the changes are minimal! perhaps we could improve the syntax to be actually user friendly, but for now it is good!
> 
> Open to suggestions. I couldn't come up with a shortened name that was understandable so I went with initials.

instead of `$twl`, how about `$wrapL` and `$wrapR`, it is just 2 more letters.

calculuschild:
I have concerns introducing an entire separate form of syntax just for this feature.

We have Markdown, we have CSS, and now this Macro thing seems out of place.

Is there no way we could do this with standard CSS?

Gazook89:
My understanding is that this *is* just regular CSS, but a shortcut--- it's just outputting additional inline styles.  So yeah, it *could* be done with regular CSS, but inline it would require using the same image url twice plus some additional properties.    It'd just be this:

```
![homebrew mug](http://i.imgur.com/hMna6G0.png){float:left;shape-outside:url(http://i.imgur.com/hMna6G0.png);shape-margin:5px;width:280px;left:-100px;position:relative;}
```
vs

```
![homebrew mug](http://i.imgur.com/hMna6G0.png){$twl-5px,width:280px,left:-100px,position:relative}
```
for the same result:
image

Maybe, if we really don't want a permanent additional syntax to worry about in brews going forward, this feature could be used just for initially *entering* a text wrap....but then expands into the full CSS which is actually saved to the brew text.  When exactly the expanding happens would need to be thought about.  Changing the image would require it in two places which isn't a *big* deal.  And seeing the full CSS allows users to see what is actually happening with the CSS.  Downside would be just the extra non-content inline styling in the editor.

calculuschild:
My main concern isn't so much making a shortcut, but rather introducing an unfamiliar `$macro` syntax. I would rather it just look like a normal CSS class "wrapLeft/wrapRight" or something, even if we have to do some extra JS behind the scenes.

ericscheid:
It's a dang shame that CSS won't do `shape-outside:url(attr(src))` or similar. 

Or a keyword of `self`.

ericscheid:
Hmm .. CSS does not support `shape-outside:self` .. but could we patch our curly-inject code to make the transformation as needed?

Like, this markdown:

````
![homebrew mug](http://i.imgur.com/hMna6G0.png){shape-outside:self; ...}
````

produces this rendering:

````
![homebrew mug](http://i.imgur.com/hMna6G0.png){shape-outside:url(http://i.imgur.com/hMna6G0.png); ...}

````

dbolack-ab:
Based on conversation, pulling this pr so the preferred alternates may be implemented.

Issue #2340, opened 2022-09-05
Modify Markdown `dl` syntax to support multiple subsequent `dt` or `dd` elements


develleoper:
## Description
The current `<dl>` syntax (i.e. `dt :: dd`) only supports a 1:1 ratio of terms to descriptions.

### Reference
- Original Discussion at [Reddit](https://www.reddit.com/r/homebrewery/comments/tdre2d/feature_request_support_dl_tag_syntax_extension/)
- The following examples follow the syntax set forth in [Kramdown](https://kramdown.gettalong.org/syntax.html#definition-lists), [Pandoc](https://pandoc.org/MANUAL.html#definition-lists), [MultiMarkdown](http://fletcher.github.io/MultiMarkdown-4/definitionlists), and [Markdown Extra](https://michelf.ca/projects/php-markdown/extra/#def-list).

#### Basic Case:
```markdown
First term
: First description
: Second description
```
```html
<dl>
  <dt>First term</dt>
  <dd>First description</dd>
  <dd>Second description</dd>
</dl>
```

develleoper:
#### Advanced Cases:
```markdown
Preceding paragraph

First term
Second term
: Description, first paragraph
  Description, second paragraph
  - Description, unordered list item
  > Description, block quote

Following paragraph
```
```html
<p>Preceding paragraph</p>
<dl>
  <dt>First term</dt>
  <dt>Second Term</dt>
  <dd>
    <p>Description, first paragraph</p>
    <p>Description, second paragraph</p>
    <ul>
      <li>Description, unordered list item</li>
    </ul>
    <blockquote>Description, block quote</blockquote>
  </dd>
</dl>
<p>Following paragraph</p>
```

calculuschild:
What would you see this be displayed as when entered into the Homebrewery? Any particular style/format from the D&D books that a multi-dd would be used for?

ericscheid:
Two paragraphs in a `<dd>`, from #181 

<img width="426" alt="two paragraphs in a dd" src="https://user-images.githubusercontent.com/2605467/188527307-6bb4ca9d-5263-48ca-8668-426f6992af91.png">

develleoper:
@ericscheid I believe the ticket and associated pull request actually use bold tags and paragraphs to accomplish that? Is this incorrect?

[...]

PR #3129, opened 2023-11-07
Add Multi-line Dictionary Definition (<dd>) rows.

dbolack-ab:
Expands the existing syntax to allow/expect the option of multiple definitions by adding any number of ``` :: Definition``` to an a DT/DD set.

Solves #2340.

**Final Syntax and Behavior**

This has been updated to a multi-description ( DD ) mode that can take multi-line input. This mode expects the term on one line, the description(s) on lines after, then a blank line before a new term. Two blank lines terminate the list.

This cannot be mixed with the existing inline style definition lists.

Example:

```
Term 1
:: Definition 1
:: Definition 2
is on multiple lines

Term 2:
:: Definition 1


```
Test case Brew at: https://homebrewery.naturalcrit.com/share/h-sHSQsRyUlz


calculuschild:
Before going further with this, I think we need to have a little more discussion about where this syntax would actually be used. As I mentioned in this comment https://github.com/naturalcrit/homebrewery/issues/2340#issuecomment-1241417784, there was a specific use case we identified for the single-line syntax we currently use. I don't think we have any use case identified for multiple `<dd>` entries yet. How should we render these? What CSS should go along with them? Does it do something that we already do just as well with plain paragraphs? I don't think any of that was settled in #2340.

As another note, this PR unfortunately breaks existing `<dl>` rendering by adding a `\n`

dbolack-ab:
> Before going further with this, I think we need to have a little more discussion about where this syntax would actually be used. As I mentioned in this comment [#2340 (comment)](https://github.com/naturalcrit/homebrewery/issues/2340#issuecomment-1241417784), there was a specific use case we identified for the single-line syntax we currently use. I don't think we have any use case identified for multiple `<dd>` entries yet. How should we render these? What CSS should go along with them? Does it do something that we already do just as well with plain paragraphs? I don't think any of that was settled in #2340.
> 
> As another note, this PR unfortunately breaks existing `<dl>` rendering by adding a `\n`

I'm not spotting this? 

```
                return `<dl>${token.definitions.reduce((html, def)=>{
-                       return `${html}<dt>${this.parser.parseInline(def.dt)}</dt>`
-                                   + `<dd>${this.parser.parseInline(def.dd)}</dd>\n`;
+                       const dds = def.dd.map((s)=>`<dd>${this.parser.parseInline(s)}</dd>`).join('\n');
+                       return `${html}<dt>${this.parser.parseInline(def.dt)}</dt>
+                               ${dds}`;
                }, '')}</dl>`;
```

Do you mean the linebreak between the close of the <dt> and the open of the first <dd> ?

calculuschild:
> Do you mean the linebreak between the close of the <dt> and the open of the first <dd> ?

Yep. Try generating a monster stat block on this PR for example. We format the `<dt>` and `<dd>` to be on one line, but that `\n` breaks all the formatting.

image

dbolack-ab:
> > Do you mean the linebreak between the close of the  and the open of the first  ?
> 
> Yep. Try generating a monster stat block on this PR for example. We format the `<dt>` and `<dd>` to be on one line, but that `\n` breaks all the formatting.
> 
> image

Weird. That is not what I remember to be the correct behavior for a DD/DT pair but it's been a spell. :)

calculuschild:
The following code crashes the page:

```
Test
:: one
```
Can we get this to work? If we are allowing multi-line syntax I would like to be able to write each DD entry its own line.

calculuschild:
Another thought:

Since we display single-DD definitions in one line, would it make sense to always display multi-DD definitions in one DD per line?

I.e

```
**Term** :: definition
```

should continue to look like:

image

but 

```
**Term** :: def1 :: def2 :: def3

or

**Term**
:: def1
:: def2
:: def3
```

Should look like this?

image

instead of this half-inline, half-multiline (current behavior)?

image

ericscheid:
> Another thought:
> 
> Since we display single-DD definitions in one line, would it make sense to always display multi-DD definitions with a separate line per DD?
> 
> Along the lines of Markdown's "prime directive": Input source should look like the output.

Good point. This is mostly a matter for css, as demonstrated here: 

https://homebrewery.naturalcrit.com/share/O6tGj3WujG1A

However, since html structures for DL/DT/DD are not arranged into semantic hierarchies (i.e. with each DD being a child element to the respective DT element), this makes writing CSS selectors to differentiate between the two cases likely impossible (we can't use `:only-child` or similar).

We could however detect the multi-DD situations during rendering and attach a css class to the DD and DT elements.

ericscheid:
> ```
> **Text**
> :: defn-one
> ```

Is that the syntax we're going with? I thought we were signalling that `**text**` was a `<dt>` by the presence of `::` on the line later in that line. I don't mind it, per se, and if detecting the `<term>` is easy enough with that syntax then great.

> If we are allowing multi-line syntax I would like to be able to write each DD entry its own line.

100% agree.

calculuschild:
> Is that the syntax we're going with?

Its currently not, but I think it makes sense to have only one `::` colon pair for each dd. Following that, if we want to be able to write each dd on its own line, it seems the extra `::` on the first line is redundant.

This also would bring us a little closer to the Pandoc style of:

```
Term
: def
: def
: def
```

It would be almost like a sibling syntax to bulleted lists in that sense. One delimiter to mark each element on the list.

Gazook89:
When typing your suggestion above, before the definition is written, you would get a vertical space first, before doing the look-behind, right?

```
Term
:
```
Is that a concern?

ericscheid:
> When typing your suggestion above, before the definition is written, you would get a vertical space first, before doing the look-behind, right?
> ```
> Term
> :
> ```
> 
> Is that a concern?

No.

1. If the edit cursor is at the end of the brew (or end of \page) then there is no visible effect in the preview.

2. If the editing is occuring mid-brew, then as soon as you press return to start a new line then nothing happens (because blank lines don't render as an element), and then when the `:` is typed then that is rendered as `<div class="blank"></div>` which appears as a blank line (and is evident becuase the following content is pushed down by that blank line) .. _but_ that "blank line" is indistinguishable¹ from the blank space that an empty DD would take up. As soon as any text is typed after the `:` then the `<div class="blank"></div>`² is replaced with `<dd>text</dd>` (i.e. the "blank line" gets filled with the definition text).
   1. Modulo any css vagaries, like `height`, `background-color`, `padding`, `margin`.
   2. Technically, there would be two `<div class="blank"></div>` elements, which then collapses back to one `<dd>text</dd>`. That's an ephemeral visual glitch, but no different from seeing other visual glitches (e.g. typing `-` on a new line after text).

shared/naturalcrit/markdown.js
```diff
@@ -237,32 +237,45 @@ const superSubScripts = {
...
 	},
 	renderer(token) {
-		return `<dl>${token.definitions.reduce((html, def)=>{
-			return `${html}<dt>${this.parser.parseInline(def.dt)}</dt>`
-			            + `<dd>${this.parser.parseInline(def.dd)}</dd>\n`;
-		}, '')}</dl>`;
+		let returnVal = `<dl>`;
+		token.definitions.forEach((def)=>{
+			const dds = def.dd.map((s)=>`<dd>${this.parser.parseInline(s)}</dd>`).join('\n');
+			returnVal += `<div><dt>${this.parser.parseInline(def.dt)}</dt>${dds}</div>`;
```
calculuschild, line 276: The `<dt>` should not be inside a `<div>` here. Can you remove that?

shared/naturalcrit/markdown.js
```diff
@@ -237,32 +237,45 @@ const superSubScripts = {
...
 		}
 	},
 	renderer(token) {
-		return `<dl>${token.definitions.reduce((html, def)=>{
-			return `${html}<dt>${this.parser.parseInline(def.dt)}</dt>`
-			            + `<dd>${this.parser.parseInline(def.dd)}</dd>\n`;
-		}, '')}</dl>`;
+		let returnVal = `<dl>`;
+		token.definitions.forEach((def)=>{
+			const dds = def.dd.map((s)=>`<dd>${this.parser.parseInline(s)}</dd>`).join('\n');
```
calculuschild, line 275: Can we change the behavior here slightly so all `<dd>`s are joined with `''` and newline `<dd>`s are also prepended with `'\n'`?

So the output looks like the markdown. i.e. preserve newlines so we can get.

image
vs
image
depending how the list is written.

For example:

```
Term :: def1 :: def2 :: def3
↓
<dl><dt>Term</dt><dd>def1</dd><dd>def2</dd><dd>def3</dd></dl>
```

and

```
Term
:: def1
:: def2
:: def3
↓
<dl><dt>Term</dt>
<dd>def1</dd>
<dd>def2</dd>
<dd>def3</dd></dl>
```

and I guess mixing inline and new lines would be:

```
Term :: def1
:: def2 :: def3
↓
<dl><dt>Term</dt><dd>def1</dd>
<dd>def2</dd><dd>def3</dd></dl>
```

This should result

shared/naturalcrit/markdown.js
```diff
@@ -237,32 +237,45 @@ const superSubScripts = {
...
+				if(currentDefinition?.dt?.length) {
+					allDefinitions.push(currentDefinition);
+					currentDefinition = [];
+				}
+				currentDefinition = {
+					dt : this.lexer.inlineTokens(match[1].trim()),
+					dd : []
+				};
+			}
+			currentDefinition.dd = currentDefinition.dd.concat(match[2].split('::').filter((item)=>item).map((s)=>this.lexer.inlineTokens(s.trim())));
```
calculuschild, line 258: There is a crash here when typing the following line

```
Test
:: one
```
Can we get this to work? If we are allowing multi-line syntax I would like to be able to write each DD entry its own line.

calculuschild (changes requested):
This looks pretty good. Just some slight changes to the behavior following some discussion with Eric. Then I will merge this.

dbolack-ab:
> Along the lines of Markdown's "prime directive": Input source should look like the output.

In my view, this is a CSS problem, not an HTML or code problem. Because a DL may be styled to be single like `<DT><DD>` or multiline `<DL>
<DD` this seems like a place where bending that particular rule is appropriate.

shared/naturalcrit/markdown.js
dbolack-ab: This should be resolved.

shared/naturalcrit/markdown.js
dbolack-ab: I've done this, though it causes.. funny behavior with the default styling which looks like it was designed solely for single term single definition, inline lists. It might be prudent to simply move that CSS from Blank to PHB/DMG if that was done purely for book emulation.

calculuschild: Thabks! Ok. Ill take a look.

shared/naturalcrit/markdown.js
dbolack-ab: It is valid per spec and was recommended in some examples, but I have removed it.

Gazook89:
I don't think that this syntax highlighting should highlight this line of only colons as `define` as if it's a definition list-- rather, it should be vertical spacing.

I only just looked cracked this open so I don't have a solution in hand, but am looking now, at least until the kid wakes up from nap.

image

Gazook89:
I have a fix for my point raised above. Was about to submit but kid woke up so it’ll have to be later tonight

calculuschild:
The latest changes have affected the Monster Stat block again:

image

dbolack-ab:
> The latest changes have affected the Monster Stat block again:

I believe I have this resolved.

Gazook89:
> I don't think that this syntax highlighting should highlight this line of only colons as define as if it's a definition list-- rather, it should be vertical spacing.

Okay, so I have this figured out, and I've pulled in the latest changes from your branch into my branch.  However, while it appears to be possible for me to rebase onto your branch in your repo without issue, the word `rebase` is scary to me, so for right now I'm just going to copy/paste my version here below.  If it all looks okay to you, go ahead and just drop it into your PR yourself.  

#### editor.js
```js
// definition lists
if(line.includes('::')){
	if(/^:*$/.test(line) == true){ return };
	const regex = /^([^\n:]*?)(::[^\n]*)(?:\n|$)/ymd;  // the `d` flag, for match indices, throws an ESLint error.
	let match;
	while ((match = regex.exec(line)) != null){
		codeMirror.markText({ line: lineNumber, ch: match.indices[0][0] }, { line: lineNumber, ch: match.indices[0][1] }, { className: 'dl-highlight' });
		codeMirror.markText({ line: lineNumber, ch: match.indices[1][0] }, { line: lineNumber, ch: match.indices[1][1] }, { className: 'dt-highlight' });
		codeMirror.markText({ line: lineNumber, ch: match.indices[2][0] }, { line: lineNumber, ch: match.indices[2][1] }, { className: 'dd-highlight' });
		const ddIndex = match.indices[2][0];
		let colons = /::/g;
		let colonMatches;
		while((colonMatches = colons.exec(match[2])) !== null){
			codeMirror.markText({ line: lineNumber, ch: colonMatches.index + ddIndex }, { line: lineNumber, ch: colonMatches.index + colonMatches[0].length + ddIndex }, { className: 'dl-colon-highlight'} )
		}
	}
}
```
#### editor.css
```css
		.dl-highlight {
			&.dl-colon-highlight {
				font-weight   : bold;
				color         : #949494;
				background    : #E5E5E5;
				border-radius : 3px;
			}
			&.dt-highlight { color : rgb(96, 117, 143); }
			&.dd-highlight { color : rgb(97, 57, 178); }
		}
```

Note, I changed the class names to something I thought was easier to understand, if more verbose.  This also adds another class, one specifically for the colons syntax, rather than just styling that is based on `:not()`s.  

Finally, it uses the `d` regex flag for `indices`, which for some reason throws an ESLint error even though it's totally fine and has been for a few years.

Result:

image

dbolack-ab:
It was requested that this be altered to:

Treat "old-style" inline as before, without multiple entries

Add multi-line input similar to Ordered/Unordered Lists to the Multi-line definition descriptions

dbolack-ab:
> Sorry, I think i found an issue--
> 
> ```
> Term 1:
> ::Definition 1
> ::Definition 2
> ```
> 
> image
> similar issue with inline definition as well. I think it should be possible to have a colon after your term.

This should now be fixed.

client/homebrew/editor/editor.jsx
```diff
@@ -151,15 +151,21 @@ const Editor = createClass({
 
 						// definition lists
 						if(line.includes('::')){
-							const regex = /^([^\n]*?)::([^\n]*)(?:\n|$)/ym;
+							if(/^:*$/.test(line) == true){ return };
+							const regex = /^([^\n:]*?)(::[^\n]*)(?:\n|$)/ymd;  // the `d` flag, for match indices, throws an ESLint error.
```
Gazook89, line 155: I *think* you can just get rid of the first colon on this line, to make this:
```regex
/^([^\n]*?)(::[^\n]*)(?:\n|$)/ymd
```
That colon was added on Jan 7th, according to Blame, and I wonder if it is a remnant from a previous syntax that was tried?

This is to address this lack of highlight:
image

```
Term: :: definition
```

Gazook89 (commented):
I tried several times to try and break it and it seems all good, with one exception in the highlighting.

shared/naturalcrit/markdown.js
```diff
@@ -288,7 +288,114 @@ const definitionLists = {
...
+				return `<dd>${this.parser.parseInline(s)}</dd>`;
+			}).join('\n');
+			returnVal += `<dt>${this.parser.parseInline(def.dt)}</dt>${dds.indexOf('\n') > -1 ? '\n' : ''}${dds}\n`;
+		});
+		returnVal = returnVal.trim();
+		return `${returnVal}</dl>`;
+	}
+};
+
+Marked.use({ extensions: [mustacheSpans, mustacheDivs, mustacheInjectInline, blockDefinitionLists, superSubScripts] });
```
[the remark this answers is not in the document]
dbolack-ab: It does. *(&^(&*^

client/homebrew/editor/editor.jsx
dbolack-ab: That mostly fixed it, but I had to throw in a `:?\s?` at the end of the first capture group to catch `term:::definition` cases.

shared/naturalcrit/markdown.js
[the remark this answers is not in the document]
dbolack-ab: No, it didn't. I managed to collapse it into one extension and didn't clean up. *doh*

This is now cleaned up and blockDefinitionLists renamed as definitionLists.

dbolack-ab:
https://homebrewery.naturalcrit.com/edit/FtExK3Y9dKYm updated for more tests.

shared/naturalcrit/markdown.js
```diff
@@ -295,37 +295,125 @@ const superSubScripts = {
...
+				return `<dd>${this.parser.parseInline(s).trim()}</dd>`;
+			}).join('\n');
+			returnVal += `<dt>${this.parser.parseInline(def.dt)}</dt>${dds.indexOf('\n') > -1 ? '\n' : ''}${dds}\n`;
+		});
+		returnVal = returnVal.trim();
+		return `${returnVal}</dl>`;
 	}
 };
 
+Marked.use({ extensions: [mustacheSpans, mustacheDivs, mustacheInjectInline, definitionLists, superSubScripts] });
```
calculuschild, line 416: You have two separate `Marked.use` blocks, with mostly the same extensions listed. Probably remove this one and add your `definitionLists` to the one below instead.

calculuschild:
This looks good! I'm just going to go through and see if the logic can be shortened up at all, then merge.

shared/naturalcrit/markdown.js
dbolack-ab: Bad merge there. All cleaned up.

dbolack-ab:
@calculuschild Is the multiline regression from your work or did I miss it?

calculuschild:
> @calculuschild Is the multiline regression from your work or did I miss it?

It's just a new test I added today for an edge case we missed. Pretty sure I know where the fix is though.

dbolack-ab:
> > @calculuschild Is the multiline regression from your work or did I miss it?
> 
> It's just a new test I added today for an edge case we missed.
> 
> Pretty sure I know where the fix is though. I'm planning to merge it in just a bit.

Okay.  The fail I'm seeing is indeed a fail that worked at last I remember.

```
      <dl><dt>Term 1</dt>
    - <dd>Definition 1 and more and more</dd>
    + <dd>Definition 1</dd>
    + <dt>and more and more</dt>
      <dd>Definition 2</dd>
      <dd>Definition 3</dd></dl>
```

calculuschild:
No, it was broken before. It comes from this test which was written incorrectly:

```
Term 1
::Definition 1
Term 2
::Definition 1
```

This should output:

```
<dl>
<dt>Term 1</dt>
<dd>Definition 1 Term 2</dd> ("Term 2" is part of dd1 multiline)
<dd>Definition 1</dd>
</dl>
```

I added an extra newline between the Terms to correct that test, then added this new test which just changes `Term 2` to `more and more`

PR #3130, opened 2023-11-07
Fix minor grammatical error in FAQ

MurdoMaclachlan:
Just a minor grammatical fix, "its" being the correct possessive as opposed to "it's" being a contraction of "it is".

Not sure what's supposed to have changed on the last line, I think GitHub must have done something automatically since I didn't touch that line.


calculuschild:
Thanks @MurdoMaclachlan ! 💯 👍 ⭐

Issue #1488, opened 2021-07-30
Use "=" in curly syntax for html attribute declarations


ericscheid:
The new curly format looks like this: `{{classname,#idname,cssproperty:cssvalue CONTENT GOES HERE}}`.

**Proposal: have the equals character `=` signify `htmlattribute=attributevalue`.**

This would apply to the braced-span, braced-block, and braced-injector variants.

---
#### injecting the `ismap` attribute onto the `img` element in this link:

```
[![alt text](https://images.io/file.png "Title attribute"){ismap=true}](https://mappy.io/)
```

#### setting the lang attribute for a block of text

```
{{lang="fr",quote
Quand on a pas ce que l’on aime, il faut aimer ce que l’on a.
}}
```

ericscheid:
This particular feature might be a while off, but in the meantime can we disallow the use of `=` in the curly declarations so as to reserve its use for this?

5e-Cleric:
Fixed by #3131

PR #3131, opened 2023-11-07
Add arbitrary tag attribute assignment in mustaches

dbolack-ab:
This adds the ability to include attribute values for any element that can be altered by a moustache.

Form:

```
{attribute=value}

example:

![homebrew mug](https://i.imgur.com/hMna6G0.png) {position:absolute,bottom:20px,left:130px,width:220px,a=b and c,g=h}

becomes:

<img class=" " style="position:absolute; bottom:20px; left:130px; width:220px;" a="b and c" g="h" src="https://i.imgur.com/hMna6G0.png" alt="homebrew mug">
```

In order to permit spaces, the pattern matches for moustache code had to remove the space character as a delimiter. I believe I have adequate compensated.

This should solve #1488


ericscheid:
Looks good, but for the `a="b and c"` variation. 

In curly-blocks and curly-spans, values with spaces need to be quoted, like this: `{{border:"1px red solid" span-text-with-red-border}}` It would be better if we didn't have two slightly different/incompatible syntaxes and that this code could also (eventually) be applied to curly-blocks and curly-spans.

calculuschild:
> In order to permit spaces, the pattern matches for moustache code had to remove the space character as a delimiter.

Unfortunately this breaks a core behavior of the curly syntax. A space is the delimiter to separate the CSS directives from any block contents, in the case of Spans in particular.

dbolack-ab:
> Looks good, but for the `a="b and c"` variation.
> 
> In curly-blocks and curly-spans, values with spaces need to be quoted, like this: `{{border:"1px red solid" span-text-with-red-border}}` It would be better if we didn't have two slightly different/incompatible syntaxes and that this code could also (eventually) be applied to curly-blocks and curly-spans.

dbolack-ab:
> > Looks good, but for the `a="b and c"` variation.
> > In curly-blocks and curly-spans, values with spaces need to be quoted, like this: `{{border:"1px red solid" span-text-with-red-border}}` It would be better if we didn't have two slightly different/incompatible syntaxes and that this code could also (eventually) be applied to curly-blocks and curly-spans.

I was so focused on my test case I completely forgot about all the other patterns.

I'm gonna have to beat on the regex more...

dbolack-ab:
> > In order to permit spaces, the pattern matches for moustache code had to remove the space character as a delimiter.
> 
> Unfortunately this breaks a core behavior of the curly syntax. A space is the delimiter to separate the CSS directives from any block contents, in the case of Spans in particular.

I looked at the "welcome" brew to see if I could spot any places where my change on the \w to \t\f\n\r breaks anything and I'm not spotting one. Am I missing anything?  All the moustache tests pass as well?

calculuschild:
It doesn't break any existing tests, but your new code only touches the single curly "injector" variant. If we were to add that syntax to the double curly "span" variant it would break. We need to ensure any syntax we add would be functional for all variants.

calculuschild:
For example:

```
{{position:absolute,bottom:20px,left:130px,width:220px,a=b and c,g=h My Span Text}}

becomes:

<span class=" " style="position:absolute; bottom:20px; left:130px; width:220px;" a="b">and c,g=h My Span Text<\span>
```

ericscheid:
The regex for `attr=value` and `attr="value with spaces"` would be _very_ similar to the regex for `csspropery:value` and `csspropery:"value with spaces"` which exists in curly-blocks and curly-spans.

calculuschild:
@dbolack-ab If you can get this to work using quotes to contain the spaces instead of removing space as a delimiter, I can give it another look.

client/homebrew/editor/editor.jsx
```diff
@@ -162,15 +162,15 @@ const Editor = createClass({
 
 						// Highlight injectors {style}
 						if(line.includes('{') && line.includes('}')){
-							const regex = /(?:^|[^{\n])({(?=((?::(?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"':{}\s]*)*))\2})/gm;
+							const regex = /(?:^|[^{\n])({(?=((?::(?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"':{}\n\r\t\f]*)*))\2})/gm;
```
calculuschild, line 165: Can you update these highlighting regex to match your current regex approach from `markdown.js`? i.e. revert back to `\s` and put the `[:=]` as needed.

Perhaps you can use this example to make sure the highlighting is correct (screenshot shows current behavior; I drew lines where the color is not right):

```
{{test,attribute=myAttr,attribute2="my Attr" text and text}}

{{text,style:myStyle,style2:"my Style" text and text}}

{{text,style:myStyle,attribute=myAttr,style2:"my Style",attribute2="my Attr" text and text}}
```

image

shared/naturalcrit/markdown.js
```diff
@@ -84,7 +84,7 @@ const mustacheDivs = {
 	start(src) { return src.match(/\n *{{[^{]/m)?.index; },  // Hint to Marked.js to stop and check for a match
 	tokenizer(src, tokens) {
 		const completeBlock = /^ *{{[^\n}]* *\n.*\n *}}/s;                // Regex for the complete token
-		const blockRegex = /^ *{{(?=((?::(?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"':{}\s]*)*))\1 *$|^ *}}$/gm;
+		const blockRegex = /^ *{{(?=((?:[=:](?:"[\w,\-()#%. ]*"|[\w\-()#%.]*)|[^"'=:{}\s]*)*))\1 *$|^ *}}$/gm;
```
calculuschild, line 87: Nitpick. Just change this to `[:=]` to match the order of the others.

tests/markdown/mustache-syntax.test.js
```diff
@@ -124,6 +124,12 @@ describe('Inline: When using the Inline syntax {{ }}', ()=>{
 		const rendered = Markdown.render(source);
 		expect(rendered, `Input:\n${source}`, { showPrefix: false }).toBe('<span class="inline-block pen" id="author" style="color:orange; font-family:trebuchet ms;">text</span>');
 	});
+
+	it('Renders an image with added attributes', function() {
```
calculuschild, line 128: Can you add test cases for `{{mustache` inline spans and blocks?

shared/naturalcrit/markdown.js
```diff
@@ -328,14 +327,21 @@ const voidTags = new Set([
...
+
+	const id         = _.remove(tags, (tag)=>tag.startsWith('#')).map((tag)=>tag.slice(1))[0];
+	const classes    = _.remove(tags, (tag)=>(!tag.includes(':')) && (!tag.includes('=')));
+	let attributes = _.remove(tags, (tag)=>(!tag.includes(':')) && (!tag.includes('#')));
+	const styles     = tags?.length ? tags.map((tag)=>tag.replace(/:"?([^"]*)"?/g, ':$1;').trim()) : [];
 
-	if(!tags)	return '"';
+	if(attributes) {
+		attributes = attributes.map((attr)=>attr.replace(/="?([^"]*)"?/g, '="$1"'));
+	}
```
calculuschild, line 339: I have an idea to simplify this slightly. I will push up a commit later after I make sure it works.

calculuschild (changes requested):
Good job here! Sorry I'm finally trying to catch up on these PRs. Just some small items.

5e-Cleric:
Should this wait for #3066 first? to get strings as attributes?

I say that, but i'm the one holding that PR back, will fix that later today if i can.

client/homebrew/editor/editor.jsx
dbolack-ab: This should be corrected.

shared/naturalcrit/markdown.js
dbolack-ab: Let me know. I tend to get sideswiped by a third party pushing a change at a PR from a personal fork. It's a weird pattern for me. :)

I banged my head on this particular bit a bunch and this was the only order of operations I found that worked.

dbolack-ab:
> Should this wait for #3066 first? to get strings as attributes?
> 
> I say that, but i'm the one holding that PR back, will fix that later today if i can.

Probably a good idea.

calculuschild (approved):

calculuschild:
I fixed a small bug and did a little cleanup. I think this is ready to merge unless you have any major concern about my changes. Passes your tests still and seems to function as expected.

dbolack-ab:
@calculuschild  Looks good to me!

Issue #1958, opened 2022-01-11
[LOCAL]:  Enable serving local images


G-Ambatte:
By modifying `server.js` slightly, we can allow users with local installs to serve images from their local directory, meaning that they will no long need to rely on services outside of their local machine (e.g. Imgur.com).

Steps:

1. Create a directory "staticImages" in the project root directory.

2. Add the following lines to `server.js`:
```
if(process.env.NODE_ENV == 'local') {
    app.use(express.static(`./staticImages`));
}
```
3. Restart the Homebrewery server.

---

As you will see in the following screenshot, my Ubuntu local install can now access images hosted in the `staticImages` folder.
![image](https://user-images.githubusercontent.com/19826659/148885979-148d4ff0-b0f8-4ad5-ae24-aeb509f5fcc8.png)

*EDIT: In the above image, my specific configuration is visible - I have symlinked `/home/hb-test/images` to `/usr/local/homebrewery/staticImages`, which is why two different paths are shown - I can upload files via WinSCP to `/home/hb-test/images` while Homebrewery can serve the files locally from `/usr/local/homebrewery/staticImages`. The above change does not include any such folder linking and putting it in place it is left as an exercise for the individual user, if that is what they should desire.*

G-Ambatte:
A slightly better approach may be along the lines of:

```
const localEnvironments = ['local','docker','other'];
if(localEnvironments.includes(process.env.NODE_ENV) {
    [...]
}
```

This allows the same static directory logic to be applied to other environments than just `local`, as deemed appropriate, and is easily extended to any additional NODE_ENV settings that might be developed in the future.

It also might be appropriate to store the content of `localEnvironments` in a config file.

PR #3132, opened 2023-11-08
Add local statics for images and typefaces

dbolack-ab:
This solves issue #1958.

This adds the endpoints /staticImages and /staticFonts when Homebrewery is run as a `local` NODE_ENV. 

staticImages is set to the path set in the HB config/default.json as hb_images *OR* $HBROOT/staticImages if not set.
staticFonts is set to the path set in the HB config/default.json as hb_fonts *OR* $HBROOT/staticFonts if not set.

Paths should be fully qualified in the config file.


Gazook89:
I only worked with the images part for now, but I couldn't get it to work?  This is likely my fault, so I probably just need a little more guidance.

I have pulled your PR here, confirmed that is what I am working with, and did `npm install` just in case.  Then, I added a folder to the root directory `.../homebrewery/staticImages` and dropped a png in there.  For now, I left the `/homebrewery/config/defaults.json` alone, relying on the built-in default you've added.

In my brew, I did this:

```md
![cat warrior](Class-Table-Frame.png)

![cat warrior](staticImages/Class-Table-Frame.png)
```

And a couple of other variants.  No dice.  I then tried adding the `hb_images` (and `HB_IMAGES`) key to the default.json to see if I could get it to work that way, again no dice.

Any tips?

ericscheid:
> ```
> cat warrior
> 
> cat warrior
> ```
> Any tips?

Those two url references would be relative to the current page location, and thus expanded to something like `https://homebrewery.naturalcrit.com/edit/:id/Class-Table-Frame.png` or `https://homebrewery.naturalcrit.com/edit/:id/staticImages/Class-Table-Frame.png` or 
`https://homebrewery.naturalcrit.com/share/:id/Class-Table-Frame.png` or 
`https://homebrewery.naturalcrit.com/new/:id/Class-Table-Frame.png`.

Try `cat warrior` to cause the img src to be an absolute url.

Here's an example of another image that should already work:
`https://homebrewery.naturalcrit.com/assets/redTriangle.png`
`red triangle`

Gazook89:
Yeah I thought I tried that as well with no luck but I will try again later today.  I had also tried changing the code to match what @G-Ambatte  had suggested in the linked issue, and then following that format, to no avail.

ericscheid:
What happens if you right-click the broken image and say "open image in new tab" .. what does the window.location look like?

G-Ambatte:
I'm getting 406 responses when attempting to get a local image file, which appears to be from the content-negotiation middleware, That said, disabling said middleware does NOT result in the image being correctly served, so I suspect that there's still something that I've missed.

Gazook89:
As noted in gitter, when i copy the images to the build folder, it works.  

In the buildHomebrew.js script:
```js
	// Move assets
	await fs.copy('./themes/fonts', './build/fonts');
	await fs.copy('./themes/assets', './build/assets');
	await fs.copy('./client/icons', './build/icons');
	await fs.copy('./staticImages', './build/staticImages');
```
(note, this in the "themes" portion, which doesn't quite sit right with me...this is sort of out of my depth, though).

and in app.js:
```js
	// Add Static Local Paths
	app.use(express.static('build/staticImages'));
```

but it does require including the directory:
```md
![alt](staticImages/image.png)
```

dbolack-ab:
> ~As noted in gitter, when i copy the images to the build folder, it works.~ see below in edit
> 
> In the buildHomebrew.js script:
> 
> ```js
> 	// Move assets
> 	await fs.copy('./themes/fonts', './build/fonts');
> 	await fs.copy('./themes/assets', './build/assets');
> 	await fs.copy('./client/icons', './build/icons');
> 	await fs.copy('./staticImages', './build/staticImages');
> ```
> 
> (note, this in the "themes" portion, which doesn't quite sit right with me...this is sort of out of my depth, though).
> 
> and in app.js:
> 
> ```js
> 	// Add Static Local Paths
> 	app.use(express.static('build/staticImages'));
> ```
> 
> but it does require including the directory:
> 
> ```
> alt
> ```
> 
> duh, of course it works...it doesn't even need the app.js bit...because it's just referencing a directory in the build, which isn't anything special. So I guess....if someone is working locally, is there anything particularly wrong with just telling them to create a path `./build/images/` and using that to store their images? Do we need have this in app.js at all?
> 
> Sorry, I think i've lost the thread.

Sorry, seem to have missed these questions.

My read on the original request was this was desired for situations where the user probably does not want or cannot  copy the static content into place or needs to point at a path outside of `/build/` such as a mounted docker volume.

I've updated the initial commit comment to be a little more clear on usage.

5e-Cleric:
After reading all the comments, i am still not clear on if this is working and how to test this, just create a folder called `staticImages` and use paths relative to that?

dbolack-ab:
Updated description

Gazook89:
> I've updated the initial commit comment to be a little more clear on usage.

I'm a dumbo.  Can you make it even more clear?  My expectation is that I can create a `staticImages` directory at the root of the Homebrewery repo, and drop `cauldron.png` in there.  Then, in homebrewery, do this:

```
![cat warrior](/staticImages/cauldron.png) {width:325px,mix-blend-mode:multiply}
```
And get a picture of a cauldron.  

This doesn't work.

ericscheid:
If I'm reading the code right ..

1. update your `config/default.json` file to specify where your images will be
2. the server will internally map `/staticImages/*` to the path specified

You probably don't want to drop a folder of images into your repo folder, as git/etc will pester you about adding those assets into the repo proper. Those are your local/private files.

Gazook89:
I have now attempted this in default.json:

```json
{
	"secret" : "secret",
	"web_port" : 8000,
	"enable_v3" : true,
	"enable_themes" : true,
	"hb_images" : "/Users/JohnJones/Documents/Graphic Design/Works"
}
```
Which is filled with a bunch of pngs including `HB-wordcloud.png`, and then in HB doing this:

```
![cat warrior](/staticImages/HB-wordcloud.png) {width:325px,mix-blend-mode:multiply}
```
And it still doesn't work.  I think this PR just needs a step-by-step of what needs to be done in the repo, and a screenshot of what it looks like in HB.

5e-Cleric:
Question, can it be done for the build process to create a folder if in local, then add that as the static automatically to the package?

dbolack-ab:
Let me walk myself through it - it's been a spell.

dbolack-ab:
This is currently working and not working. I assume this is a react issue I do not have the depth to understand.

If I place a file in the configured  path it is accessible by direct browser request. If I place that same URL in markdown it doesn't load and there is no attempt to network load.

G-Ambatte:
The content negotiation middleware is blocking the image request. Putting in an exception specifically for image requests to the local directory fixes the issue.

dbolack-ab:
> The content negotiation middleware is blocking the image request. Putting in an exception specifically for image requests to the local directory fixes the issue.

I looked at that but could not sort it out

Gazook89:
Is it worth putting in the optional lines in `default.json`, commented out, with example directories so that this feature has more visibility and is clearer to execute?  

Crap, as I was typing I realized JSON doesn't allow comments.  (*why?!?!*)

Hm...perhaps, in lieu of that, perhaps part of this PR should be updating the readme.md with clear instructions as well.  Otherwise, unless you know about this PR, you would have no way of knowing this feature exists.

Gazook89:
So for my testing, this is what is working for me (on MacOS):

# `default.json`
```json
{
	"secret" : "secret",
	"web_port" : 8000,
	"enable_v3" : true,
	"enable_themes" : true,
	"hb_images" : "/Users/[MyName]/Documents/Graphic Design/Assets/Page Elements/Frames/Homebrewery",
	"hb_fonts" : "/Users/[MyName]/Library/Fonts"
}
```
# Brew Editor
```md
...Veniam amet nulla est.

![cat warrior](/staticImages/SciFi-Screen.png) {width:325px,mix-blend-mode:multiply}
```
# Style Editor
```css
@font-face {
    font-family: 'MoneyPenney';
    src: url('/staticFonts/Webfonts/MoneyPenny-Sans.woff2') format('woff2');
    font-weight: normal;
    font-style: normal;
}
.page p {
  font-family: MoneyPenney;
}
```
# Result:

image

So i'd say that it definitely works.  Some tricky bits:  
- I had to rename my `SciFi Screen.png` to `SciFi-Screen.png`, eliminating the space.  I understand this may be obvious and that uses spaces within file names is generally frowned on, but in this day-and-age is probably pretty common.
- The @font-face declaration likely isn't immediately obvious to most users.  

I'd say that if any documentation is written, these two points should be included.

-----

Question about the Fonts portion though--- if this is just for Local use, shouldn't fonts work anywhere, if you have them installed on your machine?  I can set fonts in my local HB, without this PR, to any installed font no problem.  Is the sole advantage of this PR that maybe you don't need to have the font *installed* on your local machine, just available?

5e-Cleric:
`@font-face` is a bitch to work with, but that's life, users will have to learn it, although we could make a snippet to add a font...

Gazook89:
This is purely for local users, not online. So I don’t think we need a snippet. Just documentation in a read me.

dbolack-ab:
> Crap, as I was typing I realized JSON doesn't allow comments. (_why?!?!_)

JSON wasn't really designed to be used in a lot of the ways it is used. Comments were not a consideration.

dbolack-ab:
> * I had to rename my `SciFi Screen.png` to `SciFi-Screen.png`, eliminating the space.  I understand this may be obvious and that uses spaces within file names is generally frowned on, but in this day-and-age is probably pretty common.

I suspect you still needed to %20 the spaces in the Markdown.

> Question about the Fonts portion though--- if this is just for Local use, shouldn't fonts work anywhere, if you have them installed on your machine? I can set fonts in my local HB, without this PR, to any installed font no problem. Is the sole advantage of this PR that maybe you don't need to have the font _installed_ on your local machine, just available?

* Not providing the faces in this manner doesn't emulate outside behaviour - which may not be a big deal
* This may be a useful feature outside of local environments if you wished to use storage that wasn't populated by the build process, such as a Docker Volume. Though that would be an extension of this PR.

config/default.json
```diff
@@ -7,4 +7,4 @@
 	"enable_themes" : true,
 	"local_environments" : ["docker", "local"],
 	"publicUrl" : "https://homebrewery.naturalcrit.com"
-}
+	}
```
Gazook89, line 10: Can remove the whitespace here (the tab)

Gazook89 (changes requested):
Can the hb_images and hb_fonts config properties be included in default.json from the outset, but with values `"default"` or `null`?  And then if those values are set, it just treats it as if the property doesn't exist?

This gets to my point in one of my most recent comments on the PR:  this feature is basically hidden unless you happen to follow development and know about this PR.  Having the properties already available in the config, but "turned off", will at least hint to a user that it exists and then they can ask about it.   

To go further, a small blurb in the readme.md about usage might be helpful (or it could even just link to this PR, [perhaps my comment where I spell it out for an infants](https://github.com/naturalcrit/homebrewery/pull/3132#issuecomment-2334779523)).

dbolack-ab:
> Can the hb_images and hb_fonts config properties be included in default.json from the outset, but with values `"default"` or `null`? And then if those values are set, it just treats it as if the property doesn't exist?

I see no reason why not.

calculuschild (commented):
I don't have strong feelings on the behavior of this PR. I will approve it so when you guys are in agreement that it is done someone can merge it.

server/middleware/content-negotiation.js
```diff
@@ -1,12 +1,18 @@
...
+const isLocalEnvironment = config.get('local_environments').includes(nodeEnv);
+
 module.exports = (req, res, next)=>{
 	const isImageRequest = req.get('Accept')?.split(',')
         ?.filter((h)=>!h.includes('q='))
         ?.every((h)=>/image\/.*/.test(h));
-	if(isImageRequest) {
+	if(isImageRequest &&
+		(!isLocalEnvironment && !req.url?.startsWith('/staticImages'))
+	) {
```
calculuschild, line 11:
```suggestion
	if(isImageRequest && !isLocalEnvironment && !req.url?.startsWith('/staticImages') {
```

calculuschild (approved):

Gazook89 (approved):

calculuschild:
Ok. Has enough reviews and doesn't seem to affect normal site function, so I'll just merge this.

Issue #2171, opened 2022-05-10
add markdown superscript/subscript?


ericscheid:
https://www.reddit.com/r/homebrewery/comments/ujuvov/superscript_and_subscript/

Superscript can be done via html tags `<sup>...</sup>` and `<sub>...</sub>` but it would be nicer to have markdown style syntax.

Superscript and subscript are not currently supported in the markdown library we use. We could add it as custom syntax.

Reddit does it, like this:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/2605467/167633802-8c964425-cdf6-4f69-ae6c-36d86582990d.png">

The optional wrapping in (parens) is useful for superscripts with word breaks, but also for ending a superscript without a word break (e.g. `e^(iπ)+1 = 0` for <kbd>e<sup>iπ</sup>+1 = 0</kbd>, or `0.5 × E = mc^(2)÷2` for <kbd>0.5 × E = mc<sup>2</sup>÷2</kbd>).

Some other markdown flavors do it slightly differently (i.e. wrapping the superscript in carets, like this: `E=MC^2^`.
https://www.markdownguide.org/extended-syntax/

Subscript could be indicated via tildes `~` perhaps, like `H~(2)SO~4` for Sulphuric Acid H<sub>2</sub>SO<sub>4</sub>.

calculuschild:
Tilde `~` is already used for ~strikethrough~ text.

Gazook89:
I know that I have used and recommended to others to use the strikethrough syntax for quick & easy custom inline styling.   Set `strikethrough` in the css to whatever styling you want, and then use `~~my fancy text~~` in your brew to more concisely do what `<span class='my-fancy-style`>my fancy text</span>` would normally do.  This works precisely because strikethrough isn't typically used for anything else.

Now with v3, it's almost as simple to do `{{my-fancy-style my fancy text}}` or `{{mfs my fancy text}}`.  

Maybe subscripts can be a combination like `text^.(my subscript)` or `^_`.    Or simply just create baked in classes `sup` and `sub` so that doing `{{sup my text}}` creates superscript and similar for subscript.  It's not *new syntax* but just a convenience class.

---

A little off topic here, but maybe in the far future it's possible for users to create a list of their own 'syntax' that overrides the built in syntax.  In the user profile or a single brews metadata, they could add to a list something like:

```
++{text}++ = "custom-span"
```
[...]

PR #3133, opened 2023-11-08
Add subscript and subscript markdown tokens

dbolack-ab:
Uses ^^ for superscript and ^^^ subscript as wrappers in the same
pattern as italics and bold ( * and **, respectively)

Adds editor hot-keys and syntax highlighting. (CTRL-6/CTRL-7)

Exact values may not be ideal. Short of the suggested overloading of ~,
I didn't see a better option for the delimiter.

Solves Issue #2171


calculuschild:
There's probably room for discussion, but I would vote for single `^` and double `^^` rather than double and triple.

Otherwise I agree this is probably the cleanest approach without overriding things.

Gazook89:
One concern is that maybe single pairs would break some brews, if they are using the symbols? I think it’s fairly low chance of that.

calculuschild:
> One concern is that maybe single pairs would break some brews, if they are using the symbols? I think it’s fairly low chance of that.

Probably low chance, but we can also be specific with our syntax that you can't allow blank spaces on the end or start, which I think would account for most cases:

i.e.:

`^some text^` is valid, but `^ some text^`  and `^some text ^` are not.

calculuschild:
Also, should we account for Long strings of `^^^^^^^^^^^Some text^^^^^^^^^^^`? Is that multiple supers or multiple subs?

shared/naturalcrit/markdown.js
```diff
@@ -206,6 +206,38 @@ const mustacheInjectBlock = {
 	}
 };
 
+const superSubScripts = {
+	name  : 'superSubScripts',
+	level : 'inline',
+	start(src) { return src.match(/.*\^\^(.+)\^\^/)?.index; },  // Hint to Marked.js to stop and check for a match
```
calculuschild, line 212: This regex can be made faster by removing the starting `.*` (it doesn't matter what is before the beginning of the superscript) and making sure the central `.+` excludes newlines (so this isn't madly searching the entire document each time it sees a ^`.

I suggest we also limit the regex so the super/sub must not start with whitespace. i.e. `^some text^` is valid for superscript, but `^ some text^`  and `^some text ^` are not. This would help avoid conflicts with normal use cases of `^` and also match the existing markdown specs for `*italics*`.

shared/naturalcrit/markdown.js
```diff
@@ -206,6 +206,38 @@ const mustacheInjectBlock = {
...
+	tokenizer(src, tokens) {
+		const superRegex = /\^\^(.+)\^\^/y;
```
calculuschild, line 214: The Marked.js package we use recommends these should start with an unescaped `^` at the beginning and use the `m` regex tag to make sure these only match at the start of the current string. It's possible for the markdown parser to mistakenly consume extra characters in certain cases if it finds a match later on in the string behind other tokens.

calculuschild (changes requested):

dbolack-ab:
> One concern is that maybe single pairs would break some brews, if they are using the symbols? I think it’s fairly low chance of that.

That was the assumption I made, that it was *plausible* for there to be a line that contained two unrelated carets and could trigger a false match.

shared/naturalcrit/markdown.js
dbolack-ab: This should be resolved by changing the  (.+) to ([^\s\^][^\^]*[^\s\^]).

shared/naturalcrit/markdown.js
dbolack-ab: I made that change on the tokenizer regexes but when made on the start match it fails to match.

ericscheid:
> > One concern is that maybe single pairs would break some brews, if they are using the symbols? I think it’s fairly low chance of that.

One possible use case is where someone is using `^` as a footnote marker for cells in a table, like this:

```
...
|**Text**| More text^| Would you believe it^ | And Again* | ... |
...
`^` special note numero uno
`*` another special note
```

calculuschild:
> One possible use case is where someone is using `^` as a footnote marker for cells in a table, like this:

Fortunately, the contents of each table cell is evaluated individually, so there is no risk of "cross-cell" problems like this. (The syntax highlighting is still there, but the tables are not an issue in the output:

image

shared/naturalcrit/markdown.js
calculuschild: Yeah this only applies to the regex inside the tokenizer.

shared/naturalcrit/markdown.js
```diff
@@ -206,6 +206,38 @@ const mustacheInjectBlock = {
 	}
 };
 
+const superSubScripts = {
+	name  : 'superSubScripts',
+	level : 'inline',
+	start(src) { return src.match(/\^\^[^\s].+[^\s]\^\^/m)?.index; },  // Hint to Marked.js to stop and check for a match
+	tokenizer(src, tokens) {
+		const superRegex = /^\^\^([^\s\^][^\^]*[^\s\^])\^\^/m;
```
calculuschild, line 214: Unfortunately this approach requires a minimum of 2 characters inside the `^^` because you require a non-space at the start, and another non-space at the end.

After fiddling around on regex101.com, I think the following approach will be the fastest:
```suggestion
		const superRegex = /^\^\^(?!\s)(?=([^\n\^]*[^\s\^]))\1\^\^/m;
```

The `(?!\s)` is a negative lookahead. It will check that there is no space without actually capturing it, which fixes the 2-character minimum.

The `(?=(...))\1` uses a positive lookahead to mimic the behavior of Atomic Groups `(?>...)` which are not available in Javascript. Essentially if gets to the end of the Regex and finds no match, it won't try to backtrack to find a valid end somewhere it missed. Greatly speeds up the `^^whitespace at the end ^^` case.

The `[^\n\^]*[^\s\^]` ensures that there are no newlines or `^` inside. I also added a check to the final character so it can't be another `^`.

Of course this doesn't allow any nesting of superscript or subscript. Doing that in a predictable way is a lot more involved and can't be solved with RegEx alone. Though if we decide to add that capability in a later PR, we can just copy the source from Marked.js and use the same approach they use to handle nested asterisks.

shared/naturalcrit/markdown.js
```diff
@@ -206,6 +206,38 @@ const mustacheInjectBlock = {
 	}
 };
 
+const superSubScripts = {
+	name  : 'superSubScripts',
+	level : 'inline',
+	start(src) { return src.match(/\^\^[^\s].+[^\s]\^\^/m)?.index; },  // Hint to Marked.js to stop and check for a match
```
calculuschild, line 212: This start script still shouldn't have a `.+` in it because that means if it sees a `^^` it will scan the entire document including newlines for the largest possible string to fit inside the closing `^^`. This will become very slow. We want to limit this search to one line.

Actually, for Marked.js, this "start" parameter can just be the first few characters. We might simplify this to just this:
```suggestion
	start(src) { return src.match(/\^\^/m)?.index; },  // Hint to Marked.js to stop and check for a match
```

calculuschild:
> That was the assumption I made, that it was _plausible_ for there to be a line that contained two unrelated carets and could trigger a false match.

I think we are pretty safe that this won't happen. Maybe in a couple of _very rare_ cases.

shared/naturalcrit/markdown.js
dbolack-ab: > Of course this doesn't allow any nesting of superscript or subscript. Doing that in a predictable way is a lot more involved and can't be solved with RegEx alone. Though if we decide to add that capability in a later PR, we can just copy the source from Marked.js and use the same approach they use to handle nested asterisks.

Is this a thing that happens?

dbolack-ab (commented):
Should be taken care of now?

calculuschild (approved):

calculuschild:
I'm finally getting around to merging this! Just did a small bit of cleanup and I think it's ready. 👍 

Thanks @dbolack-ab ! 💯 🎉 ⭐

PR #3159, opened 2023-11-29
Primitives for inline variables

dbolack-ab:
This is experimental code based on some of the existing reflink pattern.

Due to the huge number of variations in link blocks, I went with double [] instead of single.

Example Brew code:

```
[a]:b
[[$b="this is a ;"]]


[[$b]]

With the quiet macro

[[a; $g="Hello";:q ]]

Without the quiet macro

[[a; $g="Hello" ]]

[[a]]

```

```
this is a ;

    this is a ;

    With the quiet macro

    Without the quiet macro

    Hello

    b
```

Just a start. Fails are few things, like, I can't seem to make it bold...


dbolack-ab:
I'm going to close this for the other PR, though I may lift some things from it later. :)

PR #3163, opened 2023-12-01
Convert BrewRenderer to function, PPR always on

calculuschild:
Converts the BrewRenderer component to a functional component, which simplifies a lot of the logic.

Also removes any "PPR" logic, since we parse all pages anyway. Still only "updates" the HTML of the current page in view + 3 pages in both directions.

Should be just as responsive as the current Master.

Remaining quirk is switching between Legacy and V3, which throws the current page out of position and might scramble page elements as it loads the new CSS. With the previous logic, we rendered dummy pages everywhere offscreen, which uses both `.phb` and `.page` in the class, so changing renderers did not affect them. Now, for the short moment while the new CSS loads, page sizes may be undefined briefly as Legacy themes use `.phb` and V3 themes use `.page`. This can lead to crashes if it shuffles the current page fully out of range of any page.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -88,7 +88,7 @@ const BrewRenderer = (props)=>{
 		}));
 	};
 
-	const shouldRender = (pageText, index)=>{
+	const shouldRender = (index)=>{
 		if(!state.isMounted) return false;
 
 		if(Math.abs(index - state.viewablePageNumber) <= 3)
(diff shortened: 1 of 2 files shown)


client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -13,254 +12,215 @@ const ErrorBar = require('./errorBar/errorBar.jsx');
...
-	},
+	};
 
-	renderPage : function(pageText, index){
-		let cleanPageText = this.sanitizeScriptTags(pageText);
-		if(this.props.renderer == 'legacy')
-			return <div className='phb page' id={`p${index + 1}`} dangerouslySetInnerHTML={{ __html: MarkdownLegacy.render(cleanPageText) }} key={index} />;
-		else {
+	const renderPage = (pageText, index)=>{
+		console.log(`renderPage ${index}`);
```
G-Ambatte, line 131: Might want to drop this debugging statement before merge to master.

PR #3165, opened 2023-12-02
Add tag filtering

### LINKED ISSUE #3164: filter brews by clicking tags
> When you click a tag [in the userpage brew card], it filters the user's documents so only documents with that tag are showing

Reading back through the history on this, we never actually created this functionality. I believe we can probably do so by making the tags automatically link to the current URL with a `?filter=${URL.encode(tag content)}`.

_Originally posted by @G-Ambatte in https://github.com/naturalcrit/homebrewery/issues/758#issuecomment-1710898940_

G-Ambatte:
This PR resolves #3164.

It adds the `updateListFilter` function to the BrewItem component, allowing it to update the filter in the upstream ListPage.

This function currently takes two arguments, `type` and `term`. In the current implementation, `type` is unused, and `term` is applied to the existing text filter.


ericscheid:
Looking great, all we need is UI feedback/status to show that the userpage is currently filtered by tags.

I'm imagining a 2nd userpage utility bar that appears when tag filtering is in effect.

image

- Clicking any of those tags should unselect that tag from the filter.
- They should be sorted and styled similar to how tags are in `.brewItem`

5e-Cleric:
This looks complete, what is holding it?

Myself, i liked the first style option better, looks cleaner and pops out way more.

calculuschild:
> This looks complete, what is holding it?

Mostly I just didn't get around to review it yet. Once I see it works, I usually try to look through the code itself to check:

- Is this code going to add a bug?
- Is there any logic that can be simplified?
- Is the new code linted?

This PR seems pretty harmless, so if you want to look it over and check any of those things we can work on getting it merged.

5e-Cleric:
Found an error, not exactly in this PR, tags are not trimmed, we could include that in this PR, `a  ` is registered as tag, obviously different from `a`

5e-Cleric:
About logic simplification, i recommend adding this property to `.brewTags span {`:
`border-color:currentColor` or `border-color:initial`, both will use the color property as their border-color, which is being set anyway.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
```diff
@@ -59,6 +59,44 @@
 		white-space: nowrap;
 		display: inline-block;
 		font-weight: bold;
```
5e-Cleric, line 61: Let's add `borderColor:currentColor` here, and delete border-color declarations along the way, since they all use the same color.

client/homebrew/pages/basePages/listPage/listPage.less
```diff
@@ -125,4 +125,73 @@
 
 
 	}
+	.tags-container {
```
5e-Cleric, line 128: In this element, there are some properties that are not valid, and some which add nothing.

- font-family is already inherited, so can go
- position:sticky does not really happen, since the page is not scrolling, only the brewlist does. This applies to all nav elements
- if position is not set, top and left can go
- a width of 100% is the default value of the property(can't find the doc that talks about this but any block-level element(flex included)'s default width is auto, which means the element will take up the available width in its containing block.
- text-align:center is not necessary, since there is no text outside of a span in this element
- z-index:1 is also unnecessary, the default works just fine.

5e-Cleric (changes requested):

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
G-Ambatte: Applied in the latest commit.

client/homebrew/pages/basePages/listPage/listPage.less
G-Ambatte: I think a lot of these are remnants of the original tag container, which was styled quite differently.
Applied in the latest commit.

5e-Cleric (approved):

5e-Cleric:
Remove that last comment and i think this is done, merging in 24 hours if nobody says otherwise.

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -136,13 +137,37 @@ const ListPage = createClass({
 		return;
 	},
 
-	updateUrl : function(filterTerm, sortType, sortDir){
+	updateListFilter : function(type, term){
+		this.updateUrl(this.state.filterString, this.state.sortType, this.state.sortDir, term);
+	},
```
calculuschild, line 142: This appears to be a function that just calls another function. Is there a way we could just use `updateUrl` directly?

calculuschild (commented):
Trusting 5e-Cleric has done his due dilligence on this. Just noticed one thing I'm not sure about.

5e-Cleric:
> Trusting 5e-Cleric has done his due dilligence on this. Just noticed one thing I'm not sure about.

Gotta be honest, not sure how this one can go sideways, or how to test it other than using it, i already reported one issue, but it is outside of the scope of this PR, and taking care of it in #3284, albeit could use some help with that.

client/homebrew/pages/basePages/listPage/listPage.jsx
G-Ambatte: This function is passed down to the BrewItem (line 86) to allow the tag onClick to update the URL while maintaining the state of the rest of the current filtering parameters. That said, the `type` parameter is never actually used - I had originally envisioned being able to set multiple parameters using this function, but that's not how I ended up actually making the code functional.
I'll take a deeper dive into it in a few hours time, once I'm free from work.

calculuschild: What i meant, is can we pass down `updateUrl` instead of passing down this other function?

G-Ambatte: Almost certainly, I'll see what I can do after work tonight.

G-Ambatte: Latest commit should have the call directly to `updateUrl`.

PR #3172, opened 2023-12-05
Eldeberry Inn Icon font to HB

### LINKED ISSUE #3171: Add Elderberry Inn Icons to HB
@Gazook89 found this awesome icon font with [The CC0 1.0 Universal!](https://creativecommons.org/publicdomain/zero/1.0/deed.en) License, and conversation has started on whether we could add this font to the tool for users to use on their brews.

The main discussion revolves around whether to use this font as a websafe font (WOFF2) or as an icon system, hosting the SVGs and using CSS to insert the icon itself.

The first method's syntax would probably look like this:

`{{font-family:"elderberry inn" &#E923;}}` or `{{icon &#E923;}}` if we made a specific class.

The second method would be more user friendly, its syntax would look more like:

Using the standard FontAwesome syntax: `{{fac, fa-chained-hand}}`  or using a specific syntax: `{{icon-chained-hand}}` or `{{i-chained-hand}}`.

Both are valid options, but style details still might be an important issue.

5e-Cleric:
Resolves #3171 

We should still talk about icon size and other easy-to-solve matters


5e-Cleric:
[HB-ElderberryInnIconReferenceList.txt](https://github.com/naturalcrit/homebrewery/files/13573802/HB-ElderberryInnIconReferenceList.txt)
That should work as a reference list

calculuschild:
Could you make an example document that shows how you would use these icons in action?

5e-Cleric:
BTW, example document is prepared with all icons, just waiting for the deployment to work

calculuschild:
> We should still talk about icon size and other easy-to-solve matters

I think if we just match the existing height of our other icons as seen in the homepage that would be sufficient.

G-Ambatte:
Just removing the `font-size` declaration from the EI styling will get it pretty close, see below:

image

5e-Cleric:
I feel like that is too small to really see some icons, but users will change their size anyway, so

5e-Cleric:
To have them completely align, the following CSS will do the trick, and its the simplest i've found:

```
.page span.ei {
  font-size:inherit; (this one is not a thing, i'll simply remove the font-size declaration from the initial commit)
  translate:0 -1.3px;
}
```

image

The black dashed outline is there to show the actual box of the different elements.

calculuschild:
Can we not just have them reuse the same CSS as the font-awesome icons?

5e-Cleric:
> Can we not just have them reuse the same CSS as the font-awesome icons?
> 
> I'm not sure why we are making a special separate style just for these?

Because they don't use the same classes, we could use an `include` or `mixin` , or whatever function less uses, but small fixes are required anyway, because both fonts are not drawn to the same standard.

There are only 7 lines of code in common with Font awesome, is it that much of a problem?

calculuschild:
The 200 character declarations are fine. My question is why do we need to add `translate: abc` instead of just using the same positioning CSS as Font-Awesome? If we use the same `vertical-align`, `font-size`, `line-height`, `padding`, etc. do they not align naturally?

5e-Cleric:
> do they not align naturally?

Using the exact same CSS, there is a difference on vertical position of (from memory) ~1.5px, so they do not align.

This could theoretically be fixed in the font files, which i doubt anyone wants to do to get rid of a single line of CSS.

Gazook89:
I think it’s worth adjusting the font file to match the heights of font awesome. Partly because a 1.3px adjustment likely won’t translate well to different font sizes (what if using these icons in a large header?). But also it would give some experience in just doing that fix— I have to believe there is a method to do a bulk change in file height across the entire font.

5e-Cleric:
> I think it’s worth adjusting the font file to match the heights of font awesome. Partly because a 1.3px adjustment likely won’t translate well to different font sizes (what if using these icons in a large header?). But also it would give some experience in just doing that fix— I have to believe there is a method to do a bulk change in file height across the entire font.

Not with any program i own, but i'll look into it.

Gazook89:
It’s likely too much to ask, but since you already started a conversation with the original artist you might ask if she’d be willing to do it for us? Or, if she had knowledge about how to do it. Again, that might be pushing our luck, but it’s an idea.

5e-Cleric:
Eh bad news, i did make a version of the font lower, to match FA, and moving the glyph did NOT align the boxes, it does align the character, but the spans are still misaligned.

Dropping that as an option.

For future reference, best way to translate the glyps is Fonforge, select all characters, then go to element, transform, and input a big negative number, like -20 for 2 px.

Right now, without transforms, either glyph or css, they look like this:

image

calculuschild:
Tried my hand at repositioning the font. This seems to align the boxes just fine (attached).
[Elderberry-Inn-Icons.zip](https://github.com/naturalcrit/homebrewery/files/13891131/Elderberry-Inn-Icons.zip)

Each font as different metrics for where the baseline is, and where the top, bottom, etc. are positioned relative to it. That's where the discrepancy comes from. I just copied over the values from the FontAwesome into this one and resized accordingly.

image

5e-Cleric:
That's great! if you send me the file i can add it to the PR and it will be ready to go! Or you can add it yourself.

calculuschild:
File is linked in my comment up there.

5e-Cleric:
With the new version in local: 
image

calculuschild:
> With the new version in local: image

Looks (mostly) fine in the deployment. 

image

Also need to remove the `translate` property to get them to align fully.

5e-Cleric:
> Also need to remove the translate property to get them to align fully.

Oh, right, forgot that

calculuschild:
Gonna merge this now. Eventually we should support some kind of auto-complete for icons using `:icon:` syntax.

image

PR #3184, opened 2023-12-13
Allow Reflinks to be brew document global

dbolack-ab:
This isolates the Global reflink code from the AlternateVariables branch.


calculuschild:
I can't seem to get this to work. A simple document like this:

```
[hello]:url

[bello]

\page

[hello]

[bello]:anotherurl
```

renders both reflinks as plain text.

calculuschild:
Another thought, is that it may be simpler to just make this into an "extension" format instead of trying to overwrite the existing Tokenizer. Extensions always have priority so we would be more free to use our own logic and not have to battle to make this compatible with the built-in renderer stuff. I may check to see if that simplifies anything here.

dbolack-ab:
> Another thought, is that it may be simpler to just make this into an "extension" format instead of trying to overwrite the existing Tokenizer. Extensions always have priority if they consume the same text, so we would be more free to use our own logic and not have to battle with the existing renderer to make this compatible. I may check to see if that simplifies anything here.

I looked at this today. As best I can determine, moving this functionality to an extension only moves the upsert function and the tokenizer.reflink code out. The rest of the hoop jumping needed to have reflinks persist still has to occur. It just doesn't seem link much of a gain.

dbolack-ab:
> I can't seem to get this to work. A simple document like this:
> 
> ```
> [hello]:url
> 
> [bello]
> 
> \page
> 
> [hello]
> 
> [bello]:anotherurl
> ```
> 
> renders both reflinks as plain text.

Hmm. Seems I missed *something* when I pulled the old work out.  Though I see little reason to keep this PR around or fix it.

dbolack-ab:
Demo/Test document.

https://homebrewery-pr-3184.herokuapp.com/edit/XmcytDtU-tkf

dbolack-ab:
This should be redundant now that alternate-alternate-alternate variables are in palce.

PR #3185, opened 2023-12-13
Create Underline Token

dbolack-ab:
Override the poorly chose underscore as emphasis to it's one true purpose! Underline!

Seriously though.  Changes _Thing_ to be an underlined Thing. Regex may need tightening up.

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -239,9 +239,24 @@ const underline = {
 	level : 'inline',
 	start(src) { return src.match(/\w_[^_]_/m)?.index;},
 	tokenizer(src, tokens) {
-		const uRegex = /^_(?!\s)(?=([^_]*[^\s]))\1\^/m;
+		const uRegex = /^_([^_]*)_/m;
+	    const match = uRegex.exec(src);
+		console.log('Looking');
+		if(match?.length) {
+			console.log('found!');
+			console.log(match);
+			return {
+				type   : 'underline',
+				raw    : match[0],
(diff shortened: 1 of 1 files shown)


5e-Cleric:
Seems to be working just fine.

- [x] syntax creates an `<u></u>` element around the text
- [x] injected styles work on it

doesn't seem to be breaking anything

5e-Cleric:
Because of the minimalistic nature of this PR, i think it should be kept open or as draft as is until we can merge (v4). With as little maintenance as possible. I wish we could merge it now.

config/default.json
```diff
@@ -7,5 +7,6 @@
 	"local_environments" : ["docker", "local"],
 	"publicUrl" : "https://homebrewery.naturalcrit.com",
 	"hb_images" : null,
-	"hb_fonts" : null
+	"hb_fonts" : null,
+	"enable_v4" : true
```
5e-Cleric, line 11: Merging this straight away would enable v4 directly, can we keep the `enable_v4` as false in default.json and true in local.json? not sure if that is how it was made last time but makes sense to me.

5e-Cleric:
## Correct behavior

image

image

- [x] Works correctly inside 


## Bugs

image
A line break seems to make the function go crazy

image
Not leaving a space after the syntax breaks it

image
IF a second underscore is placed, the syntax breaks and the older `em` syntax comes through


This is from the live deployment, granted that is not from the last commit, not sure why the deployment doesn't automatically rebuild right now.

dbolack-ab:
I'm going to rebuild this as a module now that we're moving that way.

PR #3188, opened 2023-12-14
Add Highlight token to marked

dbolack-ab:
As I am already on a blasphemy streak, this is a PR to overload the tradition <strong> mapping for \_\_thing\_\_ to \<mark>thing\</mark>.


dbolack-ab:
A side note - if this is intended to only be used for a strikeout pattern then it is redundant vs the GHM `~` based pattern.It should be used differently or the markup indicators should be saved for another purpose.

Gazook89:
I'm kind of assuming there was discussion in Gitter about this back when you first opened it, but I couldn't find it with a quick search.  

I hesitate with this because I'm not sure I see the utility in being able to `<mark>` text, *beyond* just having one more html tag we can then style.  Personally, I just use the `<del>` tag that is easily available with `~~text~~` and then style it with css.  I do think that `mark` is likely a *more semantically appropriate* tag for this, and in another environment I would fully support having more tags in the toolchest...

...but:

> or the markup indicators should be saved for another purpose.

as you mention here, we have limited prime indicators available and thus we have to be very conservative.  I don't think `<mark>` is unique/useful enough to use a good indicator on.

image

dbolack-ab:
> as you mention here, we have limited prime indicators available and thus we have to be very conservative. I don't think `<mark>` is unique/useful enough to use a good indicator on.

I tend to agree, however, I have failed, utterly, to think of a single better use for the markup that also suits the pattern.

5e-Cleric:
At the end of the day, we can use this for another feature if we think of it for v4, i say we give this a go.

calculuschild:
Taking more inspiration from Obsidian, they use `==Content==` for highlighting. May be we could just follow that precedent again here and not have to worry about overriding underscores for now (plus, if we use `_content_` for underline later, `__content__` for highlighting might be similar enough to be confusing.)

dbolack-ab:
> Taking more inspiration from Obsidian, they use `==Content==` for highlighting. May be we could just follow that precedent again here and not have to worry about overriding underscores for now (plus, if we use `_content_` for underline later, `__content__` for highlighting might be similar enough to be confusing.)

Good thought.

PR #3191, opened 2023-12-16
Navigator Node Error Fix

5e-Cleric:
From Gitter chat: "The newest Node version added a "navigator" object -- Heroku must be grabbing that version automatically -- which conflicts with our check whether the page is on the server or in a browser"

This PR only changes that keyword in a specific check, fixing deployments.


calculuschild (approved):

calculuschild:
Thanks @5e-Cleric !

Issue #234, opened 2016-11-29
Feature Request: Table of contents


tullisar:
It would be really handy if we had a method for generating a table of contents for headers in the document, similar to how [Markdeep](https://casual-effects.com/markdeep/) handles them. I find myself wanting to jump to certain pages frequently, and scrolling through the doc or the source view for headers is a bit tedious.

dmmagic:
This would be a fantastic addition!

stolksdorf:
Kerblam, it's in v2.6.0 under editor snippets. It will auto generate based on `#`, `##`, and `###` level titles, and become hyperlinks to the pages where those titles show up. Let me know if this fits the bill. Enjoy :)

PR #3198, opened 2023-12-21
Experimental language translation

### LINKED ISSUE #2624: Modify Homebrewery strings to allow alternate languages
### Your idea:

Modifying Homebrewery to use a file containing all of the text strings presented to the end user will allow the creation of string files for alternate languages, and the site can be switched between them as desired by the user.

G-Ambatte:
This PR resolves #2624.

This PR adds a `translate` function to the String prototype, which when called searches a given YAML file using the string as key, returning the YAML file's value if found, and the original text if not.


G-Ambatte:
The current state of this PR is that only the Account menu has been modified to use the new `translate` function.

5e-Cleric:
I could put this syntax into all texts, should i or do you think we should wait in case we want another syntax?

G-Ambatte:
> I could put this syntax into all texts, should i or do you think we should wait in case we want another syntax?

I've kept it as a Draft PR for now, specifically so we could let the syntax discussion play out.
That said, it's been a couple of days, and I think the following ideas are worth following up on:

1. The template file key should match the literal string being translated.
2. A `translateDefaults` function, so that calls can be `.translate()` instead of `.translate(translateOpts)`.
3. React's `useContext` hook. It seems to be ideal for this feature, but may require some refactoring to make it work.
4. Use AccountPage to hold the UI elements for language selection.

On 1. As it currently stands, if the string to be translated does not exist in the localization file, then the string itself is returned, allowing as to use the exact text as the key. That said, it doesn't HAVE to be the exact text, and where the text is particularly long, it's probably a good idea to use an Identifier instead.
However, I have not put any thought into a process for identifying strings that should be referenced by ID instead of literal text. Maybe `string.length > 20` or some other arbitrary number?
This will effect the localization template, so I'd like to lock it down before we push too much further into the process; there may be work already completed that will be affected by this decision.

On 2. This seems like a good idea - as each JSX file deals with one area of the UI, we should be able to set the defaults once per file and have it "just work" from then on. Implementation, however, is usually where things fall apart.

For 3. I think we need `useContext` in order to switch the state of the entire UI, and that appears to be the exact purpose of the feature. That said, I have only a passing familiarity with `useContext`, so I need to go back and look at how to actually implement such a thing.

Finally, 4. Ultimately, no matter what page this is implemented on, a value needs to be manipulated and stored by the page, then retrieved and applied to the data source that the `translate()` function uses.
a. AccountPage: build the UI to manipulate the value
b. AccountPage: store the value
c. Translate(): retrieve the value
d. Translate(): apply the appropriate localization data

G-Ambatte:
I intend to take a further stab at some of these as time allows in the next couple of days; however, given the time of year, time is a fleeting thing right now...

ericscheid:
> 1. The template file key should match the literal string being translated.
> 2. A `translateDefaults` function, so that calls can be `.translate()` instead of `.translate(translateOpts)`.

> On 1. As it currently stands, if the string to be translated does not exist in the localization file, then the string itself is returned, allowing as to use the exact text as the key. That said, it doesn't HAVE to be the exact text, and where the text is particularly long, it's probably a good idea to use an Identifier instead.

It is entirely reasonable to have the source string in the code be "Error #234" and the translation be "An error occured (#234). This sometimes happens because [long reasons]. You can try [stuff]". The "identifier", per se, doesn't have to be a neologism, it can be a simple plain language string (one we'd be happy getting stuck as being permanent as a key only visible in the code .. so no rude words pls).

If we do go with using strings-in-the-code as the keys-in-the-translations, we definitely should have a code-to-English translation file (and not rely on the fallback-to-key method for English UI). Otherwise we can't easily change the English UI (because it would break the keys to other langs).

Shouldn't be hard to find a volunteer to write up the English translation file ;-)

ericscheid:
Quick check .. does this PR include setting the `<html lang="%UI-LANG%">` attribute?

As of v3.10, THB does not assert a `lang` on `<html>`. It does have Open Graph tags, of which the spec defaults to `en_US` for `og:locale` .. though that should be set from the brew language, not the UI language, so that's a separate PR.

5e-Cleric:
>  does this PR include setting the <html lang="%UI-LANG%"> attribute?

We could, not sure it is necessary, but it is best practice

5e-Cleric:
Last time we talked about this, we talked about what languages we could add as a start, we agreed on:

- EN (is there value in doing EN-US and EN-GB?)
- DE (German)
- ES (Spanish)
- PT-BR (Brazilian Portuguese
- CA (Catalan)
- VA (High Valyrian) @G-Ambatte we are doing it, right?

G-Ambatte:
>     * HV (High Valyrian) @G-Ambatte we are doing it, right?

_Engos_ is tongue, but if I remember correctly can also be used to mean "language"; while _arilla_ is "beer" but might be close enough to "brews".

This is why we need to finish a template file, to determine exactly what words we actually need.

5e-Cleric:
> This is why we need to finish a template file, to determine exactly what words we actually nee

[base-localization-file.zip](https://github.com/naturalcrit/homebrewery/files/13800968/base-localization-file.zip)
You mean this one still lacks strings?

5e-Cleric:
Right now the user page takes the following syntax:
```
Username's published brews
Username's unpublished brews
```

This is not only redundant, but difficult to translate, since the `'s` is conditional to the word, and added using another function.

We propose the following change:
```
Brews by Username

Published brews
Unpublished brews
```

locale/base-localization-file.yaml
```diff
@@ -0,0 +1,237 @@
...
+  brewsOnGoogleDrive: "Brews on Google Drive:"
+  noFiles: Unable to retrieve files
+  followSteps: follow these steps to renew your Google credentials.
+  Default Save Location:
+  Homebrewery:
+  Google Drive:
+  Language Selection:
+  langSub: This option selects your language exclusively for the user Interface, and not for your brews. You can change brew language in their properties tab.
+
+htmlError:
```
5e-Cleric, line 225: I am unsure about using this nomenclature

locale/base-localization-file.yaml
```diff
@@ -0,0 +1,237 @@
...
+  Account Information:
+  username: "Username:"
+  noUser: No user currently logged in
+  lastLogin: "Last Login:"
+  Homebrewery Information:
+  brewsOnHomebrewery: "Brews on Homebrewery:"
+  Google Information:
+  linkedToGoogle: "Linked to Google:"
+  brewsOnGoogleDrive: "Brews on Google Drive:"
+  noFiles: Unable to retrieve files
```
5e-Cleric, line 217: We should move colons to outisde the yaml, right?

5e-Cleric:
ToDo: 

1. [ ] Resolve comments
2. [ ] Snippet names
3. [ ] Add Languages

Numbered, because they should follow that order.

locale/base-localization-file.yaml
```diff
@@ -0,0 +1,311 @@
...
+  unkownError: "## An unknown error occurred! \n\n We aren't sure what happened, but our server wasn't able to find what you were looking for."
+  googleError: "## An error occurred while retrieving this brew from Google Drive! \n\n Google reported an error while attempting to retrieve a brew from this link."
+  gd404: "## We can't find this brew in Google Drive! \n\nThis file was saved on Google Drive, but this link doesn't work anymore."
+  gd4042: "Note that this brew belongs to the Homebrewery account"
+  gd4043: which is
+  gd4044: your account.
+  gd4045: not your account (you are currently signed in as
+  gd4046: and you are not currently signed in to any account.
+  gd404longText: |
+    The Homebrewery cannot delete files from Google Drive on its own, so there
```
5e-Cleric, line 233: I am unsure of THIS nomenclature

locale/base-localization-file.yaml
G-Ambatte: I'd say "yes", mostly because otherwise we need to remember to include them in every single localization.

locale/base-localization-file.yaml
ericscheid: I note that gd404* are simply successive strings which get stitched together with other values in the calling context. This suggests that maybe the `translate()` function should support passing in an object of keyed values which get substituted into the translated string.

Thus, the multiple `.translate()` calls here would instead become just one call, `"GDoc file not found error".translate({'account': props.brew.account});`, with all the magic happening in the function. 

This will also help when a translation needs to literally vary based on other variables e.g. plurality. Note that this will also easily handle cases where the variation is particular to the language and so the logic can't be done outside it (e.g. handling dual number plurality in Arabic).

locale/base-localization-file.yaml
5e-Cleric: Fixed

locale/base-localization-file.yaml
5e-Cleric: I now reckon we can push the PR like this and make that change later.

5e-Cleric:
@calculuschild there is no point on deploying if we can't log in to change the language.

calculuschild:
Also no harm to deploying.

In any case you can log in by copying the cookie from the live site over to the deployment.

5e-Cleric:
> Quick check .. does this PR include setting the `<html lang="%UI-LANG%">` attribute?
> 
> As of v3.10, THB does not assert a `lang` on `<html>`. It does have Open Graph tags, of which the spec defaults to `en_US` for `og:locale` .. though that should be set from the brew language, not the UI language, so that's a separate PR.

Fixed that, the lang attribute is now set in the html tag, following the official docs syntax.

calculuschild:
Getting an error trying this on the home page:

image
image

locale/locale-es-ES.yaml
```diff
@@ -0,0 +1,321 @@
...
+accountPage:
+  Account Information: Información de cuenta
+  username: Nombre de usuario
+  noUser: Actualmente no hay ningún usuario con sesión iniciada
+  lastLogin: Último inicio de sesión
+  Homebrewery Information: Información de Homebrewery
+  brewsOnHomebrewery: Brews en Homebrewery
+  Google Information: Información de Google
+  linkedToGoogle: Cuenta vinculada a Google
+  brewsOnGoogleDrive: Brews en Google Drive
```
Gazook89, line 212: This should just be `brewsOnDrive`, or the accountpage.jsx should have the corresponding part match this `brewsOnGoogleDrive`.  Right now it is not translating.

Gazook89:
I think in the spots where we use MomentJS to set relative times ("1 year ago"), such as in the metadata viewer in the navbar and in the Recent Items dropdown, we can use translation as well.  Check out this page from MomentJS docs: http://momentjs.com/docs/#/i18n/instance-locale/

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -167,12 +169,13 @@ const ListPage = createClass({
...
 
 	renderSortOptions : function(){
+		''.setTranslationDefaults(['userpage', 'filters']);
 		return <div className='sort-container'>
-			<h6>Sort by :</h6>
-			{this.renderSortOption('Title', 'alpha')}
-			{this.renderSortOption('Created Date', 'created')}
-			{this.renderSortOption('Updated Date', 'updated')}
-			{this.renderSortOption('Views', 'views')}
+			<h6>{'sortyBy'.translate()}:</h6>
```
Gazook89, line 174: this should just be `sortBy`   (not `sortyBy`).

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -167,12 +169,13 @@ const ListPage = createClass({
 	},
 
 	renderSortOptions : function(){
+		''.setTranslationDefaults(['userpage', 'filters']);
```
Gazook89, line 172: `userpage` --> `userPage`

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -167,12 +169,13 @@ const ListPage = createClass({
...
-			<h6>Sort by :</h6>
-			{this.renderSortOption('Title', 'alpha')}
-			{this.renderSortOption('Created Date', 'created')}
-			{this.renderSortOption('Updated Date', 'updated')}
-			{this.renderSortOption('Views', 'views')}
+			<h6>{'sortyBy'.translate()}:</h6>
+			{this.renderSortOption('title'.translate(), 'alpha')}
+			{this.renderSortOption('created date'.translate(), 'created')}
+			{this.renderSortOption('updated date'.translate(), 'updated')}
+			{this.renderSortOption('views'.translate(), 'views')}
```
Gazook89, line 178: Instead of translating each of these four options individually, you can instead place a `.translate()` on line ~125:

```js
	renderSortOption : function(sortTitle, sortValue){
		return <div className={`sort-option ${(this.state.sortType == sortValue ? 'active' : '')}`}>
			<button
				value={`${sortValue}`}
				onClick={this.state.sortType == sortValue ? this.handleSortDirChange : this.handleSortOptionChange}
			>
				{`${sortTitle.translate()}`}   //  <---------  add .translate() here
			</button>
			{this.state.sortType == sortValue &&
				<i className={`sortDir fas ${this.state.sortDir == 'asc' ? 'fa-sort-up' : 'fa-sort-down'}`}></i>
			}
		  </div>;
	},
```

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -207,17 +210,18 @@ const ListPage = createClass({
 
 	renderBrewCollection : function(brewCollection){
 		if(brewCollection == []) return <div className='brewCollection'>
-			<h1>No Brews</h1>
+			<h1>{'No Brews'.translate(['userPage'])}.</h1>
 		</div>;
 		return _.map(brewCollection, (brewGroup, idx)=>{
 			return <div key={idx} className={`brewCollection ${brewGroup.class ?? ''}`}>
-				<h1 className={brewGroup.visible ? 'active' : 'inactive'} onClick={()=>{this.toggleBrewCollectionState(brewGroup.class);}}>{brewGroup.title || 'No Title'}</h1>
+				<h1 className={brewGroup.visible ? 'active' : 'inactive'} onClick={()=>{this.toggleBrewCollectionState(brewGroup.class);}}>{brewGroup.title || 'No Title'.translate()}</h1>
```
Gazook89, line 217: Right now, this does not translate "Gazook89's Published Brews" / "Gazook89's Unpublished Brews".  This could be accomplished with a change on this line, and on the userPage.jsx.  This will change the `title` of the brewCollections, so someone wiser than I (@calculuschild , @G-Ambatte for example of a non-exhaustive list) might want to confirm this doesn't pose other dangers.  

But the gist of it is:

listPage line 217
```jsx
				<h1 className={brewGroup.visible ? 'active' : 'inactive'} onClick={()=>{this.toggleBrewCollectionState(brewGroup.class);}}>{`${brewGroup.title} ` + `${brewGroup.class} brews`.translate(['userPage']) || 'No Title'.translate(['userPage'])}</h1>

```

userPage line 34
```jsx
		const brewCollection = [
			{
				title : `${usernameWithS}`,
				class : 'published',
				brews : brews.published
			}
		];
```
result:
image


Probably there is some better way of concatenating the string together for the translate function, but right now I couldn't think of one.

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -79,15 +81,15 @@ const ListPage = createClass({
 	},
 
 	renderBrews : function(brews){
-		if(!brews || !brews.length) return <div className='noBrews'>No Brews.</div>;
+		if(!brews || !brews.length) return <div className='noBrews'>{'No Brews'.translate()}.</div>;
```
Gazook89, line 84: This should have `'No Brews'.translate(['userPage'])`.

I'm not leaving these comments in any particular order-- i realize now that I probably should have made all of the comments as part of a "Review", rather than some as part of a review and some as individual comments.  Sorry about that.

Gazook89 (changes requested):
A few comments in this review, and then a couple of separate comments because I didn't think i'd do as many as I did.

Also-- Do you have any ideas on how tooltips can be translated?  It seems right now that tooltips are done entirely with LESS/CSS, going all the way to tooltip.less.  Can `translate()` work in a LESS file?

client/homebrew/pages/basePages/listPage/listPage.jsx
ericscheid: >> Right now, this does not translate "Gazook89's Published Brews" / "Gazook89's Unpublished Brews".
>
> [...] Probably there is some better way of concatenating the string together for the translate function, but right now I couldn't think of one.

I suggest the `translate()` function include an optional parameter of key:values to be merged into the translated string. such that the brewCollection title could be translated via 

`"username-published-brews".translate({"name": "Gazook89"})`

The english translation reference YAML file could then have for the `username-published-brews` key this translated value:

    ${name}'s published brews

or even this, to properly handle English possessive formation for names ending with "s":

    ${name}${name.endsWith('s') || name.endsWith('S') ? '’' : '’s'} published brews

This can handle simple substitution (e.g. inserting a name into the string), weird substitutions (e.g. irregular possessive s formation), multiple values (e.g. errorNumber, date, reference merged into a longer paragraph), and even basic math etc (e.g. `${number * factor}` or string operations (e.g. "le" or "la", according to circumstance).

And, of course, handle straight up static strings (e.g.`"Hello!": "Bon jour!"`)

----

In the `translate()` function itself, it would dynamically define an anon function that returns the filled-template string (via the `with (merge)` statement to get the merge values into the scope).

```js
String.prototype.translate = function (values) {
// exempli gratia..
// also ignoring the `groups` parameter functionality requirement
  const translationData = {
    "Names Published Brews": "${name}${name.endsWith('s') || name.endsWith('S') ? '’' : '’s'} Brassins Publiés",
    "Names Unpublished Brews": "${name}${name.endsWith('s') || name.endsWith('S') ? '’' : '’s'} Brassins Non Publiés",
    "Some string": "Jibber jabba",
    // Add more translations as needed
  };

  const key = Object.keys(values)[0];

  if (translationData[key]) {
    // Dynamically define the function
    const functionString = `with (values) return \`${translationData[key]}\`;`;
    const translationFunction = new Function('name', functionString);

    // Call the dynamically defined function
    return translationFunction(values[key]);
  } else {
    return `Translation not found for key: ${key}`;
  }
};
// CODE NOT TESTED
```

NB. this would require a refactor of code to date, since the current `translate()` function already has one optional parameter which is an unlabelled array of strings for scoping the translation lookup group). I suggest instead 

`"source-string".translate( { values: {"key": "value", "another-key": "another value"}, groups: ["array", "of", "strings"] })`

(Writing a function with an optional unlabelled parameter value is a code smell. Pass that thing inside an object with a key, please).

client/homebrew/pages/basePages/listPage/listPage.jsx
5e-Cleric: i am unfamiliar with this specific code, would we keep the same key names in the yaml files?

client/homebrew/pages/basePages/listPage/listPage.jsx
5e-Cleric: OR, and just hear me out, simply change the way the userpage displays the titles, to have the username on top of the page, and then simply:

## Published brews

## Unpublished brews

and voilà, we just saved having to build that function at all.

5e-Cleric: This still needs discussion.

5e-Cleric:
I think i have fixed most comments and reviews, except for those which require discussion or i need an answer about something, if i'm wrong, please notify me about it.

client/homebrew/pages/basePages/listPage/listPage.jsx
ericscheid: Collection titles is not the only place where we want to present a translated message that contains various values merged into it.

Gazook89: I agree with both:  we can easily sidestep the issue on these particular headers without losing anything, but we may still benefit from the functionality that eric proposes.

client/homebrew/pages/basePages/listPage/listPage.jsx
Gazook89: Yeah, the yaml's will stay the same.  

Right now in the PR, the sort buttons are rendered in the `renderSortOption` function, and the actual text that displays on the buttons is done here:

```jsx
<button ...>
	{`${sortTitle}`}
</button>
```

`sortTitle` is an arg of `renderSortOption`, and is just a string.  In the PR, right now, the string that is being passed here into this arg is already translated.   But it doesn't have to be.  Instead, you can pass the original english string into `renderSortOption`.  Then, we can translate `sortTitle` (which is just the string we are passing in).

Just for completeness, here are the two entire functions as I describe them:
```jsx
	renderSortOption : function(sortTitle, sortValue){
		return <div className={`sort-option ${(this.state.sortType == sortValue ? 'active' : '')}`}>
			<button
				value={`${sortValue}`}
				onClick={this.state.sortType == sortValue ? this.handleSortDirChange : this.handleSortOptionChange}
			>
				{`${sortTitle.translate()}`}
			</button>
			{this.state.sortType == sortValue &&
				<i className={`sortDir fas ${this.state.sortDir == 'asc' ? 'fa-sort-up' : 'fa-sort-down'}`}></i>
			}
		  </div>;
	},

// ...

	renderSortOptions : function(){
		''.setTranslationDefaults(['userPage', 'filters']);
		return <div className='sort-container'>
			<h6>{'sortBy'.translate()}:</h6>
			{this.renderSortOption('title', 'alpha')}
			{this.renderSortOption('created date', 'created')}
			{this.renderSortOption('updated date', 'updated')}
			{this.renderSortOption('views', 'views')}
			{/* {this.renderSortOption('Latest', 'latest')} */}

			{this.renderFilterOption()}
		</div>;
	},
```

(*note, both of these have very similar function names....possibly that should be addressed but I'm not concerned about it now*)

5e-Cleric:
> A few comments in this review, and then a couple of separate comments because I didn't think i'd do as many as I did.
> 
> Also-- Do you have any ideas on how tooltips can be translated? It seems right now that tooltips are done entirely with LESS/CSS, going all the way to tooltip.less. Can `translate()` work in a LESS file?

If i understand the code correctly, we use an html attribute which is then referenced in the less, so we don't need to change anything in the less file.

client/homebrew/pages/basePages/listPage/listPage.jsx
calculuschild: I agree with Eric's proposal. Languages are not always going to follow the same grammatical sequence, so we need a way for each language to place passed-in values in the correct point in the text.

server/app.js
```diff
@@ -423,10 +433,21 @@ if(isLocalEnvironment){
 const templateFn = require('./../client/template.js');
 const renderPage = async (req, res)=>{
 	// Create configuration object
+	const languageOptions = ['pt-BR','en-US', 'es-ES', 'fr-FR'];
+	const langCodeCorrect =()=> {
+		if (languageOptions.indexOf(req.language) > -1) {return req.language;}
+		return 'en-US';
+	};
+
+	const langPreference = langCodeCorrect();
```
calculuschild, line 442: I would probably just simplify this to

```suggestion
	const langPreference = languageOptions.includes(req.language) ? req.language : 'en-US';
```

locale/localization.js
```diff
@@ -0,0 +1,31 @@
...
+					obj = obj[groups[0]];
+					groups.shift();
+					continue;
+				}
+				break;
+			}
+
+			if(obj[text] && obj[text].length > 0) return `${obj[text]}`;
+			return localeData[text]?.length > 0 ? `${localeData[text]}` : `${text}`;
+		};
```
calculuschild, line 20: I would suggest we add a parameter **interpolations** to the `translate` function which is an object containing key-value pairs of any text that need to be interpolated in the text.

'UsersPublishedBrews`.translate({username : this.props.username});`

Then in the YAML file, we can just add some kind of marker (I would vote the standard `${}`):

`UsersPublishedBrews : '${username}'s published brews'`
...
`UsersPublishedBrews : 'Brews publicados de ${username}'`

```suggestion
String.prototype.translate = function(_groups = String.prototype.getTranslationDefaults(), interpolations = {}) {
    const text = this.valueOf();
    const groups = _groups ? Array.from(_groups) : [];
    let obj = localeData;
    while (groups?.length > 0) {
        if(obj[groups[0]]) {
            obj = obj[groups[0]];
            groups.shift();
            continue;
        }
        break;
    }

    let translatedText = obj[text] && obj[text].length > 0 ? `${obj[text]}` : localeData[text]?.length > 0 ? `${localeData[text]}` : `${text}`;

    // Interpolate the values
    for (let key in interpolations) {
        translatedText = translatedText.replace(new RegExp(`\${${key}}`, 'g'), interpolations[key]);
    }

    return translatedText;
};
```

locale/localization.js
```diff
@@ -0,0 +1,31 @@
...
+
+		// Add defaults
+		String.prototype.getTranslationDefaults = function() {
```
calculuschild, line 23: I will just comment that I am not a huge fan of this `groups` parameter. It seems to add a layer of complexity that isn't really needed. Can I suggest we instead just handle the different translation groups via `group.key` notation?

`'nav.save'.translate()`

locale/localization.js
ericscheid: > Then in the YAML file, we can just add some kind of marker (I would vote the standard ${}):

Why partially emulate template strings when you could implement actual template strings?

```javascript
 if (translationData[key]) {
    // Dynamically define the function
    const functionString = `with (values) return \`${translationData[key]}\`;`;
    const translationFunction = new Function('name', functionString);

    // Call the dynamically defined function
    return translationFunction(values);
  } else {
    return `Translation not found for key: ${key}`;
  }
```

calculuschild: Not opposed to that, but I don't see what the added benefit would be given the increased complexity. (The proposed code uses deprecated and highly discouraged syntax as well, so we should avoid that if going this route.)

ericscheid: The proposal supports `` `${expression}` ``, which is more than simple substitution, and thus supports `` `${name.endsWith('s') ? name : name + 's'} Published Brews` `` (for example).

So .. odd grammar rules (like English possessive `'s`), or pluralisation rules, or even simply date formatting .. can all be supported via `` `${expression}` `` but not `` ${field}``.

The general concept of dynamically compiling a function from a string is usually discouraged because you really don't want to allow user-supplied strings being compiled. In this situation though only the strings in the translation files get compiled, and they should be getting reviewed when committed.

The other issue sometimes noted is that using `new Function` constructors can impair code optimisation and is less performant. Unless we want to hardcode a bunch of functions for use by `translate()`, we're gonna have to compile on the fly.

calculuschild: The points about `new Function` are valid and something to be careful with. But my main concern is actually the `with` operator, which is deprecated and should not be used.

calculuschild: I might suggest lodash's `_.template` function which can evaluate a string with expressions like this.

https://lodash.com/docs/#template

ericscheid: Ah ..  my understanding with the `with` operator is that the main discouragment has to do with potentially causing unintended variable declarations and making it difficult to determine where variables are defined. Last I heard it wasn't actually deprecated, per se.

Nonetheless, I support the recommendation of `_.template` .. it has additional checks too (but still uses `new Function` as the fundamental technique, heh).

calculuschild: So for reference, I believe the `_.template` could be used like this:

```
String.prototype.translate = function(_groups = String.prototype.getTranslationDefaults(), interpolations = {}) {
    const text = this.valueOf();
    const groups = _groups ? Array.from(_groups) : [];
    let obj = localeData;
    while (groups?.length > 0) {
        if(obj[groups[0]]) {
            obj = obj[groups[0]];
            groups.shift();
            continue;
        }
        break;
    }
    let translatedText = obj[text] && obj[text].length > 0 ? `${obj[text]}` : localeData[text]?.length > 0 ? `${localeData[text]}` : `${text}`;

    // Interpolate the values
    var compiledTemplate = _.template(translatedText);
    translatedText = compiledTemplate(interpolations);
    return translatedText;
};
```

dbolack-ab:
Has this been shelved for another path?

5e-Cleric:
Orphaned for 2 years, closing.

PR #3173, opened 2023-12-07
Alternate user brew variables based on further discussion

dbolack-ab:
I decided to run this as a parallel PR rather than stomp the previous work.

This PR is a little more in line with previous discussions.

## Overview

This PR implements five key features.

1. Converts the per-page markdown run variables list into a document persistent list.
2. Adds a resetBrewVars() function to markdown.js  to clear the document persistent variables list
3. Assign user brew variables based on the link pattern: $[variableName](Variable Value) 
4. Express user brew variables based on the reflink pattern: $[variableName]
5. Assign page-local user brew variables that do not persist across pages

Assignment extends the reflink object by adding a (currently unused) _formatting_ property that is primarily used to signal if the object is a traditional reflink or a user brew variable. The extended plan/expectation is to use this for number formatting rules. The variable value is stored in the object's title property. The assignment does not output.

Example Brew:

```
[a]:b
$[b](this is a ;)
:
$[b]
:
$[_pageonelocal](local variable from page one?)
$[_pageonelocal]
:
[a]
:
[doesntexist]

\page

[c]:This
[c]
:
$[m](n)
$[m]
:
[a]
:
$[_pageonelocal]
:
$[b]
:
[c]:That
[c]



\page

[a]:c
[a]
:
$[g](Goodbye)
$[g]

```

[AlternateVarsSample.pdf](https://github.com/naturalcrit/homebrewery/files/13602209/AlternateVarsSample.pdf)

PR #3199, opened 2023-12-21
Yet another Brew Variables PR

calculuschild:
Building on #3173, I think this is a pretty good syntax that does everything everyone wants.

Deployed to https://homebrewery-pr-3199.herokuapp.com/

### Notes:
- Imperative definition works (i.e. variables can hold different values at different points in the document). 
- Double interpolation is not possible (i.e., composing variable names out of other variables).
- Variable assignments inside of other variable assignments is not possible.
- Variables are hoisted if not yet defined when called. Uses the latest possible contents of that variable.

### Three syntaxes:
| syntax | description |
|:-------|-------------|
| `[var]:content`  | Assigns a variable |
| `[var](content)` | Assigns a variable and outputs it |
| `[var]`          | Outputs a variable |

### Examples:
| syntax | output |
|:--------|:-------|
|`[var]: Variable Contents`  | Assigns "Variable Contents" to `var` |
| [var]: Variable Contents<br>on two<br>or three<br>or more lines | Assigns "Variable Contents\non two\nor three\nor more lines" to `var` |
| [var]: <br>\| h1 \| h2 \|<br>\|----\|----\|<br>\| c1 \| c2 \| | Assigns "\| h1 \| h2 \|\n\|----\|----\|\n\| c1 \| c2 \| to `var`  (and can later be rendered as a table) |
| $[var](Variable Contents) | Assigns "Variable Contents" to `var` *and* outputs that value |
| `$[var](I love $[var2] and $[var3])` | Assigns "I love *var2 contents* and *var3 contents*" to `var` *and* outputs that value |
| `$[var]` | Outputs *var contents* |
| `[var]` | outputs the variable contents as a link, if formatted as a valid link |
| `![var]` | outputs as an image, if formatted as a valid image
| `$[var]: Variable Contents`<br>or<br>`![var]: Variable Contents` | Identical to `[var]: Variable Contents` . Links, images, and variables all share the same global list. The `!` or `$` prefix only matters when the variable is being expressed |
| `[var](Variable Contents)`<br>or<br>`var` | Identical to assignment via `$[var](Variable Contents)`, but outputs as a link or image if contents are a valid link format |
| `$[num1 + num2 * (num3 - round(5.5))]` | If `+`, `-`, `*`,`/`,`^`,`(`, or `)` are found in the variable name, the whole variable is parsed mathematically with correct order of operations. Only parses if every variable is defined (hoisting works), and evaluate to a number, and the expression is valid. Supports `round(x)`. ||

### Math
Math currently supports the following function names: `round()`, `floor()`, `ceil()`.
It also allows for `()` nesting operations.

So yes, you can do incrementing tables like this:

```
There are $[TableNum] tables in this document. // Final value of $[TableNum] gets hoisted up:
                                               //"There are 2 tables in this document."

$[TableNum]: 0 // Initialize to 0

$[TableNum]: $[TableNum + 1] // TableNum = 1

##### $[TableRefKnights](Table $[TableNum]: Horse Type and Quality)

...

$[TableNum]: $[TableNum + 1] // TableNum = 2

##### $[TableRefDragons](Table $[TableNum]: Dragons of the Realm)
```


ericscheid:
How would one assign the string "https://example.com/" to a variable and then have it rendered as (literally) `https://example.com/` ?

For an authoring system that produces print-destined output, being unable to _print_ a url sounds like a problem.

I don't want printed documents that read as "For more details, visit our website at PublisherWebsite."

ericscheid:
I'm a bit worried that double-interpolation might be setting us up for some recursion headaches.

5e-Cleric:
> How would one assign the string "https://example.com/" to a variable and then have it rendered as (literally) `https://example.com/` ?
> 
> For an authoring system that produces print-destined output, being unable to _print_ a url sounds like a problem.
> 
> I don't want printed documents that read as "For more details, visit our website at PublisherWebsite."

```
$[var]:https://example.com/ 
$[var]:(https://example.com/)

$[var]
```
first line is just assignment, second is assingment and output, third is just output
This worked just fine to output what you ask for in the deployment.

calculuschild:
> How would one assign the string "https://example.com/" to a variable and then have it rendered as (literally) `https://example.com/` ?

Just use `$[var]:(https://example.com/)`

calculuschild:
> I'm a bit worried that double-interpolation might be setting us up for some recursion headaches.

I'm not sure what you mean by double-interpolation.

Recursion would be handled by tracking a list of the variables you are currently inside and stopp processing if a nested variable is in that list.

ericscheid:
> I'm not sure what you mean by double-interpolation.

Interpolation is replacing variable references in a string with the value of those variables.

Double interpolation is replacing variable references in a string with the value of the variable named in those variables.

```
var foo = "some string"; // a string assigned to foo
var bar = `template with ${foo}`; // → "template with some string" (single interpolation)
var qaz = "foo"; // the string "foo" assigned to qaz
var qax = `template ${${qaz}}`; // the qaz variable is double interpolated, ultimately inserting the value of $foo
```

(I've also worked with systems which supported stuff like `Set Field By Name[ variable_of_field_name, value_to_assign]`, but that's just more madness).

calculuschild:
> Double interpolation is replacing variable references in a string with the value of the variable named in those variables.

Ah ok. This PR isn't doing any double-interpolation by that definition so that shouldn't be a concern.

5e-Cleric:
Found something, inputing `$[var]:(Dungeons & Dragons)` outputs `dungeons & dragons`, losing the caps.

5e-Cleric:
`$[var]:()` will still break the brew completely.

calculuschild:
> `$[var]:()` will still break the brew completely.

Should be fixed. Now has same behavior as `$[var]:` where it just isn't parsed as a variable until it has some content, so empty `()` is not recognized.

ericscheid:
> > `$[var]:()` will still break the brew completely.
> 
> Should be fixed. Now has same behavior as `$[var]:` where it just isn't parsed as a variable until it has some content, so empty `()` is not recognized.

Does this mean that `$[var]` can't be set to be empty (e.g. `$[title]`) ?

calculuschild:
> Does this mean that `$[var]` can't be set to be empty (e.g. `$[title]`) ?

Currently yes. The regex expects both the var name `$[var]` and a value to assign. If both are not present it isn't recognized.

Could be changed easily though if we really want `''` variables.

5e-Cleric:
| syntax                                                           | assignment | output|
|:-----------------------------------------------------------------|:----------:|:-----:|
| `$[var]: C`                                                      | YES        | NO    |
|  $[var]: C<br>on two<br>or three<br>or more lines                | YES        | NO    |
|  [var]: <br>\| h1 \| h2 \|<br>\|----\|----\|<br>\| c1 \| c2 \|   | YES        | NO    |
| `$[var]:(C)`                                                     | YES        | YES   |
| `$[var]:(I love $[var2] and $[var3])`                            | YES        | YES   |
| `$[var]`                                                         | NO         | YES   |
| `[var]`                                                          | NO         | YES   |
| `![var]`                                                         | NO         | YES   |
| `[var]: C`<br>or<br>`![var]: C`                                  | YES        | NO    |
| `[var]:(C)`<br>or<br>`![var]:(C)`                                | YES        | YES   |
| `[var](C)`<br>or<br>`var`                                  | NO         | YES   |
| `$[num1 + num2 * num3 - 5.5]`                                    | NO         | YES   |

C="Variable Contents"

If i have made any mistake tell me so i can correct the table

5e-Cleric:
Right now, `$[var]:google.com` gets rendered as text, even though `[var](google.com)` gets rendered as link
Also happens when assigning and outputing, looking for https is not the way to go.

calculuschild:
> * [ ]  `$[var]:google.com` gets rendered as text, even when assigned and then output

What output are you expecting, and what did you try? To output this as a link, simply use [var]. See https://homebrewery-pr-3199.herokuapp.com/edit/w59eQSNc-119

> * [ ]  Definition syntax breaks variables, try `$[var]:whatever::something`

This expected behavior, and is because Definition Lists are "block-level". They work fine if called at a block level, but block-level tokens can't be rendered inline. See https://homebrewery-pr-3199.herokuapp.com/edit/YhTHoIqQrQV8

You would have the same problem if you try to create a table in the middle of a paragraph.

> * [ ]  Assignment still undercases variable contents, breaking image links such as `https://i.imgur.com/IwHRrbF.jpg`

I don't have this issue. Caps are retained in all cases. See: https://homebrewery-pr-3199.herokuapp.com/edit/a6eIg5HjzWIo

calculuschild:
> If i have made any mistake tell me so i can correct the table

This is all correct.

5e-Cleric:
image
This is still an issue.

5e-Cleric:
### Bugs to fix
- [ ] `[]` should always emit a link, even if generated from math.
- [ ] Invalid links such as `[link]:(invalid link contents)` should render as plain text, rather than `undefined` as text
- [ ] If the contents of a var link start and/or end with double quotes `"`, they must be parsed as `%22` so they work like markdown links


### Syntax highlighting

image

ericscheid:
> $[divided]:$[CHA - 10]

What are you expecting `#[divided]` to be assigned here? The value of CHA minus 10, or the literal string "$[CHA - 10]"?

If it is assigning the literal string, then of course `($[divided / 2])` is going to produce an error (because a string can't be divided).

(IIRC, we chose to avoid recursive string interpolation to maintain sanity)

ericscheid:
>  Invalid links such as [link]:(invalid link contents) should render as plain text, rather than undefined as text

What qualifies as an invalid link content though?

calculuschild:
> What qualifies as an invalid link content though?

Link formatting is defined according to the commonmark spec: https://github.github.com/gfm/#link-reference-definitions

When attempting to render a variable as a link, it should fail if the variable contents are not in this format.

calculuschild:
@5e-Cleric Fixed the math and the images getting lowercased. If it's easier to keep track, feel free to edit my original post or one of yours and put all the issues into the same checklist.

calculuschild:
> What are you expecting `#[divided]` to be assigned here? The value of CHA minus 10, or the literal string "$[CHA - 10]"?

`$[divided]` should be assigned the value of 4. Then `$[divided / 2]` should be 2. This is now fixed.

> (IIRC, we chose to avoid recursive string interpolation to maintain sanity)

Correct.

calculuschild:
@ericscheid 5e-Cleric has proposed a simplification to the current syntax:

Replace:

- `$[var]: C` assign
- `$[var]` lookup, then output
- `$[var]:(C)` assign, then output
- `$[var](C)` output

with

- `$[var]: C` assign
- `$[var]` lookup, then output
- `$[var](C)` assign, then output 

i.e., remove the `[var]:(C)` syntax and "upgrade" the old standard image/link syntax to also assign the value to a variable.

So `var` outputs an image like normal (and doesn't break any old brews), but now you can _also_ reuse the image later with just `![var]`.

5e-Cleric:
This math is good and all, but can't really be used for D&D calculations, without:

- Rounding numbers up or down
- multiple calculation syntax: `(x-10)/2`

calculuschild:
> This math is good and all, but can't really be used for D&D calculations, without:
> 
> * Rounding numbers up or down
> * multiple calculation syntax: `(x-10)/2`

`()` and `round(x)` are now supported.

> Other issues, i know `[var]:something::something` doesn't work, but shouldn't `something::$[var]` work?

Both of these now work.

5e-Cleric:
image

### Am i missing something?

calculuschild:
> image
> ### Am i missing something?

This has been fixed now.

5e-Cleric:
image

Also, variables are being rendered inside codeblocks, which is not desired.

calculuschild:
> image According to Calc, 2nd and third are wrong, 2nd should emit 9 and third should emit `$[round($[foo])]`
> Also, variables are being rendered inside codeblocks, which is not desired.

All of this should be fixed now. Expected results:

image

shared/naturalcrit/markdown.js
```diff
@@ -4,7 +4,40 @@ const Marked = require('marked');
...
+		add         : true,
+		subtract    : true,
+		multiply    : true,
+		divide      : true,
+		power       : true,
+		round       : true,
+		floor       : true,
+		ceil        : true,
+
+		sin   : false, cos    : false, tan    : false, asin : false, acos  : false,
```
5e-Cleric, line 24: in the future, we might want to add this as toggle logic for users to change

PR #3202, opened 2023-12-23
Improve Theme selector styling

Gazook89:
This is a small effort to improve the styling of the theme selector.

Current:
image

Proposed:
image


5e-Cleric:
Do we need to add yet another shade of grey? why not use the same shade as in the selected tab?

Gazook89:
> Do we need to add yet another shade of grey? why not use the same shade as in the selected tab?

Agreed and fixed.

calculuschild:
Thanks @Gazook89 ! ⭐ 🎉 💯

PR #3203, opened 2023-12-23
fonts as snippets

5e-Cleric:
Don't check the CSS there are some junky things so i don't have to write javascript.

but it works, its simple, and won't need more work later on.

image


G-Ambatte:
My first thought is that it can be difficult to read the names of fonts which do not have glyphs that approximate characters (like Rellanic, Iokharic, or Davek, and any future similar fonts - such as Infernal, if I ever finish that font). That said, being able to see what the font actually looks like is also important.

Perhaps adding a `title` property so that the name appears in a standardized font if you hover over them? I assume that would work for the navItems.

5e-Cleric:
Lets be honest here, if you know the name of the font you already know which one is it. But it is not so bad as an idea. I could link a title attribute to `snippet.name` for all of them.

G-Ambatte:
snippetbar.jsx, lines 227-232:
```
renderSnippets : function(snippets){
		return _.map(snippets, (snippet)=>{
			const excludedSnippetProperties = ['name', 'icon', 'gen', 'experimental', 'subsnippets'];
			return <div {..._.omit(snippet, excludedSnippetProperties)} className='snippet' key={snippet.name} onClick={(e)=>this.handleSnippetClick(e, snippet)}>
				<i className={snippet.icon} />
				<span className='name'>{snippet.name}</span>
```

This change will add any properties of the snippet object to the output element, as long as they are not in the list of excluded elements - then you need only add the `title` property to each snippet object to get this:

image

G-Ambatte:
I'm getting a duplicate key error in the console, looks like Mr Eaves is in `snippets.jsx` twice

image

calculuschild:
> My first thought is that it can be difficult to read the names of fonts which do not have glyphs that approximate characters (like Rellanic, Iokharic, or Davek, and any future similar fonts - such as Infernal, if I ever finish that font). That said, being able to see what the font actually looks like is also important.
> 
> Perhaps adding a `title` property so that the name appears in a standardized font if you hover over them? I assume that would work for the navItems - I expect that would require a small change to the navItem itself, rather than just the snippets.

What about doing something similar to the existing "Rune Table" snippets? Have the name in normal font but 1 (or 3) characters in the target font?

5e-Cleric:
> What about doing something similar to the existing "Rune Table" snippets? Have the name in normal font but 1 (or 3) characters in the target font?

That would mean adding an additional span, and breaking the syntax into normal snippets and font snippets, i am against this, after all, it just takes a few seconds to see if you inserted the right one and if not, delete it.

calculuschild:
> Actually, we already do that with the rune table:

Lol. Hence me saying 'What about doing something similar to the existing "Rune Table" snippets?'

Seems to me like the obvious solution since the work has already been done.

5e-Cleric:
> Seems to me like the obvious solution since the work has already been done.

Right, workflow would be to create an icon out of a character of the font, then display it using css like we do there.

Nothing wrong with that, but i don't see it being anything short of confusing with easy-to-read fonts such as open sans:

image

If you insist, i will, but looks silly to me.

calculuschild:
I was mainly just referring to the fantasy fonts.

Gazook89:
Consistency is key. Like the rune tables, we should have the display of a single character in each font for their snippets. It takes the place of the icon that other snippets have. 

Further, the font name should just be in the regular menu font (making the same character the single example character). Unless we make that menu two-column, with regular font paired with sample text as is done in other editors (like Affinity). 

And, I doing think we need rune tables at all if we are including the fonts this way.

G-Ambatte:
> Consistency is key. Like the rune tables, we should have the display of a single character in each font for their snippets. It takes the place of the icon that other snippets have.

> Unless we make that menu two-column, with regular font paired with sample text

It would make sense to keep the style the same (or at least similar) throughout the snippet menus. Even if we do make the font selection a special case, it should be consistent throughout that menu.
I was thinking that FontAwesome uses the `i` element's class to insert content to the `before` pseudo-element; we could probably do something similar by tagging the font snippets with the `font` class and adding styling to add 'ABC' as the `before` content. Shift the `font-family` styling from the `span` to the `i` and it should give a short display to indicate what the font looks like.

> And, I don’t think we need rune tables at all if we are including the fonts this way.

Rune tables exist in the PHB, so I think we should continue to include them, if only as a Theme snippet. However, I also think that's a separate issue, so can have it's own PR, to be approved or rejected on it's own merits.

5e-Cleric:
> It would make sense to keep the style the same (or at least similar) throughout the snippet menus. Even if we do make the font selection a special case, it should be consistent throughout that menu. I was thinking that FontAwesome uses the `i` element's class to insert content to the `before` pseudo-element; we could probably do something similar by tagging the font snippets with the `font` class and adding styling to add 'ABC' as the `before` content. Shift the `font-family` styling from the `span` to the `i` and it should give a short display to indicate what the font looks like.

That is a really good idea, i can have it ready for tomorrow.


> And, I don’t think we need rune tables at all if we are including the fonts this way.

I completely disagree with this, rune tables are a somewhat complex syntax to display very specific stuff, getting rid of those makes no sense, it would be like getting rid of the `note`, or `decriptive` snippets just because they are easy to write. Snippets are there because they are common blocks that you may need for a book. Admitedly, a script table is not as common, but it belongs there.

calculuschild:
So another thought on this.  Is it better to just have each font as a classname?

- `{{Davek ...`,

or to use natural CSS?

- `{{font-family:Davek ...`

Considering `runeTable` uses `font-family`.


> It would make sense to keep the style the same (or at least similar) throughout the snippet menus. Even if we do make the font selection a special case, it should be consistent throughout that menu. I was thinking that FontAwesome uses the `i` element's class to insert content to the `before` pseudo-element; we could probably do something similar by tagging the font snippets with the `font` class and adding styling to add 'ABC' as the `before` content. Shift the `font-family` styling from the `span` to the `i` and it should give a short display to indicate what the font looks like.

It looks like now instead of typing 'ABC' and applying the font, we have an icon for each. Either way is fine with me as long as we are consistent. Any strong opinions or is the dropdown menu ok now?

5e-Cleric:
> So another thought on this. Is it better to just have each font as a classname? or to use natural CSS?

Same thing to me, whatever you like better.

The dropdown menu is ok as is, probably better than the first draft.

calculuschild:
Ok, so I would say just change this to `font-family` (then we don't need a bunch of classes) and I'll merge this.

PR #3204, opened 2023-12-23
quick fix page size override

5e-Cleric:
A specificity conflict was forcing pages to the letter size, from this commit: [Add default page size to brewrenderer.less](https://github.com/naturalcrit/homebrewery/commit/bf30cadb68f1edaac323c43e681c2da96d2ea292)

This PR lowers specificity for that selector using [`:where`](https://developer.mozilla.org/en-US/docs/Web/CSS/:where).


calculuschild:
Thanks @5e-Cleric !

PR #3067, opened 2023-09-28
Add syntax highlighting for definition lists

Gazook89:
This PR was prompted by @ericscheid on gitter:

> How easy would it be to add some styling to the brew code for definition-term:: definition?
> I'm thinking maybe two tones of colour?

### Result:
image

### Editor HTML structure:
```html
<span role="presentation" style="padding-right: 0.1px;">
	<span class=" define term">Lorem </span>
	<span class=" define">::</span>
	<span class=" define definition"> Veniam incididunt veniam eu. Ad in qui nisi reprehenderit do mollit tempor. Culpa et anim deserunt eu sint nisi qui. Tempor in proident excepteur nisi dolore adipisicing nostrud consequat irure aute culpa. Nulla velit sint deserunt excepteur esse.</span>
</span>
```

PR #3207, opened 2023-12-24
Definition List syntax highlighting fix

Gazook89:
This continues the unresolved issue on PR #3067.  For some reason I recently thought that was fixed, but as @ericscheid points out, it isn't.

I think this PR finally fixes it.

image


client/homebrew/editor/editor.jsx
```diff
@@ -151,7 +151,7 @@ const Editor = createClass({
 
 						// definition lists
 						if(line.includes('::')){
-							const regex = /^([^\n]*?)::([^\n]*)(?:\n|$)/ym;
+							const regex = /^([^\n]*?\s)::(\s[^\n]*)(?:\n|$)/ym;
```
ericscheid, line 154: This regex requires the `::` to have white space on each side, and so something like `term::defn` won't match.

What you want is to not match any line that starts with a colon.
`/^([^:][^\n]*?)::([^\n]*)(?:\n|$)/ym;`

Gazook89: Is it the worst thing to require spaces on each side? To me that is just cleaner. I could do both too

ericscheid: It would be inconsistent with what is recognised by the parser as a DT/DD, which means the `::` won't get highlighted even when it is successfully parsed as a DT/DD. The highlighting there is particularly helpful since the text just runs along, so the lack of highlighting is egregious.

Hmm .. suggested regex still isn't right, since `:term :: defn` is parsed as a dt/dd. The thing we don't want to match is a run of colons with no other text.

Gazook89: Okay fair enough.  I have the regex now set so that it doesn't require spaces on either side of the colons.  It also checks for a colon at the start of a line (and prevents highlighting if one exists).  The regex can be seen below, and I just pushed the commit.

Another thing in this commit:  I changed how the "range" used in highlighting is derived, though in this commit I only changed one line and before going ahead with a merge I could apply the same change to other similar "range checks" for highlighting.  

Below is the code with my change, what was just committed, though you can see in the other two lines the original method.

```js
						// definition lists
						if(line.includes('::')){
							const regex = /^([^\n:]+?)::([^\n]*)(?:\n|$)/ymd;
							let match;
							while ((match = regex.exec(line)) != null){
								codeMirror.markText({ line: lineNumber, ch: line.indexOf(match[0]) }, { line: lineNumber, ch: line.indexOf(match[0]) + match[0].length }, { className: 'define' });
								codeMirror.markText({ line: lineNumber, ch: line.indexOf(match[1]) }, { line: lineNumber, ch: line.indexOf(match[1]) + match[1].length }, { className: 'term' });
								codeMirror.markText({ line: lineNumber, ch: match.indices[2][0] }, { line: lineNumber, ch: match.indices[2][1] }, { className: 'definition' });
							}
						}
```

I noticed that there was some odd highlighting behavior, and how it checked where the `span`s were applied in the editor (the spans differentiate between "term" and "definition" and apply "define" to everything *not* included in the other spans).  This was evident when your line was just `cat dog :: ` and no definition typed yet.  It would take the ` ` (space) at the end as the "definition" match (`match[2]`) and then when we want to actually apply the class name for the highlight, it would find the ` ` between `cat dog` because we just did `indexOf` on the string again.  It would apply `definition` class to that first space.

With my change, I used the `d` flag in the regex, which makes the `exec()` command return the indices of the match groups back as an array of arrays.  I then just use those indices to set the range, or bounds, of the `span`s.  

It's likely this same technique could be applied to the Injection syntax highlighting, and inline spans?  Though I'd be satisfied to just apply it to definition lists for now.

That `d` tag was neat, I didn't know about it until now.  It may be helpful with all the variables/reflinks talk.

Gazook89:
Here's a curveball-- are we even sure we want the background highlight of the colons?  I'm right now leaning towards no, we don't.  Couple of reasons:

1. No other syntax that I can think of has a background-color property, so it seems weird to use it for this.
2. Choosing one background-color is tough now that we have editor themes.  We could live with one unchanging color that looks fine in some themes, is livable in others, and looks bad in many...  

Or we could apply standard Codemirror class names to each part of the DL (term, colons, definition).  Those standard class names are generally present in every Codemirror theme, so the theme authors have decided on a color for each that works well as part of their palette.

For example, the HTML output:

image


And with the Hopscotch theme, it looks like this:

image

And the Ayu-Dark theme, without changing anything but the theme:

image

Note, that most (all?) the themes have very few spans styled with background-color, with the exception being colors for highlighted text and errors. 

One small concern I have is applying classnames that are not reflective of what the element actually is....In my example above you can see I used `cm-keyword` and `cm-def`, which aren't exactly the correct uses but it's what I got.  There are other options like `cm-variable`, which are maybe more appropriate, I don't know.

If this is a direction we want to go in, setting syntax colors based on theme rather than "globally", we could do this same application of standard class names to everything else we do custom highlighting of (curly syntax, `\page`, etc).

5e-Cleric:
Best case scenario would be if there were extra unused classes for every theme already styled so we can simply use them.

I assume that is not the case, and creating so seems pretty... complex.

The most used imo should be Darkbrewery, which is merged but not live, i would create a palette for that(maybe even rethink all those colors) and make a color for dark themes and light themes.

It's not the best, i might have to take a deeper look.

Gazook89:
I'm going slow my roll a little bit and just get this PR "Done".  Any discussion about how managing editor themes and custom syntax can happen elsewhere (either a PR that I might submit, or a Discussion in the meantime).

Right now this PR just fixes the DL highlighting.  I went ahead and used the regex `d` flag for match indices and reference those indices in the codemirror marking functions.  

_Is this something that we want unit tests for?  If so, I can *try* that, but my first attmept didn't get off the ground.  It seems silly to suggest the tests should go into another PR, but that's my suggestion as it seems that some of the highlighting functions needs to be refactored a bit to make that easier.  I could be wrong._

dbolack-ab:
Might be redundant with https://github.com/naturalcrit/homebrewery/pull/3129.

5e-Cleric:
> it also checks for a colon at the start of a line (and prevents highlighting if one exists). 

Wont this crash with #3129?

Gazook89:
I'm going to close this since @dbolack-ab has it covered in another PR.

PR #3208, opened 2023-12-25
Working Demonstration of "Next Stage Themes" Discussion topic.

dbolack-ab:
This starts demonstration work on my suggestion for the next iteration of themes with an eye on moving towards user-created sharable themes and decoupling the "official" ( not the best word ) themes from the codebase.

This pulls out the Journal, 5ePHB, and 5eDMG themes, leaving only the Blank theme in place. Theme versioning primitives are in place but cannot toggle or store versions for brews.

The code as presented runs as a replacement V3 system, though if used, this is a V4 system. I did not want to add another layer of complication to the work for something that may not move forward.

This work is aimed at Section one of my [discussion ](https://github.com/naturalcrit/homebrewery/discussions/3142) and implements I and III. II will require general consensus and be necessary to REALLY complete I ( moving fonts and assets around between themes and the "blank" theme ). IV and V require a bit more thinking and general investigation to complete.


dbolack-ab:
Item 1.iv is now implemented.

dbolack-ab:
At this point I feel like the User Brew themes are a better overall path.

PR #3218, opened 2023-12-30
Page background texture

Gazook89:
Floating this as an option for texturing the page background.  Rather than use just a 'flat' image as a background image, we could use an image with transparency so that the `background-color` property of the page can show through.  Then users could change that color and more easily change the color of the page as a whole.  This can be done with other textured items like the monster stat block, both the main background and the bars at top and bottom.

## Options

We could take the current background file and just set `background-blend-mode: multiply` and get a pretty good result.  The issue is that this current image file isn't really black and white-- there is some color range there, and it actually looks pretty good with the rough color range that we use, and adds some vibrancy to it.  But that same vibrancy, *in my opinion*, might not look great when you change the base color to other colors.  But maybe not so bad:

image

(with tan #EEE5CE as background-color)

image

(with blue #a5d9f2 as background-color)

Or we could actually take our current background image file and drain the color from it and replace it with transparency, and then do the same thing with `background-blend-mode: multiply`.  This sort of 'flattens' the whole thing, losing a little bit of depth but also looks fine.

image

(with tan #EEE5CE as background-color)

image

(with blue #a5d9f2 as background-color)

Either way, there is a good range of colors this works with (both light and dark colors).


5e-Cleric:
Posting here some backgrounds i did manually in photoshop 2 years ago, trying some stuff, looks relevant:

redscroll
yellowscroll
bluescroll
seascroll
turquoisescroll
lgbtscroll

This is to say, i think this feature needs much more work than simply adding a blending mode and a color.

Gazook89:
With blending mode you could (likely) get all those colors, including the rainbow using a linear gradient. Besides color, they all have the same texture right?

I don’t mind going into lots more options for backgrounds, but this PR is hopefully about maximizing options with one line of css and possibly replacing one file.  In fact, if reusing the same texture for the monster stat block background, reducing the file count by one.

Gazook89:
People will always be able to do their own css and link to their own images. But we shouldn’t worry about including a bunch of alternate images in the build if we can avoid it.

5e-Cleric:
> you could (likely) get all those colors

Maybe with weird combinations yes, but for some i had to make several changes to contrast, exposure, brightness, changes that are very hard to make using the filter function from css.

We can try, and it will work to some degree.

Gazook89:
>  filter function from css.

Don't want to use the filter property on `.page` because that will impact everything, rather than just the background.  So yes, getting those backgrounds might be tough.  Those backgrounds also I think have greater range between the lightest highlights and the darkest shadows, so everything is naturally more pronounced than what can be done with the current default texture (i could be wrong about the range though).  

If the other textures should be included, I think they should fit into new Themes as part of their asset folder, *if* the strictly black and white (or rather, black and transparent) version of their texture can't just be used once and color applied with a blend mode and background color.

5e-Cleric:
If you want i can create more and add them as assets as part of this draft, or apart from it.

Gazook89:
> If you want i can create more and add them as assets as part of this draft, or apart from it.

Not on this PR.  The focus with this PR is to do more with fewer (or the same) assets.  If more backgrounds are needed, I think it's worth considering taking two approaches:

- Add full color backgrounds to individual themes (so using the red paper texture would mean using the red background you have above)
- Adding black & transparent backgrounds that can be used with any color background via background blend mode. 

And if a user wants to use the red background you have above but without using the entire "Red" theme, they can continue to make their own custom theme deviations as usual by changing the background-image property.

5e-Cleric:
It is fine, i guess the main issue is that colors as they are inputed, have already a lightness or brightness value assigned to them, and depending of what blending mode we use, that can be a problem.

So i would try with `hue` as a blending mode, does that _**really**_ only take the hue?

5e-Cleric:
Alright, here is a test that should cover most of it:

https://homebrewery.naturalcrit.com/share/dxGPbliAoNLS

Gazook89:
> Alright, here is a test that should cover most of it:

Just note that I don't think it's possible to choose whether the color is above or below the background if using within the same element like this:

```css
.page {
	background-color: blue;
	background-image: url(' ... ');
	background-blend-mode: darken;
}
```

Noting it because I'm not talking about some additional div element that is getting covered.  This would be a change to the `.page` element itself.

5e-Cleric:
> > Alright, here is a test that should cover most of it:
> 
> Just note that I don't think it's possible to choose whether the color is above or below the background if using within the same element like this:
> 
> ```css
> .page {
> 	background-color: blue;
> 	background-image: url(' ... ');
> 	background-blend-mode: darken;
> }
> ```
> 
> Noting it because I'm not talking about some additional div element that is getting covered. This would be a change to the `.page` element itself.

Well the color would be first in every case i believe.

5e-Cleric:
This might be useful https://stackoverflow.com/questions/43284984/mimic-photoshop-rgb-levels-with-css-and-svg-filter

5e-Cleric:
> Noting it because I'm not talking about some additional div element that is getting covered. This would be a change to the `.page` element itself.

We could use a `:before` pseudo element of the page element, and use filters on that element, that should break just a few brews.

5e-Cleric:
> Or we could actually take our current background image file and drain the color from it and replace it with transparency, and then do the same thing with background-blend-mode: multiply. This sort of 'flattens' the whole thing, losing a little bit of depth but also looks fine.

Here are three variants with different contrasts, in case you wanna use them.

scroll bw 1
scroll bw 2
scroll bw 3


This is all using basically the levels filter in Adobe Photoshop, after some careful tweaking to get the black and white through a camera raw filter.

The levels filter allows you to see what general colors are in the image and give you control over the exact output: 
image

This grayscale variant is in out assets folder, we could use it:
parchmentBackgroundGrayscale

Gazook89:
We would take those grayscale images at replace the white with transparency (or said another way, reduce the opacity of the white).

Gazook89:
Something like that. The texture I offered as an example in my original post is only black, with varying levels of alpha. Of course, the screenshot in my comment has it overlayed on white so that it is visible. But if you go into my example brew you can see that it’s just a black textured layer, on top of a colored background. 

I used affinity and the channel mixer to create a mask. Honestly I’m not sure I remember exact steps but I think I wrote it down.

calculuschild:
Note that we already do this for monster stat blocks, and so we already have a black & white parchment texture, "parchmentBackgroundGrayscale.jpg".

Gazook89:
> Note that we already do this for monster stat blocks, and so we already have a black & white parchment texture, "parchmentBackgroundGrayscale.jpg".

This is actually different-- `parchmentBackgroundGrayscale.jpg` is black & white.   I am talking about images that are black & *transparent*.

calculuschild:
> 2\. But it lacks an artistic punch because it loses depth of color palette, and a good theme should have its own custom page background beyond just _texture over color_.

Just for fun, people could do something like add a series of gradients with different colors and then put the texture on top of that:

image

image

Gazook89:
I'm going to close this as I'm not sure it addresses an Issue, and the PR itself has minimal changes, and the discussion is a little overcharged.

PR #3252, opened 2024-01-17
Direct PDF Export

dbolack-ab:
This is an attempt to integrate jsPDF into HB for direct export.

It very much doesn't work right now for reasons I can't quite sort out other than to assume it is CSS issues that are either inobvious or incompatible with html2canvas.

Mostly here for bus insurance, etc.


dbolack-ab:
Closing to declutter.

PR #3254, opened 2024-01-18
Refactor ToC Generation

dbolack-ab:
This restructures the existing ToC generation to provide three options in preparation Widget based activities.
    
This changes the existing menu item to a menu parent with subsnippets for generating ToC that use
     - Headers # to ###
     - Headers # to ####
     - Headers # to #####
    
Current styling is set up to emulate the existing styles but probably should be examined further.


themes/V3/5ePHB/snippets.js
```diff
@@ -21,9 +21,26 @@ module.exports = [
 		view      : 'text',
 		snippets  : [
 			{
-				name : 'Table of Contents',
-				icon : 'fas fa-book',
-				gen  : TableOfContentsGen
+				name        : 'Table of Contents',
```
5e-Cleric, line 24: we might want to change those icons to something like `h1-h4`

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -27,7 +27,31 @@ const getTOC = (pages)=>{
...
+		if(!_.last(res)) add1(null, page);
+		if(!_.last(_.last(res).children)) add2(null, page);
+		if(!_.last(_.last(_.last(res).children))) add3(null, page);
+		_.last(_.last(_.last(res).children).children).children.push({
+			title    : title,
+			page     : page + 1,
+			children : []
+		});
+	};
+
```
5e-Cleric, line 40: this looks very readable, thinking if we can refactor this to make it more clear

5e-Cleric (approved):
Overall looks fine, left some non-helpful comments

themes/V3/5ePHB/snippets/tableOfContents.gen.js
dbolack-ab: What needs clarification?

5e-Cleric: I just think code should be readable to a point, and even knowing JS i have no idea what that code does, maybe add a simple comment about it?

themes/V3/5ePHB/snippets.js
5e-Cleric: I was thinking in the lines of this:
image

But those are pro icons, and we don't have access to that. I guess it is fine as it is.

ericscheid:
Now that we've sorted out the bugs from PR #3261 .. could we use just the one method for determining if a heading is included or not to avoid confusing authors?

We should also add a snippet to the Style Editor snippet menu for "Include H4, H5 in ToC".

Something like this
```
/* Include H4 and H5 headings in Table of Contents */
/* This only takes effect when the ToC is generated in the brew editor */
h4, h5 { --TOC: include; }`
```

dbolack-ab:
Closing as sorta redundant.

Issue #2826, opened 2023-05-11
Table of Contents doesn't align with CSS Counter Resets


Gazook89:
### Your idea:

If a user does a CSS Counter Reset on their page numbers, to start their page count on page 5 for example, any Table of Contents generated by HB won't take that reset into account and give the wrong numbers.  This is because the TOC generator is just counting how many page elements there are, and doesn't look at the value of the CSS Counter.

From what I can tell, there is no method of getting the value of CSS counter via javascript.  So I think it is likely that any fix for this would mean replacing the CSS Counters with a pure JS implementation.

Such a change might open more possibilities, but certainly would break existing brews, and thus likely needs to be a v4 thing.

ericscheid:
```js
      const computedStyle = window.getComputedStyle(theDiv);
      const counterValue = computedStyle.counter('my-counter');
```

Gazook89:
I'm probably missing something, and still playing with it, but here is what I get with:

```js
const computedStyle = window.getComputedStyle(theDiv, ':after');
console.log(computedStyle);
```
<details>
<summary>Output:  (expand for log)</summary>
```
CSS2Properties(368) { "accent-color" → "auto", "align-content" → "normal", "align-items" → "normal", "align-self" → "auto", "animation-delay" → "0s", "animation-direction" → "normal", "animation-duration" → "0s", "animation-fill-mode" → "none", "animation-iteration-count" → "1", "animation-name" → "none", … }
​
"-moz-animation": "0s ease 0s 1 normal none running none"
​
"-moz-animation-delay": "0s"
​
"-moz-animation-direction": "normal"
​
"-moz-animation-duration": "0s"
​
"-moz-animation-fill-mode": "none"
​
"-moz-animation-iteration-count": "1"
​
"-moz-animation-name": "none"
​
"-moz-animation-play-state": "running"
​
"-moz-animation-timing-function": "ease"
​
"-moz-appearance": "none"
​
"-moz-backface-visibility": "visible"
​
"-moz-border-end": "0px none rgb(201, 173, 106)"
​
"-moz-border-end-color": "rgb(201, 173, 106)"
​
"-moz-border-end-style": "none"
​
"-moz-border-end-width": "0px"
​
"-moz-border-image": "none 100% / 1 / 0 stretch"
​
"-moz-border-start": "0px none rgb(201, 173, 106)"
​
"-moz-border-start-color": "rgb(201, 173, 106)"
​
"-moz-border-start-style": "none"
​
[...]

PR #3258, opened 2024-01-19
A system for book sections enabling section labelling and distinct page numbering

dbolack-ab:
This is one take on how to handle the notion of a brew having multiple, distinctly numbered sections in a way that both the CSS and snippets can track the page number resetting.

This system works by adding a single word parameter to the \page token. This value is added to the <page> class list. This allows both a predefined list and the users to create their own.

These classes *should* hold no CSS properties and exist to act as triggers to reset the page counter ( phb-page-numbers - which should probably be renamed ) and reset the Section Label and the numbering type for the page counter (Roman numerals, decimal, etc  - based on the LI numbering types)

This system uses .(sectionname):nth-of-type(1) for counter resets and .(sectionname) to override two global CSS variables where (sectionname) current defaults to one of index, appendix, chapter, frontmatter, and index.

Current design weaknesses that aren't addressed 
  - There is no way to track multiple indexes or appendixes and label accordingly ( Appendix A, pg 7 ) 
  - Assumes each section type is used once 
  - Does not enforce section order
  - Behaves the same as the present scheme if the first line of the brew does not set a section style
  - Still requires the user to insert the {{pageNumber }} bug.
 
Strengths:
  - SHOULD work on both the CSS side of the document and the snippets side by binding the counter resets to a token
  - Requires nothing of users not wanting to use it.
  - Easy extension via themes 
  - Allows user extension via the brew styles tab 

Solves:

Issue #2826


Gazook89:
The deployment doesn't seem to work at this point.  It doesn't load the stylesheet and some other error.  Crashes if I type in the editor as well.

I haven't taken a look at the code yet so don't have much more to say, just a heads up.

dbolack-ab:
> The deployment doesn't seem to work at this point. It doesn't load the stylesheet and some other error. Crashes if I type in the editor as well.
> 
> I haven't taken a look at the code yet so don't have much more to say, just a heads up.

I haven't updated this one in a bit, I'm not surprised it has some issues. I'm going to move it back to draft until I can catch it back up.

PR #3259, opened 2024-01-19
New page in new tab

5e-Cleric:
Added `newTab={true}` to the new page link, as per user request via discord.

While this does not suppose any decrement in user steps, and they could just middle-click to get that, it does make sense that the new page button opened a new tab instead of reloading the one you're on.

If this gets approved, i'll merge it right away.


calculuschild (approved):
Looks good to me

PR #3260, opened 2024-01-20
Proof of concept solution for Issue 2294

dbolack-ab:
The current ToC snippet design scans the pages for Header tags and builds a list.  Unfortunately, this includes inside of blocks where the headers are used as labels that have no value to a table of contents and should not be collected.

This fix creates a new flag mustache modifier class that can be used by snippets and is ignored for rendering.

This demonstration adds four flags. !toc, !h1, !h2, and !h3. 

- !toc: Exclude this Mustache Div entirely from Table of contents generation
- !h1: Exclude `#` tokens from Table of contents generation
- !h2: Exclude `##` tokens from Table of contents generation
- !h3: Exclude '###` tokens from Table of contents generation

This could be similarly accomplished by setting behavior based on known classes ( monster, etc. ) but this makes it difficult for users to extend.


dbolack-ab:
The other solution is Waaaay better. Closing.

PR #3261, opened 2024-01-20
Table of Contents Snippet exclusion system

dbolack-ab:
Based on @ericscheid's suggestion.

This modifies the Table of Contents snippet to ignore headers that have an active CSS variable `--TOC` with a value of exclude.  To facilitate backward compatibility, CSS classes have been added or altered to have this value set along with the `monster` class in the 5ePHB theme.  Additional classes have been added for toggling on and off exclusions.

To Be Determined:

Which of this CSS work should relocate to the Blank class.


ericscheid:
This js works:

```js
 var iframe = document.getElementById('BrewRenderer');
 var iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
 var headings = iframeDocument.querySelectorAll('h1, h2, h3, h4, h5, h6');

 headings.forEach(function (heading) {
   let text = heading.innerText;
   let page = heading.closest('.page,.phb').id;
   let excluded = getComputedStyle(heading).getPropertyValue("--TOC");
   console.log({page, text, excluded});
   // add to the snippet output 
   }
```

calculuschild:
Getting an error trying this on the home page:

image
image

calculuschild:
@ericscheid Is the current behavior acceptable to you? Default exclusions with manual inclusion and exclusion tags?

themes/V3/5ePHB/style.less
```diff
@@ -796,6 +796,28 @@
...
+.monster,
+.noToC,
+.toc { --TOC: exclude; }
+
+// Manual Inclusion classes
+
+.addToC,
+.addh1, .addh2, .addh3, .addh4, .addh5, .addh6, .addh1h2 h1, .addh1h2 h2,
+.addh1h3 h1, .addh1h3 h2, .addh1h3 h3,  .addh1h4 h1, .addh1h4 h2, .addh1h4 h3, .addh1h4 h4, 
+.addh1h5 h1, .addh1h5 h2, .addh1h5 h3, .addh1h5 h4, .addh1h5 h5 { --TOC: include; }
```
calculuschild, line 814: I'm not clear what the purpose is of the `.addh1hX` classes? Can you elaborate?

themes/V3/5ePHB/style.less
```diff
@@ -796,6 +796,28 @@
...
+.page:has(.insideCover), 
+.page:has(.partCover), 
+.monster,
+.noToC,
+.toc { --TOC: exclude; }
+
+// Manual Inclusion classes
+
+.addToC,
+.addh1, .addh2, .addh3, .addh4, .addh5, .addh6, .addh1h2 h1, .addh1h2 h2,
```
calculuschild, line 812: Why so many classes here that all do the same thing?

ericscheid:
> Getting an error trying this on the home page:

1. Firstly, why is there a heading element inside a `<div class="page">` where the latter doesn't have an `id` attribute? What use case is causing this error??

2. amend to use optional chaining .. and a `null` will be fed to `parseInt` instead of throwing an access error.
   `   let onPage = parseInt(heading.closest('.page,.phb').id?.replace(/^p/,′);`

3. amend the line a bit further where `onPage` is checked to instead test against `NaN` (which `parseInt(null)` produces)
   `     if(!onPage.isNaN(myValue)) {`

ericscheid:
> @ericscheid, @5e-Cleric , @Gazook89 , @G-Ambatte Is the current behavior acceptable to you? Default exclusions with manual inclusion and exclusion tags via CSS? If the behavior is agreed upon I will just check over for code quality and bugs.

The snippet js starts with the assumption of `h1,h2,h3,h4,h5,h6` (currently just `h1..h4`) and then the general exclusion rules are applied. The `h5` and `h6` have been added to the current ToC snippet, and then excluded in the css, so as to support any author that might want to later include those levels of headings.

The use of exclusion rules is short and simple to modify and covers the desired exclusions (h3 and h4 inside .monster, various cover pages, h5, h6). There's an argument for also excluding `.monster h2` or simply `.monster` to mimic the base PHB theme. I would lean towards current THB compatibility than militant PHB mimicry, mostly because it's easier to understand what needs to change in the css exclusion rules (unlike if one needed to modify `.monster {--TOC: exclude}` to instead only exclude `h3` and `h4` but to also now include `h2`).

The philosophy of using inclusion tags (i.e. assert via manually editing brew.text in all the curly-blocks) is fraught with tedium. There is a usecase for a generic `.addToC {--TOC: include}` rule, both for the occasional usage for exceptions, but also to demonstrate what value other than `exclude` can be used. The rule would need to have stronger specificity though, or use `!important`.

themes/V3/5ePHB/style.less
dbolack-ab: Individual and group toggles, per Header.

I don't think they work like I had in my head. I'm gonna slim this down and assume we will document how to build custom styles to work as desired.

dbolack-ab:
> 1. Firstly, why is there a heading element inside a `<div class="page">` where the latter doesn't have an `id` attribute? What use case is causing this error??

My guess would be `heading.closest('.page,.phb')` - that the 5ePHB theme isn't being used.

dbolack-ab:
> The use of exclusion rules is short and simple to modify and covers the desired exclusions (h3 and h4 inside .monster, various cover pages, h5, h6). There's an argument for also excluding `.monster h2` or simply `.monster` to mimic the base PHB theme. I would lean towards current THB compatibility than militant PHB mimicry, mostly because it's easier to understand what needs to change in the css exclusion rules (unlike if one needed to modify `.monster {--TOC: exclude}` to instead only exclude `h3` and `h4` but to also now include `h2`).

This particular implementation has no inclusions for .monster based on the described problem.  Should this be revised to include h1 and h2 as present occurs?  I think since this impacts a snippet it is far more permissible to behave differently.

dbolack-ab:
I am lost on what fixes are needed at this time.

themes/V3/5ePHB/style.less
```diff
@@ -795,6 +795,16 @@
 // *****************************
 // *      TABLE OF CONTENTS
 // *****************************/
+
+// Default Exclusions
+.page:has(.frontCover), 
+.page:has(.backCover), 
+.page:has(.insideCover), 
+.page:has(.partCover), 
+.monster,
```
calculuschild, line 804: Similar issue here. The `--ToC` tag is applied to `.monster`, but the generator only checks the header elements, so this still appears in the TOC.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -28,31 +28,33 @@ const getTOC = (pages)=>{
...
-					add1(title, pageNum);
+	const iframe = document.getElementById('BrewRenderer');
+	const iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
+	const headings = iframeDocument.querySelectorAll('h1, h2, h3');
+
+	_.each(headings, (heading)=>{
+		const onPage = parseInt(heading.closest('.page,.phb').id?.replace(/^p/, ''));
+		const ToCExclude = getComputedStyle(heading).getPropertyValue('--TOC');
+		if(ToCExclude != '"exclude"') {
+			if(!isNaN(onPage)) {
```
calculuschild, line 39: Page numbers appear to be off-by-1 (items on page 2 are labeled page 3, etc.)

themes/V3/5ePHB/style.less
```diff
@@ -795,6 +795,16 @@
 // *****************************
 // *      TABLE OF CONTENTS
 // *****************************/
+
+// Default Exclusions
+.page:has(.frontCover), 
+.page:has(.backCover), 
+.page:has(.insideCover), 
+.page:has(.partCover),
```
calculuschild, line 803: These have the `--ToC` tag, but it doesn't seem to properly exclude them; they still appear in the table of contents. I suspect this is because the tag is being applied to the `.page`, but the table of contents generator only looks for the tag on the header elements.

calculuschild (changes requested):
The CSS is applied, but the exclusion doesn't appear to be happening.

themes/V3/5ePHB/style.less
ericscheid: A reminder here for future readers that CSS is used here in a funky way -- instead of applying in realtime to control styling of elements, it is instead first applied  during authoring, then the toc-generation snippet is run, which produces the toc brewtext. Further, the toc-generation snippet isn't looking at style rules directly applied to heading elements, it is checking to see if they are affected by them (by virtue of the cascade of CSS). This is done via `getComputedStyle(heading).getPropertyValue('--TOC');`

(btw, this also means the CSS can be altered after TOC generation and it won't be applied to the previously generated TOC — the current styling CSS could appear to be lying.)

That said .. if the emperical test of running the snippet is that the exclusions are not being applied .. then something is indeed broken.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
ericscheid: Are you looking at physical page numbers, or logical page numbers?

calculuschild: The first page of the document is page 1. Clicking the links in the TOC takes you to the page after.

For example, if I have

`# Chapter 1` on page 2, the TOC will say it is page 3, even though I dont have a page 3.

themes/V3/5ePHB/style.less
calculuschild: From what I can see, the CSS is not being inherited down to children elements.

ericscheid: I see otherwise .. right-click the heading in the console and choose Store as Global Variable, and then in console do `getComputedStyle(temp1).getPropertyValue('--TOC')`.

The property is computed.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
ericscheid: I'm seeing something different, but also wrong.

My document is

1. front cover
2. table of contents
3. some page 3
4. another page 4

but .. the TOC looks like this:

```
- ### [{{ GATE THUNDER}}{{ 2}}](#p2)
  - #### [{{ Shadowy Depths}}{{ 2}}](#p2)
- ### [{{ Table Of Contents is on Physical Page 2}}{{ 3}}](#p3)
- ### [{{ Physical Page 3 Heading 1}}{{ 4}}](#p4)
  - #### [{{ Heading 2}}{{ 4}}](#p4)
  - #### [{{ Heading 2}}{{ 4}}](#p4)
    - [{{ a very long heading that might wrap very poorly}}{{ 4}}](#p4)
- ### [{{ Physical Page 4}}{{ 5}}](#p5)
  - #### [{{ Sparklemuffin Peacock Spider}}{{ 5}}](#p5)
- ### [{{ Physical Page 5}}{{ 6}}](#p6)
  - #### [{{ Heading 1}}{{ 6}}](#p6)
  - #### [{{ Heading 2}}{{ 6}}](#p6)
  - #### [{{ Heading 2}}{{ 6}}](#p6)
- ### [{{ Physical Page 6}}{{ 7}}](#p7)
  - #### [{{ Heading 1}}{{ 7}}](#p7)
  - #### [{{ Heading 2}}{{ 7}}](#p7)
  - #### [{{ Heading 2}}{{ 7}}](#p7)
```

Here, each _displayed_ page number does correspond to the _linked_ page number .. however the page _number_ value in both cases is off by 1.

calculuschild: This is the same thing I'm seeing. The generated link and displayed number are equal, but they are both wrong by 1.

ericscheid: The `onPage = parseInt(heading.closest('.page,.phb').id?.replace(/^p/, ''))` step is working correctly (tested via Store as Global Variable in console.).

That value is passed in to functions `add1`, `add2`, `add3`.

Looking at those functions reveals the problem:
```
	const add1 = (title, page)=>{
		res.push({
			title    : title,
			page     : page + 1, // BUG IS HERE
			children : []
		});
	};
```

I believe this is a hold over from when toc entries were found by searching through raw brew text that had been split into pages into a zero-indexed array. The page number is now being calculated from the `id="pNNN"` of the nearest `<div class="page">` element.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -28,31 +28,33 @@ const getTOC = (pages)=>{
...
-					const title = line.replace('# ', '');
-					add1(title, pageNum);
+	const iframe = document.getElementById('BrewRenderer');
+	const iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
+	const headings = iframeDocument.querySelectorAll('h1, h2, h3');
+
+	_.each(headings, (heading)=>{
+		const onPage = parseInt(heading.closest('.page,.phb').id?.replace(/^p/, ''));
+		const ToCExclude = getComputedStyle(heading).getPropertyValue('--TOC');
+		if(ToCExclude != '"exclude"') {
```
ericscheid, line 38: This line is testing the wrong value. The value of `ToCExclude` would be `exclude`, not `"exclude"`.

Remove the double quotes.

themes/V3/5ePHB/style.less
ericscheid: The reason the toc-snippet is failing to exclude is because line 38 has `if(ToCExclude != '"exclude"') {` .. which is testing the `getComputedStyle` with `"exclude"` .. but the stylesheet is using `exclude` (no quotes).

themes/V3/5ePHB/style.less
ericscheid: The `--TOC` property *is* cascading to child elements of `< div class="monster"` ... however the test on line 38 is checking for the wrong value (it is testing against `"exclude"`, it should test against `exclude`).

dbolacksn:
@ericscheid fixes applied.

ericscheid:
> @ericscheid fixes applied.

Tested - numbering is now correct, exclusions are excluded, and curly-injected inclusions are included.

💯

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -22,37 +22,39 @@ const getTOC = (pages)=>{
...
-	_.each(pages, (page, pageNum)=>{
-		if(!page.includes("{{frontCover}}") && !page.includes("{{insideCover}}") && !page.includes("{{partCover}}") && !page.includes("{{backCover}}")) {
-			const lines = page.split('\n');
-			_.each(lines, (line)=>{
-				if(_.startsWith(line, '# ')){
-					const title = line.replace('# ', '');
-					add1(title, pageNum);
+	const iframe = document.getElementById('BrewRenderer');
+	const iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
+	const headings = iframeDocument.querySelectorAll('h1, h2, h3');
```
ericscheid, line 33: This PR has a working method for excluding various headings. Currently, this snippet also always excludes the H4 and H5 headings by skipping them entirely. PR #3254 proposes an alternative method. Having both methods can only lead to confusion and complicate diagnosing author problems.

Recommend we change this to `.querySelectorAll('h1, h2, h3, h4, h5')`  and add `h4, h5, ` to the list of default exclusions in the css. If an author does want h4 or h5 headings in their TOC, they can add overrides to the `brew.style`. 

We should also add a snippet to the Styles snippet menu for "Include H4, H5 in Table of Contents".

themes/V3/5ePHB/style.less
```diff
@@ -789,6 +789,16 @@
...
 // *****************************/
+
+// Default Exclusions
+.page:has(.frontCover), 
+.page:has(.backCover), 
+.page:has(.insideCover), 
+.page:has(.partCover), 
+.monster,
+.noToC,
+.toc { --TOC: exclude; }
```
ericscheid, line 800: Add `h4, h5, ` to the default exclusions, to be compatible with current behaviour.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -22,37 +22,39 @@ const getTOC = (pages)=>{
...
-				if(_.startsWith(line, '## ')){
-					const title = line.replace('## ', '');
-					add2(title, pageNum);
+				if(heading.tagName == 'H2') {
+					add2(headingText, onPage);
 				}
-				if(_.startsWith(line, '### ')){
-					const title = line.replace('### ', '');
-					add3(title, pageNum);
+				if(heading.tagName == 'H3') {
```
ericscheid, line 47: We need cases for calls to `add4` and `add5` here.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -22,37 +22,39 @@ const getTOC = (pages)=>{
 		if(!_.last(_.last(res).children)) add2(null, page);
 		_.last(_.last(res).children).children.push({
 			title    : title,
-			page     : page + 1,
+			page     : page,
 			children : []
 		});
 	};
```
ericscheid, line 28: We need functions for `add4` and `add5` here.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
dbolack-ab: I added h4, h5, and h6 here. I seem to have missed when in this timeline we added H6, but there you go.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
dbolack-ab: I added h4, h5, and h6 here. I seem to have missed when in this timeline we added H6, but there you go.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
dbolack-ab: I added h4, h5, and h6 here. I seem to have missed when in this timeline we added H6, but there you go.

The other PR has a replacement for this section of code that is recursive - though I need to test it still. Will resolve that out in #3254.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
ericscheid: Oh, I completely forgot html headings actually go to 6 levels.

themes/V3/5ePHB/style.less
```diff
@@ -831,6 +831,9 @@ h6,
 			text-decoration : none;
 			&:hover { text-decoration : underline; }
 		}
+		h1, h2, h3, h4, h5, h6 {
+			--TOC: exclude !important;
```
ericscheid, line 835: Uncomfortable with this `!important`.

The higher specificity of `.toc h1` (etc) should be sufficient for the `exclude` to apply.

dbolack-ab: Can certainly drop it if its unneeded

dbolack-ab: Dropped

themes/V3/5ePHB/snippets.js
```diff
@@ -21,9 +21,34 @@ module.exports = [
...
+						icon : 'fas fa-dice-five',
+						gen  : dedent `{{tocH5
+							}}`,
+					},
+					{
+						name : 'Include H4-H6',
+						icon : 'fas fa-dice-six',
+						gen  : dedent `{{tocH6
+							}}`,
+					},
```
calculuschild, line 50: I figured it out eventually, but it was not obvious to me how these are meant to be used. Users will probably need more help. My thoughts along these lines:

1) We should probably include the `addToC` class to this list of snippets.
2) Adding `toH6` (or any of these) directly to a default-excluded element (monster stat block, for example) is a little weird, because it will include *only* the h4-h6 headers from that element. Perhaps these snippets by default should also behave like `addToC` (or just directly include `addToC` in the snippet), because if you are adding H4-H6, you probably want to include H1-H3 too.
3) Given 3, it might be more clear to users if we rename these:

> Include block in Table of Contents
> `{{addToC }}`

> Include up to H4.
> `{{addToC,tocH4 }}`

> Include up to H5.
> `{{addToC,tocH5 }}`

> Include up to H6.
> `{{addToC,tocH6 }}`

Not sure... Thoughts? @dbolack-ab @ericscheid

calculuschild (changes requested):

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -1,19 +1,19 @@
 const _ = require('lodash');
 const dedent = require('dedent-tabs').default;
 
-const getTOC = (pages)=>{
+const getTOC = ()=>{
```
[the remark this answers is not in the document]
dbolack-ab: My expectation was that it would be committed before this which is why I didn't copy it over. I can, though

themes/V3/5ePHB/snippets.js
dbolack-ab: Honestly, I had a hard time thinking about how to label these and label them concisely.  For the classes, the notion was that the # in tocH# was the lowest included header level.

These look good to me. @ericscheid ?

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -2,77 +2,88 @@ const _ = require('lodash');
...
+		console.log(curDepth);
+		if(curDepth > 5) return; // Something went wrong.
+		if(curDepth == targetDepth) {
+			child.push({
+				title    : title,
+				page     : page + 1,
+				children : []
+			});
+		} else {
+			console.log(typeof child);
```
calculuschild, line 16: remember to get rid of the `console.log()`s when this PR is done.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -2,77 +2,88 @@ const _ = require('lodash');
...
+			console.log(_.last(child));
+			console.log(_.last(child).children);
+			recursiveAdd(title, page, targetDepth, _.last(child).children, curDepth+1,);
+		}
 	};
 
 	const res = [];
+
 	_.each(pages, (page, pageNum)=>{
 		if(!page.includes("{{frontCover}}") && !page.includes("{{insideCover}}") && !page.includes("{{partCover}}") && !page.includes("{{backCover}}")) {
```
calculuschild, line 31: These checks for all the different covers are redundant if we already have the CSS `--exclude` for them below, right?

calculuschild:
I think the actual CSS exclusion logic has disappeared somehow. We have the CSS classes, but they no longer affect the ToC generation.

themes/V3/5ePHB/style.less
```diff
@@ -786,6 +786,31 @@
...
+h4,
+h5,
+h6,
+.page:has(.frontCover), 
+.page:has(.backCover), 
+.page:has(.insideCover), 
+.monster,
+.noToC,
+.toc { --TOC: exclude; }
+
```
Gazook89, line 800: you have the label `default exclusions`, so likely best to also label the `manual inclusions` in comment.

themes/V3/5ePHB/snippets.js
```diff
@@ -21,9 +21,41 @@ module.exports = [
 		view      : 'text',
 		snippets  : [
 			{
-				name : 'Table of Contents',
-				icon : 'fas fa-book',
-				gen  : TableOfContentsGen
+				name        : 'Table of Contents',
+				icon        : 'fas fa-book',
+				gen         : TableOfContentsGen,
```
Gazook89, line 26: add `experimental : true` to the top level item, i think.  For the "Beta" tag.  I don't know if it makes much sense to put it on every sub item in the menu, since it's all very much linked together.

themes/V3/5ePHB/snippets.js
```diff
@@ -21,9 +21,43 @@ module.exports = [
...
+				subsnippets  : [
+					{
+						name         : 'Table of Contents',
+						icon         : 'fas fa-book',
+						gen          : TableOfContentsGen,
+						experimental : true
+					},
+					{
+						name : 'Include in ToC up to H3',
+						icon : 'fas fa-dice-six',
```
calculuschild, line 37:
```suggestion
						icon : 'fas fa-dice-three',
```

calculuschild:
Two more things to tweak:

- PartCover is still including the H2 header. The only thing on a partCover page that should appear in the ToC is the H1 header: {Part X)
- In general, H3's should appear in the ToC with no header:
-
```
# Test h1

## Test h2

### Testy test h3
```
becomes:
```
- ### [{{ Test h1}}{{ 7}}](#p7)
  - #### [{{ Test h2}}{{ 7}}](#p7)
    - [{{ Testy test h3}}{{ 7}}](#p7)
```

This PR has changed H3's to appear as  `    - #### [{{ Testy test h3}}{{ 8}}](#p8)`

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -2,77 +2,66 @@ const _ = require('lodash');
...
+		if(ToCExclude != 'exclude') {
+			recursiveAdd(heading.innerText.trim(), onPage, headerDepth.indexOf(heading.tagName), res);
 		}
 	});
 	return res;
 };
 
+
+const ToCIterate = (entries, curDepth=0)=>{
+	const levelPad = ['- ###', '  - ####', '    - ####', '      - ####', '        - ####', '          - ####'];
```
calculuschild, line 44: After The first two levels, any other nested ToC levels should have no `#` attached (displayed as plain text)

```suggestion
	const levelPad = ['- ###', '  - ####', '    - ', '      - ', '        - ', '          - '];
```

This change would get us mostly there, but there is one more catch. Indentation level also depends on the previous indentation level. So an H3 following an H1 should only be indented once.

See this example for what we are looking for (as implemented in the current live site). The H3 entries "Homebrew D&D made easy" and "Editing and Sharing" are only indented once since there is no intermediate H2. However the later H3 entries "Helping out", "Bugs, Issues, Suggestions?", etc. do have an intermediary H2, so they should be indented twice.

image

themes/V3/5ePHB/style.less
```diff
@@ -786,6 +786,32 @@
...
```
calculuschild, line 801: Since `partCover` should *only* include the H1 header in the ToC, perhaps we can add a rule like this:

```
.page:has(.partCover) {
  --TOC: exclude
  &h1 {
    --TOC: include
  }
}
```

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -2,77 +2,66 @@ const _ = require('lodash');
...
-			_.each(g1.children, (g2, idx2)=>{
-				if(g2.title !== null) {
-					r.push(`  - #### [{{ ${g2.title}}}{{ ${g2.page}}}](#p${g2.page})`);
-				}
-				if(g2.children.length){
-					_.each(g2.children, (g3, idx3)=>{
-						if(g2.title !== null) {
-							r.push(`    - [{{ ${g3.title}}}{{ ${g3.page}}}](#p${g3.page})`);
-						} else { // Don't over-indent if no level-2 parent entry
-							r.push(`  - [{{ ${g3.title}}}{{ ${g3.page}}}](#p${g3.page})`);
```
calculuschild, line 70: The previous behavior makes sure H3 headers aren't indented too much if there is no H2 header in between. That is, each entry in the ToC should only be indented one level deeper than the parent ToC entry. Else you can run into cases where the Markdown parser fails to interpret the list.

Example:

```
- H1
  - H2
    - H3
    
and
    
- H1
  - H3    <- only indented once, since no H2 above
    - H6  <- only indented one more, since no H4, H5 above

because

- H1
    - H3        <- This entry will appear too-far indented
          - H6  <- This entry will not parse at all
```

5e-Cleric:
to test this are we supposed to use the first option only, or any?
image

5e-Cleric:
Not sure if this has been talked, is the type of element in the ToC set via the type of header it references, or by the position in the ToC? i have an h1 followed by h3, and the h3 gets the same styling as an h2 would.

dbolack-ab:
> Not sure if this has been talked, is the type of element in the ToC set via the type of header it references, or by the position in the ToC? i have an h1 followed by h3, and the h3 gets the same styling as an h2 would.

By the position.

dbolack-ab:
> to test this are we supposed to use the first option only, or any? image

THat is correct. How can we make this more obvious?

5e-Cleric:
> > to test this are we supposed to use the first option only, or any? image
> 
> THat is correct. How can we make this more obvious?

I don't understand this answer, which is it? The first option, or the latter ones??

dbolack-ab:
> > > to test this are we supposed to use the first option only, or any? image
> > 
> > 
> > THat is correct. How can we make this more obvious?
> 
> I don't understand this answer, which is it? The first option, or the latter ones??

Oh! I misread  The first one. The latter ones drop the "add this level to the ToC" tags.

calculuschild:
Ok, merging this!

Issue #2832, opened 2023-05-12
heading IDs are not unique vs manual html or curly-blocks


ericscheid:
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

Heading markdown (`# A Heading`) gets rendered as html `Hn` elements with an `id` attribute derived from the heading text (`<h1 id="a-heading">A heading</h1>`). Marked will ensure the value of the `id` attribute is unique among all heading elements it renders .. but does not consider any other elements with the same `id`. 

Other elements with the same `id` could come from manually inserted html (`<table id="myTable">...`) and even from curly-blocks and curly-spans (`{{#myID ...}}`) and also curly-injectors (`*BOLD*{#myID}`).

The heading IDs should be unique to the full render, not just other headings.

Additional: any `<a name="value"></a>` should also be considered and avoided, since links to those fragment targets compete for the same namespace.

[Marking as bug because heading ids should be unique. Rare, esoteric, and obscure, and can be worked around.]

### Code

```gfm
{{#bq,note
this is a note with the id `bq`
}}

## BQ

Above is a `H2` which will get rendered as `<h2 id="bq">BQ</h2>`. This is not unique, given the curly-block with the id `bq` 

## BQ

Another `BQ` heading, but this time `marked()` will assign `id="bq-1"` because it has already seen the id `bq` in it's processing [of headings].
```

ericscheid:
No rush on this bug. 

Right now, even the TOC generator does not use the heading IDs for linking, it just uses page numbers. So, this bug will only appear for people knowingly and manually creating links to anchors. Hopefully they'll figure out the workaround (i.e. assign unique IDs to manually inserted html et al).

ericscheid:
This issue will become (slightly) more important if these get solved: #1923 #2830

5e-Cleric:
Important to consider that repeated IDs or names will not break anything really, it will just prevent links and scripts to get the element, but should not break a brew.

G-Ambatte:
It's possible that we could inject a step right before the rendered Markdown gets pushed into the BrewRenderer display element which builds a list of all element IDs in the HTML document and then runs a deduplication process.

5e-Cleric:
Another note, headings IDs are not unique unless they are the same tag, if you have a `h1` and a `h5` with the same name, they will have the same ID.

ericscheid:
[...]

PR #3262, opened 2024-01-22
 Simple solution for global unique header IDs

dbolack-ab:
Solves Issue #2832.

 Trusts that MarkedGFMHeadingId will make the ids unique per page and header level and then massages them to have what should be a global uniq name via the hooks: postprocess  system.

```
# Header 1
# Header 1
# Header 1
## Header 2
### Header 3
#### Header 4

```

```
<h1 id="p0-H1-header-1-1">Header 1</h1>
<h1 id="p0-H1-header-1-2">Header 1</h1>
<h2 id="p0-H2-header-2">Header 2</h2>
<h3 id="p0-H3-header-3">Header 3</h3>
<h4 id="p0-H4-header-4">Header 4</h4>
```


dbolack-ab:
Code tests are failing due to an issue with jest and dom that I do not have enough knowledge of to solve.

ericscheid:
The useful thing for putting `id` attributes on headings is that an author can then link to those headings.

See this brew for an example: https://homebrewery.naturalcrit.com/share/SyXKaZmQM
Every maneuvre listed on the 1st page links to the relevant section in the text.

This only remains useful if the `id` values don't change. 

Adding or removing a page early in the brew will cause all the headings in this solution to change. And not in an obviously broken way either (i.e. clicking the link to `#p7-H3-header-3` will still link to a header with that `id` .. despite the intended destination now being located on p8 or p6).

dbolack-ab:
Okay. I'll go to plan b.

calculuschild:
@ericscheid This now keeps the original ID from users, but still applies a page number and counter: `#p1-myID-1`.

Is this any better for your use case, or do you have an alternate preferred behavior? I understand this may still be an issue when deleting or adding pages because the `#pX` will change, but at least `myID` will remain intact.

ericscheid:
> @ericscheid This now keeps the original ID from users, but still applies a page number and counter: `#p1-myID-1`.
> 
> Is this any better for your use case, or do you have an alternate preferred behavior? I understand this may still be an issue when deleting or adding pages because the `#pX` will change, but at least `myID` will remain intact.

This is no better. The `id` still changes, and only exact matches are used for linking.

The numbering will also change if headings are inserted or deleted or rearranged on a given page.

calculuschild:
> The numbering will also change if headings are inserted or deleted or rearranged on a given page.

Well, this is going to be true of any solution that creates unique identifiers for duplicated names. Rearranging the headings would produce unique ids in a different order, no?

ericscheid:
> > The numbering will also change if headings are inserted or deleted or rearranged on a given page.
> 
> Well, this is going to be true of any solution that creates unique identifiers for duplicated names. Rearranging the headings in text would always correspond to a rearrangement of the unique IDs being generated, no?
> 
> Unless I'm missing something, what naming scheme would you propose that is immune to reordering of the text?
> 
> Or maybe if you don't mind making a short example document showing what rearrangements of headers you expect to be stable? So we are all working from the same requirements.

The issue isn't the instability of "unique identifiers for duplicated names", but rather that even for non-duplicated names the ids are unstable. For example .. this document

```
## foo
## bar 
## baz
```

should have the same ids as 
```
## baz
## bar 
## foo
```
and even this
```
## foo
### bar 
#### baz
```

calculuschild:
I see
 And ideally this should persist across pages, no?

```
### foo

\page

## bar
## baz
```

ericscheid:
> I see, and ideally this should persist across pages, no?

Absolutely.

Bottom line: the fix for one specific case shouldn't break a more general case.

dbolack-ab:
> > @ericscheid This now keeps the original ID from users, but still applies a page number and counter: `#p1-myID-1`.
> > Is this any better for your use case, or do you have an alternate preferred behavior? I understand this may still be an issue when deleting or adding pages because the `#pX` will change, but at least `myID` will remain intact.
> 
> This is no better. The `id` still changes, and only exact matches are used for linking.
> 
> The numbering will also change if headings are inserted or deleted or rearranged on a given page.

There is no way to prevent this other than by the user forcing an id via { }. 

Even if we could maintain a global list without massaging the IDs  ( similar to how we have in the PR for reflinks )  nothing guarantees the integrity of auto-generated header ID reflinks staying valid. It is an unreliable method of internal document linking and should be discouraged. Even in a single-page document.

ericscheid:
> There is no way to prevent this other than by the user forcing an id via { }.

Expanding the scope of a problem from specific to general cannot be excused by saying there's already a problem with an edge case.

PR #3263, opened 2024-01-22
[FEATURE]: Vault page (search published brews) (alt title Archive)

5e-Cleric:
This is just me trying to create the most wanted function, find published brews by title.

This PR creates a new Page, called the vault page, it also creates a new API endpoint to get files based of title, with multiple additional filters.

As per last update, has 2 separate request that are run simultaneously in first load of query, then only one of them is run with every reload. This avoids timeouts.

### Added files and their functions:
- `vaultPage.jsx`
- `vaultPage.less`
- `vault.api.js`
- `vault.navitem.jsx`

### Modified files:
- `homebrew.jsx`: added Route for the page
-  `brewItem.jsx`: modified author display to show 'no authors';
- `app.js`:  added vault error handling
- `homepage.jsx` & `userpage.jsx`: added nav item to vault
- `splitPane.jsx` : hide arrows if prop `hideMoveArrows`

### For a next PR
- [x] Turn to functional component
- [ ] Use $search or $facet methods
- [x] Use splitPane component
- [ ] Use listPage component
    - [ ] add Splitpane inside listpage
    - [ ] move query to outer page and just send the results
    - [ ] move filter and sort options to backend at least for Vault, due to pagination.

### ToDo in this PR
- [x] Change Page size to page count (name)
- [x] Page count from input to select
- [x] Disable button programatically
- [x] Fix CSS for background not filling the page correctly
- [x] unable selection

### Testing:

```
- [ ] Full functionality
  - [ ] Title search, both renderers
  - [ ] Author search, both renderers
  - [ ] Title & Author search, both renderers
  - [ ] All aforementioned, with v3 only
  - [ ] All aforementioned, with Legacy only
  - [ ] Testing advanced syntaxes ("word", -word, are there any more?)
- [ ] full functionality from URL directly
- [ ] Correct pagination controls
- [ ] Correct disabling of buttons
- [ ] Correct validation of inputs
- [ ] Request times under half a second
```

I don't know what else to test.


5e-Cleric:
toDo list:
- [x] move from admin.api to its own
- [x] use ~listpage.jsx and~ brewitem.jsx
- [ ] add filtering functions
   - [x] renderer
   - [x] cloned or not (-CLONE should do it)
   - [x] excluding words?
   - [x] views?
   - [x] authors?
   - [ ] theme
- [x] implement agnostic UI
- [x] response brew count
- [x] page size query option

5e-Cleric:
Adding here talk from gitter chat:

brews: null is getting pushed to console twice on every update, but the search only occurs on Enter/button click
Correction: it hits the console twice on a page refresh
I assume this is due to the this.lookup() in componentDidMount, so it runs a search as soon as the page loads


Ah .. here's the way to search for a whole word: (^|[^a-z])DM($|[^a-z]).

A ssearch for [PF2e] takes a long time and returns a huge result of which very few actually contain the string "[PF2e]".
This is because it's actually doing a regex search for any title that contains, in any order, any of the characters p,f,2,e.
Oops.

So .. proposal: instead of defaulting to regex (with a failback of exact string match if invalid regex), do a exact string match by default, with the exception being if the search string is indicated to be a regex by wrapping with /.

Thus: Search [PF2e] will find titles containing that exact string; /[PF2e]/ will do a regex search.

This is the same way in-brew search/replace also supports regex.
No need for a UI widget (e.g. checkbox) to indicate if regex applicable.

ericscheid:
Very exciting, lets keep this moving..

1. searching for a term that does has zero matches gives the false explanation of
   > ### I'm sorry, your request didn't work
   > Your search is not enough specific, too many brews meet this criteria for us to forward them.

2. The error message itself is black text on the dark blue backgound. (#2C3E50)


image

5e-Cleric:
Yes, i am aware, however, i am right now trying to figure out pagination ( i know how it works, i just keep hitting error after error(my normal working process)).

client/homebrew/pages/archivePage/archivePage.jsx
```diff
@@ -0,0 +1,218 @@
...
+			</div>
+		);
+	},
+
+
+	renderForm : function () {
+		return (
+			<div className='brewLookup'>
+				<h2>Brew Lookup</h2>
+				<label>Title of the brew</label>
```
ericscheid, line 144: Move the `<input>` element to be a child of the `<label>` element.

server/archive.api.js
```diff
@@ -0,0 +1,52 @@
...
+
+const archive = {
+	archiveApi : router,
+	/* Searches for matching title, also attempts to partial match */
+	findBrews  : async (req, res, next)=>{
+		try {
+			const title = req.query.title || '';
+			const page = parseInt(req.query.page) || 1;
+			console.log('trying page ', page);
+			const pageSize = 10; // Set a default page size
```
ericscheid, line 13: Change to `const pageSize =  parseInt(req.query.size) || 10;`.

It can remain a "hidden" feature (as an aid in testing), with _maybe_ a UI element added later.

Probably best to do this, just in case we don't want users supplying a negative page size
```
const minPageSize = 1; 
const pageSize = Math.max(parseInt(req.query.size) || 10, minPageSize);
```

Probably best to have a more reasonable minumum page size too .. otherwise we might end up with a huge mass of pagination links (e.g. `pageSize = 1`).

server/archive.api.js
```diff
@@ -0,0 +1,52 @@
...
+const router = require('express').Router();
+const asyncHandler = require('express-async-handler');
+
+const archive = {
+	archiveApi : router,
+	/* Searches for matching title, also attempts to partial match */
+	findBrews  : async (req, res, next)=>{
+		try {
+			const title = req.query.title || '';
+			const page = parseInt(req.query.page) || 1;
```
ericscheid, line 11: Enforce a minimum of `1` to avoid shenanigans

`const page = Math.max(parseInt(req.query.page) || 1, 1);`

ericscheid:
Clicking the pagination links is funky..

1. the desired page of results is displayed :yes:
2. the pagination indicator for the newly displayed is highlighted :yes:
3. the window.location.href improperly updates
   - if clicking the pagination page number links, window.location is not updated at all
   - if navigating by the `<<` or `>>` buttons, window.location has the previously displayed page number, not the newly navigated to page number

(Unrelated: hitting (e.g.) https://homebrewery-pr-3263.herokuapp.com/archive?title=dagger&page=5 does not load the 5th page of results — _it does not load any page of results_. This makes testing this aspect impossible at the moment.)

server/archive.api.js
5e-Cleric: That is a good idea, but i am not sure on how to implement the pagesize control, and i don't think we need it.
I'm trying to come up with a website that does something like that.

5e-Cleric:
> 3. 👎 the window.location.href improperly updates ..
>    
>    * if clicking the pagination page number links, window.location is not updated at all
>    * if navigating by the `<<` or `>>` buttons, window.location has the previously displayed page number, not the newly navigated to page number

That is an interesting matter, not sure why that happens, i might not update the page state properly before calling updateUrl

> (Unrelated: 👎 hitting (e.g.) https://homebrewery-pr-3263.herokuapp.com/archive?title=dagger&page=5 does not load the 5th page of results — _it does not load any page of results_. This makes testing this aspect impossible at the moment.)

Hate to say it, works fine on my end, reached even the 6th page.

server/archive.api.js
5e-Cleric: Done, hope you like how i did it.

server/archive.api.js
```diff
@@ -0,0 +1,63 @@
...
+				text     : 0,
+				textBin  : 0,
+			};
+			const brews = await HomebrewModel.find(titleQuery, projection)
+                .skip(skip)
+                .limit(pageSize)
+                .maxTimeMS(5000)
+                .exec();
+
+			const totalBrews = await HomebrewModel.countDocuments(titleQuery, projection).maxTimeMS(5000);
```
ericscheid, line 44: Do we need to pass in `projection` here, as we want to count all documents matching `titleQuery` and all those documents will have a `title` and the `published` flag is part of `titleQuery`. That is, there is nothing in `projection` that says "don't count documents that don't have these fields".

5e-Cleric: you're absolutely right

server/archive.api.js
```diff
@@ -0,0 +1,63 @@
...
+            //console.log('Total pages: ', totalPages);
+			return res.json({ brews, page, totalPages, totalBrews});
+		} catch (error) {
+			console.error(error);
+            if (error.response && error.response.status === 503) {
+                return res.status(503).json({ errorCode: '503', message: 'Service Unavailable' });
+            } else {
+                return res.status(500).json({ errorCode: '500', message: 'Internal Server Error' });
+            }
+		}
```
ericscheid, line 57: OK, so here are my notes on improving title search...

Goals:

1. default to plain text search
  a) is faster than regex search. Because index.
  b) fewer errors from strings that are actually bogus regex patterns (e.g. unbalanced `[]` or `()`)
  c) avoid plain text searches (e.g. `[PF2e]` being handled as regex and getting unexpected results (e.g. `[PF2e]` as a regex search matches all titles that contain at least one of `P`, `F`, `2`, or `e` characters)

2. enable regex search by requiring the user to fence their regex pattern with `/`

3. for plain text title search, split the search string into words and do an AND search (so `WORD1 and WORD2` will match titles that contain both words, in any order, with any intervening words)

4. actually, split the plain text search string into operands, which are either space delimited words, or quoted strings. So we can do searches like `word1 word2 "quoted string"`.

5. detect if a plain text search operand has a `-` prefix, and interpret that as an exclude operand

So .. to start with #1 and #2...

```
  const title = req.query.title || '';
  const page = Math.max(parseInt(req.query.page) || 1, 1);
  const minPageSize = 6; 
  const pageSize = Math.max(parseInt(req.query.size) || 10, minPageSize);  
  const skip = (page - 1) * pageSize;

  const brewsQuery = {
    $or: [],
    published : true
  };

  if (req.query.v3 === 'true') { ... }
  if (req.query.legacy === 'true') { ... }
  
  if (title.startsWith('/') && title.endsWith('/')) {
    const titleConditionsArray = [
      {  "title": { $regex: title.slice(1, -1), $options: "i" } }
    ];  
  } else {
    // do plain text search
    const titleConditionsArray = buildTitleConditions (title); 
  }

  function buildTitleConditions(inputString) {
    return [ { "title": { $text: { $search: inputString, $caseSensitive: false } } } ];
  }
  
  const titleQuery = {
    $and: [
      brewsQuery,
      ...titleConditionsArray
    ]
  };

// ... now do the HomebrewModel.countDocuments() and HomebrewModel.find() and etc

  const projection = { ... };
  // ...

```


For #3, #4, #5 lets build out a fuller buildTitleConditions()

```
function buildTitleConditions(inputString) {
    const regex = /(-?"[^"]+"|-?\S+)/g;
    const matches = inputString.match(regex) || [];

    const finalQueries = [];

    for (const matchedString of matches) {
        const isExcluded = matchedString.startsWith('-');
        const baseString = isExcluded ? matchedString.slice(1) : matchedString;

        const unquotedMatch = baseString.startsWith('"') && baseString.endsWith('"')
            ? baseString.slice(1, -1)
            : baseString;

        const baseQuery = { "title": { $text: { $search: unquotedMatch, $caseSensitive: false } } };

        if (isExcluded) {
            finalQueries.push({ $not: baseQuery });
        } else {
            finalQueries.push(baseQuery);
        }
    }

    return finalQueries;
}

```

G-Ambatte:
Preserving @calculuschild's comment from Gitter about Mongo searching, as apparently search capabilities differ, depending hosting:

---

So you should be able to do something like this:

```
{ $search: {
        "index": "default", // optional, defaults to "default"
        "text": {
            "query": "dragon",
            "path": "title",
            "fuzzy": {
                "maxEdits": 1,
                "prefixLength": 0,
                "maxExpansions": 50
            }
        }
    }
}
```

**maxEdits:** Maximum number of single-character differences allowed to be considered a match. Value can be 1 or 2. The default value is 2.
**prefixLength:** Number of characters at the beginning of each term in the result that must exactly match. The default value is 0.
**maxExpansions:** The maximum number of variations to generate and search for. This limit applies on a per-token basis. The default value is 50.

server/archive.api.js
5e-Cleric: Maybe i am wrong about this, but isn't titleConditionsArray defined inside an if, and tried to use down below, where it is 
not defined?

5e-Cleric: This is implemented now, works fine in local, with some adjustments.

5e-Cleric:
Searching works again, after lots of tweaks, the index read is "title_text", will try creating a separate PR to try the $search stage method.

Gazook89:
Not sure if you are looking for this kind of input yet or not, but I'm going to shoot my shot here.  Pretty much all of these bulleted items i think are "minor nitpicks" that can probably be addressed fairly easily one-by-one (though there is some overlap).  

A. Can you revise your first comment on this PR to clearly explain what is added, how to access it, and what might yet need work either in this PR or a future PR? It was immediately clear to me how to access this Archive page-- I thought i'd find it in the User menu, or on the Share page, or something. I suppose that will come later, but for now it needs to be mentioned up top that this page is only accessible via the address bar.

B. I think you should get rid of `Your results, my lordship`.  I only saw this when testing on `local`, I don't see it on the deployment, so maybe not a problem?  I'd just be sure it's gone.

C. Pretty much all of the text that isn't in an input should use CSS to make it non-selectable/not use the "text cursor" on hover.

D. The "whenever you want, just start typing" area might be improved by just showing a default result ("most viewed", "most recent", etc).  At minimum, I wouldn't use "whenever you want, just start typing..." as text in this area.  This gives the impression that I should either type in that area of the window, or can start typing without focus on *any* input.  If not doing a default search result as suggested, this text could just say "showing no results" or similar.

E.  I'm not even sure that the Archive page needs to differ from the User Brew List page that much-- the brew list page already has a filter/sort bar which is relevant to Archive results as well. No other page has a big banner at the top declaring the page name.  And the two side-by-side portions, if kept, could likely benefit from splitPane as well.

F.  I don't think a number input is the best option for "number of results", and the current attributes are not great (mystifying was the first word that came to mind).  Instead, `<select>` menu with options like just 15 & 30, or 15, 30, 50.  

```html
<label>Results per page<input
    id="size"
    type="number"
    min="6"
    step="2"
    max="30"
    name="pageSize"
    value="12"
/></label>
```

G.  If more renderer versions come out, the URL parameters will have to be updated since it is handled like this: `&v3=true&legacy=true`.  Is it possible to just send something like `&renderer=legacy,v3` so that additional future renderers can just be added to the list? 

H.  I prefer `count` over `size` for the param setting the number of results returned.

I.  Going further on the URL, is it better to arrange the parameters like this:  `/archive?title=monster&renderer=legacy,v3&count=12&page=1`?    To me, the `page` parameter is the most *dynamic* of the set-- having it at the end makes it easier to change if using the address bar (or scraping results).

J.  Do these parameters have defaults?  It seems like all of them besides `title` should have defaults so that a simple query like `http://localhost:8000/archive?title=monster` would work.  I think sensible defaults with the current params could be: 
- renderer:  `legacy,v3`
- count: `15`
- page: `1`

K. Maybe `disable` the Search button (and style it as such) if it's not going to function-- for example, pressing Search when no search text has been entered does nothing.

L.  Perhaps we can load the mongo 'stop word' list and check the search string against it before submitting.  If a user tries to search just "and" it's going to return no results....so we can just prevent the submission preemptively.  Probably there is some other validation we can do, like making sure the search string is at least 3 characters or something.

M.  This might be resolved by another suggestion above (using splitPane or matching the overall form of the user brew list), but currently the blue background in the resultsContainer doesn't extend to the bottom of the screen, and neither does the associated scrollbar.  Several CSS methods to fix this.


Finally, one bigger question that is beyond a nitpick:   Is there a benefit to having this separate archive page in addition to a user brew list?    Or could there be a single page for the two--- navigating to the "users own brews" would just set an `author` parameter on the search that matches the username.  And via the UI you could remove or add authors to your search query.   If this Archive page is available to non-users (or users not logged in), which is another question, then going to the Archive would obviously not default to the current user's brews.


Anyway, if you want help with any of the points above I can submit PR's on your PR.  Otherwise this is all just up for discussion.

PS.  I think the end result of this PR is huge, so good work 🚀

5e-Cleric:
## Point by point

> A. Can you revise your first comment on this PR to clearly explain what is added, how to access it, and what might yet need work either in this PR or a future PR? It was immediately clear to me how to access this Archive page-- I thought i'd find it in the User menu, or on the Share page, or something. I suppose that will come later, but for now it needs to be mentioned up top that this page is only accessible via the address bar.

Sure, no problem, Calc asked that the nav change came in another PR

> B. I think you should get rid of `Your results, my lordship`. I only saw this when testing on `local`, I don't see it on the deployment, so maybe not a problem? I'd just be sure it's gone.

I changed it for Your searched returned these results, but i should probably change that, I wanted to make it visually different from edit and homepage, but i guess i could go with the theme.

> C. Pretty much all of the text that isn't in an input should use CSS to make it non-selectable/not use the "text cursor" on hover.

I don't believe there is a single page where we have done that:

image

But i can make that no problem

> D. The "whenever you want, just start typing" area might be improved by just showing a default result ("most viewed", "most recent", etc). At minimum, I wouldn't use "whenever you want, just start typing..." as text in this area. This gives the impression that I should either type in that area of the window, or can start typing without focus on _any_ input. If not doing a default search result as suggested, this text could just say "showing no results" or similar.

I don't personally like `showing no results` if no search is done, that response is reserved for an empty search.
But i agree it should be changed.

> E. I'm not even sure that the Archive page needs to differ from the User Brew List page that much-- the brew list page already has a filter/sort bar which is relevant to Archive results as well. No other page has a big banner at the top declaring the page name. And the two side-by-side portions, if kept, could likely benefit from splitPane as well.

I am not using listpage.jsx for some reason i don't remember, i think i needed to refactor some logic in listpage to work correctly, will go over it again.

I will ask this on the gitter chat to make sure everyone agrees, but using splitPane makes sense, i just don't want to get caught another month just modifying the component to fit. Actually, i'd rather we do this in another later PR.

> F. I don't think a number input is the best option for "number of results", and the current attributes are not great (mystifying was the first word that came to mind). Instead, `<select>` menu with options like just 15 & 30, or 15, 30, 50.

Eh.. not sure there is much difference but i can make that, would be more similar to how it was made in other websites.

> G. If more renderer versions come out, the URL parameters will have to be updated since it is handled like this: `&v3=true&legacy=true`. Is it possible to just send something like `&renderer=legacy,v3` so that additional future renderers can just be added to the list?
> 
> H. I prefer `count` over `size` for the param setting the number of results returned.
> 
> I. Going further on the URL, is it better to arrange the parameters like this: `/archive?title=monster&renderer=legacy,v3&count=12&page=1`? To me, the `page` parameter is the most _dynamic_ of the set-- having it at the end makes it easier to change if using the address bar (or scraping results).

Good suggestions, will take a look at that, they don't look like problems to me, but something more left to choice and taste... and i'm not sure if i can make the renderer one, don't know much about queries

> J. Do these parameters have defaults? It seems like all of them besides `title` should have defaults so that a simple query like `http://localhost:8000/archive?title=monster` would work. I think sensible defaults with the current params could be:
> 
> * renderer:  `legacy,v3`
> * count: `15`
> * page: `1`
>

both renderers have a default of true, count has a default of 10, page has a default of 1:

```
title: this.props.query.title || '',
legacy: this.props.query.legacy !== 'false',
v3: this.props.query.v3 !== 'false',
pageSize: this.props.query.size || 10,
page: parseInt(this.props.query.page) || 1,
```

> K. Maybe `disable` the Search button (and style it as such) if it's not going to function-- for example, pressing Search when no search text has been entered does nothing.

good idea!

> L. Perhaps we can load the mongo 'stop word' list and check the search string against it before submitting. If a user tries to search just "and" it's going to return no results....so we can just prevent the submission preemptively. Probably there is some other validation we can do, like making sure the search string is at least 3 characters or something.

I can whip up a simple regex pattern for the input, but i suck at doing your super complex validation that you have done in other files, could use a hand with that. We could also do it in another PR

> M. This might be resolved by another suggestion above (using splitPane or matching the overall form of the user brew list), but currently the blue background in the resultsContainer doesn't extend to the bottom of the screen, and neither does the associated scrollbar. Several CSS methods to fix this.

thought i had fixed that, damn CSS

> Finally, one bigger question that is beyond a nitpick: Is there a benefit to having this separate archive page in addition to a user brew list? Or could there be a single page for the two--- navigating to the "users own brews" would just set an `author` parameter on the search that matches the username. And via the UI you could remove or add authors to your search query. If this Archive page is available to non-users (or users not logged in), which is another question, then going to the Archive would obviously not default to the current user's brews.

I'd rather that not be the case, that would be a huge component to load just for the userPage, and the different routes... idk
If done, in another PR

> Anyway, if you want help with any of the points above I can submit PR's on your PR. Otherwise this is all just up for discussion.

i'd appreciate that, could you take a look at M and possibly G? should be fairly easy without knowing the code.

> PS. I think the end result of this PR is huge, so good work 🚀

client/homebrew/pages/archivePage/archivePage.jsx
```diff
@@ -0,0 +1,443 @@
...
+                            placeholder="v3 Reference Document"
+                        />
+                    </label>
+                    <small>
+                        Tip! you can use <code>-</code> to negate words, and{' '}
+                        <code>"word"</code> to specify an exact string.
+                    </small>
+                    <label>
+                        Results per page
+                        <select name="pageSize" id="size">
```
Gazook89, line 207: This still seems like way too much granularity in options.  Just because it's possible to return many different amounts, it seems needless to offer it.  

And I would say that '6' is a weird number to increment by.  I understand that an *even* number is good, because currently it's returning as predominantly a 2 column list.  But that may not always be the case (for example, if the screen is narrow or we later change how this list is displayed).  And why would anyone only ever want 6 as a max number of results back?  Is someone really getting any benefit from choosing between 6 and 10 results, would a difference greater than 4 break the bank?

If an even number is desired, stick to 10's....but actually 20's.   Options being `20`, `40`, `60`, dropping `60` if it is too much of a toll on the server.  

Sidenote-- I am doing a search for "druid" which returns 222 brews, or 23 pages with 6 brews each.  If i change the `size` parameter in the address bar manually to `60`, it returns only 6 results at a time *but the "pages" portion under the results only shows 4 pages*.  So somewhere the app knows I am only supposed to 4 pages of 60 results each, but returns only 6 brews on each (for 24 brews total).  The param value in the URL should be the source of truth for everything in the query, I think.

ericscheid:
> G. If more renderer versions come out, the URL parameters will have to be updated since it is handled like this: &v3=true&legacy=true. Is it possible to just send something like &renderer=legacy,v3 so that additional future renderers can just be added to the list?

A more common code pattern is to simply have multiple checkboxes with the same `name` attribute, such that when submitted you get `&renderer=V3&renderer=legacy`.

> Is there a benefit to having this separate archive page in addition to a user brew list? Or could there be a single page for the two--- navigating to the "users own brews" would just set an author parameter on the search that matches the username. And via the UI you could remove or add authors to your search query. If this Archive page is available to non-users (or users not logged in), which is another question, then going to the Archive would obviously not default to the current user's brews.

A hiccup to this plan is that we are planning on supporting folders and so forth on the /userpage

client/homebrew/pages/archivePage/archivePage.jsx
```diff
@@ -0,0 +1,443 @@
...
+                    <div className="welcome">
+                        <h1>Welcome to the Archive</h1>
+                    </div>
+                    <div className="flexGroup">
+                        <div className="form dataGroup">
+                            {this.renderForm()}
+                        </div>
+                        <div className="resultsContainer dataGroup">
+                            <div className="title">
+                                <h2>Your searched returned these results</h2>
```
Gazook89, line 432: This line could report the number of brews found.

- "230 brews matched your query"
- "230 brews returned"
- "Your search returned 230 brews"
- etc

Also note, at minimum, there is a typo, should be "Your search returned these results"

client/homebrew/pages/archivePage/archivePage.jsx
```diff
@@ -164,6 +164,16 @@ const ArchivePage = createClass({
         );
     },
 
+    validateInput: function (e) {
+        const textInput = e.target
+        const submitButton = document.getElementById('searchButton');
+        if (textInput.value.length > 3) {
```
Gazook89, line 170: I know I suggested `3`, but two thoughts:  

1)  this is `greater than 3` so it's actually requiring 4, which precludes any searches for "cat" or "dog" which is maybe too tight. But, ...
2)  probably the most important limitation is that the input has `greater than 0` length.  Just has to have something in it.

Gazook89:
> A more common code pattern is to simply have multiple checkboxes with the same name attribute, such that when submitted you get &renderer=V3&renderer=legacy.

Yeah I saw that too when searching around a bit.  Is that burdensome when we have `&renderer=legacy&renderer=v3&renderer=v4&renderer=v5`?  

I don't know how searching by tags exactly works, but lets say it's just passing the tags to params in a similar fashion....should we have `&tag=foo1&tag=foo2&tag=foo3&tag=foo4` etc?  

I'm okay with the suggestion, just thinking out loud in case it's worth considering.

client/homebrew/pages/archivePage/archivePage.jsx
Gazook89: The validation on the Title input that disables/enables the Search button doesn't seem to check if it's valid when the page reloads.  The same valid text is in the field, but because validation only happens at keyUp, it shows invalid--- likely validation method needs to happen at componentDidMount stage.

Same deal if I submit a search with valid Title-- the component reloads, maintains it's valid text, but the Search button is disabled.  Likely same solution as above.

client/homebrew/pages/archivePage/archivePage.jsx
```diff
@@ -164,6 +164,16 @@ const ArchivePage = createClass({
...
+            submitButton.disabled = false;
+        } else {
+            submitButton.disabled = true;
+        }
+    },
+
     renderForm: function () {
```
Gazook89, line 177: Regarding the "form":

I'm noticing now that the Search button doesn't actually do anything for me when I click on it.  I can trigger a search by hitting Enter in the Title input, but not by clicking on the button (and it isn't disabled).  No errors in console.

I think this has to do with the `onClick` function, but I checked out your branch to work on my local machine and just couldn't figure out what was up.  I did manage to fix it by converting the whole "form" div to an actual HTML `<form>` element, and changing the Search button into a `<button type='submit' onSubmit='....'>` and removing the onClick attribute (the same functions go into onSubmit).  It's hard to be sure it works on local, though, as it seems I can't actually perform the search.

ericscheid:
> > A more common code pattern is to simply have multiple checkboxes with the same name attribute, such that when submitted you get &renderer=V3&renderer=legacy.
> Yeah I saw that too when searching around a bit. Is that burdensome when we have `&renderer=legacy&renderer=v3&renderer=v4&renderer=v5`?

I don't know how node/react handles plurality, but other platforms will usuaily ingest such as an array for handling. (PHP does that automatically if of the form `&arg[]=1&arg=2&arg[]=3`)

> I don't know how searching by tags exactly works, but lets say it's just passing the tags to params in a similar fashion....should we have `&tag=foo1&tag=foo2&tag=foo3&tag=foo4` etc?

Yes. Whichever way we send the query parameters they'll need to be ingested into an array, and then when doing the search of the tags array we can't use `$in` and we'll want to do as case insensitive, so we'll need to use regex and the array element test of `$elemMatch`.

client/homebrew/pages/archivePage/archivePage.jsx
5e-Cleric: 6 is the only number of results that fit into my screen without needing to scroll, that's where that number comes from. But i'm not too attached to the idea.

5e-Cleric: good catch with the page number, it is still getting them from the input. I'll fix this.

client/homebrew/pages/archivePage/archivePage.jsx
5e-Cleric: Damn it, changed the class to an id last night, forgot to change the appropiate selectors for both CSS and js. Should be using refs anyway, but will fix this now to keep testing.

5e-Cleric: actually that can't be it, not sure what the problem is here.

client/homebrew/pages/archivePage/archivePage.jsx
5e-Cleric: The problem with this is i'm supposed to update state if the form is updated, then get from the state, but the setstate method is not fast enough to keep up, and i get the last search instead, not sure how to account for that.

client/homebrew/pages/archivePage/archivePage.jsx
5e-Cleric: I'd rather we don't do that, since the total count takes many seconds to load, there is a counter at the bottom right of the page for this, not disrupting.

client/homebrew/pages/archivePage/archivePage.jsx
5e-Cleric: Many functions depend on title not being empty, but this should be consistent, i agree

5e-Cleric: Fixed ( i think)

5e-Cleric:
This is ready for a merge. Everything not finished will be done in a separate, later PR.

dbolack-ab (approved):
This looks really good for this first stage.

5e-Cleric:
At this point, i'll keep adding things until i can merge, but the PR is ready.

Gazook89:
I think before this gets merged, it should include some way to navigate to the vault page in the navbar.  I'm not sure where it should go at this point-- in my opinion, we don't have a great logical place for it except yet another button in the top nav bar.  That should suffice for now.  I do not think it should go in the Account menu-- it shouldn't be necessary to be logged in to search for brews.

For an icon?:  `<i class="fa-solid fa-dungeon"></i>`
image

either that or `fa-search`.

Gazook89:
Are you interested in adding more search fields?  I am thinking additional fields like "Author" and "Tags", so you could search by specific fields, or a combination of fields.

For example, do a search right now for `theme`:

image

Let's say I don't like that *sor_robertson* guy and don't want to see his brews.  Right now, in the search box, I might try to modify my search like `theme -"sor_robertson"`, expecting that I could use the negation to remove his brews. 

That doesn't work.

Or, let's say I like that *Gazook89* guy, and I just want to see his brews...obviously, i could just use the address bar to look at his user page, but let's pretend this is my first day on the internet.  Doing a search for `Gazook89` only returns one result...a cloned brew by some other user with "Gazook89" in the title.

I might also want to search for brews by specific *tags*, like "meta:theme", or "system:pf2".  Currently, those don't do anything (though I know there are some brews that have those tags).

I think as far as layout goes, you would have the main search more visually prominent, and then the additional fields in a `<detail>` accordion, or just below the main search box.

Gazook89:
Trying to break my comments here into separate chunks-- 'nav', 'features',-- and now 'styling'.  Also, I thought I could run this PR locally but doesn't seem to be working for me.  I knew it obviously wouldn't have access to the production database but I thought I'd be able to view the page at least?  Let me know if i'm just doing it wrong.

For styling:  What do you think about just matching the styling of the Metadata/Properties editor?  Gray background, the inline labels, square corners on inputs, button colors and fonts, etc.   I was hoping to run locally and quickly test how easy it would be to just link the same stylesheet, and how much effort it would be to change the html structure so it would work.

v3/legacy buttons don't need to have "yes"/"no" options with a handwriting font-- they could simply be checkboxes.

There should a min-width set on the overall form now that we are using split pane (granted, the Properties editor doesn't have one, yet).

I might suggest changing "Brew Lookup" to "Vault", and then a subheader like "Search for brews" or "brew search" or similar.  For some reason, I think "Lookup" is a weird word/phrase that might not translate well.  Regardless, i think "Vault" should be more prominent than just being in the navbar.

5e-Cleric:
> I think before this gets merged, it should include some way to navigate to the vault page in the navbar. I'm not sure where it should go at this point-- in my opinion, we don't have a great logical place for it except yet another button in the top nav bar. That should suffice for now. I do not think it should go in the Account menu-- it shouldn't be necessary to be logged in to search for brews.
> 
> For an icon?: `<i class="fa-solid fa-dungeon"></i>` image
> 
> either that or `fa-search`.

At some point, specially for mobile, we should consider an icon-only navigation, or a big dropdown.
On the topic of nav items, i think changelog and faq should have more nav items.

5e-Cleric:
> Are you interested in adding more search fields? I am thinking additional fields like "Author" and "Tags", so you could search by specific fields, or a combination of fields.
> 
> For example, do a search right now for `theme`:
> 
> image
> Let's say I don't like that _sor_robertson_ guy and don't want to see his brews. Right now, in the search box, I might try to modify my search like `theme -"sor_robertson"`, expecting that I could use the negation to remove his brews.
> 
> That doesn't work.
> 
> Or, let's say I like that _Gazook89_ guy, and I just want to see his brews...obviously, i could just use the address bar to look at his user page, but let's pretend this is my first day on the internet. Doing a search for `Gazook89` only returns one result...a cloned brew by some other user with "Gazook89" in the title.
> 
> I might also want to search for brews by specific _tags_, like "meta:theme", or "system:pf2". Currently, those don't do anything (though I know there are some brews that have those tags).
> 
> I think as far as layout goes, you would have the main search more visually prominent, and then the additional fields in a `<detail>` accordion, or just below the main search box.

I haven't tried multiple field search, i am not sure if it would be easier or harder on the server, but i can try. We have plenty of space to accomodate more fields without the need of an accordion or advanced search option.

First search option should be tags, so i will focus on that.

Gazook89:
I made a branch on your branch but haven't pushed it to GH yet.  But here is the current status:

image

Small variations from the Properties editor include the stacked label/input rather than side-by-side.  I thought this might condense better at low widths.  Personally I think the Properties editor could/should do this, too, but that's another day.

I also understand that a more prominent title on the vault might be desired, so here is a variation:

image

Another thing I'm wondering is if the background should be a lighter gray, for this *and* for the Properties editor, to increase contrast with the text.

image

5e-Cleric:
> Small variations from the Properties editor include the stacked label/input rather than side-by-side. I thought this might condense better at low widths. Personally I think the Properties editor could/should do this, too, but that's another day.
> 
> I also understand that a more prominent title on the vault might be desired, so here is a variation:
> 
> Another thing I'm wondering is if the background should be a lighter gray, for this _and_ for the Properties editor, to increase contrast with the text.
> 
> Edit: I neglected to add styling for `<code>` bits, so those don't currently stand out from the rest of the text at all...but was just an oversight and can be done.

Don't have strong feelings about this... I don't really enjoy the overly long inputs, even if they work better at smaller widths.. and would prefer to stay away from the default appearance of the checkboxes, as i am sure you will have noticed. I like the title, specially if we include it with the icon as suggested in gitter. And those are `<small>` not code. Overall, thank you for paying attention to the project :)

Gazook89:
>  And those are <small> not code.

I believe it's `small` with the syntax marked by `code`, right?  Anyway, in my branch i switched out `small` anyway, for just a styled div:

```html
<div id='title-tip' class='tip'>
						<i className='fas fa-info'></i> you can use <code>-</code> to negate words, and <code>"word"</code> to specify an exact string.
					</div>
```

And the input itself uses `aria-describedby` to point to that element:

```jsx
<div className='field'>
					<label for='title-search'>Title</label>
					<input
						id='title-search'
						className='value'
						ref={titleRef}
						type='text'
						name='title'
						defaultValue={title}
						onKeyUp={validateForm}
						pattern='.{3,}'
						title='At least 3 characters'
						onKeyDown={(e)=>{
							if(e.key === 'Enter') {
								if(!searchButtonRef.current.disabled) {
									loadPage(1, true, true);
								}
							}
						}}
						aria-describedby='title-tip'
					/>
					<div id='title-tip' class='tip'>
						<i className='fas fa-info'></i> you can use <code>-</code> to negate words, and <code>"word"</code> to specify an exact string.
					</div>
				</div>
```

Gazook89:
> don't like normal checkboxes

I have preference for normal checkboxes, but on a whim I put this together, which could be achieved with a normal checkbox altered with just CSS (using background images):

image

Thoughts?

5e-Cleric:
I actually prefer the sword one better, far easier to see. We could try to animate that, but we could turn mad trying.

FYI right now the search is disabled (just the button and enter key) if both fields are empty, or if both renderers are unmarked. Although i could join the renderers so you need to chose one...

5e-Cleric:
Disabling "Author as owner" check, it is not worth the trouble for those slim changes in the results IMO. Keepind the functionality just in case we need it later.

5e-Cleric:
Situations to test:

| Tested | Title | Author | v3 | Legacy |
|:------:|:-----:|:------:|:--:|:------:|
- [ ] |  ✅ |   ❌   |✅ |   ❌   | 
- [ ]   |  ✅  |   ❌   |✅ |   ✅   |
- [ ]   |  ✅  |   ❌   |❌ |   ✅   |
- [ ]   |  ❌  |   ✅   |✅ |   ❌   |
- [ ]   |  ❌  |   ✅   |✅ |   ✅   |
- [ ]   |  ❌  |   ✅   |❌ |   ✅   |
- [ ]   |  ✅  |   ✅   |✅ |   ❌   |
- [ ]   |  ✅  |   ✅   |✅ |   ✅   |
- [ ]   |  ✅  |   ✅   |❌ |   ✅   |

Also:

 - [ ] all variants with title using exact words, or negated words


(I tried making this into a table, but checkboses don't work inside tables)

5e-Cleric:
Timeout issues solved by applying a compound index of title and published status.

Left toDo are:

- Cleanup
- Styling (can be done in a separate PR)

dbolack-ab:
My testing:

- [ ] Full functionality
  - [ ] Title search, both renderers
  - [ ] Author search, both renderers
  - [ ] Title & Author search, both renderers
  - [ ] All aforementioned, with v3 only
  - [ ] All aforementioned, with Legacy only
  - [ ] Testing advanced syntaxes ("word", -word, are there any more?)
- [ ] full functionality from URL directly
- [ ] Correct pagination controls
- [ ] Correct disabling of buttons
- [ ] Correct validation of inputs
- [ ] Request times under half a second

dbolack-ab:
Negation does not work in any of my tests. Made sure to try a long nonsense word to not hit the stop words list.


Times when *no* brews are found can be weirdly long.

What input validation is expected?

server/vault.api.js
```diff
@@ -0,0 +1,129 @@
...
+
+const buildBrewsQuery = (legacy, v3) => {
+    //first off, no need to include renderers, if all are valid
+    if (legacy === 'true' && v3 === 'true') return { published: true };
+
+    const renderers = [];
+    if (legacy === 'true') renderers.push('legacy');
+    if (v3 === 'true') renderers.push('V3');
+
+    const brewsQuery = { published: true };
```
dbolack-ab, line 46: Might be better to declare this first and return it on line 40 on chance this default grows further

5e-Cleric: good point, also simplifying the syntax.

5e-Cleric:
Updated all those comments to be in a tips and tricks post below the form:

| Before | After |
|:------:|:-----:|
|image|image|

Also updated the header form to use codebold, to make it less plain

5e-Cleric:
> What input validation is expected?

Good point, not sure there is any other than a simple pattern.

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+
+const buildTitleConditions = (title) => {
+    if (!title) return {};
+    return {
+        $text: {
+            $search: title,
+            $caseSensitive: false,
+        },
+    };
+};
```
calculuschild, line 15: Could be slimmed down and remains legible

```suggestion
const buildTitleConditions = (title) => {
    if (!title) return {};
    return {$text: {$search: title, $caseSensitive: false}};
};
```

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+    if (legacy === 'true' && v3 === 'true') return { published: true };
+
+    if (legacy === 'true' && v3 !== 'true') {
+        brewsQuery.renderer = 'legacy';
+    } else if (v3 === 'true' && legacy !== 'true') {
+        brewsQuery.renderer = 'V3';
+    }
+
+    return brewsQuery;
+};
```
calculuschild, line 49:
```suggestion
const buildBrewsQuery = (legacy, v3) => {
    const brewsQuery = { published: true };

    if (legacy && !v3)
        brewsQuery.renderer = 'legacy';
    else if (v3 && !legacy)
        brewsQuery.renderer = 'V3';
 
    return brewsQuery;
};
```

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+
+const vault = {
```
calculuschild, line 51: Why is this inside a `vault` object? Just for organization?

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+
+    console.error(`Error in ${functionName}:`, error);
+
+    return res.status(status).json({
+        errorCode: status.toString(),
+        message: `Error in function ${functionName}: ${message}`,
+    });
+};
+
+const buildBrewsQuery = (legacy, v3) => {
```
calculuschild, line 38: Would it make more sense to rename this "buildRendererConditions" and just put the `published : true` down in the `combinedQuery` since it's always there by default? Let this function handle just one field?

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+    const message =
+        status === 503 ? 'Service Unavailable' : 'Internal Server Error';
+
+    console.error(`Error in ${functionName}:`, error);
+
+    return res.status(status).json({
+        errorCode: status.toString(),
+        message: `Error in function ${functionName}: ${message}`,
+    });
+};
```
calculuschild, line 36: Instead of a whole separate error handler, can we just utilize the central error handler we already have in `app.js`? That's what `asyncHandler()` is for. It passes error objects along. All you need to do is throw the error object and it will land there and be sent. You can customize any of the properties in the object. The whole error object will be logged, it already has logic to handle the response status, and if you add an `HBErrorCode` we can track where it came from.

See:

https://github.com/naturalcrit/homebrewery/blob/c41116e6c8eadc418b1dcf799dd3a83c74d5bcae/server/homebrew.api.js#L136-L138

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+                .exec();
+
+            console.log(
+                'Query in findBrews: ',
+                JSON.stringify(combinedQuery, null, 2)
+            );
+            return res.json({ brews, page });
+        } catch (error) {
+            console.error(error);
+            return handleErrorResponse(res, error, 'findBrews');
```
calculuschild, line 91: For example:

```suggestion
            throw {...err, message: "Error finding brews in Vault search", HBErrorCode: '99');
```

server/vault.api.js
```diff
@@ -0,0 +1,128 @@
...
+    } else if (v3 === 'true' && legacy !== 'true') {
+        brewsQuery.renderer = 'V3';
+    }
+
+    return brewsQuery;
+};
+
+const vault = {
+    findBrews: async (req, res) => {
+        try {
```
calculuschild, line 53: Generally, we have preferred `await.catch()` instead of `try... catch`. Not a huge deal, but may be worth putting it in a consistent format.

See:

https://github.com/naturalcrit/homebrewery/blob/c41116e6c8eadc418b1dcf799dd3a83c74d5bcae/server/homebrew.api.js#L260-L265

calculuschild (changes requested):
Reviewed the `vault.api.js` code. Will do `vaultPage.jsx` next. All the other changed files look fine to me.

client/homebrew/pages/homePage/homePage.jsx
```diff
@@ -76,6 +76,7 @@ const HomePage = createClass({
 				}
 				<NewBrewItem />
 				<HelpNavItem />
+				<VaultNavItem />
```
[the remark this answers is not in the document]
5e-Cleric: Have thought about this, it currently is in homepage and userpage, i think that is fine, specially for now. If we were to simplify our navigation as i have suggested, we could have it everywhere.

Related to this, do you think the link should open a new page or navigate in the same page?

server/app.js
```diff
@@ -497,6 +503,11 @@ app.use(async (err, req, res, next)=>{
 		res.status(err.status || err.response?.status || 500).send(err);
 		return;
 	}
+	if(err.originalUrl?.startsWith('/vault/')) {
+		// console.log('vault error');
+		res.status(err.status || err.response?.status || 500).send(err);
+		return;
+	}
```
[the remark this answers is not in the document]
5e-Cleric: This is like the first piece of code i wrote when i separated this from admin, i think you are absolutely right.

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,473 @@
...
+const Account = require('../../navbar/account.navitem.jsx');
+const NewBrew = require('../../navbar/newbrew.navitem.jsx');
+const HelpNavItem = require('../../navbar/help.navitem.jsx');
+const BrewItem = require('../basePages/listPage/brewItem/brewItem.jsx');
+const SplitPane = require('../../../../shared/naturalcrit/splitPane/splitPane.jsx');
+
+const request = require('../../utils/request-middleware.js');
+
+const VaultPage = (props) => {
+    const [title, setTitle] = useState(props.query.title || '');
```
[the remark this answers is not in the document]
5e-Cleric: Does our linter handle jsx? because this code has been written, rewritten, and past through many places, i can take a look at my prettier config to see if i can change those to tabs.

server/vault.api.js
5e-Cleric: I do prefer it spanning a few lines, like my CSS, if i can choose here

server/vault.api.js
5e-Cleric: applied

server/vault.api.js
5e-Cleric: replaced with trows

server/vault.api.js
5e-Cleric: It would make sense too, yeah

server/vault.api.js
5e-Cleric: That is a good excuse, remember this PR is my first contact with backEnd work, i guess no vault object is required since the whole file is vault, and they could all use router. Want that change?

5e-Cleric: good point, the entire file is vault, no need for that.

server/vault.api.js
5e-Cleric: had to ask the toaster for a difference between the two, alright

5e-Cleric:
Per request from @dbolack-ab , usernames containing emails are hidden to prevent user info leakage:

image

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+					console.log('error at loadPage: ', error);
+					setError(
+						`${
+							error.response
+								? error.response.status
+								: error.message
+						}`
+					);
+					updateStateWithBrews([], 1);
+				}
```
calculuschild, line 99: Suggest using `await.catch`  on these spots too. Something like:

```suggestion
				const response = await request.get(
					`/api/vault?title=${title}&author=${author}&v3=${v3}&legacy=${legacy}&count=${count}&page=${page}`
				).catch((error)=>{
					console.log('error at loadPage: ', error);
					setError(`${error.response
						? error.response.status
						: error.message}`
					);
					updateStateWithBrews([], 1);
				});
				
				if (response.ok)
					updateStateWithBrews(response.body.brews, page);
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+		//Different searches use the update or total props to make only the necessary queries and functions
+
+		if (!validateForm()) {
+			return;
+		}
+
+		setSearching(true);
+		setError(null);
+
+		const performSearch = async ({ title, author, count, v3, legacy }) => {
```
calculuschild, line 76: This is re-defining `performSearch` and `loadTotal` every time `loadPage` is called. These functions should be moved outside of `loadPage`.

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+	);
+
+	const validateForm = () => {
+		//form validity: title or author must be written, and at least one renderer set
+		const { current: titleInput } = titleRef;
+		const { current: legacyCheckbox } = legacyRef;
+		const { current: v3Checkbox } = v3Ref;
+		const { current: authorInput } = authorRef;
+
+		const isTitleValid = titleInput.validity.valid && titleInput.value;
```
calculuschild, line 175: Why not just
```suggestion
		const isTitleValid = titleRef.current.validity.valid && titleInput.value;
```

on all of these?

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+		const isFormValid =
+			(isTitleValid || isAuthorValid) && isCheckboxChecked;
+
+		return isFormValid;
+	};
+
+	const disableSubmitIfFormInvalid = () => {
+		const { current: submitButton } = searchButtonRef;
+		submitButton.disabled = !validateForm();
+	};
```
calculuschild, line 189: Similarly

```suggestion
	const disableSubmitIfFormInvalid = () => {
		submitButtonRef.current.disabled = !validateForm();
	};
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+						})}
+					/>
+				</button>
+			</div>
+			<legend>
+				<h3>Tips and tricks</h3>
+				<ul>
+					<li>
+						You can only search brews with this tool if they are
+						published
```
calculuschild, line 289:
```suggestion
						Only <b>published</b> brews are searchable via this tool
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+					</li>
+					<li>
+						Usernames are case sensitive, make sure you are writing
+						it correctly
```
calculuschild, line 293:
```suggestion
						Usernames are case-sensitive,
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+					</li>
+					<li>
+						You can use <code>-</code> to negate words, assuming
+						there is any word not negated, and <code>"word"</code>
+						to specify an exact string.
```
calculuschild, line 298:
```suggestion
						Use <code>"word"</code> to match an exact string, and
						<code>-</code> to exclude words (at least one word
						must not be negated) .
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+					</li>
+
+					<li>
+						Some words like a, after, through, itself, or here, are
+						ignored in searches, make sure your search has relevant
+						words. The full list can be found &nbsp;
+						<a href="https://github.com/mongodb/mongo/blob/0e3b3ca8480ddddf5d0105d11a94bd4698335312/src/mongo/db/fts/stop_words_english.txt">
+							here
+						</a>
```
calculuschild, line 307:
```suggestion
						Some common words like "a", "after", "through", "itself", "here", etc.,
						are ignored in searches. The full list can be found
						<a href="https://github.com/mongodb/mongo/blob/0e3b3ca8480ddddf5d0105d11a94bd4698335312/src/mongo/db/fts/stop_words_english.txt">
							here
						</a>
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+					</li>
+				</ul>
+			</legend>
+			<small></small>
```
calculuschild, line 311: This `<small>` is empty?

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
+require('./vaultPage.less');
+
+const React = require('react');
+const { useState, useEffect, useRef } = React;
+const cx = require('classnames');
```
calculuschild, line 5: Shouldn't need to use `cx`. This was for very old React.

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,478 @@
...
+	//state author
+	const [author, setAuthor] = useState(props.query.author || '');
+	const [legacy, setLegacy] = useState(props.query.legacy !== 'false');
+	const [v3, setV3] = useState(props.query.v3 !== 'false');
+	const [count, setCount] = useState(props.query.count || 20);
+	const [page, setPage] = useState(parseInt(props.query.page) || 1);
+	const [brewCollection, setBrewCollection] = useState(null);
+	const [totalBrews, setTotalBrews] = useState(null);
+	const [searching, setSearching] = useState(false);
+	const [error, setError] = useState(null);
```
calculuschild, line 29: Most of these states are not used anywhere that I can see, but the same names are used in several functions as parameters. Please rename these or the ones in the function parameters so it is clear when the state is being used and when it is a normal function variable.

calculuschild (changes requested):
More cleanup focusing on `vaultPage.jsx`

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,473 @@
...
+                        defaultValue={author}
+                        onKeyUp={disableSubmitIfFormInvalid}
+                        onKeyDown={(e) => {
+                            if (e.key === 'Enter') {
+                                if (!searchButtonRef.current.disabled) {
+                                    loadPage(1, true, true);
+                                }
+                            }
+                        }}
+                        placeholder="Gazook89"
```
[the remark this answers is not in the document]
5e-Cleric: Of course authors are case sensitive, the only case insensitive here is the title, and because we have is set that way.

client/homebrew/pages/vaultPage/vaultPage.jsx
5e-Cleric: States are derived from props and used to fill inputs with values from the query. I suppose i could add them directly to the inputs as `defaultValue={props.query.xxx}`

client/homebrew/pages/vaultPage/vaultPage.jsx
5e-Cleric: replaced with cleaner use of classes

client/homebrew/pages/vaultPage/vaultPage.jsx
5e-Cleric: fixed

server/vault.api.js
```diff
@@ -0,0 +1,119 @@
...
+			console.error(error);
+			throw {...err, message: "Error finding brews in Vault search", HBErrorCode: '99'}; 
+		});
+};
+
+const findTotal = async (req, res) => {
+	const title = req.query.title || '';
+	const author = req.query.author || '';
+
+	const brewsQuery = buildRendererConditions(req.query.legacy, req.query.v3);
```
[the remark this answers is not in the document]
5e-Cleric: You suggested this name, but yes, sure

server/vault.api.js
```diff
@@ -0,0 +1,119 @@
...
+	};
+};
+
+const buildAuthorConditions = (author) => {
+	if (!author) return {};
+	return { authors: author };
+};
+
+const buildRendererConditions = (legacy, v3) => {
+	const brewsQuery = {};
```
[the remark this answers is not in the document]
5e-Cleric: renamed all those as `rendererConditions`, `titleConditions`, and `authorConditions`, i would keep it that way.

client/homebrew/pages/vaultPage/vaultPage.jsx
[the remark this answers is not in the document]
ericscheid: Author values in the database are case sensitive.

We should treat the user's search criteria as case insensitive though.

5e-Cleric:
Applied all suggested changes, please do check and answer if possible to the unresolved ones.

server/vault.api.js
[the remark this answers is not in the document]
calculuschild: I mean `brewsQuery` should just be `renderer`

server/vault.api.js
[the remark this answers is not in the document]
calculuschild: You misread.

You can keep `buildRendererQuery`>. But the variable `brewQuery` should now be `rendererQuery`

client/homebrew/pages/vaultPage/vaultPage.jsx
calculuschild: My main concern is we have states and function arguments with the same name, so it is difficult to read which one is being used.

calculuschild: For example

```
        const updateUrl = (title, author, count, v3, legacy, page) => {
```

This function has several parameters which identical names to the state variables. Either these or the states should probably be renamed so it's not ambiguous.

server/vault.api.js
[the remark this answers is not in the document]
5e-Cleric: ah, right, yes

client/homebrew/pages/vaultPage/vaultPage.jsx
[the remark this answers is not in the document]
Gazook89: Can you use a different placeholder other than my account name-- perhaps just `homebrewery`?

Or, i suggest no placeholders at all.  I don't think there is much gained by them in this case (the required input isn't complicated).  And they styling of placeholders isn't sufficiently different enough from normal text in the input, so there is a chance a user may think the fields are already filled (the gray is fairly similar to black text).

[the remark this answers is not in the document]
5e-Cleric: > Author values in the database are case sensitive.
> 
> We should treat the user's search criteria as case insensitive though.

Been looking into that, because it is not a text search we should have a [case insensitive index for this to work](https://www.mongodb.com/docs/manual/core/index-case-insensitive/). Not sure that is the way to go. Maybe for a later PR.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx
```diff
@@ -141,11 +141,15 @@ const BrewItem = createClass({
...
-  					</>))}
+				<i className='fas fa-user'/> 
+				{brew.authors?.map((author, index) => (
+					<React.Fragment key={index}>
+						{author === 'hidden'
+							? <span title="Username contained an email address; hidden to protect user's privacy">{author}</span>
+							: <a href={`/user/${author}`}>{author}</a>
+						}
+					</React.Fragment>
+				))}
```
calculuschild, line 152:
```suggestion
					<i className='fas fa-user'/> 
					{brew.authors?.map((author, index) => (
						<React.Fragment key={index}>
							{author === 'hidden'
								? <span title="Username contained an email address; hidden to protect user's privacy">{author}</span>
								: <a href={`/user/${author}`}>{author}</a>
							}
						</React.Fragment>
					))}
```

calculuschild:
@5e-Cleric 

Could we move the email filtering to occur in `vault.api.js` as part of fetching the brews, so the emails are already filtered out by the time they arrive at the client?

5e-Cleric:
> @5e-Cleric
> 
> Could we move the email filtering to occur in `vault.api.js` as part of fetching the brews, so the emails are already filtered out by the time they arrive at the client?

uh, sure, perhaps i should implement a stronger pattern too. To catch only emails

calculuschild:
Something is going on with the storage icon. All results are showing the NaturalCrit logo, even if stored on Google Drive, and the NaturalCrit logo is not transparent. (Desert Maw is stored on Google Drive)

image

I actually think we don't need to show the storage location in the Vault, but with the recent issues with thumbnail over text, I'm wondering what is happening that the existing CSS for `brewItems` works on the userpage but not in the Vault.

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,401 @@
...
+	};
+
+	const performSearch = async (title, author, count, v3, legacy, page)=>{
+		updateUrl(title, author, count, v3, legacy, page);
+
+		const response = await request.get(
+			`/api/vault?title=${title}&author=${author}&v3=${v3}&legacy=${legacy}&count=${count}&page=${page}`
+		).catch((error)=>{
+			console.log('error at loadPage: ', error);
+			setError(`${error.response
```
calculuschild, line 68: Just noticed. `setError` here is basically erasing most of the error object. You lose the `HBError` property for example.

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,401 @@
...
+	};
+
+	const loadTotal = async (title, author, v3, legacy)=>{
+		setTotalBrews(null);
+
+		const response = await request.get(
+			`/api/vault/total?title=${title}&author=${author}&v3=${v3}&legacy=${legacy}`
+		).catch((error)=>{
+			console.log('error at loadTotal: ', error);
+			setError(`${error.response
```
calculuschild, line 86: Same here.

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -0,0 +1,401 @@
...
+		if(error) {
+			const errorText = ErrorIndex({ brew })[brew.HBErrorCode.toString()] || '';
+			console.log('render Error: ', error);
+
+			return (
+				<div className='foundBrews noBrews'>
+					<h3>Error: {errorText}</h3>
+				</div>
+			);
+		}
```
calculuschild, line 345: A few things here:

- `brew` will crash because it isn't defined. You don't need to pass in `{brew}` or anything if the error you are reading doesn't use them.

- Use `error.HBErrorCode` instead, assuming you fix `setError` above to not remove it.

- Don't need the `console.log` here. It is already logged above when you `setError`

calculuschild (approved):
Ok, I have made all my tweaks.

Just one more set of issues I found that crash the page when trying to render an error message. Feel free to merge after fixing.

5e-Cleric:
All fixed again

calculuschild:
Gonna merge now..... Good work @5e-Cleric . This is a big one.

PR #3264, opened 2024-01-23
Fix of admin page

5e-Cleric:
I have only fixed the brew lookup function and the stats function, will update more in the future


5e-Cleric:
All functions updated to use promises instead of callbacks.
Some queries were fixed to be able to execute them more than once, as [indicated by mongo](https://mongoosejs.com/docs/migrating_to_6.html#duplicate-query-execution).

Mostly done by the toaster, not a backend engineer.

PR #3267, opened 2024-01-26
Clone a brew's tags when cloning a brew.

dbolack-ab:
Per Element conversation.


calculuschild:
Easy enough. Looks good! Thanks @dbolack-ab ⭐ 🎉 💯

Issue #3231, opened 2024-01-09
Brew title parsed as markdown on Error pages. e.g. images will be rendered.


calculuschild:
Found this when creating a test brew where the title was auto-generated from the first line, which happened to be an image.

Looks like brew titles are being parsed as Markdown when displayed in error pages:

![image](https://github.com/naturalcrit/homebrewery/assets/5878534/b20e9cb0-d6b2-4b60-857e-52758854b9ea)

This should only really affect users trying to access their own document, but is definitely not the expected behavior. I could see potential avenue for abuse if people are trying to load scripts/malicious links into the title and then distributing their edit page around.

PR #3271, opened 2024-01-26
Fix for Brew title parsed as markdown on Error pages. e.g. images will be rendered.

dbolack-ab:
This resolves issue #3231 by performing a non-destructive lazy sanitize on HTML tags that might exist in the props.brew.brewTitle when passed to an error page.

### DIFF
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -2,7 +2,12 @@ const dedent = require('dedent-tabs').default;
 
 const loginUrl = 'https://www.naturalcrit.com/login';
 
+const lazySantizize = (s)=>{
+	return s.replace('<', '&lt;');
+};
+
 const errorIndex = (props)=>{
+	const cleanBrewTitle = lazySantizize(props.brew.brewTitle);
 	return {
 		// Default catch all
 		'00' : dedent`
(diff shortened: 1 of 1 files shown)


calculuschild:
This might fix a raw HTML image, but unfortunately this doesn't fix the issue in #3232. The problem is that the title is *markdown* at that point, so `<` isn't quite enough.

I think a more universal approach might be `encodeURIComponent` or something like we to when generating file names for download: https://github.com/naturalcrit/homebrewery/blob/da11089fffb2776302e54c603dc34de2f405ebb3/server/app.js#L194-L199

Or just handle more characters as we do when sanitizing for the "raw" source page:
https://github.com/naturalcrit/homebrewery/blob/da11089fffb2776302e54c603dc34de2f405ebb3/server/app.js#L179-L183

See:
https://homebrewery-pr-3271.herokuapp.com/edit/Fk06enjm68C0

dbolack-ab:
> This might fix a raw HTML image, but unfortunately this doesn't fix the issue in #3232. The problem is that the title is _markdown_ at that point, so `<` isn't quite enough. See: https://homebrewery-pr-3271.herokuapp.com/edit/Fk06enjm68C0
> 
> I think a more universal approach might be `encodeURIComponent` or something like we to when generating file names for download:
> 
> https://github.com/naturalcrit/homebrewery/blob/da11089fffb2776302e54c603dc34de2f405ebb3/server/app.js#L194-L199
> 
> Or just handle more characters as we do when sanitizing for the "raw" source page:
> 
> https://github.com/naturalcrit/homebrewery/blob/da11089fffb2776302e54c603dc34de2f405ebb3/server/app.js#L179-L183

I went with adding more characters. I have to wonder though, why is the title getting parse?

calculuschild:
> I have to wonder though, why is the title getting parse?

The whole error text is written in Markdown and then parsed for display.

You know what though.... This whole thing would maybe be simpler if we just put `<pre>` tags around the title, which prevents Markdown parsing.... I'm dumb.

dbolack-ab:
> > I have to wonder though, why is the title getting parse?
> 
> The whole error text is written in Markdown and then parsed for display.
> 
> You know what though.... This whole thing would maybe be simpler if we just put `\<pre\>` tags around the title, which prevents Markdown parsing.... I'm dumb.

Oh, lordy, that's so much more elegant.
 I'm closing this. :)

PR #3277, opened 2024-01-29
Possible fix for multiple inline divs in a block

dbolack-ab:
This runs a split on the html input on inline divs and then parses each singly.

This solves the issue of 

`<div>1</div><div>2</div><div>3</div><div>4</div><div>5</div>`

producing 

```
<div>
<p>1</p>
</div>
<div>
2
</div>
<div>
3
</div>
<div>
4
</div>
<div>
5
</div>
```

However the results instead are 
```
<div>
<p>
1
</p>
</div>
<div>
<p>
2
</p>
</div>
<div>
<p>
3
</p>
</div>
<div>
<p>
4
</p>
</div>
<div>
<p>
5
</p>
</div>
```

If the paragraph wrappers are undesired, we either need to not run parse on the div contents *or* strip the paragraph wrapper.


G-Ambatte:
The tests are failing because the final `</div>` is being duplicated, at least some of the time:

image

image

I haven't had a chance to look at the actual code yet to see what might be causing this duplication.

dbolack-ab:
This does not fix nested divs

calculuschild:
Does this have an issue it's linked to?

Our current behavior of parsing markdown inside block-level divs already deviates from the Commonmark standard. This alters it even more (and suggests we should also accommodate `<span>` and other structures?

If we have no specific user case that needs this behavior, I might suggest it's better to leave this alone. We should be encouraging users to avoid raw HTML in V3 anyway.

dbolack-ab:
Closing pending discussion.

PR #3278, opened 2024-01-31
Add type attribute to all stylesheet import link tags

5e-Cleric:
Initially started to fix a MIME error with FF:

> The stylesheet https://homebrewery.naturalcrit.com/homebrew/cm-themes/default.css was not loaded because its MIME type, “text/html”, is not “text/css”.

I just went with it and fixed all of them.


calculuschild (commented):
Not sure why, but font awesome icons aren't appearing

calculuschild:
Thanks @5e-Cleric !

PR #3284, opened 2024-02-06
Tag trim

5e-Cleric:
Right now this PR passes the uniqueness test in stringArrayEditor, but it should not pass, i need to fix that, somehow.

Added console logs to see what is happening


5e-Cleric:
Could use someone's help in figuring out why trimming where i did it makes it pass the unique tag test.

Gazook89:
Can you add more info as to what this PR is trying to solve, and what you mean about not passing the 'uniqueness test'?

calculuschild:
> 1. The Properties Editor could show "invalid/red input" if there is ever a tag that is `a duplicate` + `trailer space`....so with tags `test` and `test `, it would turn the input red and prevent adding the second tag. This works by adding the `trim()` method to the `valueIsValid` function:

I would vote for this approach. I think `validations.js` is currently indented solely to detect specific patterns, but not for uniqueness against other values. It makes sense to just update the uniqueness check we already have.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -62,16 +62,26 @@ const MetadataEditor = createClass({
 	},
 
 	handleFieldChange : function(name, e){
+		
+		const trimmedValue = (()=> {
+			if(e.target.value.length === 0) {
+				return e.target.value;
+			}
+			return [...e.target.value.slice(0, -1),e.target.value[e.target.value.length - 1].trim()]; // remove trailing whitespace from last item in list
+		})();
```
calculuschild, line 71: I think this can be simplified to this:

```
if(e.target.value.length >  0) {
  e.target.value[e.target.value.length - 1] = e.target.value[e.target.value.length - 1].trim();
}
```

Alternatively, we may just want to trim all the values anyway, because you can have the same uniqueness problem if you edit a tag earlier in the list by adding spaces:

```
const trimmedValues = e.target.values.map((str) => str.trim());
```

See I can add a space to the first tag and the value will not be trimmed.
image

5e-Cleric:
I can go ahead and implement that, and i could add a simple regex pattern to the input, that allows for way easier styling than via js.

```
input:invalid {
    background:red;
}
```
In case you've never heard of that, `:valid` and `:invalid` are a thing, and work great with the `pattern` attribute.

calculuschild:
> i could add a simple regex pattern to the input, that allows for way easier styling than via js.

I think any pattern checking is already handled with `valueIsValid()` and `validators.js`. I would prefer we keep any of that logic in one place. I think we even used `invalid` at the start, but we found that just limiting to regex was not enough, hence the custom solution.

Gazook89:
@5e-Cleric I have a full rework of the stringArrayEditor in PR #3469.  I can fold the issue/answer of this draft PR into that one, and close this one--- is that okay with you?

calculuschild:
That's fine.

5e-Cleric:
> @5e-Cleric I have a full rework of the stringArrayEditor in PR #3469.  I can fold the issue/answer of this draft PR into that one, and close this one--- is that okay with you?

Please do.

PR #3289, opened 2024-02-08
New Brew from Local file

5e-Cleric:
This adds a functionality to create a new brew in the new page using contents from a local `txt` file.

Like the cloning, this erases the previous contents of a new brew.

### ToDo
- [x] fix nav styling to fit everything
~- [ ] change alert for a nice popup error message~


client/homebrew/navbar/newbrew.navitem.jsx
```diff
@@ -1,12 +1,95 @@
...
-		color='purple'
-		icon='fas fa-plus-square'>
-		new
-	</Nav.item>;
+const BREWKEY  = 'homebrewery-new';
+const STYLEKEY = 'homebrewery-new-style';
+const METAKEY  = 'homebrewery-new-meta';
+
+const NewBrew = () => {
+    const inputRef = useRef(null);
```
calculuschild, line 12: Seems this `inputRef` isn't necessary.

5e-Cleric:
You were right, that was over-complicating stuff, went for a more traditional approach for the simulated click.

client/homebrew/navbar/newbrew.navitem.jsx
```diff
@@ -1,12 +1,95 @@
...
+
+const NewBrew = () => {
+
+    
+    const [brew, setBrew] = useState({
+        text: '',
+        style: ''
+    });
+
+    const splitTextStyleAndMetadata = (brewContent) => {
```
calculuschild, line 19: Any chance we can import this function to avoid duplicating it? Maybe extract it from `app.js` into a `/shared/helpers.js` file or something so both the server and the client can use it.

client/homebrew/navbar/newbrew.navitem.jsx
```diff
@@ -1,12 +1,95 @@
...
-		new
-	</Nav.item>;
+const BREWKEY  = 'homebrewery-new';
+const STYLEKEY = 'homebrewery-new-style';
+const METAKEY  = 'homebrewery-new-meta';
+
+const NewBrew = () => {
+
+    
+    const [brew, setBrew] = useState({
```
calculuschild, line 14: I don't think this is needed either. Once the brew is uploaded, we immediately redirect to `/new`, so there's no need to save it to state.

calculuschild (changes requested):
Functionality seems good. Just some cleanup requests.

client/homebrew/navbar/newbrew.navitem.jsx
```diff
@@ -1,12 +1,95 @@
...
+				} else {
+                    alert('This file is invalid, please, enter a valid file');
+                }
+            };
+            reader.readAsText(file);
+        }
+    };
+
+    return (
+        <Nav.dropdown>
```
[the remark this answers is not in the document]
5e-Cleric: done

client/homebrew/navbar/newbrew.navitem.jsx
5e-Cleric: helpers sounds extremely vague
what about `fileProcessing.js` or something like that

5e-Cleric: can we do this in a posterior PR? log it in as an issue, then merge this?

client/homebrew/navbar/newbrew.navitem.jsx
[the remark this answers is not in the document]
calculuschild: Please update the icons as well.

client/homebrew/navbar/newbrew.navitem.jsx
calculuschild: `helpers` is a pretty standard practice for small functions that need to be shared between multiple areas, and isn't limited to just file processing. `helpers.js` would be a file that potentially contains many small functions that are needed in different places but don't deserve a whole file all to themselves. Alternative names I see a lot are "utilities" or "tools".

> can we do this in a later, separate PR? log it in as an issue, then merge this?

I would rather we just address this now; adding one problem (code duplication) to fix another problem seems like an anti-pattern, especially where the fix is quite easy.

5e-Cleric:
Should be all done and fixed!

client/homebrew/navbar/newbrew.navitem.jsx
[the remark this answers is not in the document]
5e-Cleric: solved

calculuschild:
Looks good! Merging now!

PR #3291, opened 2024-02-09
Small navigation fixes

5e-Cleric:
Changing padding and icon positioning to allow for longer texts with correct wrapping of icons and words

also fixing three misalignments:

misalignment


Gazook89:
An alternate method is to use flex and align-items.

shared/naturalcrit/nav/nav.less
```diff
@@ -87,10 +87,15 @@ nav{
 				position         : relative;
 				display          : block;
 				width            : 100%;
-				vertical-align   : middle;
-				padding          : 8px 5px;
+				padding			 : 5px 20px 5px 12px;
```
calculuschild, line 90: While you are unifying the alignments, let's also keep the top and bottom margins the same (change 5px to 8px to match the top-level button)

shared/naturalcrit/nav/nav.less
```diff
@@ -87,10 +87,15 @@ nav{
 				position         : relative;
 				display          : block;
 				width            : 100%;
-				vertical-align   : middle;
-				padding          : 8px 5px;
+				padding			 : 5px 20px 5px 12px;
 				border           : 1px solid #888;
 				border-bottom    : 0;
+				i:before {
```
calculuschild, line 93: Is there a possibility of doing this on the `i` element itself instead of the `before`?

Also, if you want everything to line up with the top-level header, perhaps it makes sense to just merge this with the `.navItem i` up on line 65 so we don't have two different places that define how the icons work.

calculuschild (changes requested):
I like the idea of this PR. However I do notice that this does really squeeze the dropdown menus, especially "Need Help?" which forces "Report Issue" onto 2 lines, and "Share", which forces "Post to Reddit" onto _3_ lines which looks bad.

I wonder if it might work better to have some of the dropdowns set their own width, kind of like the Snippets, so we don't have weird cases like that?

shared/naturalcrit/nav/nav.less
calculuschild: Also this large 20px right padding makes any menu items without icons squeezed over in an odd way.

image

5e-Cleric: will add icons

5e-Cleric (commented):
All adressed

calculuschild (changes requested):
After fixing #3292 it looks like this PR has a merge conflict. Need to make the same CSS update here or rebase this on the new master.

Also, I apologize for this, but the new appearance looks a little funny to me with the really wide buttons. I didn't think about this before. Auto-save in particular looks a little weird being off-center.

image

What if we did more like this? Rather than set a min-width for each top nav item, we just set the `.navDropdown` style to `width: max-content` and `min-width: 100%`. It would allow dropdowns to grow wider as needed (like the Snippets), but be at least as wide as the top button. For example:

image

This would also simplify the CSS because no need to set individual width values for each button.

Gazook89:
One trick is that the last menu item, the Account menu, might stick off the page.  Of course, you could have the last menu stay the way it is, too.

5e-Cleric:
This is why dropdowns are usually not that common, why most pages have a big dropdown with sections, rather than very small dropdowns.

but hey, if the CIA can do it, so can we:
image

PR #3292, opened 2024-02-11
Combine nav and navbar styles

Gazook89:
This PR tidies up the navbar styling by moving the styles in the `/shared/` directory in `nav.less` to the `/client/` directory, into the `navbar.less` file.  This reduces overlap, and hopefully minimizes confusion when making changes to the navbar.  Also, in the process, it became clear that there is a little bit of refactoring (particularly around the Save menu) that could bring things together better.  However, beyond grouping 'like' rules together,  **no actual changes were made to the processed CSS in this PR.**

- moves all the nav.less styling to navbar.less in the `client` directory
- deletes nav.less
- changes the nav.jsx import of styles to navbar.less
- stylelint navbar.less
- added a couple comments about easy future changes.

-----

Next Steps:
1. Save menu refactor so that it isn't using it's own dropdown container structure
2. Simplify styling of `navItem` so it is consistent across different item/menu types
3. Replace common color hex codes with variables
4. Likewise with fonts


client/homebrew/navbar/navbar.less
```diff
@@ -1,272 +1,350 @@
...
+				}
+				&.recent {
+					position : relative;
+					.navDropdown .navItem {
+						#backgroundColorsHover;
+						.animate(background-color);
+						position         : relative;
+						box-sizing       : border-box;
+						display          : block;
+						max-height       : ~"calc(100vh - 28px)";  // I don't think is correct syntax, but leaving it in for now... (Gazook89)
```
5e-Cleric, line 267: if you are worried about the calc itself, it is valid, substracting 28px from the total height of the window. https://developer.mozilla.org/en-US/docs/Web/CSS/calc

the `~""` is invalidating the statement, so i don't get it either.

5e-Cleric (approved):

client/homebrew/navbar/navbar.less
5e-Cleric: Gitter today:
> From memory, that's to maintain the calc() as CSS, because otherwise LESS will try to execute it's own calc function
A quick Google shows that there is an option from v1.4 onwards that will disable this behaviour, but changes the required format for all actual LESS calculations

So we can leave that right there

5e-Cleric:
Merging in 24 hours unless someone disagrees

calculuschild:
> Merging in 24 hours unless someone disagrees

I haven't been able to test this but if you have verified nothing looks broken go ahead.

5e-Cleric:
i went to re review it, found an issue, small, but we lost the orange color in the save button, is that affected by this PR?

calculuschild (approved):

PR #3296, opened 2024-02-12
Fix lang on print

5e-Cleric:
Should fix hyphenation on print pages


calculuschild:
Easy fix. Thanks @5e-Cleric ! 🎉 💯 ⭐

Issue #3275, opened 2024-01-29
Framed monster block background image doesn't appear in the PDF


RKuerten:
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

When generating a PDF, the .monster.frame is always generated with the background color, ignoring the background image.

I tested with different background images, with different background-blend-modes (mainly normal, lighten, darken, and multiply), the result is always the same (no background image for the monster block in the PDF).

![Print difference](https://github.com/naturalcrit/homebrewery/assets/30464993/b15b4bbf-9af4-4e6a-81b7-40802f476c82)
Browser vs Generated PDF

Browsers tested: 
Brave V 1.62.153 w/ Chromium 121.0.6167.85 64-bits
Chrome V 121.0.6167.86 64-bits

Gazook89:
I just test with a new brew, featuring only a framed stat block, with Chrome on Mac (viewed in Preview), and had no issue.  Hrm.

Are you using the default PDF printer, or something else (CutePDF etc)?  

Is this true with a brand new brew, with no custom css?  Or, can you provide the Share link to your brew that isn't working?

RKuerten:
I can't share the brew publicly because it is a product yet to be released on the DMsGuild. I'm using the default printer (on Windows, btw). 

I've tested with a brand new brew, and it works. It also works by copying a single page of the brew with a stat block with the custom css. The brew is not that big (30 pages), but there's basically an image in every page along side the stat blocks. Could be some memory issue.  

Once I add the pages (cover page & toc + intro page) that should be before the first page with a stat block, the background disappears. If delete both pages, only the first stat block has the background, the others default to the background color.

RKuerten:
I've circumvented the problem by adding the background texture as a before element in the .monster.frame blocks. The code for it looks like this (for .wide blocks, it is equal except for the width, which is 102.1%). 

```
.page .monster.frame::before  {
  background-color     : #F7F0D4;
  background-image     : url(https://i.imgur.com/owy8gdY.jpg);
  background-blend-mode: normal;
  background-size      : 100% 100%;
  background-repeat    : no-repeat;
  background-position  : top center;
  content              : "";
  position             : absolute;
  width                : 104.1%;
  height               : 100%;
  top                  : 0;
  left                 : -7px;
  z-index              : -3;
}
```

I'm closing this issue.

5e-Cleric:
New brew just created:
[...]

PR #3297, opened 2024-02-12
Changed background-attachment: fixed on the monster frame

RKuerten:
Fix for #3275


5e-Cleric:
Merging this, small, nonbreaking change, thanks @RKuerten

PR #3299, opened 2024-02-13
popup update 3.11

5e-Cleric:


Gazook89 (changes requested):
You should be sure to clean up the capitalization.  "Google Drive", "Google Photos".  And fix "reccomend" --> "recommend".  imgur is correct, though...no capitalization needed.

And I would finish it with "...such as imgur, imgbb, or Google Photos." but that is my opinion.

calculuschild:
Easy review. Thanks @5e-Cleric !

PR #3303, opened 2024-02-17
Use a brew as a theme, prep work for user sharable theming.

dbolack-ab:
This PR explores the idea of User themes being derived from brew documents. It does not address the notion of permissions ( beyond normal brew access ) or tagging a brew as a theme document. The only limitation it places is on the csstheme endpoint not including the theme if it uses the legacy renderer.

This has been implemented in three different ways to allow for comparison and discussion

- /api/css/:id : This returns the style frontmatter of the referenced document as a text/css document. 
- /api/theme/:id : This returns an object with the reference'd object's theme and style frontmatter.
- /api/csstheme/:id : This returns the stylye frontmatter of the referenced document as a text/css document and adds the theme as an @import ( if not using the legacy renderer )

I prepared the PR this way to make comparing and contrasting the different methods easier.

## Update

Based on revisiting earlier conversations on theme parenting, I have removed the first two options and renamed the path for the third as /css/. 

I have added an additional option for parental recursion on static themes. 

 - /api/css/[renderer]/[theme] will import the parent theme ( if present ) via the API path and the theme via its static path and . For example, /api/css/V3/5eDMG will result in 
 ```
 @import /api/css/V3/5ePHB
 @import /themes/V3/5eDMG/style.css
```

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -255,7 +255,7 @@ const api = {
 				res.set('Content-Type', 'text/css');
 			}
 			const parentTheme = themeParent ? `@import /api/css/${req.params.engine}/${themeParent}\n` : '';
-			return res.status(200).send(`${parentTheme}@import /themes/${req.params.engine}/${req.params.id}\n`);
+			return res.status(200).send(`${parentTheme}@import /themes/${req.params.engine}/${req.params.id}/style.css\n`);
 		}
 	},
 	updateBrew : async (req, res)=>{
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -11,6 +11,19 @@ const { nanoid } = require('nanoid');
 
 const { DEFAULT_BREW, DEFAULT_BREW_LOAD } = require('./brewDefaults.js');
 
+const themes = require('../themes/themes.json');
+
+const isStaticTheme = (engine, themeName)=>{
+	if(!themes.hasOwnProperty(engine)) {
+		return undefined;
+	}
+	if(themes[engine].hasOwnProperty(themeName)) {
+		return themes[engine][themeName].baseTheme;
+	} else {
+		return undefined;
+	}
+};
+
 // const getTopBrews = (cb) => {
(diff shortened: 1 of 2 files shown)


5e-Cleric:
the initial idea was to have a this as an alternative to brew themes, so the idea would be to store this style in the db as a new kind of object, and pull that in the brew theme selector.

I am aware that such a functionality is outside of the scope of this particular PR, but we can make this follow a syntax compatible with other brew themes.

calculuschild:
> the initial idea was to have a this as an alternative to brew themes, so the idea would be to store this style in the db as a new kind of object, and pull that in the brew theme selector.
> 
> I am aware that such a functionality is outside of the scope of this particular PR, but we can make this follow a syntax compatible with other brew themes.
> 
> 

The initial idea was to be able to select normal brews to import their Style Tab contents as Themes, so no need for a separate kind of DB objects.

ericscheid:
For reference
- https://github.com/naturalcrit/homebrewery/issues/1097
- https://github.com/naturalcrit/homebrewery/pull/3075

dbolack-ab:
> > The initial idea was to be able to select normal brews to import their Style Tab contents as Themes, so no need for a separate kind of DB object. See #1899 under the "Custom Themes" heading in the original comment.
> 
> Addtionally, the `brew.text` could be used for both documentation and example for the theme.

That is part of my intent for brews that are intended to be shared styles assuming a style tag and an eventual UI for selecting (personal and favorited) styles.

ericscheid:
> > Addtionally, the `brew.text` could be used for both documentation and example for the theme.

The neat thing is that the `brew.text` can contain `<style>...</style>` content for any documentation-specific styling, avoiding loading that into the `brew.style` property (and consequently loaded by other brews).

This could even include `<style>@import url('/css/theme-documentation.css')</style>` for a consistent look across multiple theme documentations =)

dbolack-ab:
#3321 was based on this as a possible path. As that has been approved, I'm closing this as a redundancy.

PR #3316, opened 2024-02-22
Fix sanitizeScriptTags when content is null

### LINKED ISSUE #3315: Sanitize Script Tags crashes when content is null/empty
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

When there is no content to be sanitized, the sanitize script tags function fails to handle the null condition, crashing the entire website to a white screen.

### Code

_No response_

G-Ambatte:
This PR resolves #3315.

This PR modifies the sanitizeScriptTags function to properly handle the null case, which will preventing website crashing to a white screen when the content window is empty.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -104,8 +104,9 @@ const BrewRenderer = (props)=>{
 
 	const sanitizeScriptTags = (content)=>{
 		return content
-			.replace(/<script/ig, '&lt;script')
-			.replace(/<\/script>/ig, '&lt;/script&gt;');
+			?.replace(/<script/ig, '&lt;script')
+			.replace(/<\/script>/ig, '&lt;/script&gt;')
+			|| '';
 	};
 
 	const renderPageInfo = ()=>{


calculuschild:
Just noting here for future reference: The error is caused by the addition of this line:

https://github.com/naturalcrit/homebrewery/blob/4feeaee7e779391bbfe4de61c786864b408ddac9/client/homebrew/brewRenderer/brewRenderer.jsx#L155

This line is to make sure when rendering pages, always render the currently-edited page first (which may not be in view on the brewRenderer preview). This ensures that any changes that have cross-page effects (e.g. variables or links) will be parsed first, so the latest values will propagate correctly to the pages in view.

Unfortunately, updating the "current editor page" happens slightly out of sync here:

Codemirror sees a change -> sends this change up to the top at `handleTextChange()` in `editPage.jsx` (or `newPage` or `homePage`) and gets the currently edited page from `editor.jsx` -> propagates new text down to `editor.jsx` and `brewRenderer.jsx`.

As you can see, `editor.jsx` is being asked to calculate the latest cursor position before it it has received the latest text.

So while this PR does fix the crashing (I will merge this for now), it might be worth moving `getCurrentPage()` to come directly from Codemirror in `codeEditor.jsx` instead which will be more accurate.

PR #3321, opened 2024-02-23
Enable User Brew theme selection

dbolack-ab:
This adds the UI elements to the metadata selector for users to select brews tagged as `theme` or `Theme` as a brew's theme.

It also swaps in use of the CSS endpoints ( with small changes ) from https://github.com/naturalcrit/homebrewery/pull/3303

Added Endpoints:

*/theme/:renderer/:id/*  - This endpoint provides a JSON object consisting of snippets (`.snippets`) and styles (`.styles`) from the referenced combination of *renderer* and *id*. Both members return an array of contents from the referenced theme and its entire parental tree from oldest to newest.  

    The '.styles' member is an array of strings. Each generation should lead with a comment naming the source followed by any style content.

    The `.snippets` member is an array of snippets menu object(s). Empty generations will yield an empty array member.

Testing: 
 - [x] Loads a Legacy Brew correctly
 - [x] Loads a Blank Themed Brew correctly
 - [x] Loads a 5ePHB Themed Brew correctly
 - [x] Loads a 5eDMG Themed Brew correctly
 - [x] Shows Traditional and User Brews on the theme selector with a V3 brew.
 - [x] Loads a brew using a User Brew Theme correctly
 - [x] Hot loads changes between User Brew and Traditional themes in metadata editor
 - [x] Hot loads changes between renderers (should be forced to Legacy 5ePHB theme if Legacy is chosen)
 - [x] Gracefully fails when a User brew theme'd brew loads and its theme is no longer available - should show error in theme selector
 - [x] Correctly shows inherited Traditional Theme **snippets**
 - [x] Supports nested User Themes in combination with nested Traditional Themes (Ex: Current brew -> User Theme 1 -> User Theme 2 -> User Theme 3 -> 5eDMG -> 5ePHB -> Blank)
 - [x] Loads themes saved on Google Drive
 - [x] Displays current brew's style tags on top of any inherited themes.


5e-Cleric:
I need either access to your repo to check this in local, or a deployment, probably both.

dbolack-ab:
Test theme: https://homebrewery.naturalcrit.com/share/qr7i-TXHV8yf
Example "using" it. https://homebrewery.naturalcrit.com/share/cPtkGUBcVldo

5e-Cleric:
Checking this PR, the PHB theme is imported twice if the brew has the same theme as base, we should check for that redundancy

Tbh, CHROME should remove that redundancy, no idea why it is not

dbolack-ab:
> Checking this PR, the PHB theme is imported twice if the brew has the same theme as base, we should check for that redundancy
> 
> Tbh, CHROME should remove that redundancy, no idea why it is not.

It's not hitting it via the same path, I think. If it is, I'm unsure if it *really* is loading it twice or if that graph is unhelpfully showing every load attempt even if it was pulled from cache.

5e-Cleric:
> It's not hitting it via the same path, I think. If it is, I'm unsure if it _really_ is loading it twice or if that graph is unhelpfully showing every load attempt even if it was pulled from cache.

Absolutely no idea how to figure out that. But we should be able to fix it either way, if the style is set to custom not loading another.

dbolack-ab:
@5e-Cleric Is that looking moer correct now?

5e-Cleric:
- [x] Brew themes with the same name appear separately in the list
 - [ ] Brew themes in new brew (this is not necessarily bad, in my opninion its okay if its not in this PR, we can add it later)
 - [ ] Can we change the name of the generated stylesheet from the edit link to the Name of the brew? or to the share link
 - [x] there are no repeated styles

Overall looks great!

dbolack-ab:
> * [ ]  Brew themes in new brew (this is not necessarily bad, in my opninion its okay if its not in this PR, we can add it later)

Not sure I follow this?

> * [ ]  Can we change the name of the generated stylesheet from the edit link to the Name of the brew? or to the share link

That's a good idea. I've done that.

Gazook89:
> Brew themes in new brew 

I suspect he means that currently you can't set the brew theme from /new/, because we don't have any buttons for the Style or Properties editors in the /new/ page at all.

Gazook89:
Right now, the image associated in each theme in the dropdown shows broken links for user themes.  It might be messy/hard for users to control, but maybe we can check to see if the saved user theme has a "preview image" saved, and use that  (cropped with css).  And if no image, then just return 'null' and display no image.

image

(faded icons due to a linear gradient fade we have).

5e-Cleric (approved):
Looks good to me, let's just remove console logs and i say this is a go!
Thanks @dbolack-ab for your great work on this! 🎉

Gazook89:
On the latest deployment (2 hours ago) I see that something was updated-- the texture on each dropdown item now shows the PHB texture.  But it seems to do this regardless of what is used as the thumbnail in the theme brew.  For example, here is what the deployment is showing me:

image

But I would think that we could show a piece of the thumbnail (the same as what is shown in the Preview box that shows on hover):

image

(please excuse that this isn't a one-to-one comparison...the second image is from my local branch changes)

for reference, here is what the Preview box looks like when hovering over that last option:
image

5e-Cleric:
A theme from a brew with `renderer:legacy` can be selected but the css does not arrive.

If you could in this PR, make sure legacy renderer brews are ignored completely, that would be nice.

5e-Cleric:
> > * [ ]   Brew themes in new brew (this is not necessarily bad, in my opninion its okay if its not in this PR, we can add it later)
> 
> Not sure I follow this?

`/new` has theme dropdown, but custom themes are not displayed:

Can't provide example right now as the whole system seems to be failing.

5e-Cleric:
What do we need to bring this one to the end?

calculuschild:
I still need to give this one a deep review for code quality but if the functionality is working it should be mostly ready. This will be next on the stack when I get #3261 merged which should be soon.

5e-Cleric (approved):
Overall nice changes, will further test, but this looks good

client/homebrew/editor/metadataEditor/metadataEditor.less
```diff
@@ -230,14 +237,23 @@
...
+						left: 0;
+						overflow: hidden;
+						> img {
+							mask-image			    : linear-gradient(90deg, transparent, black 20%);
+							-webkit-mask-image	: linear-gradient(90deg, transparent, black 20%);
+							position			    : absolute;
+							right				      : 0;
+							top					      : 0px;
+							width				        : 50%;
+							min-height: 100%;
```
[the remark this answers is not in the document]
dbolack-ab: Fixed. 

Do we need to see about some less linting

G-Ambatte:
Something's gone awry with the merge, the NotificationPopup is appearing twice on the deployment:
image

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -221,11 +244,16 @@ const BrewRenderer = (props)=>{
 					tabIndex={-1}
 					style={{ height: state.height }}>
 
-					<link href={`/themes/${rendererPath}/Blank/style.css`} type='text/css' rel='stylesheet'/>
+					<ErrorBar errors={props.errors} />
+					<div className='popups'>
```
G-Ambatte, line 248: This is a duplicate of the lines 230-233.

5e-Cleric (approved):
Looks good to me.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -188,9 +188,32 @@ const BrewRenderer = (props)=>{
 		document.dispatchEvent(new MouseEvent('click'));
 	};
 
-	const rendererPath  = props.renderer == 'V3' ? 'V3' : 'Legacy';
-	const themePath     = props.theme ?? '5ePHB';
-	const baseThemePath = Themes[rendererPath][themePath].baseTheme;
+	let rendererPath  = props.renderer == 'V3' ? 'V3' : 'Legacy';
+	let baseRendererPath = props.renderer == 'V3' ? 'V3' : 'Legacy';
+	const blankRendererPath = props.renderer == 'V3' ? 'V3' : 'Legacy';
```
calculuschild, line 193: Seems like all 3 of these end up being the same value. Can we just use one variable for this? Themes should be restricted to working with a matching renderer engine anyway. I.e., a V3 brew cannot use a Legacy Theme.

server/homebrew.api.js
```diff
@@ -37,14 +68,50 @@ const api = {
...
+			Brew : {
+
+			}
+		};
+
+		const brews = await HomebrewModel.getByUser(username, true, fields, { tags: { $in: ['meta:theme', 'meta:Theme'] }, shareId: { $ne: id }, renderer: { $ne: 'Legacy' } });
+
+		if(brews) {
+			for await (const brew of brews) {
+				userThemes.Brew[`#${brew.shareId}`] = {
```
calculuschild, line 92: Can I get a brief overview of how the `#` plays into everything? I'm having trouble following what its purpose is.

client/homebrew/brewRenderer/brewRenderer.jsx
dbolack-ab: I don't recall doing that for a specific reason at the time. I may have kept them distinct in case they proved to need be in the future.

Reviewing the code, there was some redundancy and one of the consts only existed to be more readable. Collapsed that.

server/homebrew.api.js
dbolack-ab: It is used to be a clear indicator that the theme is a user theme, so code doesn't try to load `\themes\V3\editId`

calculuschild: Is there ever a case where such an endpoint would be generated? All static themes are going to be behind `\v3\themeName` and all user themes are going to be `\shareId`, right?

dbolack-ab: There are places where themes are referenced by just the name (prop, front matter, etc.) with the expectation of being joined with the renderer field in the same structure vs places where they are not.  (IE `{renderer: 'V3', theme: '5ePHB'}` vs `/V3/53PHB`. 

I was also assuming that it was plausible for future rendering versions ( V4, etc. ) and that theme names could match with editID specs.

There were three obvious solutions to be certain a theme was a brew theme:

1. Add a "brew" value to the options for renderer - not future proof.
2. Add a new front-matter property/prop to indicate a brew theme 
3. "Tag" the theme name to indicate a brew theme in the places where themes are expressed in their string path form.
4. Walk the 'static' themes list to verify that the combination of renderer and theme was absent. Not horrible.

2 and 3 are the best choices as when HB is looking at the front matter metadata to load it currently only knows the renderer and a string that should represent a string. 4 isn't horrible.

I chose 3 because it is non-looping, adds no data structures, and is visually obvious.  I also wasn't always confident I'd have access to the global static themes structure. It doesn't take much to validate and acts as a smidge of a check bit. I don't think it would be hard to switch to 2 or 4, but I don't see a good enough reason to rework that little bit.

calculuschild: Ok, I see where you're coming from now. We can keep it as-is for now.

I would eventually like to change his though; perhaps moving to option 2 or 4. The main reason being option 3 introduces an unique handling of brew IDs endpoints, where all the other endpoints (\share, \edit, \download, \source) use the ID directly without needing a `#` marker. It's a potential maintenance issue trying to remember that one endpoint has this different schema and remembering all the places that `#` is being handled.

I will note that `renderer` (Markdown version) and `theme` are separate entities. So 1. above wouldn't make sense anyway to select a brew as a renderer. As it stands now, *all* themes from user brews must use V3 renderer.

dbolack-ab: While they are separate entities, there is no reason to expect that all Rendering Engine versions will have the same themes, they're pretty coupled.

If there is a preference for 2 or 4, let's just rip that bandaid now.

calculuschild: > While they are separate entities, there is no reason to expect that all Rendering Engine versions will have the same themes, they're pretty coupled.

Exactly. But mainly, we don't want to overwrite the `renderer` field with a shareID because they are different values. Same reason we don't overwrite the `title` field with the `shareID`. They are two identifiers that serve different purposes.

> If there is a preference for 2 or 4, let's just rip that bandaid now.

I think I would prefer 4 for this. If `theme` is found in the list of static themes, load that. Otherwise assume it is a brew Id and try to load that.

dbolack-ab: And done.

You'll also get to test the "the brew went away!" code if you have something saved with a brew theme. :)

server/homebrew.api.js
```diff
@@ -281,35 +274,33 @@ const api = {
...
-	getBrewThemeWithCSS : async (req, res)=>{
+	//Return CSS for a brew theme, with @include endpoint for its parent theme if any
+	getBrewThemeCSS : async (req, res)=>{
 		const brew = req.brew;
 		splitTextStyleAndMetadata(brew);
 		res.setHeader('Content-Type', 'text/css');
-		const themePath = themes[_.upperFirst(req.brew.renderer)].hasOwnProperty(req.brew.theme) ? `/css/${req.brew.renderer}/${req.brew.theme}` : `/css/${req.brew.theme}`;
-		// Drop Parent theme if it has already been loaded.
-		// This assumes the continued use of the V3/5ePHB and V3/Blank themes for the app.
-		const parentThemeImport = ((req.brew.theme != '5ePHB') && (req.brew.theme != 'Blank')) ? `@import url(\"${themePath}\");\n\n`:'';
```
dbolack-ab: This was in here because the editor loads this theme and double loading was causing issues.

client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -86,6 +89,17 @@ const NewPage = createClass({
 		document.removeEventListener('keydown', this.handleControlKeys);
 	},
 
+	// Loads the theme bundle and parses it out. Called when the iFrame is first mounted, and when a new theme is selected
+	fetchThemeBundle : function(renderer, theme) {
+		fetch(`${window.location.protocol}//${window.location.host}/theme/${renderer}/${theme}`).then((response)=>response.json()).then((themeBundle)=>{
```
G-Ambatte, line 94: As per `homePage.jsx:line 50`.

client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -18,18 +18,37 @@ const SharePage = createClass({
...
+		this.fetchThemeBundle(this.props.brew.renderer, this.props.brew.theme);
 	},
 
 	componentWillUnmount : function() {
 		document.removeEventListener('keydown', this.handleControlKeys);
 	},
 
+	// Loads the theme bundle and parses it out. Called when the iFrame is first mounted, and when a new theme is selected
+	fetchThemeBundle : function(renderer, theme) {
+		fetch(`${window.location.protocol}//${window.location.host}/theme/${renderer}/${theme}`).then((response)=>response.json()).then((themeBundle)=>{
```
G-Ambatte, line 43: As per `homePage.jsx:line 50`.

G-Ambatte (commented):
I've run my eyes over the code, but I haven't had the opportunity to actually compile and run the PR locally yet.

G-Ambatte:
The test failure appears to be due to insufficient test coverage of `homebrew.api.js`; probably Jest determining that the new routes are insufficiently tested?

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -130,21 +133,34 @@ const EditPage = createClass({
...
-
 	},
 
 	hasChanges : function(){
 		return !_.isEqual(this.state.brew, this.savedBrew);
 	},
 
+	// Loads the theme bundle and parses it out. Called when the iFrame is first mounted, and when a new theme is selected
+	fetchThemeBundle : function(renderer, theme) {
+		fetch(`${window.location.protocol}//${window.location.host}/theme/${renderer}/${theme}`).then((response)=>response.json()).then((themeBundle)=>{
```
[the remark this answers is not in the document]
dbolack-ab: The documentation I read seemed to think if you lead with a '/' it would treat it as a file path though a relative path would not.

I dunno. I've always erred on the side of being explicit when possible.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -130,21 +133,34 @@ const EditPage = createClass({
...
+			themeBundle.joinedStyles = themeBundle.styles.map((style)=>`<style>${style}</style>`).join('\n\n'); //DOMPurify.sanitize(joinedStyles, purifyConfig);
```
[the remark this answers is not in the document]
dbolack-ab: Debugging, mostly.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -125,10 +124,8 @@ const BrewRenderer = (props)=>{
 	};
 
 	const renderStyle = ()=>{
-		if(!props.style) return;
 		const cleanStyle = props.style; //DOMPurify.sanitize(props.style, purifyConfig);
-		//return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: `<style>@layer styleTab {\n${sanitizeScriptTags(props.style)}\n} </style>` }} />;
-		return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: `<style> ${cleanStyle} </style>` }} />;
+		return <div style={{ display: 'none' }} dangerouslySetInnerHTML={{ __html: `${props.themeBundle.joinedStyles} \n\n <style> ${cleanStyle} </style>` }} />;
```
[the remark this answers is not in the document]
dbolack-ab: Correct

client/homebrew/pages/editPage/editPage.jsx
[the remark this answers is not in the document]
calculuschild: I did try putting it all into one `<style>` block, but an issue arises if any themes use `@import` statements for fonts, etc. Valid CSS must have all `@import`s at the top of the `<style>` block before any other styling. Any `@import`s that come after normal CSS styling get ignored, at least in the current version of Chrome.

Each theme being in its own `<style>` block means we don't have to worry about the second or third theme having an `@import` show up in the middle of the block and suddenly fonts are missing.

client/homebrew/pages/homePage/homePage.jsx
```diff
@@ -34,12 +34,28 @@ const HomePage = createClass({
...
 
 	editor : React.createRef(null),
 
+	componentDidMount : function() {
+		this.fetchThemeBundle(this.props.brew.renderer, this.props.brew.theme);
+	},
+
+	// Loads the theme bundle and parses it out. Called when the iFrame is first mounted, and when a new theme is selected
+	fetchThemeBundle : function(renderer, theme) {
+		fetch(`${window.location.protocol}//${window.location.host}/theme/${renderer}/${theme}`).then((response)=>response.json()).then((themeBundle)=>{
```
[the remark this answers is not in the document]
calculuschild: > Can this be modularized?

Probably.... though there might be some trickiness getting the `this.state` to work. Might have to just pull that part out and set the state after the function returns.

dbolack-ab:
Added A possible fix for G-Ambatte's concerns

server/homebrew.api.js
```diff
@@ -209,6 +250,81 @@ const api = {
...
+			// Reverse the order of the arrays so they are listed oldest parent to youngest child.
+			styles   : completeStyles.reverse(),
+			snippets : completeSnippets.reverse()
+		};
+
+		res.setHeader('Content-Type', 'text/json');
+		return res.status(200).send(JSON.stringify(returnObj));
+	},
+	//Return CSS for a brew theme, with @include endpoint for its parent theme if any
+	getBrewThemeCSS : async (req, res)=>{
```
calculuschild, line 304: So, with the rework to use `/theme`, is there a purpose for these separate `/css` endpoints anymore?

I don't think they are really needed anymore. *BUT*, now that `/theme` gets the whole chain of themes, maybe the niche for `/css` could be simplified to just retrieve the current brew styling and not the whole chain. Basically what https://github.com/naturalcrit/homebrewery/pull/3075 does (but support fetching styles from staticThemes too).

I'm curious what @ericscheid thinks, as he raised the original issue about a CSS endpoint https://github.com/naturalcrit/homebrewery/issues/1097

ericscheid: Yes, #3075 is desired. It's a simple way to patch in one or more modules of css styling.

calculuschild: And to be clear, you don't think that need is filled by this PR's `/theme` endpoint?

5e-Cleric:
> I've run my eyes over the code, but I haven't had the opportunity to actually compile and run the PR locally yet.

I'm confused, is there a reason you don't want to test in the deployment?

client/homebrew/pages/editPage/editPage.jsx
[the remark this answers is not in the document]
5e-Cleric: Should add here as far as i have read, there is no intention to change how the import at-rule behaves, it will probably never work after other CSS.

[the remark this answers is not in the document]
dbolack-ab: Given that the alternate is floating all of the @imports to the top, this seems best.

5e-Cleric:
image
### Gracefully fails might be overstating it

I'm guessing the original intention is to make it back to default as phb or blank...

calculuschild:
> I'm guessing the original intention is to make it back to default as phb or blank...

I would say if it fails to load a theme, just keep the previous theme.

dbolack-ab:
> image
> 
> ### Gracefully fails might be overstating it
> I'm guessing the original intention is to make it back to default as phb or blank...

No. This was my intent.  

Falling back to a seemingly arbitrary theme is bad form.

dbolack-ab:
I don't understand what is meant by this checkbox.

` Supports User Themes inheriting other User Themes and Traditional Themes`

If you're looking at verifying "nested" themes, how deep do you need to see to consider this passed?
if not that, what is being looked for here?

5e-Cleric:
> I don't understand what is meant by this checkbox.
> 
> ` Supports User Themes inheriting other User Themes and Traditional Themes`
> 
> If you're looking at verifying "nested" themes, how deep do you need to see to consider this passed?
> if not that, what is being looked for here?

I did not make that checkbox, i thought you made it.

5e-Cleric:
> > I'm guessing the original intention is to make it back to default as phb or blank...
> 
> No. This was my intent.  
> 
> Falling back to a seemingly arbitrary theme is bad form.

So the brew failing into no theme is expected behaviour? Because that seems unnacceptable to me.

calculuschild:
> > I don't understand what is meant by this checkbox.
> > 
> > ` Supports User Themes inheriting other User Themes and Traditional Themes`

I added this one. Yes, it means nested user themes, including a transition to Traditional themes. I would say this passes if we can go at least 3 levels of user brews deep, and end with 3 Traditional Themes (including Blank):

Current brew -> User Theme 1 -> User Theme 2 -> User Theme 3 -> 5eDMG -> 5ePHB -> Blank

Gazook89:
Regarding the missing theme error, this seems like a good use of the error notifications we have-- pop an error that gives the id of the missing theme and any other info that would help resolve it.

Regarding what is displayed for the brew (or if the brew is shown at all), i'm not sure.  Maybe just default to the Blank theme, which is more agnostic than defaulting to PHB, since there is a solid chance that if someone is extensively using themes they want to *avoid* looking like the PHB.

dbolack-ab:
> > > I'm guessing the original intention is to make it back to default as phb or blank...
> > 
> > 
> > No. This was my intent.
> > Falling back to a seemingly arbitrary theme is bad form.
> 
> So the brew failing into no theme is expected behaviour? Because that seems unnacceptable to me.

Let me get this straight.  

You think it would be better to reassign the theme of a Brew to an arbitrary brew instead of removing theming and informing the user?

As a user I would find that infuriating.

dbolack-ab:
> > So the brew failing into no theme is expected behaviour? Because that seems unnacceptable to me.
> 
> I agree with you. It is bad design if an invalid or missing option just visually breaks the website. It should at least show some kind of default or fallback value.

No styling *is* a fallback value. It's the penultimate fallback.

calculuschild:
> You think it would be better to reassign the theme of a Brew to an arbitrary brew instead of removing theming and informing the user?

Not reassign their selection to an arbitrary theme. Just dont show a broken webpage if the one they pick doesn't exist. The app should have some robustness against errors like this.

Same reason we put the brewRenderer in an iFrame. If a user messes up their HTML, allowing the whole webpage to crash would be bad design.

dbolack-ab:
> > You think it would be better to reassign the theme of a Brew to an arbitrary brew instead of removing theming and informing the user?
> 
> Not reassign their selection to an arbitrary theme. Just dont show a broken webpage if the one they pick doesn't exist. The app should have some robustness against errors like this.
> 
> Same reason we put the brewRenderer in an iFrame. If a user messes up their HTML, allowing the whole webpage to crash would be bad design.

Unless some of your rearrangement has broken existing behavior ( hadn't last I checked ) it does not break the site. I'm unsure where this assertion comes from.

> If No Styling is the penultimate, then I'm curious what you would consider the ultimate fallback after that?

There is none. No styling drops you back at the default behavior for your browser 

> No, I dont mean change the selection out from under the user. But don't just show a broken page either. That would drive people crazy, and give the impression our site has just broken.

Again with this broken site thing.

calculuschild:
The site isn't literally broken, but this *looks* like something broke.

image

calculuschild:
> Regarding the missing theme error, this seems like a good use of the error notifications we have-- pop an error that gives the id of the missing theme and any other info that would help resolve it.
> 
> Regarding what is displayed for the brew (or if the brew is shown at all), i'm not sure. Maybe just default to the Blank theme, which is more agnostic than defaulting to PHB, since there is a solid chance that if someone is extensively using themes they want to _avoid_ looking like the PHB.

I have added a proper popup if a theme, or any theme in the inheritance chain, fails to load. In this case, the brewRenderer falls back to displaying an unstyled **Blank** theme (but does not change the user's selection in the menu).

image

calculuschild (approved):

dbolack-ab:
> > Regarding the missing theme error, this seems like a good use of the error notifications we have-- pop an error that gives the id of the missing theme and any other info that would help resolve it.
> > Regarding what is displayed for the brew (or if the brew is shown at all), i'm not sure. Maybe just default to the Blank theme, which is more agnostic than defaulting to PHB, since there is a solid chance that if someone is extensively using themes they want to _avoid_ looking like the PHB.
> 
> I have added a proper popup if a theme, or any theme in the inheritance chain, fails to load. In this case, the brewRenderer falls back to displaying the unstyled **Blank** theme (but does not change the user's selection in the menu).
> 
> image

This lines up with my expectations and desires though it looks like we have to take a *slightly* more convoluted route to get there than I had in my head.

5e-Cleric:
After deployment of v3.14 bugs were reported, specially heroku error pages, due to timeout errors on load of undefined themes. Patch was rolled back. After fixing, was deployed again now, crashes seem to be avoided, but snippet bar seems to vanish for `/new` and homepage. Furthermore, custom theme brews appear to fail to show at theme dropdown.

PR #3323, opened 2024-02-23
Fix syntax highlighting on sub and superscript

dbolack-ab:
I did not originally test this very robustly, it seems.


5e-Cleric:
image

CORRECT

<br/>
<br/>
image

INCORRECT
<br/><br/>
image

INCORRECT

dbolack-ab:
> image
> INCORRECT image
> 
> INCORRECT

Looking at both of these I'm struggling to see how they could ever be correct. These are not really valid uses...

5e-Cleric:
> Looking at both of these I'm struggling to see how they could ever be correct. These are not really valid uses...

wym? what if i want to write a simple equation?
image
this works, but does not get correctly highlighted

calculuschild:
@dbolack-ab I poked a little at this and think I might have fixed it. Here's your warning that I pushed to your PR.

@5e-Cleric I believe your last test cases work now. Can you verify?

5e-Cleric:
Everything works, but i do have something to comment, line size changes when there is a sub, or a sup, if there are both, the line gets even bigger:

image

Not saying this is something bad, just commenting on it in case you haven't seen it.

5e-Cleric (approved):

calculuschild:
Thanks for working on this one, @dbolack-ab !

PR #3327, opened 2024-02-25
Add administrative brew locking function

### LINKED ISSUE #3326: Administrative Locking
### Your idea:

On occasion, the administrators of the Homebrewery website will need to lock a brew down, so that it cannot be accessed by anyone - not even the authors. Specifically, I am thinking of legal action, such as a DMCA takedown request.

My suggestion is that we create the ability to add a new property to the document's stub in the local Homebrewery database, `lock` or something similar.

```
lock: {
  active: true,
  code: 451,
  message: 'This brew has been locked due to DMCA Takedown Request. Please contact the Administrators to resolve this activity.'
}
```

During the `getBrew` process, if the `lock` property exists and `lock.active` is true, an error should be thrown of type `lock.code` with an error message of `lock.message`.

Mock up of final error page:

image

I would also expect that a lock page would need to be added to the Admin pages, to allow administrators to apply and remove such locks without needing to access the database directly.

G-Ambatte:
This PR is work towards #3326.

This PR adds a check to the `getBrew` function that throws an error if the brew stub has a `lock` object with a `state` property set to 1 or 2.
State 1: locked, inaccessible.
State 2: locked, only accessible via Edit page.

The intended purpose of this function is to allow administrators to prevent access to brews that have identified issues - for example, DMCA take down requests.


server/homebrew.api.js
```diff
@@ -54,6 +54,13 @@ const api = {
 				});
 			stub = stub?.toObject();
 
+			if(stub.lock?.state) {
+				// State 1 : Locked for everything
+				// State 2 : Edit only
+				if(stub.lock.state == 1 || (stub.lock.state == 2 && accessType != 'edit'))
+					throw { HBErrorCode: '100', code: stub.lock.code, message: stub.lock.message, brewId: accessType === 'edit' ? stub.editId : stub.shareId, brewTitle: stub.title };
```
calculuschild, line 61: What is `lock.code` meant to be?

calculuschild (commented):
This seems fine, and I would be ok merging this to form a basis for the next steps, since it shouldn't break anything. Do you want me to merge it now or keep tinkering on it?

server/homebrew.api.js
ericscheid: `lock.code` contains the http response code to be sent back .. e.g. `451` (Unavailable for Legal Reasons) or `403` (Forbidden).

calculuschild: Ah. Of course.

G-Ambatte:
> This seems fine, and I would be ok merging this to form a basis for the next steps, since it shouldn't break anything. Do you want me to merge it now or keep tinkering on it?

I'll keep tinkering for now, I've used `lock.state` as an integer because I anticipated having multiple states (such as `completely locked`, or `locked for editing only`). If it will stay as simple as it is, then it may be more legible to future developers if I were to use named boolean states - such as `locked` and `editable`.

5e-Cleric:
Not sure if this is useful, but there is a `423 - Locked` error code.

G-Ambatte:
MongoDB entry:
image

Share Page:
image

Edit Page:
image

G-Ambatte:
Includes a locked brew test:
image

Gazook89:
I haven’t taken a strong look at this beyond just get comments/screenshots here, so pardon if I missed it:

i think this needs two things:  a list of possible reasons the brew is locked (just a static thing, not anything actually related to a specific brew), and a path to contacting “administrators” like a link to the Reddit mod mail or whatever. 

Right now it just looks overly harsh, with no indication of what the issue could be and no further help to get it fixed. Part of this is styling (big red letters) and part of it is content. 

Further, it might just be best to say “unavailable” rather than locked, or even just use the same error page as the other “no brew” error. I’m not sure we need to tell all viewers that an author is getting DMCA’d (when we aren’t the ones really arbitrating/deciding that)

PR #3330, opened 2024-02-27
Remove background-attachment property from .monster.frame

RKuerten:
Fix for #3275 as requested by @calculuschild


calculuschild:
Thanks @RKuerten !

PR #3339, opened 2024-03-03
Add S3 storage backing for the Brew Text

dbolack-ab:
Requires these ENV vars:

AWS_REGION
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY

use enable_s3 in config/default.json to enable
use s3_bucket in config/default.json to set the bucket name.


Gazook89:
I don't follow along with every conversation in Gitter, but also for posterity sake, can you just give a quick bullet point on what this solves or offers?  Is it a cost reducing thing over the current storage?

dbolack-ab:
> I don't follow along with every conversation in Gitter, but also for posterity sake, can you just give a quick bullet point on what this solves or offers? Is it a cost reducing thing over the current storage?

The primary impact would be potential cost reductions by moving the Brew text in Mongo-stored brews to S3 storage which is significantly cheaper and *should* only cost storage cost for the S3 instance if done in the same zone as the Mongo instance. Bucketeer (the heroku.com managed module) would be more.

The secondary impact would be in the sizing of the Mongo instance. Even if it is not reduced now, this would let it delay growth for a much longer time.

calculuschild:
My recollection was that S3 was more or less equal cost-wise, since our Mongo database is already in the same region as our Heroku server.

We should track down the numbers we found in the Gitter chat.

Also, keep in mind splitting our storage into a separate Mongo *and* S3 means twice the server calls, which might end up being more expensive anyway.

calculuschild:
I might suggest we put a pause on this, *but* if we ever decide to provide some kind of image hosting, it seems like S3 is probably the best option, so it will be good to have a working example of how to connect to S3 servers, etc.

dbolack-ab:
Three concluding thoughts - 

1) This may be a feature more useful for !naturalcrit folks who have set up an environment differently.
2) Other S3 options may be more performant, but yeah.
3) Should work just fine with Bucketeer.

dbolack-ab:
I'm going to close this. I think there's simply no interest at this time.

Issue #3344, opened 2024-03-06
VERY minor Table of Contents widget edit


Kaiburr-kath-hound:
### Renderer

Both

### Browser

Chrome

### Operating System

MacOS

### What happened?

The "Table of Contents" generation button actually titles the page "Table Of Contents", with the word "Of" being capitalized. I believe the word "of" is never capitalized in any title.

Obviously super minor, but I believe every 5e book only uses the word "Contents" anyways, rather than "Table of Contents".

There's a fun minor issue, if anyone is bored and you run out of actual bugs to work on haha.

### Code

```gfm
{{toc,wide
# Table Of Contents


}}

<!---
Instead of "Table of Contents". I excel at pedantry.
--->
```

RKuerten:
Closing this issue as #3345 got merged. Thanks for opening the issue!

PR #3345, opened 2024-03-06
Updated Table of Contents snippet

RKuerten:
Fix for #3344 
I changed it to "Contents" following the 5e books (I checked and they use only "Contents" instead of "Table of Contents"). 
If "Table of Contents" is preferred, I can simply change it.


5e-Cleric (approved):

calculuschild:
Thanks @RKuerten ! I love easy PRs! 💯 ⭐ 🪄

PR #3348, opened 2024-03-06
Font fix, df to icon fonts, remove duplication

5e-Cleric:
- moved DiceFonts to icon fonts folder
- altered stylesheets import paths to be relative
- removed redundant imports


Gazook89 (changes requested):
It is likely an unfounded concern, but I think we should avoid `spaces` inside folder names.  `/icon fonts/` --> `/icon_fonts/` or `/iconFonts/`.   I'm not sure if we have a settled convention for folder names (and I don't have VS Code open right now and too lazy to check).  

`dicefont.less` should likely be `diceFont.less` as well.

5e-Cleric:
> It is likely an unfounded concern, but I think we should avoid `spaces` inside folder names. `/icon fonts/` --> `/icon_fonts/` or `/iconFonts/`. I'm not sure if we have a settled convention for folder names (and I don't have VS Code open right now and too lazy to check).

## Choose your option:

image


We got:

- only one word
- camelCase
- underscore_naming
- spaces
- One word starting with caps

calculuschild:
I think I'm going to merge this as-is.

But it is a good point that our folder naming scheme should probably be consistent. In general I would prefer camelCase, which seems to be what 90% of our files are using already. Someone can make that another PR.

PR #3359, opened 2024-03-19
Fix crash for DL, disallow block tokens as DT, add test

calculuschild:
- [x] Fixes crash for multilne DLs where DT was empty.
- [x] Disallows Block-level tokens as the DT of a multiline DL

Example:
```
{{descriptive      
Some description
}}         // invalid DT, would cause a crash
:: definiton

### Header // invalid DT
:: definition

- List     // invalid DT
:: definition

Not header // valid DT
:: definition 
```

Could any of you quickly verify that crashes are gone, and that weird use-cases of ``:: some text for hanging indent continue to work? For example, https://homebrewery.naturalcrit.com/share/k-HWy8tLJ_Gl has several `::` in unusual places to get the hanging-indent effect, and headers/etc. immediately before should still render properly.

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -337,12 +337,14 @@ const definitionListsMultiline = {
 		const definitions = [];
 		while (match = regex.exec(src)) {
 			if(match[1]) {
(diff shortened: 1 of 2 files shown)


G-Ambatte (approved):
Looks good to me!

Specifically:
- no longer able to reproduce the BrewRenderer Preview crash
- to my understanding, all changes make sense to prevent error conditions from arising

PR #3361, opened 2024-03-20
Add definition list test: ensure inline has priority over multiline

G-Ambatte:
This PR adds a test to ensure that inline definition list declarations are given priority over multiline definition list declarations, as per the Gitter chat (captured below):

image

This PR takes the example, as given by @calculuschild and adds it as a test to the file `definition-lists.test.js`.


dbolack-ab (approved):
Looks good.

PR #3368, opened 2024-03-20
Usernames in error pages being displayed as anchors to their userpage

5e-Cleric:
Also fixes a tiny bug, if a user has an email syntax username such as `a@gmail.com`, the username would be displayed as an anchor with a `mailto` attribute.


ericscheid:
Alas, this does not fix - an email string is still detected by marked.parse

image

You'll want to replace any `@` character with `\@` ..

```
return [${author}](/user/${author.replace(/@/g, "\\@")})
```

calculuschild:
Simpler solution here:

https://github.com/naturalcrit/homebrewery/pull/3271#issuecomment-1915730204

And the related issue where titles are being parsed as links/images:

https://github.com/naturalcrit/homebrewery/issues/3231

ericscheid:
We'd still want the authors to be linked to their user page though.

Can be done by `return <a href="/user/${author}">${author}</a>`

That won't get parsed either.

5e-Cleric:
I tested this with an email adress and it did work?

ericscheid:
Doh .. the brew.text I tested with was simply `/user/email@example.com`, where the raw email is detected as a substring and linkified.

The brewtext of `[email@example.com](/user/email@example.com)` is the link markup, and gets linked as a relative https link.

In my defense .. that mistake is plainly visible in the image in my earlier comment.

5e-Cleric:
We should probably fix this as calc suggested, but this works fine

calculuschild:
Looks good. I guess making intentionally into a relative link avoids the mailto result. I'll merge this now.

Thanks @5e-Cleric !

PR #3379, opened 2024-03-27
Change Marked extension priority order

### LINKED ISSUE #3378: Multiline Definition List term consumes Mustache Block opening text
### Your idea:

Repro code:
```
{{color:red
:: 1
}}
```

Expected output:
image

Current output:
image

---

I suspect that a tweak of the Marked extension priority order may be all that is required to correct this issue.

G-Ambatte:
This PR resolves #3378.

By changing the order of the Marked extensions, the priority is changed so definition list terms will no longer consume preceding mustache block definitions.
image


dbolack-ab (approved):
This works when I add it locally.

calculuschild:
Thanks @G-Ambatte!

PR #3383, opened 2024-04-01
Changes to terminal output on server start

Gazook89:
This simply modifies the terminal output when running the build + start scripts in the repo.  It adds the URL of the server, so in local development the user will get "http://localhost:3000" which they can `cmd`/`ctrl` `click` to open the local Homebrewery in their browser. And, it just adds a smidgen of color to highlight that URL and the Google Service Account error.

It could be improved if I could find out how to take the nodejs errors like "DEP0040: ...punycode..." and further style them or move them so that the Service Account error would come after them.  I looked and I looked and couldn't find how to intercept those errors (in a meaningful way).

image

### DIFF
--- a/server.js
+++ b/server.js
@@ -12,8 +12,9 @@ DB.connect(config).then(()=>{
 		const cyan = '\x1b[36m'; // Cyan color
 		const underline = '\x1b[4m'; // Underlined style
 
-		console.log(`\n\tserver on port: ${PORT}`);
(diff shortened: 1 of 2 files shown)


G-Ambatte:
> It could be improved if I could find out how to take the nodejs errors like "DEP0040: ...punycode..."

I don't get that error, currently running NodeJS v20.11.1 and NPM v10.2.4. What versions are you running at the moment?

calculuschild:
Thanks @G-Ambatte

PR #3389, opened 2024-04-03
Align snippet menu icons and names

Gazook89:
(edit:  sorry Eric, there is no pre-existing issue that this addresses)

This PR does a small tweak of the CSS for the snippet menus in an effort to align the left edge of the snippet names across the entire menu.  It sets a min-width for the icons.  It also reduces the size of the font snippet "icons"....partially to make it easier to align them, and also because I think the larger current font size looks a little silly (they are much larger than other fontawesome icons).  I left the font size change as a separate commit, though, because maybe someone likes the clarity of the larger font.   Overall, I think there are some other improvements to the menus that could be done, but these are were easy changes to do.

| before | after |
|:---:|:----:|
| image | image |

I also think that we can leave off "remake" off the font names-- seems like irrelevant information for *most* users.  Or maybe, put that part in a span and style it so it's a lighter gray.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.less
+++ b/client/homebrew/editor/snippetbar/snippetbar.less
@@ -130,6 +130,8 @@
 					height       : 1.2em;
 					margin-right : 8px;
 					font-size    : 1.2em;
+					min-width: 25px;
+					text-align: center;
 					& ~ i {
 						margin-right : 0;
 						margin-left  : 5px;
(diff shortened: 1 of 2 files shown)


calculuschild:
I agree we can remove "remake".

Otherwise looks good.

5e-Cleric:
Agreed, this looks good, can we get a screencap of the other misaligned menus, just to be sure? phb for example.

5e-Cleric (approved):

Gazook89:
> can we get a screencap of the other misaligned menus

image

PR #3394, opened 2024-04-04
1px adjustment to Editor height

Gazook89:
This removes a `+ 1` from the editor height calculation, so that the editor height reaches to the bottom of the window rather than leave a 1px gap, visible here:

image

image

I'm not sure if that 1px was there for a reason, but it was set 3 years ago when adding the style editor.

after the fix:

image


calculuschild:
Thanks @Gazook89 !

Issue #3298, opened 2024-02-13
Add share link to error page 401 if brew published


5e-Cleric:
As the published status boolean indicates the availability of the brew to the public, it would make sense to add a share link to the error 401 page if a brew is published.

calculuschild:
After looking at @ericscheid 's [comment](https://github.com/naturalcrit/homebrewery/issues/3086#issuecomment-2041501408) on #3086, I think he has a point that this probably isn't necessary. A user who has an edit link should probably have the share link as well, whether the brew is published or not.

5e-Cleric:
If this is generally agreed, then the issue is done for, but there is still one page without share link, the PR will still be of use.

calculuschild:
Which page is missing a share link?

PR #3399, opened 2024-04-07
Add share link to error pages ONLY IF PUBLISHED

5e-Cleric:
as i suggested in #3298, share page links should only be granted from published brews.


calculuschild:
Based on discussion in the related issues, I would suggest we keep the new share link being added, but just temive the checks for published status.

calculuschild:
Merged! Thanks @5e-Cleric

PR #3407, opened 2024-04-14
Convert errorPage.jsx to functional component.

Gazook89:
This PR simply converts the errorPage.jsx to a functional component and does no other change.  

I think in the future there is an opportunity to re-evaluate the semantics of the html elements but will leave that for another day (and likely most of that is in the errors list file).


5e-Cleric (approved):

Gazook89:
@calculuschild just pinging to see if you can review and merge this when you get a chance.

calculuschild:
Looks good to me! Thanks @Gazook89 !

PR #3090, opened 2023-10-09
License snippet dropdown

5e-Cleric:
Will fix #1976 

Licenses included:

- [x] WotC Fan Content Policy disclaimer
- [x] OGL1.0a
- [ ] OGL1.1/2.0*
- [x] ORC*
- [ ] CC/MIT
- [x] GPL

As per right now, the entire license text is included in each snippet, with a very simplistic format.
Texts are very long, so we might want to look into an alternative.

PR #3420, opened 2024-04-20
RPG Awesome Redux

Gazook89:
This branch just does the same thing as the RPG Awesome repo, but more closely aligns with the code used in the other icon fonts and isn't an NPM package.  It was a 'proof of concept' to see what it takes to get the icons from Game-Icons.net rolled into a font package. 

I used icomoon.io as suggested in the RA repo (actually the companion repo listed in it's readme), but skipped everything in the actual RA repo....

Anyway, point being, this process could now easily include other symbols from Game-Icons.net.   Process would be:

1. download svg from gi.net
2. import folder of svgs to icomoon.io and download as font file
3. some light regex to change some class names around
4. get the font listed in the related files (markdown.js, editor.less, etc)

theoretically, this could be done for ALL of game icons library— they have a download file that includes their whole library, and icomoon.io does batch upload and download. 

image

(this is a PR on your open branch again...this one might be easier to just merge into your branch, or you could download the woff and .less file and just fold into your code if you prefer, if approved at all).


calculuschild:
@Gazook89 I just merged the Emojis PR. Do you want to rebase this on top of master? There are a few conflicts since some of the files over there were changed.

calculuschild (changes requested):
@Gazook89 I just merged the Emojis PR. Do you want to rebase this on top of master? There are a few conflicts since some of the files over there were changed. Otherwise I think this is basically fine. Just need to handle the merge conflicts.

Gazook89:
@calculuschild i did the merge, pretty sure I got it right but did have to fix a couple things in a later commit (duplicated a file, and missed some references to my font).  It works.

In Gitter I did raise a license issue which might be a downer.  Basically, I didn't realize when I started that *each* glyph requires attribution per game-icons.net, and that the attribution is different between different glyphs.  Options I see are 1) seek alternate license, 2) mass attribution of all authors on a FAQ-like page and encourage users to attribute in their brews, or 3) drop the PR.    Or, some system where the creator of each icon is held in the gameIcons.js file, turning the 'values' of each icon from strings to objects (`{ name: 'vase', author: 'lorc' }`) and then use the `author` prop to fill in a `data-author` attribute on the HTML element...and then use that somehow....or if any the editor finds any gameIcon icons present, it can create a notification to the user to add attribution.

calculuschild:
Just noting here I have reached out to game-icons via email to ask about the licensing question on May 16, 2024. No response yet. However, I believe something like this may be sufficient:

> This document uses the 'GameIcon' font, which includes icons created by various artists under CC 3.0 and public domain licenses, sourced from Game-Icons.net. A list of contributing authors can be found at https://game-icons.net/about.html#authors .

calculuschild:
I am going to work on merging this. I am fairly confident the licensing thing won't be an issue if we link to the authors page on Game-Icons. However, we should add an easy method for users to insert the appropriate attribution/licenses for any fonts/icons included in the document. Perhaps as a separate PR in the vein of #3090

PR #3421, opened 2024-04-21
Add lock routes

G-Ambatte:
This PR contributes towards Issue #3326.

This PR adds the following routes to the Admin API:

- [x] `/admin/lock` : `GET` returns the number of documents in the collection that have the `lock.locked` property set to `true`.
- [x] `/admin/lock/:id` : `POST` takes a brew ID and a JSON `lock` object; updates the identified brew with the properties of the `lock` object.
- [x] `/admin/unlock/:id` : `PUT` takes a brew ID; removes the lock from the identified brew
- [x] `/admin/lock/reviews` : `GET` returns a collection of documents that has the `lock.locked` set to `true` and `lock.reviewRequested` exists.
- [x] `/admin/lock/review/request/:id` : `PUT` updates the identified brew's `lock` object with a `reviewRequested` property (current date/time). **NON-ADMIN ROUTE**
- [x] `/admin/lock/review/remove/:id` : `PUT` updates the identified brew's `lock` object to remove the `reviewRequested` property without removing the entire `lock`.


5e-Cleric:
We could use two columns in the lock tab there

G-Ambatte:
Still working on functionality, visibility will come next.

G-Ambatte:
I believe that these latest changes have finished the first pass at functionality.

My thinking for the process is roughly as follows, at this stage I've used the DMCA Takedown Request as the example inciting action.
image

G-Ambatte:
There's still a ton of debugging commands that will need to be removed, and some tuning of the actual lock data functions - I believe that I can probably just rewrite the queries to look for the existence of the `lock` object itself, rather than the `lock.locked` property.

G-Ambatte:
On opening a locked brew for editing:
image

On clicking "REQUEST REVIEW":
image

On the new "LOCKS" tab of the Admin page:
- Lock Count:
image
Clicking "REFRESH" causes the lock count function to run and update the total.

- Brews Awaiting Review:
image
Clicking the button causes the locks with review function to run and update the table. Clicking a row in the table causes the Share ID to be copied to the clipboard (NYI: words explaining this on the page).

- Lock Brew:
image
The four fields are required, attempting to submit a lock with an empty field results in the following:
image
The information on the right are suggested codes (from [this Github comment](https://github.com/naturalcrit/homebrewery/pull/3382#issuecomment-2111009074) and information about the messages

- Unlock Brew:
image
Entering the Share ID of a locked brew and clicking the button results in the lock being removed from that brew, if one exists.

- Clear Review Request
image
Entering the Share ID of a locked brew and clicking the button results in the review request being removed from a locked brew. Possibly this should also have an option to update the `editMessage` to advise the user why their changes are insufficient to have the lock completely removed.

G-Ambatte:
Still need to implement some feedback from the Lock Brew method, to indicate success or failure. Otherwise I think most of the functionality is in a good place, although the styling could probably still use some improvement.

dbolack-ab:
I feel like this is a place we need tests.

5e-Cleric:
I'm afraid to report the number of locks doesn't seem to work for me.

Also, is lock review request supposed to work yet?

Lock removal works fine, and i could set another lock, but doesn't seem to block in share page any longer.

G-Ambatte:
It should work - the change to `homebrew.api:getBrew` should throw an error on any locked brew, the only exception being a valid edit attempt, in which case the lock notification prompt should fire. 

I've updated to the latest master, I'll have a look soon.

G-Ambatte:
Looks like it's getting back 503 Service Unavailable errors, possibly related to DB request timeouts.
Might be a case of pushing `locks` to it's own collection, rather than adding to the `homebrews` collection.

5e-Cleric:
Have in mind the main brew count function in admin returns always a 503, you might be reading that one as well. Better to disable that automatic query on page load for the moment.

G-Ambatte:
The addition of indexes on the database have greatly increased the speed of the lock functions.

5e-Cleric:
Well, with that, the lock count works, and i can see lock review requests.

5e-Cleric:
- [x] `/admin/lock` : `GET` returns the number of documents in the collection that have the `lock.locked` property set to `true`.
- [ ] `/admin/lock/:id` : `POST` takes a brew ID and a JSON `lock` object; updates the identified brew with the properties of the `lock` object.
- [ ] `/admin/unlock/:id` : `PUT` takes a brew ID; removes the lock from the identified brew
- [ ] `/admin/lock/reviews` : `GET` returns a collection of documents that has the `lock.locked` set to `true` and `lock.reviewRequested` exists.
- [ ] `/admin/lock/review/request/:id` : `PUT` updates the identified brew's `lock` object with a `reviewRequested` property (current date/time). **NON-ADMIN ROUTE**
- [ ] `/admin/lock/review/remove/:id` : `PUT` updates the identified brew's `lock` object to remove the `reviewRequested` property without removing the entire `lock`.

client/homebrew/utils/request-middleware.js
```diff
@@ -1,6 +1,7 @@
 const request = require('superagent');
+const version = require('../../../package.json').version;
 
-const addHeader = (request)=>request.set('Homebrewery-Version', global.version);
+const addHeader = (request)=>request.set('Homebrewery-Version', version);
```
calculuschild, line 4: Not sure the purpose of this change. `global.version` works.

server/admin.api.js
```diff
@@ -138,6 +139,150 @@ router.get('/admin/stats', mw.adminOnly, async (req, res)=>{
...
+				return result;
+			});
+		return res.json({
+			count
+		});
+	} catch (error) {
+		console.error(error);
+		return res.json({ status: 'ERROR', detail: 'Unable to get lock count', error });
+	}
+});
```
calculuschild, line 158: Might I suggest the following, and a similar pattern to the other try/catch blocks?

```suggestion
router.get('/api/lock/count', mw.adminOnly, async (req, res)=>{
	const countLocksQuery = { lock : { $exists: true } };
	const count = await HomebrewModel.countDocuments(countLocksQuery);
		.catch (error) {
			console.error(error);
			return res.json({ status: 'ERROR', detail: 'Unable to get lock count', error });
		}
	
	if (count)
		return res.json({count});
});
```

server/admin.api.js
```diff
@@ -138,6 +139,150 @@ router.get('/admin/stats', mw.adminOnly, async (req, res)=>{
...
+		brew.lock = lock;
+		brew.markModified('lock');
+
+		await brew.save();
+
+		// console.log(`Lock applied to brew ID ${brew.shareId} - ${brew.title}`);
+		return res.json({ status: 'LOCKED', detail: `Lock applied to brew ID ${brew.shareId} - ${brew.title}`, ...lock });
+	} catch (error) {
+		console.error(error);
+		return res.json({ status: 'ERROR', error, message: `Unable to set lock on brew ${req.params.id}` });
```
calculuschild, line 185: We may want to have the errors utilize our central error handling middleware. I.e., `throw` the error here, and the error handler in `app.js` will organize the error into a common format (include a stack trace, etc.) and send it.

server/admin.api.js
G-Ambatte: This pattern is applied in the latest commits.

server/admin.api.js
G-Ambatte: I was having an issue here transitioning to use `throw`, but it should be sorted in today's commits.

client/homebrew/utils/request-middleware.js
G-Ambatte: I swear, `global.version`  _didn't_ work when I wrote it... and now it does.

¯\\\_(ツ)\_/¯

Changed in the latest commit.

G-Ambatte:
## State of the PR:

All of the routes in the original scope are now functional, and errors that occur during processing are handled using the existing error handling systems. The Lock Notification "Request Review" button correctly updates the brew's lock object.

### Lock Count & Brews Await Review

- [ ] Potential Improvement: list all current locked brews (as per list all brews awaiting review)

image

### Lock Brew

- [ ] Potential Improvement: Rename "Edit" and "Share" messages to "Private" and "Public" messages
- [ ] Potential Improvement: Add functionality to load and update details of an existing lock (current functionality will not overwrite)

image


### Unlock Brew & Clear Review Request

- [ ] Potential Improvement: Add confirmation step prior to performing removal

image

G-Ambatte:
Table for locked brews has been added.

image

5e-Cleric:
Could you perhaps update the original post to show which routes/endpoints are ready and what not?

G-Ambatte:
> Could you perhaps update the original post to show which routes/endpoints are ready and what not?

The short answer is that all routes are now functional. I've updated the original post to reflect.

5e-Cleric:
### A few things
- Pressing refresh in getting the total number of locked brews (updateReviewCount)  is returning a 503 ( i realize there is little you can do about it) 
- Perhaps we should add the author in the locked brew error page
- What is the override button? / What happens if an override is not clicked, and you lock again the same brew?

G-Ambatte:
@5e-Cleric:

>     * Pressing refresh in getting the total number of locked brews (updateReviewCount)  is returning a 503 ( i realize there is little you can do about it)

In the deployment, I showing that the new routes are returning 304 Not Modified. If I disable the cache, then I get 200 OK, so perhaps a cache disable header should be added.
`/admin/stats/` is running on every refresh, regardless of what tab is actually selected, and seems to be consistently returning a 503. To the best of my knowledge, that API call has not been modified by this PR.

>     * Perhaps we should add the author in the locked brew error page

That might be worthwhile.

>     * What is the override button? / What happens if an override is not clicked, and you lock again the same brew?

OverWRITE:
- when true, overwrites the current lock on the brew if one exists
- when false, returns an error if a lock already exists on the brew

G-Ambatte:
Update Lock Brew interface, including Overwrite option:

image

server/app.js
```diff
@@ -2,10 +2,11 @@
 // Set working directory to project root
 import { dirname }       from 'path';
 import { fileURLToPath } from 'url';
-import packageJSON       from './../package.json' with { type: 'json' };
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 process.chdir(`${__dirname}/..`);
+
+import packageJSON from '../package.json' with { type: 'json' };
```
calculuschild, line 9: Just checking why this import was moved

server/admin.api.js
```diff
@@ -162,6 +164,185 @@ router.get('/admin/stats', mw.adminOnly, async (req, res)=>{
...
+	});
+
+}));
+
+router.post('/api/lock/:id', mw.adminOnly, asyncHandler(async (req, res)=>{
+
+	const lock = req.body;
+
+	const overwrite = lock.overwrite || false;
+	lock.overwrite = undefined;
```
calculuschild, line 213: Not sure why we need a separate overwrite variable and then clear lock.overwrite?

calculuschild (approved):
Looks pretty good to me.

In a future PR I would like to move to functional components but that is not blocking.

Better to merge as soon as you feel you are done so this doesn't get stuck any longer.

server/admin.api.js
```diff
@@ -162,6 +164,185 @@ router.get('/admin/stats', mw.adminOnly, async (req, res)=>{
...
+
+	brew.lock = undefined;
+	brew.markModified('lock');
+
+	await brew.save()
+		.catch((error)=>{
+			throw { name: 'Cannot Unlock', message: 'Unable to clear lock', shareId: req.params.id, status: 500, HBErrorCode: '65', error };
+		});
+
+	// console.log(`Lock removed from brew ID ${brew.shareId} - ${brew.title}`);
```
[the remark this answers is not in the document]
G-Ambatte: Removed in the latest commit

server/admin.api.js
G-Ambatte: It wasn't necessary, in the latest commit I've removed this variable and used the property on the lock object throughout.

server/app.js
G-Ambatte: That was moved during some troubleshooting; I've reverted it now.

client/homebrew/pages/errorPage/errors/errorIndex.js
```diff
@@ -194,13 +194,47 @@ const errorIndex = (props)=>{
 
 		**Brew ID:**  ${props.brew.brewId}
 		
-		**Brew Title:** ${escape(props.brew.brewTitle)}`,
+		**Brew Title:** ${escape(props.brew.brewTitle)}
+		
+		**Brew Authors:** ${escape(props.brew.brewAuthors.length ? props.brew.brewAuthors.join(',') : 'No listed authors')}`,
```
5e-Cleric, line 199: I recommend we stick to the current syntax in other errors, which also happens to return a link to the author's page.

```suggestion
		**Brew Authors:**  ${props.brew.authors?.map((author)=>{return `[${author}](/user/${author})`;}).join(', ') || 'Unable to list authors'},
```

G-Ambatte:
> Is this one good to go? Feel free to merge if yes @5e-Cleric

Should be good to go; the only major thing left is coming up with a way for an administrator to be able to view the content of a locked brew, but that feels like it's big enough to be a PR on it's own.

So unless someone can find any major urgent bug, let's get this one merged and closed out.

5e-Cleric:
Looks good to me, merging! I love pressing the big button.

PR #3428, opened 2024-04-25
drop shadow filter to text stroke

5e-Cleric:
This PR fixes the long standing issue with text stroke, as chrome finally fixed paint-order.

Now we don't need the 12 drop shadows to fake it. Although it could look better. An svg filter might be preferred:

image


calculuschild:
I think this does look better actually. Nice crisp lines instead of the rounded corners and the fuzzy edges.  This biggest artifact I see is on the left serif of the T, which doesn't seem to appear on the right one. Might be a small  in the font itself, which we can try to fix.

We had a similar issue with the last time we tried this (when we discovered `paint-order` wasn't working) and changing the CSS might have revealed some characters we missed. Essentially some line definitions lines overlapping or protruding in a way that isn't visible until you apply a stroke.

https://github.com/naturalcrit/homebrewery/pull/2061#issuecomment-1066040732

Luckily I wrote down the adjustment steps a few comments below that so I can try tweaking the T and see if the weird prong goes away. But that just might be how it is, and I think I'm ok with that.

themes/V3/5ePHB/style.less
```diff
@@ -543,21 +543,17 @@
 		color          : white;
 		text-shadow    : unset;
 		text-transform : uppercase;
-		filter         : drop-shadow(0 0 1.5px black) drop-shadow(0 0 0 black)
-			drop-shadow(0 0 0 black) drop-shadow(0 0 0 black)
-			drop-shadow(0 0 0 black) drop-shadow(0 0 0 black)
-			drop-shadow(0 0 0 black) drop-shadow(0 0 0 black);
+		-webkit-text-stroke: 8px black;
```
calculuschild, line 546: Can we set these to use `cm`?

calculuschild (changes requested):

themes/V3/5ePHB/style.less
5e-Cleric: done

calculuschild:
I think this is good. Thanks @5e-Cleric !

PR #3432, opened 2024-04-26
Fix icon fonts

5e-Cleric:
Fixing misstypes with Elderberry Inn icon font


calculuschild:
Just skimming through, I noticed that there were some other issues. Can you sort them out?

- \E939 appears twice
- \E925 is out of order
- \E95D is missing
- \E93A is missing

5e-Cleric:
3A and 39 are the same error, fixed 25, and frankly i didn't include 5D on purpose because frankly i don't think it should be there, its not even in **their icon reference list**: https://www.elderberry-inn.com/icons

icon E95D:
image

There is also E95A missing, because it is a copy of E956

Gazook89:
We need the farting icon

5e-Cleric:
> We need the farting icon

I refuse on principle, add it yourself.

Gazook89:
So to be sure I understand:  E95D ('toot-toot') is in the ttf/otf font file, but is explicitly excluded from the woff2 file?

I was going to add it to the elderberryInn.less, but it doesn't work, so I assume the font isn't even in the woff2 file and just want to be sure I'm not overlooking something else.

5e-Cleric:
The capture i used was directly from the .woff2 file, it is there, not sure why would that not work.

Gazook89:
Hm, I'll have to retry.  And I had thought your earlier screencap was of the TTF/OTF because for a minute I thought FontForge didn't open WOFFs for some reason.  I know better, but just forgot.

calculuschild:
Ok, merging this now. Thanks @5e-Cleric for catching these errors!

Issue #546, opened 2017-11-06
URGENT: Possible vulnerability to injection attacks


jimcullenaus:
Problem: inserting arbitrary JavaScript into documents is possible, and it will execute. It also seems to cause the entire rest of the page to not render. The person with edit permission to the document cannot then *remove* the JS, because the edit field is also blanked out.

Solution: disable `<script>` tags. Do not allow execution of any JavaScript entered by the users.

For an example, try creating a new blank document, entering some dummy text, and then pasting this at the bottom:

`<script>
document.title = "The title";
</script>`
Or, see [this document](http://homebrewery.naturalcrit.com/share/SkHdIxW00W) (and its source, which is available).

Note that the title in your web browser changes.

dekelpilli:
Was just about to report this. In addition to what is specified here, it seems to mess with the user page of whoever owns the page. The following is a screenshot of my user page after I added a script to one of my brews -- https://i.imgur.com/0cdimPi.jpg -- these are all images I've used in my brews. Below the normal page, it also has some plaintext version of some of my brews (\ns, \ts, and all) -- https://i.imgur.com/4q61IaH.png

Also, the screenshots might look yellow, that part is on my end, not homebrewery's.

goopscoop:
I solved this in a different project with the `sanitize-html` node module. Super easy to setup. Want me to take care of this for you?

kkragenbrink:
Of course, the problem is there's also legitimate reasons to add Javascript, such as zoomable images, or the like.  This is not as easy to fix as simply disabling script tags, but I'll look into it.

jimcullenaus:
I cannot see any legitimate reason to use JavaScript in a homebrewery document. Homebrewery is for building up pages that look high-quality like the D&D published books. There's no room for dynamic content in that. All it does is increase the attack surface.

Sansd20:
I concur with @jimcullenaus. While the homebrewery documents do have the ability to link, the over all use case is primarily to create static documents.

If you are needing Javascript for your use case, you might want to fork this project (if that is possible) and come up with your own solution.

AlexeySachkov:
Should be fixed by #811

calculuschild:
Closing because this appears to be solved.

5e-Cleric:
Reopening because a closing script tag still breaks the website, see #3434

dbolack-ab:
Is this resolved by the DOMPurify addition?

5e-Cleric:
[...]

PR #3434, opened 2024-04-27
Add DOM Purify to BrewRenderer

### LINKED ISSUE #3293: Vulnerability via js on brew
We should remove any possibility of Javascript execution inside brews, that means removing inline event handlers like `onClick`, attributes that allow the `javascript:` protocol, clean svg files, and clean iframes from any javascript.

G-Ambatte:
This PR adds DOM Purify to the BrewRenderer, and uses it to sanitize all Page and Style content immediately before it is added to the component output (using `dangerouslySetInnerHTML`).

This PR extends the work done should close #3293.


5e-Cleric:
Will this fix #3293 ?

G-Ambatte:
From #3293:

> We should remove any possibility of Javascript execution inside brews, that means removing inline event handlers like onClick, attributes that allow the javascript: protocol, clean svg files, and clean iframes from any javascript.

Testing:
- [x] element `on` methods (`onClick`, `onError`, etc) - removed
- [x] JavaScript in HREF property - property is removed
- [x] `<iframe>` - tag and content is completely removed
- [x] `<script>` inside `<svg>` - `<script>` removed, `<svg> remains

So this appears to resolve the issues raised in #3293, although as suggested in that issue, a Content Security Policy may also be needed for defense in depth. I found a really good slide set from Cure53 which discusses the issue here (https://www.usenix.org/sites/default/files/conference/protected-files/enigma16_slides_heiderich.pdf), or if academic papers are your thing, this conference paper (https://link.springer.com/chapter/10.1007/978-3-319-66399-9_7).

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -14,6 +14,9 @@ const NotificationPopup = require('./notificationPopup/notificationPopup.jsx');
 const Frame = require('react-frame-component').default;
 const dedent = require('dedent-tabs').default;
 
+const DOMPurify = require('dompurify');
+const purifyConfig = { FORCE_BODY: true, SANITIZE_DOM: false };
```
G-Ambatte, line 18: `FORCE_BODY: true` - required to leave `<style>` tags in output
`SANITIZE_DOM: false` - required to leave manual IDs in output

If anyone can find a combination of other options to produce this functionality, we can switch to them.

From the DOMPurify source:
`FORCE_BODY` : "/* Decide if all elements (e.g. style, script) must be children of document.body. By default, browsers might move them to document.head */"
`SANITIZE_DOM` : "/* Output should be free from DOM clobbering attacks? This sanitizes markups named with colliding, clobberable built-in DOM APIs. */"

Enabling `FORCE_BODY` seems fine, but disabling `SANITIZE_DOM` seems like it might be excessive - I'm still trying to determine exactly what this parameter does.

5e-Cleric:
@calculuschild could use a deployment for this.

5e-Cleric:
> JavaScript in HREF property - property is removed

Does this mean that the href prop is removed ***only when*** javascript inside is detected?

G-Ambatte:
> > JavaScript in HREF property - property is removed
> 
> Does this mean that the href prop is removed ***only when*** javascript inside is detected?

That is my recollection from testing. Looks like there's a deployment up now so you should be able to test.

5e-Cleric:
This piece of code is what i used to test this:

```
<a href="javascript:alert('This is a JavaScript injection via href attribute')">Click me</a>
___

<img src="javascript:alert('This is a JavaScript injection via src attribute')">

___

<form action="javascript:alert('This is a JavaScript injection via action attribute')">
<input type="submit" value="Submit">
</form>

___

<div style="background-color: red; color: white; width: 100px; height: 100px;" onclick="alert('This is a JavaScript injection via inline event handler')">
Click me
</div>

___

<div onmouseover="alert('This is a JavaScript injection via inline event handler')">Hover over me</div>

___

<style>
.test { background-image: url('javascript:alert("This is a JavaScript injection via CSS background-image")'); }
</style>
<div class="test">Test</div>

___

<div data-code="javascript:alert('This is a JavaScript injection via data attribute')">Test</div>

___

<div id="parent">
<button id="child">Click me</button>
</div>
<script>
document.getElementById('parent').addEventListener('click', function(event) {
if (event.target.id === 'child') {
console.log('This is JavaScript executed via event delegation');
}
});
</script>

```

Using this code in the live site results in a complete crash, but in this deployment seems to work just fine.

#3446 might add some cases we want to check.

Is there any Js injection method we haven't tested yet?

5e-Cleric (approved):
Looks alright, we are more secure than before. Nice PR

G-Ambatte:
Added a small change to `template.js` to stop `</script>` tags from breaking everything.

5e-Cleric:
Related to this, as mentioned, `</script>` can be input in the title of a brew. If the only content of a brew or the starting content of the brew is that string, it will be used as title, therefore, the entire userpage is broken, and the entire editpage is broken too, which makes it very difficult to fix. the only fix i have found is to go into this deployment as it shares the db, and login manually, then deleting the brew from userpage or editpage.

G-Ambatte:
I have removed the `</script>` sanitization from this PR, it was becoming a much bigger job than just a single additional line of code - it will require it's own PR.

ericscheid:
> Related to this, as mentioned, `</script>` can be input in the title of a brew. If the only content of a brew or the starting content of the brew is that string, it will be used as title, therefore, the entire userpage is broken, and the entire editpage is broken too, which makes it very difficult to fix. the only fix i have found is to go into this deployment as it shares the db, and login manually, then deleting the brew from userpage or editpage.

This is a different issue though — we are emitting strings from brews without proper encoding, and they are being treated as html elements by the browser's html parser.

For example, [this line in template.js](https://github.com/naturalcrit/homebrewery/blob/2f5081156d99d10fb227121833ddfc5f972a710a/client/template.js#L21C1-L22C1):
`     <title>${title.length ? `${title} - The Homebrewery`: 'The Homebrewery - NaturalCrit'}</title>`

Here, if the title is the string `A < B && TRUE` then that emits as literally that in the html source .. but it should be `A &lt; B &amp;&amp; TRUE`. The browser will then render that as the string `A < B && TRUE`. The same should happen if the title string is `<script>..</script>`. It should not be delivered to the browser as if it is actual html code to be parsed. We do the encoding properly in other places (e.g. brew cards on userpage).

image

G-Ambatte:
The `</script>` issue is a separate issue, as recorded in #546. This PR, while similar, addresses the issue of active content being parsed by the BrewRenderer, not the HTML template. As such, this PR can be merged as is and a second PR raised to resolve the issues with `template.js`.

client/homebrew/brewRenderer/brewRenderer.jsx
G-Ambatte: After further research, I believe that `SANITIZE_DOM: false` is the only way to allow manual IDs to persist, DOM clobbering removes the ability to set IDs as it can be an attack vector, but I don't believe that applies in our case.
Marking as resolved.

calculuschild:
Thanks @G-Ambatte ! 💯 ⭐ 🎉

PR #3446, opened 2024-05-01
Make Curly Injectors append properties, not overwrite

calculuschild:
Precursor to the Emojis PR: #3405 

Currently, applying curly injectors to an object fully overwrites any inline CSS properties:

```
{{myClass,#myID,myStyle:red Test}}{color:blue}
```

Results in:

```
<span style="color:blue">Test</span>
```

This PR simply modifies the behavior of curly injectors to append to, rather than overwrite the properties of the preceding token.

```
<span id="myID" class="myClass" style="myStyle:red; color:blue">Test</span>
```

Generally should make little difference, but would be handy for elements like emojis that rely on preserving their initial CSS classes: `:an-emoji:{color:blue}`

Likely has some edge case bugs to work out.


ericscheid:
This is good. It also helps that html/css will treat duplicate `style` rules as not an error, simply overriding. 

Thus `{{color:blue,font-weight:bold BLUE-BOLD}}{color:red}` 
will get rendered as 
`<span style="color:blue; font-weight:bold; color:red;">BLUE-BOLD</span>`, 
but is effectively simply
`<span style="font-weight:bold; color:red;">BLUE-BOLD</span>`.

My only concern is that this appending method should only be exercised for insertion of styles and classes, and not for `id` or attributes (e.g. `{{some-attribute=foo BAR}}`. For `id` and _attribute_, the injected values should replace the existing values, but only if specified — that is, don't replace with blanks. (It's late and I'm going cross-eyed so haven't fully read the code)

ericscheid:
> Likely has some edge case bugs to work out.

As the say: "I see you are trying to solve a problem by using regular expressions.Now you have two problems."

Instead of using regex to extract the attributes of a html element encoded as a string, might I suggest we use the DOM API. I've previously done something like this...

```js
function extractAttributeNames(htmlString) {
    var tempDiv = document.createElement('div');
    tempDiv.innerHTML = htmlString;
    var element = tempDiv.firstChild;
    
    var attributeNames = [];
    if (element) {
        for (var i = 0; i < element.attributes.length; i++) {
            attributeNames.push(element.attributes[i].name);
        }
    }
    
    return attributeNames;
}

// Example usage:
var htmlString = '<div id="main" class="container" data-custom="123">Hello, World!</div>';
console.log(extractAttributeNames(htmlString));

var htmlString2 = '<span data-foo="style=color:red"></span>';
console.log(extractAttributeNames(htmlString2));

```

*however* .. I note that that code now faults with "This document requires 'TrustedHTML' assignment." 😢 

IIRC, we have a `SetInnerHTMLDangerously` method or similar which should get around this, right?

calculuschild:
> Instead of using regex to extract the attributes of a html element encoded as a string, might I suggest we use the DOM API. I've previously done something like this...

I started with something like this, but it crashes since we call Marked in a few places serverside without access to the DOM.  The regex parsing of this seems to be working for now. I will revisit this if I come across anything that the regex isn't able to handle.

Gazook89:
This conversation sparked some memory I had about an issue with injections, and I found this old merged PR #2774 about additional unit tests. A bunch of the tests were set to ‘fail’ because of issues related to appending vs replacing. 

Do these tests pass now, without the “fail” modifier?

and I think I had another issue with injections that I never did figure out, but I’m still looking for it to see if it’s relevant here.

calculuschild:
> Do these tests pass now, without the “fail” modifier?

Not sure, but that is a good goal. Since we already have those tests, I will try to make them pass or update them as necessary.

calculuschild:
> Maybe test these cases?
>
> <div id=foo>BAR</div>{classy}
> <div id="foo" class="classy">BAR</div>{classier}
> <div id="foo" class="classy">BAR</div>{class=classiest}
> <div data-bar="the id=QAZ">text</div>{class=qaaz,#qux}

Injector syntax doesn't work on raw HTML, because Markdown doesn't work at all in the same line as raw HTML.
I have added the equivalent tests for curly divs though.

Note explicitly removed `class`, `style`, and `id` attributes from being injected. It causes too many shenanigans and complexity mixing with the other class, style, and id syntax. As it stands now injecting `class` like this will just be ignored. (We can revisit it later if we want to add that option. Not immediately sure what value it would have).

5e-Cleric (approved):
Looks good to me, functionality working as intended in all tested cases.

ericscheid:
> Note explicitly removed `class`, `style`, and `id` attributes from being injected. It causes too many shenanigans and complexity mixing with the other class, style, and id syntax. As it stands now injecting `class` like this will just be ignored. (We can revisit it later if we want to add that option. Not immediately sure what value it would have).

I would expect that injecting `class=some-class` would operate exactly the same as injecting `any-attribute=some-value`. That is, the attribute would be assigned the value (instead of appending).

PR #3447, opened 2024-05-02
Add edit link to share page if user is author

5e-Cleric:
edit link is forwarded checking if the user is author in app.js, then checking if there is an edit link in sharePage.jsx, pretty standard stuff


calculuschild:
The idea seems fine skimming over the code, but I can't actually get the button to appear on any of my brews.

PR #3458, opened 2024-05-06
Add style tab styling for Darkbrewery

5e-Cleric:
image


5e-Cleric:
Actually, Talking to @G-Ambatte we realized more than HALF of the original styles of darkbrewery were not being applied, for a missed class that changed when codemirror themes were introduced, so the theme is barely existing.

Will do a full work to change that.

5e-Cleric:
been too long, no longer interested in the theme anyway, may pick it up later on

PR #3460, opened 2024-05-07
Search method in archive

5e-Cleric:
Branched from #3263


5e-Cleric:
Closing this in favor of #3263, since i have been able to fix all problems in that one and this PR doesn't even work.

Comment for future devs:
`$search` is supposed to be better, but can only be tested in a deployment or live, which makes it a pain, look into `$facet` as well.

PR #3464, opened 2024-05-09
Modify PrintPage to use new BrewRenderer methods

### LINKED ISSUE #3435: Multi-page variable use doesn't work when parsed to PDF
Here i describe two variable bugs reported by @TheDungeonMathster in the Discord Of Many Things:

## 2. Multi-page variable use doesn't work when parsed to PDF
This code works fine in the brew but fails in the PDF render
```
[foo]:bar

$[foo]

\page

$[foo]
```

G-Ambatte:
This PR resolves #3435.

This PR modifies the Print Page to use the BrewRenderer to render the pages for printing. This ensures that the Print Page output will always match the Share and Edit Page visual outputs - previously it used a duplicated section of code that was not updated when the BrewRenderer was changed.

Some modifications have also been made to the BrewRenderer, including the option to not render inside an iFrame (defaults to iFrame on), and a method to indicate to Print Page when the print event should fire (i.e. on completion of document rendering).

iFrame rendering would be preferable but while it is possible to create functions to print the iFrame content directly, I have been unable to get printing via Ctrl+P or selecting Print from the browser menu to work correctly. For this reason, the brew renders outside of a frame on Print Page.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -56,6 +56,7 @@ const BrewRenderer = (props)=>{
 		lang              : '',
 		errors            : [],
 		currentEditorPage : 0,
+		renderAllPages    : false,
 		useIFrame         : true,
 		frameMounted      : ()=>{},
 		...props
@@ -161,7 +162,7 @@ const BrewRenderer = (props)=>{
 		renderedPages[props.currentEditorPage] = renderPage(rawPages[props.currentEditorPage], props.currentEditorPage);
 
 		_.forEach(rawPages, (page, index)=>{
-			if((!props.useIFrame || isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
+			if((props.renderAllPages || isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
 				renderedPages[index] = renderPage(page, index); // Render any page not yet rendered, but only re-render those in PPR range
 			}
 		});
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -165,7 +165,7 @@ const BrewRenderer = (props)=>{
 		renderedPages[props.currentEditorPage] = renderPage(rawPages[props.currentEditorPage], props.currentEditorPage);
 
 		_.forEach(rawPages, (page, index)=>{
-			if((isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
+			if((!props.frame || isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
 				renderedPages[index] = renderPage(page, index); // Render any page not yet rendered, but only re-render those in PPR range
 			}
 		});
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -56,7 +56,7 @@ const BrewRenderer = (props)=>{
 		lang              : '',
 		errors            : [],
 		currentEditorPage : 0,
-		frame             : true,
+		useIFrame         : true,
 		frameMounted      : ()=>{},
 		...props
 	};
@@ -77,7 +77,7 @@ const BrewRenderer = (props)=>{
 	}
 
 	useEffect(()=>{ // Unmounting steps
-		!props.frame && frameDidMount();
+		!props.useIFrame && frameDidMount();
 		return ()=>{
 			window.removeEventListener('resize', updateSize);
 		};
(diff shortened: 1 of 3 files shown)


5e-Cleric:
will review this as soon as i get my hands on a computer

calculuschild (changes requested):

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -52,6 +52,8 @@ const BrewRenderer = (props)=>{
 		lang              : '',
 		errors            : [],
 		currentEditorPage : 0,
+		frame             : true,
```
calculuschild, line 55:
```suggestion
		useIFrame         : true,
```

G-Ambatte: Applied in the latest commit.

client/homebrew/pages/printPage/printPage.jsx
```diff
@@ -52,60 +49,30 @@ const PrintPage = createClass({
...
+		if(this.props.query.dialog) window.print();
 	},
 
 	render : function(){
-		const rendererPath = this.state.brew.renderer == 'V3' ? 'V3' : 'Legacy';
-		const themePath    = this.state.brew.theme ?? '5ePHB';
-		const baseThemePath = Themes[rendererPath][themePath].baseTheme;
-
-		return <div>
-			<Meta name='robots' content='noindex, nofollow' />
```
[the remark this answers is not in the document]
G-Ambatte: Good catch, I've added this back in the latest commit.

client/homebrew/pages/printPage/printPage.jsx
```diff
@@ -52,60 +49,30 @@ const PrintPage = createClass({
...
-				pageText += `\n\n&nbsp;\n\\column\n&nbsp;`; //Artificial column break at page end to emulate column-fill:auto (until `wide` is used, when column-fill:balance will reappear)
-				return (
-					<div className='page' id={`p${index + 1}`} key={index} >
-						<div className='columnWrapper' dangerouslySetInnerHTML={{ __html: Markdown.render(pageText) }} />
-					</div>
-				);
-			});
-		}
-
+	frameMounted : function() {
```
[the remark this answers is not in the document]
G-Ambatte: I've applied this change in the latest commit, PR should be updated now.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -152,7 +162,7 @@ const BrewRenderer = (props)=>{
 		renderedPages[props.currentEditorPage] = renderPage(rawPages[props.currentEditorPage], props.currentEditorPage);
 
 		_.forEach(rawPages, (page, index)=>{
-			if((isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
+			if((props.renderAllPages || isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
```
G-Ambatte, line 165: I realized that using `!props.frame` as a check to render a page was particularly opaque, so I have introduced a new prop `renderAllPages` to make it clearer that all pages will be rendered from the Print Page.

G-Ambatte:
I'm getting a weird case where headings are not rendered at the time the print dialog is first called, so they are missing from the output. 
image

However, closing the print dialog and pressing Ctrl+P or even reloading the page works fine.
image

I'm still digging into it.

calculuschild:
> Serious question, why did we agree on keeping this instead of printing from `/edit` or `/share`?

I'm not sure I understand the question. Printing HTML to PDF prints the entire page, so you need to somehow generate a separate window/frame. As far as I know you can't just send a single element to the printer.

5e-Cleric:
> > Serious question, why did we agree on keeping this instead of printing from `/edit` or `/share`?
> 
> I'm not sure I understand the question. Printing is possible from both /edit and /share via the "get PDF" button.
> 
> Printing HTML to PDF prints the entire page, so you need to somehow generate a separate window/frame. As far as I know you can't just send a single element to the printer.

try this:

In the brew style tab:
```
@media print {
	.brewRenderer {
		height: 100%;
		overflow-y: unset;
		.pages {
			margin: 0px;
			&>.page {
			box-shadow: unset;
			}
		}
	}
}
```

In the dev tools console:

window.frames['BrewRenderer'].contentWindow.print()

calculuschild:
Well if that works, that would simplify things a lot. Does it still work with variables across pages? Work with Legacy brews? I'm ok doing that if we can eliminate the whole `/print` page.

5e-Cleric:
First one for sure, second haven't tried, should work fine

5e-Cleric:
Issue #2742 tracks printing without print page

calculuschild:
Wow, I don't remember that issue at all, but I guess I was there. Looks like the main concern was keeping existing `/print` links working, but now that it's been a while I don't know how necessary that is. I used a few `/print` links to share minimal error examples with Google, but those ones have been solved. The other use case from @ericscheid was for the `visualping` tool which periodically scans a webpage for visual differences so you can be alerted to updates. It wouldn't work in `/share` because it only loads the outer frame or something.

G-Ambatte:
As I recall, the print function didn't trigger a full render so PPR would break the print. I think that's no longer an issue with the changes to Brew Renderer, so the logic may no longer apply.

calculuschild:
Yeah, looks like the `@media print {` approach does work with variables and legacy brews. If we are in agreement, I say we close this PR and move instead to that as a simpler alternative.

calculuschild:
I am pulling new PR together for this now.

Issue #3462, opened 2024-05-07
The StringArrayEditor component returns error if you `ESC` out of it


Gazook89:
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

The StringArrayEditor, used for the Tags input, throws an error when you hit `esc`:

```
Uncaught TypeError: Cannot set properties of undefined (setting 'editing')
    at ReactClassComponent.closeEditInput (bundle.js:80216:33)
    at ReactClassComponent.handleValueInputKeyDown (bundle.js:80211:12)
    ....
```

It is this bit:
```js
	closeEditInput : function(index) {
		const valueContext = this.state.valueContext;
		valueContext[index].editing = false;
		this.setState({ valueContext, updateValue: '' });
	},
```

Gazook89:
I have a fix but not sure it's 100% great yet.  But at the same time, I have another question about this:

I would like to set `,` as a key used to "submit" the current tag and move to the next tag input, the same as the `Enter` key currently.  I don't think this is super controversial, but please let me know if it is.  

Further, I am wondering if `Tab` should do the same.  My reflex is that yes, `Tab` should submit the current tag and move to the next.  This would require using `Esc` to a currently non-existent container around the tags before the user could then use `Tab` again to move to the next field ("systems").  I am going to look into what specs are relevant for this kind of tag entry.   I'm also likely fine with not doing this and just leaving `Tab` as an option to move to the next field ("systems") without submitting the current tag. 

*example of tags inside another container*:
<img width="478" alt="image" src="https://github.com/naturalcrit/homebrewery/assets/58999374/23bf94a6-2ec5-4e2a-92c3-8cc12de79cc9">

Finally, i think maybe that `Tab`-ing and `Shift + Tab`-ing across the tags should activate the 'edit input' state of each tag, rather than sort of skipping them like it does now.

Gazook89:
to go further away from original reported issue, i think the stringArrayEditor could be styled a little better to match the Homebrewery UI.  For example, get rid of the rounded corners.

dbolack-ab:
Is this still open?

5e-Cleric:
Indeed, it is still a thing, #4215 details all the tag-related issues, the list is long and the component needs reworking, which is ideally happening in #3469

Issue #2333, opened 2022-09-04
Tweak tags schemas


ericscheid:
## 1. make data-entry of tag prefixes case-insensitive 

If someone were to enter `Type:Adventure` then the current tags data-entry widget flags that input as invalid, which could be confusing. The *input* should permit case-insensitive text, especially since we're ignoring case when doing searches (e.g. on the user-page).

The submitted tag-type should then be normalised to lowercase. (We use the tag-type to assign a styling class to the tag element on the userpage, so it needs to be standardised there since class names are case sensitive.)

<img width="371" alt="image" src="https://user-images.githubusercontent.com/2605467/188323866-aa3be681-8465-49dd-99d7-31bea59144fb.png">

Can this be done by just tacking on the `i` flag at the end of the regex?

https://github.com/naturalcrit/homebrewery/blob/215b64f5a662efdcb3a6205e2c21caea78e44b9d/client/homebrew/editor/metadataEditor/metadataEditor.jsx#L201

```js
<StringArrayEditor label='tags' 
  valuePatterns={[/^(?:(?:group|meta|system|type):)?[a-z0-9][a-z0-9 \/.\-]{0,40}$/i]}
```

(and also slimming `A-Za-z` to just `a-z` since we have the case-insensitive flag in there now)

We should likely also normalise the casing of known tags though to match the `datalist` options.

## 2. allow ampersand in tag

<img width="224" alt="image" src="https://user-images.githubusercontent.com/2605467/188324091-d6738542-d113-4df0-8775-c4bfcbd3cbaa.png">

There will likely be other suggestions for various punctuations from reddit, but at least the `&` is an obvious candidate.

## 3. allow non-ASCII characters to support non-English language

<img width="212" alt="image" src="https://user-images.githubusercontent.com/2605467/188324176-bac70dfe-6907-460f-ad88-da8c51040fd0.png">

We should however then disallow various characters, particularly punctuation characters we want to reserve for future use

Or, we run the tag through `_.deburr( )` before pattern matching (i.e. dumbify alphabetics).

## 4. Drop the `group:` prefix

The `group:` prefix originated in early discussion of tags, when it was thought we could utilise the tag system for grouping brews into folders on the user-page. The idea of folders has been developed further and using `group:tag-string` looks to be the wrong direction. We also don't have any coding at all that does grouping of brews on user-pages, so dropping it now doesn't disable anything. We can always add it back if that is the way we want to go.

## 5. Use `<datalist />`for standardised options

[...]

Issue #3166, opened 2023-12-03
emit tag classname as lowercase on userpage


ericscheid:
Tags can be entered when enditing a brew as `<tag-type>:<tag-name>`, where `<tag-type>` is a limited enumerated set of values. That tag data-entry UI also enforces lowercase for `<tag-type>`.

It is possible however to circumvent that UI (e.g. edit the google-doc source).

The `<tag-type>` is emitted on the userpage by setting the className of the tag display element: `<span class="system">C&C</span>`. This is so different tag-types can be styled differently (e.g. "system" tags in blue, untyped tags as yellow).

However, css classnames are case sensitive. We should enforce the schema design constraint not only in the data-entry UI, but also in the emitting of tags onto the userpage. (Postel's Law)

The relevant code to fix is on line 132 here:

https://github.com/naturalcrit/homebrewery/blob/be72b029bfd267507b1d784cdfc32d5dc62c3c88/client/homebrew/pages/basePages/listPage/brewItem/brewItem.jsx#L127-L134

G-Ambatte:
`return <span key={idx} className={matches[1]?.toLowerCase()} onClick={()=>{this.updateFilter('tag', tag);}}>{matches[2]}</span>;`

ericscheid:
While stray spaces on tags do cause problems elswhere, they don't cause a problem with class names as `class=" system"` is ==  to `class=" system"` as far as html/css cares.

5e-Cleric:
> While stray spaces on tags do cause problems elswhere, they don't cause a problem with class names as `class=" system"` is == to `class=" system"` as far as html/css cares.

I know, just saying we should trim when inputing tags

dbolack-ab:
Did #3735 fix this? 

Does this relate to the recent discussion about Stubs vs codefenced metadata?

PR #3469, opened 2024-05-10
StringArrayEditor revision and update

Gazook89:
This is basically an implementation of what I wrote in #3466.  It restructures the DOM of the StringArrayEditor and adds more functionality, such as better keyboard interaction.  It removes what I consider "extra buttons" which I don't see used in many other tag editors/inputs.  The styling is yet unfinished, but even at this stage is more inline with existing HB styling.

image

### Keyboard
| context | action | result |
|---|---|---|
| new text in new-tag-input, or editing existing tag | `,` \| `Enter` | submits text as tag and focus new-tag-input |
| focus within tag-container (unless in `input`) | `ArrowLeft` \| `ArrowRight` | moves focus left/right through tags and new-tag-input, within tagContainer |
| focus on existing tag | `Enter` \| `Space` | open tag and focus input for editing |
| focus within `input` in tag or new-tag-input | `Escape` | blur/exit input without submitting, move focus to tag |
| focus on existing tag | `Delete` | delete existing tag |
| focus on element prior to tag-container | `Tab` \| `Shift + Tab` | move focus to new-tag-input, skipping existing tags (instead, use arrow keys to navigate thru individual tags) |

### Things I want to hammer out:

- [x] fix: need to take datalist options as a prop-- currently its static and shows same list for tags and authors.
- [x] fix: if an input inside an existing tag is cleared out (everything is deleted) and then you move to the next tag, by pressing `Enter` or `,`, the original text is reset and the tag isn't deleted.
- [x] fix:  if using `Delete` to delete characters after text cursor, the whole tag is deleted.
- [x] fix:  prevent 'resetting' of input text when you hit `Space` while editing an existing tag.
- [x] do not use `.invalid` styling when input field is empty, but still prevent entry of empty tags
- [x] allow entry of *tag schemes* (`meta:`, `systems:`, etc) using either upper or lowercase, but convert to lowercase when submitted.
- [x] if using backspace in existing tag and no text remains in input, it throws a non-breaking error in console...should prevent that.  
- [x] fix: only render datalist if datalist prop exists
- [x] factor in the issue mentioned in draft PR #3284
- [x] fix: the `cleanValue()` method on stringArrayEditor is too specific to one instance of the component (brew tag editor), and has no application for other instances (like 'authors' input).  Need to figure out how to pass that type of 'value manipulation' in as a prop?  The crux is that the current 'validator' is lenient on the tag scheme case-- it can be entered lowercase or uppercase-- but we actually only *accept* lowercase here, and `cleanValue()` was used to convert it to lowercase on submission.  So maybe a `validator` prop *and* a `acceptedValue` prop?
- [ ] check accessibility (semantic html, aria roles)

### Things I might want to do in this PR, but maybe not:
- [ ] convert to functional component
- [ ] preserve width of tag when switching between "editing" and "not editing"

### Things I might do if this PR is merged:
- [ ] rework the css for metadataEditor, to reduce duplicate or unused styles.  Possibly separate out the styles for each component (or at least some of them, and utilize css variables to control in metadataEditor.less)

### Related Issues:

- [x] fixes https://github.com/naturalcrit/homebrewery/issues/2333
- [x] fixes https://github.com/naturalcrit/homebrewery/issues/3462

&nbsp;

- https://github.com/naturalcrit/homebrewery/issues/2335
- https://github.com/naturalcrit/homebrewery/issues/2357
- https://github.com/naturalcrit/homebrewery/issues/3166

### Relevant PR:
https://github.com/naturalcrit/homebrewery/pull/3469


Gazook89:
I misclicked and submitted this as draft before I could really write anything (the current text is just from first commit). 

I have better explanation typed up on computer at home, but I had to run to playground for a bit before I could finish.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -109,35 +126,64 @@ const StringArrayEditor = createClass({
...
+		} else if(_.includes(['ArrowLeft'], event.key)){
+			event.target.previousElementSibling?.focus();
+		} else if(_.includes(['ArrowRight'], event.key)){
+			event.target.nextElementSibling?.focus();
+		}
+
+
+	},
+
+	renderDatalist : function(){
```
calculuschild, line 147: Concerning the datalist:

- The list is enormous. Is there any way to trim it down initially with just the major categories or something, and then transition to the full list once the user starts typing?
- For some reason, I initially had the impression that these were the *only* allowed options (maybe because the huge list looks so large it might be exhaustive? Maybe I am trained that the little "dropdown" arrow on the textbox means I have to choose one from the list?). Clearly that is not the case because we also allow custom tags, but I wonder if there is some way to better imply that these are only suggestions.
  - Better placeholder text?
  - Remove the dropdown arrow thing?
  - Somehow label the datalist with a title saying "common/suggested tags"
  - Curate or reorder the pre-selected options?
- On the other hand, the "tag-scheme"s *are* restricted. Seems like there should be some clear way to convey to the user "here is a list of allowed tag-schemes, and under those tag schemes, here are some recommended tags, but you can also make your own tag using one of these tag-schemes (or even none at all)."
- If you select `type:` (or any of the "tag-scheme"-only options, I would expect the datalist to now immediately show me the sub-categories of `type:...`, but instead I just get a red "invalid" box. It's not until I type some more that the rest of the suggestions appear.

image

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -109,35 +126,64 @@ const StringArrayEditor = createClass({
...
+					<input type='text' className={`value tag-input ${context.editing ? 'visible' : 'hidden'} ${this.valueIsValid(this.state.updateValue, i) || (this.state.updateValue == '') ? '' : 'invalid'}`} placeholder={this.props.placeholder}
 						value={this.state.updateValue}
 						onKeyDown={(e)=>this.handleValueInputKeyDown(e, i)}
-						onChange={(e)=>this.setState({ updateValue: e.target.value })}/>
-					{<div className='icon steel' onClick={(e)=>{ e.stopPropagation(); this.closeEditInput(i); }}><i className='fa fa-undo fa-fw'/></div>}
-					{this.valueIsValid(this.state.updateValue, i) ? <div className='icon steel'  onClick={(e)=>{ e.stopPropagation(); this.updateValue(this.state.updateValue, i); }}><i className='fa fa-check fa-fw'/></div> : null}
+						onChange={(e)=>this.setState({ updateValue: e.target.value })}
+						onBlur={()=>this.closeEditInput(i)}
+						list={this.props.options?.length > 0 ? `${this.props.id}__tags-precoordinated` : ''}/>
+					{this.renderDatalist()}
```
calculuschild, line 168: The datalist is already rendered below under the `field` renderer, and this adds a separate copy for each tag. Does it need to be duplicated for each tag like this?

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -109,35 +126,64 @@ const StringArrayEditor = createClass({
...
+
 	render : function() {
-		const valueElements = Object.values(this.state.valueContext).map((context, i)=>context.editing
-			? <React.Fragment key={i}>
-				<div className='input-group'>
-					<input type='text' className={`value ${this.valueIsValid(this.state.updateValue, i) ? '' : 'invalid'}`} autoFocus placeholder={this.props.placeholder}
+		const valueElements = Object.values(this.state.valueContext).map((context, i)=>{
+			return <React.Fragment key={i}>
+				<div aria-label={`tag ${context.value}`} className={`tag ${context.editing ? 'editable' : ''}`} tabIndex={-1} onKeyDown={(e)=>this.handleTagKeyDown(e, i)}>
+					<span className={`tag-text ${context.editing ? 'hidden' : 'visible'}`} key={i} onClick={()=>this.editValue(i)}>{context.value}</span>
```
calculuschild, line 161: If you click on an existing tag to edit it, you can't start typing right away. You have to click again to enter the text box. Is it possible to focus on the textbox with one click?

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -109,35 +126,64 @@ const StringArrayEditor = createClass({
...
+			event.preventDefault();
+			this.editValue(index);
+			setTimeout(()=>{
+				event.target.querySelector('input').focus();
+			}, 0);
+		} else if(_.includes(['Delete'], event.key)){
+			this.removeValue(index);
+		} else if(_.includes(['ArrowLeft'], event.key)){
+			event.target.previousElementSibling?.focus();
+		} else if(_.includes(['ArrowRight'], event.key)){
```
calculuschild, line 140: You can arrow-right all the way to the final text entry box, but you can't arrow-left from the text entry box back to the other tags.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -109,35 +126,64 @@ const StringArrayEditor = createClass({
...
+
+	},
+
+	renderDatalist : function(){
+		if(this.props.options?.length > 0){
+			return <datalist id={`${this.props.id}__tags-precoordinated`}>
+				{this.props.options.map((option)=>{
+					return <option value={`${option}`}></option>;
+				})}
+			</datalist>;
```
calculuschild, line 153: Is it possible to display a different value in the datalist options text? Just thinking, it may be nice to show something like `meta: ...` to indicate that `meta:` by itself isn't a valid option. Then when you click, it just applies `meta:` and keeps the datalist open to now show the suggestions to fill out the tag.

Or instead of `...`, the label shows some example sub-tags to explain what is meant:

Edit: Seems you can add text to the `<option>` element to display sub-text:

```
<option value="type:">(adventure, class, setting, etc.)</option>
```

image

Maybe combine this with my earlier comment to show a simplified list when the textbox is first empty so they just see a high-level list, but then use the full list after the user starts typing.

calculuschild: At some point, would it be worth it to tweak our existing dropdown menu thing so we have full customizability?

calculuschild (changes requested):
I like where this is going, but mainly the huge datalist seems a little unwieldy. Not sure how much control we have but I think we can make some tweaks.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
```diff
@@ -29,6 +32,11 @@ const StringArrayEditor = createClass({
 		};
 	},
 
+	componentDidMount : function() {
+		this.newTagInput = React.createRef();
+		this.uid = _.uniqueId();
```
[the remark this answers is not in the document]
Gazook89: Ah yes, that is a remnant from an earlier change, i'll remove that.

Gazook89:
It seems like most of the issue you have is with the datalist (justifiably). Rather than respond to each comment in the review, I'm just going to respond here.  

I pretty much just went with what was outlined in #2335, particularly in terms of what options should be in the datalist but also the tiny bit of discussion about how such a long list should be handled.  I agree the list is too long, and that such a long list can imply "choose one of these options" particularly without any prompt.  

I like the suggestions above but have to think about them a little more.  

In the meantime, I had another idea:  what if we just used simple tags that aren't "compounds" of "meta:" + "blah" for the *input*, instead just using different tag input fields for those categories and then turn them into compound tags after entry...so the UI would look something like this:

image

Is this more intuitive, that there are categories of tags?  And from this, each input area could have a datalist specific to that category.  In our system, the tags would be stored as we have them now, `meta:theme` etc, and everywhere else that tags are displayed they would show like `meta:theme`, but in this particular UI where users are editing their tags it would be more structured.

----

Also mentioned in the linked Discussion is the tag input component from Mantine.dev which might save some hassle.  Another one I was looking at is [accessible-components/tag-input,](https://github.com/accessible-components/tag-input) preview [here](https://tag-input.netlify.app/).  Unfortunately Radix doesn't have a pre-made tag input nor do they really have a pre-made combobox.  

---

I'll knock out a couple of the easy things you mentioned, push them up here, and by then hopefully have more thoughts on how to do the things asked above (or more ideas).

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
ericscheid: No, we only need one `<datalist>` element. Elsewhere (another PR?) the matter of separating `<datalist>` as it's own component function has been discussed. (It could also be used for simple `<input>` elements too, not just StringArrayEditors. We don't have a specific need for that yet.)

client/homebrew/editor/metadataEditor/tagDatalist.json
```diff
@@ -0,0 +1,55 @@
...
+		"type:Subclass",
+		"type:Feat",
+		"type:Item",
+		"type:Location",
+		"type:Monster",
+		"type:NPC",
+		"type:Race",
+		"type:Rule",
+		"type:Spell",
+		"system:D&amp;D 5e",
```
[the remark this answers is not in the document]
ericscheid: Looks good. The initial list was written in valid html, not written as a json object.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
calculuschild: I'm pretty sure `client/components/combobox.jsx` that we use for the Languages selector is already this.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
ericscheid: I've done a great deal of one-on-one usability testing over the last decade, including specifically with introducing datalist auto-completes to users. Many users are already familiar with type-ahead auto-complete from many environments (though not all users, of course). One general principle I found over and over is that users bring expectations of UI behaviour from other sites with them .. which means deviating from standard behaviour (no matter how well intended) can result in more confusion, frustration, and undermining of confidence.

I did find that a long initial droplist encourages free-form typing, which then triggers filtering of the droplist. Whereas a short droplist tends to result in selecting one of the predefined options, almost to the exclusion of free-form typing.

Please do review the individual options in the dataliast. I tried to ensure that there were no confusingly-similar synonyms, and that each option in each scheme had roughly the same level of abstraction.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
Gazook89: Can you try again?  I am able to use the arrow key to move from the final input to the previous tag without issue.

*However, you are correct if the final input has any text inside of it*-- the final input has to be cleared before you can use arrow keys to move out of it.  I'm pretty sure this was a conscious choice I made.  If someone wanted to move all the way to the beginning of their text using arrow keys, and held down the arrow key to do so, I wouldn't want to move them to the previous tag because they held down the key too long.   If necessary, I could detect if they are holding down the arrow key (rather than rapidly pressing it), and only prevent moving to the previous tag if true.  But I just didn't think it necessary at the time.

calculuschild:
> Is this more intuitive...

I see what you mean but I think the added clutter negates most of the benefit.

client/homebrew/editor/stringArrayEditor/stringArrayEditor.jsx
calculuschild: > However, you are correct if the final input has any text inside of it-- the final input has to be cleared before you can use arrow keys to move out of it.

Ah, this is what I am seeing then. I was just playing around moving left and right and when I entered the final input I couldn't get back out.

I agree with your assessment that holding down the left arrow shouldn't fully leave the box. Maybe if you can detect that the cursor is in the first character position and only transition left in that case.

Gazook89:
yeah, and it would be a problem if for some reason more 'special' categories were added...you'd have to keep adding more inputs.  

Another idea is to make each tag two inputs-- a dropdown for the tag scheme (meta, type, etc) and then a scheme-specific combobox.  The left button throws a dropdown, and the chosen value determines which datalist is presented in the right-hand side combo input. 

image

Further, it could still allow typing `meta:theme` in the right hand side for convenience, and on 'submission' would split the tag scheme portion to the left hand side (if valid).  My example above should have a 'blank' option, as well.  For regular tags, just type in the right-hand side.

The Author's StringArrayEditor/TagInput would still need to be more basic, so I'm not sure if the proposal above would just be a different component, or a configuration of the same component.  But just throwing ideas out.

ericscheid:
> In the meantime, I had another idea: what if we just used simple tags that aren't "compounds" of "meta:" + "blah" for the _input_, instead just using different tag input fields for those categories and then turn them into compound tags after entry...so the UI would look something like this:

Additional to the clutter aspect, this creates the risk that a user will enter a tag into the wrong field. Compare the combo field, where typing "ad" filters the droplist to `type:Adventure` and also `system:AD&D` in the same droplist.

calculuschild:
> Another idea is to make each tag two inputs-- a dropdown for the tag scheme (meta, type, etc) and then a scheme-specific combobox. The left button throws a dropdown, and the chosen value determines which datalist is presented in the right-hand side combo input.

This is kind of along the lines of what I was thinking when I wrote "Is there any way to trim it [the datalist] down initially with just the major categories or something, and then transition to the full list once the user starts typing?" I.e. make the datalist a two-step process, first picking the scheme and second filling out the tag name. Though I only pictured it as one single textbox.

I am intrigued by this "dual-tag" approach. I say we give it a shot.

Gazook89:
Discussed briefly in gitter:  going to move forward with working on something like the below.  Likely I will use a library like FloatingUI to handle the anchor positioning of menus (skipping CSS Anchor API for now since it is only available in Chrome and I don't want broken menus in firefox).  

image

Styling isn't locked down at this point.

calculuschild:
@Gazook89 Is this PR paused for now? Wondering if I should move it out of my priority list if this needs to simmer some more.

Gazook89:
Pause it for now-- I just haven't been as plugged in to VSC over the summer as I've got other projects going on (both house stuff and other things).  I wouldn't be bummed if someone branched off this and got it over the finish line, either (but I'll come back to it eventually too).

Gazook89:
This PR is going to close soon, in favor of a new branch.  I would keep this PR, to keep the comments and such, if anyone confidently knew if it was possible to basically take my new branch and drop it in to replace the existing branch for this PR.  But I'm not going to worry about that.

The new branch will be a functional component, but otherwise will have all the same targets as this branch did (and hopefully be more clear).  The add, remove, and edit functions on the new branch are working as of right now, but I haven't published it yet.

I just want to give a heads up in case anyone has any last comments to throw here.

5e-Cleric:
We can close after referencing the branch, that should be enough.

ericscheid:
> This PR is going to close soon, in favor of a new branch.

This PR is now closed, but do we have a link to the replacement branch/PR?

5e-Cleric:
> > This PR is going to close soon, in favor of a new branch.
> 
> This PR is now closed, but do we have a link to the replacement branch/PR?

Yeah, it was already merged, as referenced its #3735

Issue #241, opened 2016-12-04
[FEATURE]: Scroll Linking


dmmagic:
I'm not terribly familiar with Github, so I don't know where to suggest feature requests. Apologies if this is out of place.

It would be great if you could click a lock button to link the code on the left with the output on the right. So when you scroll to a page on the right, it moves to the appropriate section of code on the right -- and vise versa. And then you could click the lock again to unlink them for independant scrolling.

stolksdorf:
I will probably not implement this. To do it well will be complicated and will probably lag the UI while scrolling.

dmmagic:
Totally understand. Maybe something for the future. :)

kilkwik:
If linking is complicated, is there a possible way to make a quick scroll hotkey... sometimes I get lost when I am looking through what I have created on the formatted side and i want to quickly find out where the code is written. Something that would generally take me to the right spot would be all I need. 

Since I am using Chrome, I'll usually try and cntl+f on a unique word that is on the right hand side, and then use the arrow keys to move to the right side. It's functional as a workaround. 

Clearly a P2/P3 request... but would probably save me some time considering I am working up something larger than a couple of pages. :)

Gazook89:
5 year old issue, but requested again today on the reddit. Quick google search has [this SO question,](https://stackoverflow.com/questions/29255744/make-side-by-side-markdown-preview-scroll-with-its-editor) which contains a working [example of a markdown editor with live scroll linking](https://panwriter.com/).

There is also [Remarkable](https://github.com/jonschlinkert/remarkable), with fairly smooth scrolling.  [Live Demo](https://jonschlinkert.github.io/remarkable/demo/)

Also, this is similar to #846 or addresses the same basic problem (being lost in the source pane).

calculuschild:
In my mind I imagine rather than smooth scrolling between the two, it would detect when you have entered the next page, and then the opposing view would glide into that position. That way we match page-by-page rather than trying to sync line-by-line if that makes sense.

G-Ambatte:
Once the functionality for #2162 is in place, then I think we should be able to add a basic Scroll Linking feature, similar to that described here.

dbolack-ab:
Do we want this to be

1.  a continuous scroll 
2.  a "Jump to my cursor hot key" ?  
3.  Both where a able to be toggled on and off?

Putting this on my radar for the release after next.

Gazook89:
Certainly if doing (1), it should be able to be toggled. 

[...]

PR #3483, opened 2024-05-18
Add Jump-To hotkeys

dbolack-ab:
This binds CTRL/META-SHIFT-LEFTARROW to brewJump and CTRL/META-SHIFT-RIGHTARROW to sourceJump.

Solves part one in https://github.com/naturalcrit/homebrewery/issues/241

edit: Reflect change in hotkey binding.

 - [x] Agree on which key binds to which jump.
 - [x] Verify brewJump Binding on PC
 - [x] Verify SourceJump Binding on PC
 - [ ]  Verify brewJump Binding on MAC
 - [ ] Verify SourceJump Binding on MAC
 - [ ] Verify brewJump Binding on a Linux
 - [ ] Verify SourceJump Binding on a Linux


5e-Cleric:
Those hotkeys are already in use in chrome
For me they open downloads, and dev tools console respectively.

calculuschild:
Now, we don't *need* to add more hotkeys at all. The original issue only suggested it as a compromise if we couldn't get scroll linking to work, but the other PR takes care of that.

dbolack-ab:
> Now, we don't _need_ to add more hotkeys at all (I would actually prefer we don't). One person in the original issue only suggested it as a lesser alternative if we couldn't get scroll linking to work, but the other PR takes care of that.

As a primarily hot key user, I know *I* would use the keys more than the buttons and live scrolling would annoy the crap out of me.

calculuschild:
Fair enough. Just didn't want to add a feature if nobody actually wanted it.

Gazook89:
I think hotkeys for navigating the page is a good idea. And this counts as navigating the page (it is changing the view). 

I do think we can remove hotkeys for many snippets (and I made most of those hot keys so no one should be offended). The snippet menus themselves should have better keyboard navigation (cough radix cough), but in the meantime if there is a snippet hot key that would be better used for something else, I think it’s fine to claim it. 

I intend to work on snippet menu very soon, 1) to convert to function components and 2) look at keyboard navigation and/or radix.

5e-Cleric:
I'm sorry but:

https://homebrewery.naturalcrit.com/share/1Em9UpHcoKws3xv65lWCRGzPDr3xPAWzvp9-Fp6cxI0Ac

Ctrl-M is used to input spans

client/homebrew/editor/editor.jsx
```diff
@@ -79,6 +85,19 @@ const Editor = createClass({
 		};
 	},
 
+	handleControlKeys : function(e){
+		if(!(e.ctrlKey || e.metaKey)) return;
+		console.log(e);
+		const X_KEY = 88;
+		if((!e.shiftKey) && (e.keyCode == X_KEY)) this.brewJump();
```
calculuschild, line 92: `CTRL + x` is "cut". We definitely don't want to overwrite that.

G-Ambatte: As a suggestion:
`CTRL + ALT + RIGHT` => move brew preview to match editor position
`CTRL+ ALT + LEFT` => move editor to match brew preview position

I don't believe these combinations currently have any existing use, please let me know if I'm mistaken

G-Ambatte:
`CTRL+SHIFT+RIGHT` moves the editor, on the _left_ half of the screen
`CTRL+SHIFT+LEFT` moves the preview, on the _right_ half of the screen

To me, this feels backwards - that it would be more intuitive if the operation were reversed.

---

This is, of course, a matter of opinion.

dbolack-ab:
> `CTRL+SHIFT+RIGHT` moves the editor, on the _left_ half of the screen `CTRL+SHIFT+LEFT` moves the preview, on the _right_ half of the screen
> 
> To me, this feels backwards - that it would be more intuitive if the operation were reversed.

I really don't have a strong opinion one way or the other.

dbolack-ab:
Does anyone have strong feelings over which arrow should impact which direction? 

 @calculuschild @G-Ambatte @ericscheid @5e-Cleric

calculuschild:
I would say the arrow should indicate the direction of info transfer.  I.e. pointing right is sending the position from left to right.

Gazook89:
With the :has() selector, or one of the other newer ones, could we style each pane based on which "scroll arrow" is being hovered?

This is kind of funky idea, but if a user is hovering the "scroll preview to match editor" button, the preview pane would show a faint, but visible, double-sided arrow overlaid on it.  And vice versa.

Quick mockup, for hovering the "scroll preview to match editor":
image

I've spent about 0.5 seconds thinking of the css for this, but it seems like it should be possible (or maybe it's a JS thing).  Maybe it even points which direction the pane will scroll....(but then you could have another argument about which way an arrow should point to indicate the page will scroll "up")

Obviously this doesn't solve the immediate question about which way arrows point, but I think it would make it very clear to the user what is about to happen, both on the 'first try' and every subsequent try.

client/homebrew/editor/editor.jsx
```diff
@@ -74,6 +76,10 @@ const Editor = createClass({
 
 	componentDidUpdate : function(prevProps, prevState, snapshot) {
 		this.highlightCustomMarkdown();
+		if(this.props.jumpSource) {
+			this.sourceJump();
+			this.setState({sourceJump: false});
+		}
```
calculuschild, line 82: Is this bit of code actually doing anything? Does `this.props.jumpSource` ever get set? And does `this.state.sourceJump` ever get used?

calculuschild:
Looks good. Gonna merge this.

PR #3484, opened 2024-05-18
Live Mirror Editor position in Preview pane.

dbolack-ab:
Extends https://github.com/naturalcrit/homebrewery/pull/3483 by adding live scroll following the editor window.

Toggled with Lock icon on center pane. Defaults to locked.

Catches mouseclicks, CTRL-HOME, CTRL-END. Also triggers on arrow keys and enter.

Done as a separate PR in case this is a really crap solution but 3483 isn't.

Resolves part 2 of #241 

To Verify:

 - [x] Verify "Scroll Lock" Toggle changes state and behavior
 - [x] Verify "Scroll Lock"  behavior persists after closing and reopening Homebrewery. ( Load a page, toggle the lock, close the window, reopen Homebrewery. )


5e-Cleric:
I believe a checkbox in the snippet bar would be in order. Might be time to have some settings in the editor pane.

calculuschild:
I think it would be more user friendly to just add a third button alongside the other two "sync" buttons on the divider bar. The original issue reporter suggested a "lock" icon, and I think something like that makes sense. Keep all the "sync" buttons near each other.

dbolack-ab:
> I think it would be more clear to just add a third button alongside the other two "sync" buttons on the divider bar. Keep all the "sync" buttons near each other.
> 
> The original issue reporter suggested a "lock" icon.

Try not to laugh but I did not realise that's what those icons were. :)

dbolack-ab:
Updated to match parent and switch hotkey to scrolllock because. Needs platform testing.

calculuschild:
If you click the lock button on one brew and then open a new brew, the behavior remains but the button shows unlocked. Button needs to load the correct icon based on the currently-selected option when a page is opened..

calculuschild:
I'm not able to get actual scroll sync to work, it appears to only sync based on cursor position in the text editor.

For example, when I scroll up and down on the brew preview side, it does not scroll the editor into the same page and vice/versa. Not sure if I am missing something but I haven't had time to look at the code yet in case it is a bug.

dbolack-ab:
> For example, when I scroll up and down on the brew preview side, it does not scroll the editor into the same page and vice/versa. Not sure if I am missing something but I haven't had time to look at the code yet to check where the bug might be.

Ah. You know, that's a mode I hadn't thought of. I only ran it one way - based on editor position.  Let me look at that.

dbolack-ab:
> If you click the lock button on one brew and then open a new brew, the behavior remains but the button shows unlocked. Button needs to load the correct icon based on the currently-selected option when a page is opened..

Open a new brew in a new window?

calculuschild:
> > If you click the lock button on one brew and then open a new brew, the behavior remains but the button shows unlocked. Button needs to load the correct icon based on the currently-selected option when a page is opened..
> 
> Open a new brew in a new window?
> 

For example if I am in the /new page and start working on a brew and click the lock, then I save it so it opens an /edit page. The lock behavior remains but the button doesn't match the actual state.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -117,6 +118,20 @@ const SplitPane = createClass({
 					onClick={()=>this.setState({ moveBrew: !this.state.moveBrew })} >
 					<i className='fas fa-arrow-right' />
 				</div>
+				<div id='scrollToggleDiv' className='arrow unlock'
+					style={{ left: this.state.currentDividerPos-4 }}
+					onClick={()=>{
+						this.setState({ liveScroll: !this.state.liveScroll });
+						if(document.getElementById('scrollToggle').classList.contains('fa-unlock')) {
+							document.getElementById('scrollToggle').className = 'fas fa-lock';
```
5e-Cleric, line 126: I would recommend setting as a constant the button and parent, as they are used twice, to increase readability.

`const toggle = document.getElementById('scrollToggle');`

But also, as we are using react, we could use react to our benefit, in this case, having this entire function outside the html itself, but part of the component, in a separate function, and using Refs to find the object even if the DOM is not loaded.

Gazook89:
Quick thoughts:

A. I wonder if the animated scrolling is motion sickness inducing.  I never thought about that with the existing two 'sync' buttons, but that is probably because those are quick, explicit triggers that happen once.  This 'link' button means that animations are triggering a lot more--- and with a less explicit trigger.  I don't know if this concern is justified.  If it is, I'm not sure if the best solution is to get rid of the animation and just snap to the correct page?  Or have that tied to a User Account setting ('prefers reduced motion') or to the browser/user-agent preferences via CSS ([@media (prefers-reduced-motion)](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)).  

B.  I know this was asked in gitter and is just a preference thing, but I think the "lock" should show "locked" *when it is currently locked*, and not the "future state".  Maybe another option to consider is to change the icon onHover.  

C.  This isn't part of your PR and doesn't need to be, but I have long thought those snippet bar buttons are odd looking-- they are the only rounded element in all of Homebrewery, the only ones with borders, the only ones with shadows (besides pages) and their font-size is larger than any other UI label when in fact they are the least important item on the screen.  Here is a simple vision:

image

This reduces the overlap clutter, reduces the visual dominance, increases cohesiveness, and gives us a starting point for more "preview pane" controls.  Most importantly for this PR, the button also has a visual cue that it is *toggled*-- the darkened background color.  This *matches the behavior of the "current editor" toggle just to the left of it*.

calculuschild:
I would recommend we leave any styling changes for a separate PR (where everyone can debate as much as they like) and just let this one focus on getting the function.

dbolack-ab:
> I would recommend we leave any styling changes for a separate PR (where everyone can debate as much as they like) and just let this one focus on getting the function.



> I would recommend we leave any styling changes for a separate PR (where everyone can debate as much as they like) and just let this one focus on getting the function.

:+1:

Gazook89:
> For example if I am in the /new page and start working on a brew and click the lock, then I save it so it opens an /edit page. The lock behavior remains but the button doesn't match the actual state.

I think the current setting just needs to be saved in localStorage, and loaded when the component mounts.  This is assuming that the setting should persist across brews and loads (which i think it should).

dbolack-ab:
> > For example if I am in the /new page and start working on a brew and click the lock, then I save it so it opens an /edit page. The lock behavior remains but the button doesn't match the actual state.
> 
> I think the current setting just needs to be saved in localStorage, and loaded when the component mounts. This is assuming that the setting should persist across brews and loads (which i think it should).
> 
> _Edit_: maybe sessionStorage is better here-- this way if you have multiple tabs open, the same setting isn't applied across all of them.

I wonder if this should be a document setting or a client setting....

5e-Cleric:
> I wonder if this should be a document setting or a client setting....

If it was me, this would be a user setting, but set in localStorage instead of DB, for an easier implementation.

>Edit: maybe sessionStorage is better here-- this way if you have multiple tabs open, the same setting isn't applied across all of them.

As long as they don't reload manually the page i don't think the re-render would update the lock, would it?

Gazook89:
The scroll-lock should be saved per client, not per brew.  If you open a few brews, it could be very weird to have the behavior of your editor change with each.

If you use sessionStorage, you could still have a few brews open in different tabs and use different scroll-lock behavior in each, since sessionStorage isn't shared across tabs/windows.

> As long as they don't reload manually the page i don't think the re-render would update the lock, would it?

I'm not sure I understand this question.  If the user reloads the page, it keeps the same sessionStorage value. [ From the MDN article:](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)

> A page session lasts as long as the tab or the browser is open, and survives over page reloads and restores.

5e-Cleric:
> > As long as they don't reload manually the page i don't think the re-render would update the lock, would it?
> 
> I'm not sure I understand this question. If the user reloads the page, it keeps the same sessionStorage value. [ From the MDN article:](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)

What i mean is, if they change the lock state in one tab, the rest of the tabs will be unlocked even if in the same session until they get a manual reload, would they not?

Gazook89:
Each tab is its own session

dbolack-ab:
> The scroll-lock should be saved per client, not per brew. If you open a few brews, it could be very weird to have the behavior of your editor change with each.

I'm beginning to wonder if the tools I worked with over the years were atypical. A setting akin to this would be part of the Project/Document configuration and not part of the General tool config. :)

5e-Cleric:
We could use a toDo list to know how the progress of the project, the basic functionality is in place, at least around editor cursor movements, but we still need the other way around, and to find the bugs that will inevitably follow.

dbolack-ab:
> We could use a toDo list to know how the progress of the project, the basic functionality is in place, at least around editor cursor movements, but we still need the other way around, and to find the bugs that will inevitably follow.

I'm not sure I'm following you here?

5e-Cleric:
> > We could use a toDo list to know how the progress of the project, the basic functionality is in place, at least around editor cursor movements, but we still need the other way around, and to find the bugs that will inevitably follow.
> 
> I'm not sure I'm following you here?


Too many thoughts at the same time to properly write a comment...

1. This still needs to be able to lock scrolling, and scroll the editor when preview is scrolled.

2. if there is anything left to do in this PR, we could use a checkbock list to know when to do a full review.

dbolack-ab:
> 1. This still needs to be able to lock scrolling, and scroll the editor when preview is scrolled.

It does lock scrolling from the editor to the preview.  That is the scope.

> 2. if there is anything left to do in this PR, we could use a checkbock list to know when to do a full review.

Work is complete as best I can determine.  There may be testing scenarios that did not occur to me.

5e-Cleric:
This PR only works some of the times, perhaps i am testing wrong?

https://github.com/user-attachments/assets/9895a572-0215-4e88-8013-cd5d9191499b

dbolack-ab:
> This PR only works some of the times, perhaps i am testing wrong?
> 

The lock is tied to the cursor position. Though I may have just discovered a fix to that.

dbolack-ab:
I've bodged in scroll linking when the scrolling is done via scrollbars.

I'm not convinced I did things the best react-y way so there may be a desire to rework this. I deliberately disabled the smooth-scrolling on the scrollbar-based jumps as it was causing timing issues and seemed unnecessary.

5e-Cleric:
> > This PR only works some of the times, perhaps i am testing wrong?
> 
> The lock is tied to the cursor position. Though I may have just discovered a fix to that.

even then, in my video you can see some positions just did not trigger a scroll.

That for cursor, for the current version, i think one scroll lock is triggering the other, as in scrolling in the editor triggers the automatic scroll in the preview which in turn triggers the automatic scroll of the editor, producing some jitter.

dbolack-ab:
> > > This PR only works some of the times, perhaps i am testing wrong?
> > 
> > 
> > The lock is tied to the cursor position. Though I may have just discovered a fix to that.
> 
> even then, in my video you can see some positions just did not trigger a scroll.
> 
> That for cursor, for the current version, i think one scroll lock is triggering the other, as in scrolling in the editor triggers the automatic scroll in the preview which in turn triggers the automatic scroll of the editor, producing some jitter.

That should be corrected with the last update. There were timing issues that caused it to ping-pong. I added some locking behaviour that cleaned that up.

5e-Cleric:
Seems to be working as expected, have not been able to find any bug so far.

Is it just me or there is much code from the previous iteration?

calculuschild:
Testing just now:

- Sync still only seems to work left to right. Did it work right to left at some point?

https://github.com/user-attachments/assets/2fb3b3b5-7c5a-491a-9b30-1c480e09bbda

- If you click the edge of the button so the cursor is hanging over the edge into the editor or brew renderer EVERYTHING BREAKS.

https://github.com/user-attachments/assets/bb7d0f22-ad3f-4897-a95d-c3cbd16b71d9

- For unity with the other buttons, I would reword the tooltip something like "sync Editor and Preview locations"

5e-Cleric:
This PR is obly supposed to mirror the editor position, as the title defines

dbolack-ab:
I am finding a couple of regressions here.

dbolack-ab:
> Testing just now:
> 
> * Sync still only seems to work left to right. Did it work right to left at some point?
> 
>  Recording.2024-09-07.001903.mp4 
> * If you click the edge of the button so the cursor is hanging over the edge into the editor or brew renderer EVERYTHING BREAKS.
> 
>  Recording.2024-09-07.002336.mp4 
> * For unity with the other buttons, I would reword the tooltip something like "sync Editor and Preview locations"

It did. Something new hung a new onScroll method so I've bodged them together.

dbolack-ab:
Regressions should be fixed.

5e-Cleric:
@dbolack-ab all sorts of junkiness going around, please revert.


https://github.com/user-attachments/assets/684edacc-a180-4576-9664-65dfb95a329d

- when scrolling up in preview nothing happens
- When scrolling up in editor, scrolls up preview, which in turn scrolls up editor... recursion happening.

- the width of the editor scrolls the preview(expected, not sure if avoidable)

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -42,6 +43,22 @@ const SplitPane = createClass({
...
+		});
+		const toggle = document.getElementById('scrollToggle');
+		const toggleDiv = document.getElementById('scrollToggleDiv');
+		if(loadLiveScroll) {
+			toggle.className = 'fas fa-lock';
+			toggleDiv.className = 'arrow lock';
+		} else {
+			toggle.className = 'fas fa-unlock';
+			toggleDiv.className = 'arrow unlock';
+		}
```
calculuschild, line 61: In React, we don't want to modify the DOM directly like this. This can get React's shadow copy of the DOM out of sync and unable to detect changes.

Instead of managing the Class changes here, you would rely on your `liveScroll` state and change the `render` output  below based on the state.

So the flow should be:

Click button -> change the component state -> Render the updated component based on the state value.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -89,6 +106,21 @@ const SplitPane = createClass({
...
+		const toggleDiv = toggle.parentElement;
+		if(flipLiveScroll) {
+			toggle.className = 'fas fa-lock';
+			toggleDiv.className = 'arrow lock';
+		} else {
+			toggle.className = 'fas fa-unlock';
+			toggleDiv.className = 'arrow unlock';
+		}
+		window.localStorage.setItem('liveScroll', String(flipLiveScroll));
+		this.setState({ liveScroll: flipLiveScroll });
```
calculuschild, line 122: Same thing about not modifying the DOM here. Sometimes it's ok to _read_ the contents of the DOM. But even then, there is a React way to do it.

So similar thing. Here, the logic would be:

Click toggle button -> toggle the state -> Render the updated component based on the state value.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -15,12 +15,13 @@ const SplitPane = createClass({
...
-			isDragging        : false,
-			moveSource        : false,
-			moveBrew          : false,
-			showMoveArrows    : true
+			currentDividerPos  : null,
+			windowWidth        : 0,
+			isDragging         : false,
+			moveSource         : false,
+			moveBrew           : false,
+			viewablePageNumber : 0,
```
calculuschild, line 23: `viewablePageNumber` doesn't seem to be used in `splitPane`

calculuschild:
General thought reading through all of this: There is a whole lot of handling for all the different possible ways a page could be changed (scrolling, arrows, END/HOME hotkey, etc). There has got to be an easier way to just globally trigger when the current page has changed.

On the Editor side, we can probably take advantage of the codeMirror [viewportChange](https://codemirror.net/5/doc/manual.html#event_viewportChange) event.

For the BrewRenderer side, we should be able to use the same trigger that the "viewer toolbar" uses to display the current page.

I think just hooking into those existing triggers would cut the complexity of this PR by about half.

client/homebrew/editor/editor.jsx
```diff
@@ -292,8 +356,11 @@ const Editor = createClass({
 	},
 
 	brewJump : function(targetPage=this.getCurrentPage()){
```
calculuschild, line 358: Instead of setting and unsetting a global `scrollingJump` everywhere, I think it might make more sense to just add a parameter to `brewJump()` and `sourceJump()` that determines if the scroll is smooth or instant, with a default of smooth. So something like

```suggestion
	brewJump : function(targetPage=this.getCurrentPage(), smooth=true){
```

dbolack-ab:
Pulling this for a full rework.

shared/naturalcrit/splitPane/splitPane.jsx
dbolack-ab: I don't remember adding that one. is it maybe debris from merges?

calculuschild:
Going to merge this now. Possibility for slight behavior tweaks in the future but I think this checking all the boxes now.

Unfortunately closing the PR deleted the deployment, and Heroku can only deploy PRs from other people's branches when the PR is first created (the menu for manual deployment can't see branches outside of the core Naturalcrit repository). However this will now be integrated into staging https://homebrewery-stage.herokuapp.com/new if people want to fiddle around with it.

PR #3485, opened 2024-05-18
Preserve cursor position in editors

Gazook89:
This PR modifies the editor so that it preserves the cursor position in each the Brew and Style editors.  It saves the locations in localStorage, and loads the related position when switching between editors and across page loads.  

One side effect is that it *does* focus the Brew editor on page load.   I don't think that is necessarily the worst thing, but I'm not sure that *this* change should be doing that.  But until there is a more focused effort on how `focus` moves about the page, I think it's fine to leave as-is.

Another thing I thought about while working on this:  it might be nice if when you open an editor (changing view or loading the page), the cursor did one or two stronger flashes so you can more easily spot where your cursor is when you first jump into a view.  Not likely going in this PR, or any, but mentioning it in case anyone has further thoughts.

I feel like the commit messages (and expandable descriptions) are clear enough.  i did try JSDoc annotations as well.


5e-Cleric:
Need a deployment to test

dbolack-ab:
> One side effect is that it _does_ focus the Brew editor on page load. I don't think that is necessarily the worst thing, but I'm not sure that _this_ change should be doing that. But until there is a more focused effort on how `focus` moves about the page, I think it's fine to leave as-is.

This comes across as a feature, not a bug, to me.
 
> Another thing I thought about while working on this: it might be nice if when you open an editor (changing view or loading the page), the cursor did one or two stronger flashes so you can more easily spot where your cursor is when you first jump into a view. Not likely going in this PR, or any, but mentioning it in case anyone has further thoughts.

This seems like a good idea for a future PR.

5e-Cleric:
> Another thing I thought about while working on this: it might be nice if when you open an editor (changing view or loading the page), the cursor did one or two stronger flashes so you can more easily spot where your cursor is when you first jump into a view. Not likely going in this PR, or any, but mentioning it in case anyone has further thoughts.

In CodeMirror, as far as i know, the cursor is an html div element:

image

Therefore it would be fairly simple to add styling to make this possible, added and then removed by js.

Gazook89:
Yeah, not hard, I was already messing with it but still going to leave it for another PR where I’ll present some options and solicit other ideas.

5e-Cleric:
image

Not seeing any localstorage item popping up, neither across loads.

dbolack-ab:
Duplidcumenting here from chat:

 Seems to work after I had my session loaded. Though I'm not sure if it didn't trigger right initially? I didn't see any local storage added. I switched to styles and it crashed. Reloaded to see if I could duplicate it and it then had the values. Worked fine. Reloaded. Deleted the local storage values, switched panes, crash. Looks like it doesn't gracefully handle the values not being in local storage.

dbolack-ab (approved):
This looks good to me.

5e-Cleric:
Does this work for either of you? perhaps i am testing wrong

Gazook89:
I believe this should work now, and is updated to the `master` branch as well.  I should give it another once-over but I'm headed to bed and then busy day tomorrow.  If someone wants to review it and find the bumps before I do, please do.

5e-Cleric:
Okay, seems to work on homepage fine. Testing in `\new` doesn't seem to work, not sure why, and the metadata editor makes the PR crash

Gazook89:
Ah okay. New makes sense because there isn’t a style editor. Obviously I didn’t test it well enough. Will fix

Gazook89:
Pushed a fix that checks if there is an editor instance before storing cursor position.  So should be fixed now.  

Please keep breaking it.

Gazook89:
Number 1 seems tricky. 

Number 2: how many ids do we want to be concurrently tracking cursor position across? 10 most recent documents?

calculuschild:
We do have some logic already that saves scroll position when switching between the text and style tab. I haven't looked at this code yet, but are we making sure not to re-invent what we have already?

Gazook89:
> We do have some logic already that saves scroll position when switching between the text and style tab. I haven't looked at this code yet, but are we making sure not to re-invent what we have already, or can the two be made to share some logic since thet seem to be related?

Worse!  It actually fights the scroll saved scroll position.  When you switch back, it sets focus in the editor which means it pulls the cursor into view.  So if I set the cursor near the end of a document, and then scroll up to the top, and then switch to Style editor and back, the scroll position will now be further down the document such that the cursor is just in view.

So, might have to rethink that.  Good call out.

Gazook89:
I'm going to close this PR at this point, though it will continue to live locally on my machine (not in `master`, just as a forgotten branch).  

I don't think it's tremendously useful, and there are some considerations to be considered first:

- When switching to an editor, should prior scroll position take precedence over prior cursor position, if you can't have both?
- Can both be preserved, where the cursor position is preserved even if it's outside the current scroll view?

And as a reminder to myself, calculuschild pointed out [this older merged PR](https://github.com/naturalcrit/homebrewery/pull/1763) which made me think that possibly the cursor position is already preserved (along with scroll position), but that switching documents/editors just isn't setting the focus in the editor after the switch.  I'll play with that a little later, locally.

PR #3490, opened 2024-05-21
Lint a bunch of things

calculuschild:
Several files have been updated without linting. Handling this here so PRs don't need to be obfuscated behind large amounts of spacing changes, etc.


Gazook89:
Oh boy, looking forward to merging master in….

calculuschild:
This should mostly only be files that are not currently under PR.

PR #3491, opened 2024-05-22
Print directly from Edit/Share/New pages

calculuschild:
Implements #2742
Supersedes #3464 , as it solves the same issue and removes the need for `/print` endpoint entirely. 

Appears to work fine with legacy/V3, cross-page variables, etc. All functional features in the brewRenderer are printed correctly.

#### Open Tasks:

- [x] When scrolled to the bottom of a long (7+ page) document, after printing, scrolling back up reveals that the top few pages have disappeared. Resizing the page, etc. makes them reappear. Seems to be due to the 	`overflow-y: unset;` line in the `@media print` CSS somehow?

image

- [x] Fix hotkey printing when iframe is focused https://github.com/naturalcrit/homebrewery/pull/3491#issuecomment-2126182602
- [x] Need to fully remove the `/print` page and associated routing code.
- [x] Printing via hotkey from iframe focus  should only work on Edit/Share/New


5e-Cleric:
> When scrolled to the bottom of a long (7+ page) document, after printing, scrolling back up reveals that the top few pages have disappeared. Resizing the page, etc. makes them reappear. Seems to be due to the overflow-y: unset; line in the @media print CSS somehow?

That doesn't make any sense, @media print is very clear at what it does

calculuschild:
Try it. Removing that line of CSS prevents the issue.

G-Ambatte:
I don't seem to be able to reproduce the issue, either on the deployment or in my local build. Can you share a link to the test doc on the deployment for further experimentation please?

calculuschild:
> I don't seem to be able to reproduce the issue, either on the deployment or in my local build. Can you share a link to the test doc on the deployment for further experimentation please?

It's just a blank document with 7 pages.

```
\page

\page

\page

\page

\page

\page
```

Scroll to the bottom. Print. Close print dialog. Scroll to the top. The top 2 pages are invisible.

Gazook89:
Doesn't happen for me, in chrome or firefox.  Tested with just the blank pages and with additional stuff, in the deployment.  I can try checking out to local if necessary

calculuschild:
Hmmmm. Well if it's just me.... Let me just try clearing my cache and everything. Maybe Chrome is just acting up. I will try to record a video as well if it's still there.

G-Ambatte:
We might want to catch `CTRL-P` from the iFrame.

When Brew Renderer is focused and `CTRL-P` pressed:
- on Share Page:
image
- on Edit Page:
image

Not 100% on how we might achieve that yet, though - potentially a future improvement if it proves too difficult for this iteration.

calculuschild:
I tested the issue again just now. It occurs for me in local and in the deployment, on both Chrome and Edge, but not Firefox.


https://github.com/naturalcrit/homebrewery/assets/5878534/fc804f1f-eb14-46f1-aa97-09db0047f5c9

G-Ambatte:
WTF

It didn't happen
It didn't happen
...I changed page size to Letter from A4, and it happened

calculuschild:
> We might want to catch CTRL-P from the iFrame.

I think I have a solution for this.

5e-Cleric:
Okay, who made a pact with a fey? who was it?

calculuschild:
Ok, so.... after lots of fiddling, I think I have a workaround, but it's dumb.

It appears that triggering the print dialog forces a reflow of the brewRenderer, and somehow it is using the `@media print` criteria for determining if an element is off the page for painting. Triggering a re-render in React isn't enough, because all that does is update the DOM, and the DOM hasn't changed.

If I add these lines to the printing step, it basically forces a repaint of the DOM, which works, but is really dumb.

```
window.frames['BrewRenderer'].contentWindow.print();
let node = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0);
node.style.display='none';
node.offsetHeight; // accessing this is enough to trigger a reflow
node.style.display='';
```

5e-Cleric:
> It is using the `@media print` criteria for determining if an element is off the page for painting. 

How can you know that? and are we sure about it?

There might be an alternative to the @media print

what if we apply css right before printing via js and then remove it later?

calculuschild:
> How can you know that? and are we sure about it?

The only thing I know for sure is that `overflow-y: unset` (or `overflow-y: visible`) in the `@media print` dialog is required to cause the issue, and that a DOM repaint fixes the issue.

calculuschild:
> what if we apply css right before printing via js and then remove it later?

I tried this as well. This does mostly the same thing as my solution, since since removing the CSS causes a repaint when the dialog closes. But it adds two downsides:

1) While the print dialog is up, the webpage in the background looks like the print output
2) This in turn causes you to lose your scroll position because the elements have shifted around

calculuschild:
Ok, I think the two major issues are fixed. Missing pages are handled and CTRL-P is now captured in the iframe.

G-Ambatte:
> Ok, I think the two major issues are fixed. Missing pages are handled and CTRL-P is now captured in the iframe.

Confirmed; my reproduction method no longer produces the issue.

5e-Cleric:
Shouldn't this PR delete the print page altogether apart from adding this function?
Or are we keeping it for the other mentioned reasons?

calculuschild:
> Shouldn't this PR delete the print page altogether apart from adding this function?
> Or are we keeping it for the other mentioned reasons?

Look up at the original post. It's already one of the open tasks.

5e-Cleric (approved):
Looks good to me, what is left?

Maybe changing the names for some methods? printPage to printFunction?

calculuschild:
Changed the function to `printCurrentBrew()` so it isn't confused with "page" components.

PR #3492, opened 2024-05-24
Fix admin vulnerability to Brute Force

5e-Cleric:
Added express-rate-limit package and implemented rate limiting for admin API login attempts

Have NOT tested this as of yet, feel free to do so. Should be a limit of 10 attempts per day.


5e-Cleric:
It works!

image

dbolack-ab (approved):
Seems straightforward.

G-Ambatte:
Here's a suggested tweak for `admin.api.js`:

```
// Define rate limiter options
const loginLimiter = rateLimit({
	timeWindow : 24 * 60 * 60 * 1000, // 24 hours window
	max        : 10, // limit each IP to 10 requests per timeWindow
	handler    : ()=>{throw { HBErrorCode: '100', code: 429, message: 'Too many requests' }; }
});
```

which results in:
image

I've just re-used HBErrorCode 100 on the proof of concept, we'd need to give this particular error its own code to use the Error Page. We might also want to do something similar for the `Access Denied` result.

G-Ambatte:
On the actual error message, I think `Too many failed login attempts, try again later` is probably better than specifically stating that the ban is tied to their IP address.

5e-Cleric:
I'd rebuild the UI for this, it is not a brew that is being accessed, but the admin page. But yeah i can work with that, good point.

Gazook89:
There are some spelling errors and awkward sentences that I think could be improved.  I'm no wordsmith (I use 10 words where will 1 will do), but here is my take:

<details>
<summary>

## 401
</summary>

```
## Authorization Required
You need to provide correct administrator credentials to access this page.

:

If you should have access but are not able to log-in, please reach out via [reddit mod mail](https://old.reddit.com/message/compose/?to=/r/homebrewery) or in the [hb_issues channel of the Discord of Many Things server](https://discord.gg/domt).
```
</details>

<details>
<summary>

## 403
</summary>

```
## Access Denied
You need to provide correct administrator credentials to access this page.  *Too many unsuccessful attempts will lock you out for a period of time.*

:

If you should have access but are not able to log-in, please reach out via [reddit mod mail](https://old.reddit.com/message/compose/?to=/r/homebrewery) or in the [hb_issues channel of the Discord of Many Things server](https://discord.gg/domt).
```
</details>

<details>
<summary>

## 403
</summary>

```
## You have run out of attempts
You have failed to provide correct credentials to access the page too many times.  Please wait a bit before retrying again.

:

If you should have access but are not able to log-in, please reach out via [reddit mod mail](https://old.reddit.com/message/compose/?to=/r/homebrewery) or in the [hb_issues channel of the Discord of Many Things server](https://discord.gg/domt).
```
</details>

Gazook89:
Likely conversation about error codes has already happened-- my eyes usually gloss over when it comes up-- so feel free to wave this away:

Is `403 - Forbidden` the correct error code to return if the credentials are bad?  My reading from [MDN] is that 403 is for when a user of a system is logged in, but tries to access a resource they don't have permission to access.  Like if a logged-in homebrewery user tried to navigate to the /admin/ pages and we didn't require an extra password, they just needed a special permission on their account which they didn't have.   (The flip side of this is if the developers had a special *permission* on their normal homebrewery account that allowed them access to the admin pages, rather than them having different username/password just to access admin pages).  

I think just returning the same `401 - Unauthorized` error after authentication failed would be fine, in which case the `401` error would just be what I wrote for the `403` error in the previous comment (basically just telling them at the outset about the possibility of lockout).

Gazook89 (changes requested):
I tend to leave a bunch of normal comments outside of a review and then have nothing left to say in the Review portion, so just to recap:

- Fix spelling and grammatical errors.  Suggested text was left in previous comment.
- Use `401` error for both the initial attempt at admin page, and after log-in failures.
- Remove console.log, and a few of the extra whitespace lines particularly in error index.

server/admin.api.js
```diff
@@ -1,29 +1,40 @@
...
+            const [username, password] = Buffer.from(req.get('authorization').split(' ').pop(), 'base64')
+                .toString('ascii')
+                .split(':');
+            if (process.env.ADMIN_USER === username && process.env.ADMIN_PASS === password) {
+                return next();
+            }
+            throw { HBErrorCode: '403', code: 403, message: 'Access denied' };
+        }
+    ]
 };
```
calculuschild, line 38: Use tabs for indentation

client/homebrew/pages/errorPage/errorPage.jsx
```diff
@@ -5,9 +5,11 @@ const Markdown   = require('../../../../shared/naturalcrit/markdown.js');
 const ErrorIndex = require('./errors/errorIndex.js');
 
 const ErrorPage = ({ brew })=>{
+	
 	// Retrieving the error text based on the brew's error code from ErrorIndex
 	const errorText = ErrorIndex({ brew })[brew.HBErrorCode.toString()] || '';
-
+	console.log(errorText);
```
calculuschild, line 11: Not sure we need to reprint the same error in the console if it's already being displayed to the user in the page.

calculuschild (dismissed):
I agree with all of @Gazook89 's review items. Otherwise just formatting cleanup and merge conflicts.

5e-Cleric:
Yes, this PR was not my top most priority, but it is more so now.

calculuschild (changes requested):
Need to be able to retry the password. Not sure why it goes straight to the error page after the first attempt.

5e-Cleric:
> After the first attempt, any page reloads go straight to the 401 error page, but the password input dialog never reappears. Not sure why that is though...
> 
> So after 10 page reloads you have run out of chances, even though you only got 1 chance.

Okay, this is me being dumb, i implemented this in two ways, and just one is needed.

I added the rate limiter first, because it is what my initial idea was.

then i though: oh this error is so ugly, it should get an error page

But now, the error page itself should block any bruteforce by returning a page even though the wrong credentials have been used, and rate limiter is not really used unless you reload.

We can go two different ways, remove the error page, and show ugly errors with rate limiter blocking after 10 tries, or actually make a sign in page for this, displaying errors accordingly.

calculuschild:
At this point I think I prefer just the error page and no rate limiter. I don't think we need to put a lot of effort into a fancy login page for what should be only used by admins.

5e-Cleric:
> Choosing between the options, I think I prefer just the error page and no rate limiter. I don't think we need to put a lot of effort into a fancy login page for what should be only used by admins.

To clarify, the password dialog will not reappear until reload. Would beat the purpose of the error page.

ericscheid:
A reminder that an error page might well not even be seen by an attacker doing brute force — they are unlikely to be sitting at a browser typing in passwords over an over. They are more likely using a script and `curl` to try a great many possibilities .. thus the usefulness of a rate limiter.

5e-Cleric:
> A reminder that an error page might well not even be seen by an attacker doing brute force — they are unlikely to be sitting at a browser typing in passwords over an over. They are more likely using a script and `curl` to try a great many possibilities .. thus the usefulness of a rate limiter.

My bruteforce script returns the password whenever it reaches a 200 http code, which an error page provides, i'm sure some more sophisticated approach at hacking us can be taken, but i do not know how to.

Right now, this is stopping the only script i have for testing, so unless you can provide a more robust option, we will be proceding.

package.json
```diff
@@ -100,6 +100,7 @@
     "expr-eval": "^2.0.2",
     "express": "^4.21.0",
     "express-async-handler": "^1.2.0",
+    "express-rate-limit": "^7.2.0",
```
calculuschild, line 103: We can remove this package now if it's not being used.

5e-Cleric: You added it to the project for the 429 issues, right?

calculuschild: Took it back out because it didn't make a difference.

calculuschild: I think this is the only thing. Remove that, fix any conflicts, and I will merge.

calculuschild (approved):

package.json
5e-Cleric: I will remove it as soon as i get home, but i won't be able to merge without another review..

Gazook89 (approved):
Just some remarks on messaging and a little bit of whitespace cleanup.

Marking 'approved' so as to not hold it up.

5e-Cleric:
Small PR, works great, thank you for the reviews and helpful comments! Specially @Gazook89, i know i am not the best at writing messages.

PR #3498, opened 2024-05-27
[FEATURE]: Add default page size setting

5e-Cleric:
Opened this as an upstream PR by a miss click, but shouldn't be a difference

This feature does not check for account, unlike the default save location, because it is not deemed so important.

Checks the new page style code for a page size, if a page size is already defined, it doesn't add anything. if there isn't and the default  is A4, adds the A4 snippet via simple concatenation.


5e-Cleric:
To test: 

- Login in deployment using your session cookie
- Navigate to your new page, notice if the page size is changed in the style tab
- Navigate to Account Page
- Change default size to A4
- Navigate back to new page, now page size will be set to A4 and the A4 page size snippet will be inserted at the top of the CSS

5e-Cleric:
Thinking about it, i proabbly should insert the snippet at the end of the CSS code, otherwise, @import won't work.

Gazook89:
If this is just adding a snippet automatically when you go to `/new/`, why would there be an `@import` in there already?  Unless you are doing "clone to new", the only thing that would be in the Style Editor at the beginning is the default example CSS.

What happens if someone does "clone to new"?  Presumably you wouldn't want to clone to new and then find that the document page size has been adjusted to your preference, right?

5e-Cleric:
> If this is just adding a snippet automatically when you go to `/new/`, why would there be an `@import` in there already? Unless you are doing "clone to new", the only thing that would be in the Style Editor at the beginning is the default example CSS.
> 
> What happens if someone does "clone to new"? Presumably you wouldn't want to clone to new and then find that the document page size has been adjusted to your preference, right?

It has been some time since i made this, but i believe i made it pretty secure to make sure i don't invalidate any import or font face statement and still works, i would have to check to be sure. Requires testing.

Also, i consider this a p3 priority at most.

Gazook89:
Just testing this for a second, I noticed that it is dropping 'undefined' in the first line when I create new from blank:

Output:
```
undefined

/* A4 Page Size */
.page{
    width  : 210mm;
    height : 296.8mm;
}
```

5e-Cleric:
I should probably move this to be inserted not into the editor but somewhere else, maybe in the blank theme?

It makes more sense, not sure why i did it this way.

Gazook89:
I'm not sure what you mean about putting it into the Blank theme?  If you mean you would alter the page size via the Blank theme files somehow, I am not sure that is the best route (or at least necessary at this time).  We shouldn't want to hide the fact that the page size is getting altered from our default-- adding the actual text of the snippet into the Style Editor where it is visible to users is good.  It tells them that something happened, and is a good signal to them that the dimensions can be further changed.

Maybe in the future the page dimensions could be changed via a menu in the Properties Editor, but I'm not sure there is any advantage to that.  

I don't know if my previous comments are what you inspired your response-- but if so, I should be clear that I don't think there is anything wrong with inserting the snippet in the Style editor.

dbolack-ab:
I'm super not-keen on having to detect this via a regex. Is there any other way we can do this?

ericscheid:
Is the problem being solved "user normally creates A4 brews, has to manually insert page size snippet into any new brew" ?

We already currently insert example css code for new brews (and not cloned brews) .. why not have that example css be the account preferred page size snippet?

So, instead of this:

```css
/*=======---  Example CSS styling  ---=======*/
/* Any CSS here will apply to your document! */

.myExampleClass {
	color: black;
}
```

have it insert this if `account && account.pagesize=="A4"`:
```css
/* A4 Page Size */
.page {
	width  : 210mm;
	height : 296.8mm;
}

/*=======---  Example CSS styling  ---=======*/
/* Any CSS here will apply to your document! */

.myExampleClass {
	color: black;
}
```
and insert this otherwise:
(i.e. `!account || account.pagesize == "US Letter" `)
```css
/* US Letter Page Size */
.page {
	width  : 215.9mm; /* 8.5in */
	height : 279.4mm; /* 11in */
}

/*=======---  Example CSS styling  ---=======*/
/* Any CSS here will apply to your document! */

.myExampleClass {
	color: black;
}
```

----

(I've not messed with the "example css" just now, though obviously it should also be reviewed in this new context, e.g. do we even need it anymore?)

(Also .. why do we not have a US Letter Page Size snippet?)

dbolack-ab:
Regarding the injection method - ASSUMING - that the `\page` mustaches and/or master pages get approved - would it not be better to do the dimensional injections via classes in the `\page` mustache and make sure that the Blank template has dimensional definitions for the page sizes?

5e-Cleric:
> Regarding the injection method - ASSUMING - that the `\page` mustaches and/or master pages get approved - would it not be better to do the dimensional injections via classes in the `\page` mustache and make sure that the Blank template has dimensional definitions for the page sizes?

Change the snippet injection of `\page` to already include the page size? I'm not completely against the idea, but i have a very strong point against it. Users should be able to change the size of pages with ease, and it is simpler to add one line in the style tab than find and replace hundreds of them (one per page).

dbolack-ab:
> Change the snippet injection of `\page` to already include the page size? I'm not completely against the idea, but i have a very strong point against it. Users should be able to change the size of pages with ease, and it is simpler to add one line in the style tab than find and replace hundreds of them (one per page).

My thinking was more:

1. Set all the page sizes the metadata tab will list in the Blank theme as classes.
2. Programatically inject the page size class name in the same part of code that we inject in a `\page` mustache corresponding to the metadata tab selection.

instead of scanning the styles or doing a :has() looking for a page class.

5e-Cleric:
> > Change the snippet injection of `\page` to already include the page size? I'm not completely against the idea, but i have a very strong point against it. Users should be able to change the size of pages with ease, and it is simpler to add one line in the style tab than find and replace hundreds of them (one per page).
> 
> My thinking was more:
> 
> 1. Set all the page sizes the metadata tab will list in the Blank theme as classes.
> 2. Programatically inject the page size class name in the same part of code that we inject in a `\page` mustache corresponding to the metadata tab selection.
> 
> instead of scanning the styles or doing a :has() looking for a page class.

What? scanning? this is for a *default* page size, we will not be scanning the code for anything.

client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -55,19 +56,31 @@ const NewPage = createClass({
...
-			// brew.description = metaStorage?.description || this.state.brew.description;
+		
+			// Check if a page size is already defined in brew.style
+			const pageSizeRegex = /\.page\s*{[^}]*\b(width|height)\b[^}]*}/;
+			if (!pageSizeRegex.test(brew.style)) {
+				// No page size defined, apply the default size if sizeStorage is 'A4', otherwise the default letter stays
+				if (sizeStorage === 'A4') {
+					brew.style = brew.style + '\n\n/* A4 Page Size */\n.page{\n    width  : 210mm;\n    height : 296.8mm;\n}\n';
+				}
+			}
```
calculuschild, line 76: If this PR is only adding page size to `/new`, I would say it should specifically only add page size to EMPTY `/new` brews, so it doesn't surprise users with a sudden size change in the background if they return to a work in progress, or clone an existing brew. (i.e., imagine your personal setting is A4, but you clone a brew without a page size. The cloned brew will appear broken and it may not be obvious why the text content doesn't fit anymore).

This would also simplify this step, as we don't need to scan for an existing size. Just apply the size when a brand new brew is created.

calculuschild (approved):
I would change this to only apply a default page size on a *brand new* brew, not modify existing content that happens to be on `/new`.

calculuschild (changes requested):
I would change this to only apply a default page size on a *brand new* brew, not modify existing content that happens to be on `/new`.

5e-Cleric:
my perspective in this issue has changed, this should be changed in the theme .less, or some other similar file, not in the brew itself, it should be set in internal files and then overriden by the user in brews if necessary as it is now. If you agree i will close this and start from scratch

calculuschild:
> it should be set in internal files and then overriden by the user in brews if necessary as it is now. If you agree i will close this and start from scratch

How do you picture this working in the theme.less files? I'm not sure I follow how this would be implemented.

5e-Cleric:
> > it should be set in internal files and then overriden by the user in brews if necessary as it is now. If you agree i will close this and start from scratch
> 
> How do you picture this working in the theme.less files? I'm not sure I follow how this would be implemented.

not a question of how it'd be done, but of philosophy, it would be more appropiate. 

About the how, relatively simple:

1. get the size value from the setting
2. hardcode it in `.pages` or `.page` as a CSS custom property (`--pageDefaultSize`)
3. use that CSS var in the Less, we can even provide a css fallback in the var declaration for accounts without it.

like so:
```less
.pages {
    width: var(--pageDefaultWidth, 215.9mm);
    height: var(--pageDefaultHeight, 279.4mm);
}
```

5e-Cleric:
This is a stupid idea, instead custom CSS for all brews should be implemented. Closing.

PR #3499, opened 2024-05-28
Viewer tools

5e-Cleric:
AS per right now, only adds a zoom feature.

Creates toolbar in a separate file: toolbar.jsx

Zoom is set in toolbar, then passed to brewRenderer, and then set there in a style tag.

Two problems right now:

- [x] parent (brewRenderer) scrolls way past the element when scaling child down.
- [ ] transform origin is set on the center of the pages element, rather than the individual pages top center, which means it translates up or down the page. Should be set on top center of the individual page.

image


5e-Cleric:
Using `zoom` instead of `scale` sacrifices animatability for fixing overflow issues, which are inavoidable via CSS only with scale:

Based on docs about Overflow related to transform
> The [scrollable overflow area](https://drafts.csswg.org/css-overflow-3/#scrollable-overflow-region) is the union of:
> - The [scroll container](https://drafts.csswg.org/css-overflow-3/#scroll-container)’s own [padding box](https://drafts.csswg.org/css-box-4/#padding-box).
> - All [line boxes](https://drafts.csswg.org/css-inline-3/#line-box) directly contained by the [scroll container](https://drafts.csswg.org/css-overflow-3/#scroll-container).
> - The border boxes of all boxes for which it is the containing block and whose border boxes are positioned not wholly in the [negative scrollable overflow region](https://drafts.csswg.org/css-overflow-3/#negative-scrollable-overflow-region), accounting for transforms by projecting each box onto the plane of the element that establishes its [3D rendering context](https://drafts.csswg.org/css-transforms-2/#3d-rendering-context). [[CSS3-TRANSFORMS]](https://drafts.csswg.org/css-overflow-3/#biblio-css3-transforms)

From which i gather that the scroll container's scrollable overflow area is calculated based on the original size of the child element. This means that even though the scaled-down child element might visually occupy less space, the scrollable overflow area is determined by the original size of the element.

Zoom does change the natural size, affecting the layout, which gets rid of most problems inside the scrollable parent:

> To apply zoom, the [used value](https://www.w3.org/TR/css-cascade-5/#used-value) of a CSS property is pre-multiplied (before any other steps in the used value stage) by the used value of [zoom](https://drafts.csswg.org/css-viewport/#propdef-zoom) for the element. It also multiplies the [natural size](https://www.w3.org/TR/css-images-3/#natural-size) of all replaced elements, background images, and nested frames

5e-Cleric:
@calculuschild can i get a deployment?

calculuschild:
Deployment up.

5e-Cleric:
Converted to draft until i figure out the page works.

5e-Cleric:
Hmm as the scroll function scrolls to the top of the page, and the function getting the page gets if the footer is visible, page logic is flawed at the moment.

Gazook89:
Can I suggest moving any viewer tools down to the existing info bar we have at the bottom right of the preview pane?  Alternatively, eliminating the info bar at the bottom and placing that info in the same top bar you have.  I think the information could be condensed into one spot, but I'm not sure I have any real preference on where it is displayed.  My gut tells me that it shouldn't need to span the whole width unnecessarily-- unless we have enough tools/buttons that requires it.

Here is a mockup:
image


I also wouldn't have a strictly text input field here.  I understand this is early stages and you are just putting together basic functionality, so it's not a criticism, I just want to get it out there early.

I suggest some approaches, either singularly or some combination:

- +/- buttons to increment by fixed amounts (like 10% jumps)
- a `select` input, with options ranging from 50% to 200%...in my mind, options would would be `50`, `75`, `100`, `125`, `150`, `200`
- a combobox with the same options as the above, but with textual input as well if you want a custom amount.
- Or, use a slider/range input, with the same basic range.  It can utilize the `steps` attribute to have 'presets' at the values listed above as well.
- possibly a 'reset' button that is shown when the zoom is set to anything other than 100%.

5e-Cleric:
> Can I suggest moving any viewer tools down to the existing info bar we have at the bottom right of the preview pane? Alternatively, eliminating the info bar at the bottom and placing that info in the same top bar you have. I think the information could be condensed into one spot, but I'm not sure I have any real preference on where it is displayed. My gut tells me that it shouldn't need to span the whole width unnecessarily-- unless we have enough tools/buttons that requires it.
> 
> Here is a mockup-- note, I think we could probably get away without the word "zoom", and have just the +/- icons and the input box: image I also wouldn't have a strictly text input field here. I understand this is early stages and you are just putting together basic functionality, so it's not a criticism, I just want to get it out there early.
> 
> I suggest some approaches, either singularly or some combination:
> 
> * +/- buttons to increment by fixed amounts (like 10% jumps)
> * a `select` input, with options ranging from 50% to 200%...in my mind, options would would be `50`, `75`, `100`, `125`, `150`, `200`
> * a combobox with the same options as the above, but with textual input as well if you want a custom amount.
> * Or, use a slider/range input, with the same basic range.  It can utilize the `steps` attribute to have 'presets' at the values listed above as well.
> * possibly a 'reset' button that is shown when the zoom is set to anything other than 100%.

This i agree with, a slider or select is preferable. I have never used our combobox, so probably a slider. I will probably go for something akin to MS word's approach. Not sure to what degree we would need a reset button if we include a visual cue as to where 100% zoom is.

5e-Cleric:
added a range slider instead of a numeric input

Gazook89:
The slider doesn't change value when you press either button on the sides, just fyi.  Also, range inputs can have a `<datalist>` with `<option>` child elements, which set little 'ticks' on the line.  When you move the slider close to them, they 'lock' onto those ticks-- so if you set an option at 100%, it would be easy for users to get back to 100% zoom.   [See here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#adding_tick_marks).

You can also set the `title` attribute on the slider to equal the `value`-- so when you hover over the slider, the tooltip tells you what zoom level it is at.

You might also consider `step` attribute, and setting it to increments of 10. But I'm not even sure if I'm favor of that; just mentioning it in case you hadn't considered it.

----

Regarding the number input you have, for selecting page:    What if it was a dropdown menu, with all the headings in the brew?  Incrementing one by one on page numbers isn't terribly helpful when you can just scroll or use the scrollbar -- but if there was "flat" (non-indented, non-formatted) "ToC" that allowed to you jump to specific portions of the brew, that would be great.  In many cases, it may mean that users don't feel the need to create a Table of Contents in their brew (if they don't plan to really publish). We already have a method of getting all the headers for our TOC generator, so this would be similar, but rather than outputting formatted md we would just slot them into a datalist.

A numeric input box could replace the current "page number" down in the bottom right of the preview pane, so you could still type a page number in if you knew you needed to jump to page 200.

----

All said, something like the below.  The exact same design could be placed at the top of the preview pane instead (which may in fact make a dropdown menu easier if we aren't using a UI library with smart dropdown placement).  

image

5e-Cleric:
> Regarding the number input you have, for selecting page: What if it was a dropdown menu, with all the headings in the brew? Incrementing one by one on page numbers isn't terribly helpful when you can just scroll or use the scrollbar -- but if there was "flat" (non-indented, non-formatted) "ToC" that allowed to you jump to specific portions of the brew, that would be great. In many cases, it may mean that users don't feel the need to create a Table of Contents in their brew (if they don't plan to really publish). We already have a method of getting all the headers for our TOC generator, so this would be similar, but rather than outputting formatted md we would just slot them into a datalist.

> A numeric input box could replace the current "page number" down in the bottom right of the preview pane, so you could still type a page number in if you knew you needed to jump to page 200.

This does not sound like a good idea, i have a 90 page document which has a ToC that spans 200 lines **In the markdown**, this would create a select too big to handle, even with internal scrolling.

We will implement navigation via headers in that other PR, with another UI, but this one should not go that far IMHO. I guess we could go for only h1 or h2, but still sounds like a bad idea to me. I would like more people's opinion on this, though.

Gazook89:
I would say that it’s fine to skip the header quick navigation menu as I described, though I do think it provides value even (and especially) with large brews. 

The important bit is that the number input doesn’t have increment/decrement buttons, and that it exists within or replaces the framework of the “page number / total pages” display we already have.

5e-Cleric:
>The important bit is that the number input doesn’t have increment/decrement buttons

Wym?
image

5e-Cleric:
Added custom zoom level tooltip, synced zoom buttons to state input, and updated the styling

Gazook89:
> > The important bit is that the number input doesn’t have increment/decrement buttons
> 
> Wym?

I mean to suggest that it *shouldn't* have the increment buttons.  I just don't think they add much value.  Who wants a small button to increase the page number one at a time, placed directly above another small button that does exactly the opposite (decreases the page number)?  In accessibility terms, it seems like a problematic set of buttons.   And, it just seems like if anyone wants to move page by page, they would scroll either with the scroll wheel or the scrollbar.     Finally, leaving the increment buttons there takes as much space as the number itself.  

Per this [SO](https://stackoverflow.com/a/66759105/14335838) answer, it may be best to use a `type='text'` input.  They give some reasons for it that I think are basically overkill for our particular use case (i doubt anyone mistakenly types in non-numeric characters in this case).  But importantly here, the `text` type won't include the increment buttons.

Alternatively, you could use CSS to hide those buttons, but that seems possibly troublesome with different browsers.

Similarly, for the same reasons, I don't think we need separate buttons for changing the page one by one.  However, if I had to choose between the _increment buttons inside the input_  versus the _increment buttons on the sides_, I would go with the latter option, but decrease the space used for them.   Like this:

image

-----

> Added custom zoom level tooltip, synced zoom buttons to state input, and updated the styling

I don't see the tooltip working in the last deployment.  It should just be done like this (but with dynamic value, of course):

```html
<div class="tool" title="100%">...</div>
```

I am not sure if it's preferable to have the title attribute on the `tool` div or on the input itself...I think on the input itself is probably more "correct".   But here I am opting for the container div because it is easier to trigger the tooltip since the element area is larger; you do not have to hover directly on the slider.

5e-Cleric:
> I don't see the tooltip working in the last deployment. It should just be done like this (but with dynamic value, of course):

Did you wait for the newest deployment to load? It is a pseudo element with `content:attr(value)` on the input, that loads when you hover over its parent (to give more hoverable space to see it).
image

5e-Cleric:
> I mean to suggest that it _shouldn't_ have the increment buttons. I just don't think they add much value. Who wants a small button to increase the page number one at a time, placed directly above another small button that does exactly the opposite (decreases the page number)? In accessibility terms, it seems like a problematic set of buttons. And, it just seems like if anyone wants to move page by page, they would scroll either with the scroll wheel or the scrollbar. Finally, leaving the increment buttons there takes as much space as the number itself.
> 
> Per this [SO](https://stackoverflow.com/a/66759105/14335838) answer, it may be best to use a `type='text'` input. They give some reasons for it that I think are basically overkill for our particular use case (i doubt anyone mistakenly types in non-numeric characters in this case). But importantly here, the `text` type won't include the increment buttons.
> 
> Similarly, for the same reasons, I don't think we need separate buttons for changing the page one by one. However, if I had to choose between the _increment buttons inside the input_ versus the _increment buttons on the sides_, I would go with the latter option, but decrease the space used for them. 

I don't really care about the page navigation buttons, but it is usual to have them, and we don't need that many tools in the bar. Same reason why every button is so big, there is no real need for space.

I will give a try at moving this UI with the page numbering, although page zoom there doesn't feel right, some programs do it this way. 

Do you think we should capture the `Ctrl + mouse wheel Up/Down` event, and make it only change zoom on preview, or should we keep it as is using the UA native functionality?

Gazook89:
Just tried it again, I don't see that zoom level element.  [giphy](http://www.giphy.com/gifs/4LXOMBYCw1V5rrjuLG).  I don't see any new elements/pseudo-elements in the Inspector or similar changes.

-----

When I think of zoom levels I think of MS Office apps, with the zoom down in bottom right.  But of course, that's one program (or suite).  Obviously browsers do it different, and Apple office-like software puts it in the top left of toolbars.

image


BUT-- it's fine if the toolbar is at the top rather than bottom right.  I just think that the stuff in our existing toolbar should be part of this new toolbar you are creating.  Put it in either place, but we don't need both.

----

> Do you think we should capture the Ctrl + mouse wheel Up/Down event, and make it only change zoom on preview, or should we keep it as is using the UA native functionality?

I don't think we should hijack the browser zoom functionality.  I think that is probably an accessibility problem.  Someone might have stronger feelings about this, though.

-----
Just as an observation, some options like this for zoom might be cool, but can be done in other ways and in later PR's, if at all-- from Apple "Pages" app:

image

5e-Cleric:
> Just tried it again, I don't see that zoom level element. [giphy](http://www.giphy.com/gifs/4LXOMBYCw1V5rrjuLG). I don't see any new elements/pseudo-elements in the Inspector or similar changes.

Is this a tablet or similar touch screen device? because that would explain it.

5e-Cleric:
> Just as an observation, some options like this for zoom might be cool, but can be done in other ways and in later PR's, if at all-- from Apple "Pages" app:

I would do it with icons like Acrobat:
image

There is a menu with their names, but you can have them without like this. And the icons are pretty self explanatory.

I can add the two pages vs 1 page option, but our CSS to view pages is pretty weird, so i am not sure it will work with any page size.

Gazook89:
Yeah I wouldn’t worry about the extra options right now, except to think of theoretical placement of them if they are added in future PR. 

Not on touch screen. Desktop chrome and Firefox. I am now out of the house, but later I can really dig into it to see what’s up. Would be good to get third eyes on it too

5e-Cleric:
> Yeah I wouldn’t worry about the extra options right now, except to think of theoretical placement of them if they are added in future PR.
> 
> Not on touch screen. Desktop chrome and Firefox. I am now out of the house, but later I can really dig into it to see what’s up. Would be good to get third eyes on it too

Firefox might be the issue, will check if it is any css that i used

Gazook89:
I think there is something up with the deployment itself.  I don't know what, though.

I pulled this down to my local dev environment and was able to use it (the pull down of the branch didn't quite work as expected but there was simple hurdle that I think was unrelated to the actual branch).  

## datalist
Your datalist doesn't need an `option` for every possible value.  It unnecessary to have an option for every "10" if you have the `step="10"` attribute, which limits the selection to multiples of 10 already.  

Here is that portion, revised to be much simpler:
```html
				<datalist id='zoomLevels'>
                    <option value='100' />
				</datalist>
```

You don't need to create a pseudo-element for the tick marks, either.  They come with the range slider for free, but you cannot use `appearance: none`.  I understand that `appearance: none` is likely set so you could further customize the slider (make the thumb square, for example).  My [reading on the topic](https://css-tricks.com/sliding-nightmare-understanding-range-input/) is that styling ranges is still pretty complicated to get right in every browser.  But I do think the browser styling is mostly fine, especially since you can easily change the "accent" color...Simplifying the CSS does this:

```css
&.slider {
			color: #D3D3D3;
			accent-color: #d3d3d3;
  
			&::-webkit-slider-thumb, &::-moz-slider-thumb {
				cursor           : pointer; 
			}

			&:hover::after {
				position         : absolute;
				bottom           : -30px;
				left             : 50%;
				z-index          : 1;
				display          : grid;
				place-items      : center;
				width            : 4ch;
				height           : 1.2lh;
				pointer-events   : none;
				content          : attr(value);
				background-color : #D3D3D3;
				border           : 1px solid #A1A1A1;
				border-radius    : 5px;
				transform        : translate(-50%, 50%);
			}
		}
```
### Chrome
image

### Firefox
image

(I do think the -moz- thumb is a little on the large size, for what it's worth...)


-----

Finally, is there a reason to not update the zoom `onChange`, rather than `onMouseUp`?  Unless there is some downside (maybe performance?), I would much rather the zoom is reflected in the Preview immediately, so that I can get the size *just right* before letting go of my mouse.

5e-Cleric:
> Finally, is there a reason to not update the zoom onChange, rather than onMouseUp? Unless there is some downside (maybe performance?), I would much rather the zoom is reflected in the Preview immediately, so that I can get the size just right before letting go of my mouse.

Its just that if i don't use useChange, the console emits an error, javascript expects onchange for some reason.

Gazook89:
> Its just that if i don't use useChange, the console emits an error, javascript expects onchange for some reason.

Hm, I'm not having any issue at least on local.  I didn't look super closely at the code when i did this earlier, i just switched `onMouseUp` to `onChange` and then commented out the other `onChange` attribute (since you can only have one).  I have no idea if the first/original one is really needed-- I truly have not looked closely at it. But this works, no errors:

```jsx
<input
					className='slider'
					type='range'
					name='zoom'
					list='zoomLevels'
					min={minZoom}
					max={maxZoom}
					step={zoomStep}
					value={state.zoomInput}
					// onChange={(e)=>handleInputChange(e.target.value, 'zoom')}
					onChange={(e)=>{
						const newZoomLevel = parseInt(e.target.value, 10);
						if(newZoomLevel !== state.zoomLevel) {
							setState((prevState)=>({
								...prevState,
								zoomLevel : newZoomLevel,
								zoomInput : newZoomLevel,
							}));
							updateZoom(newZoomLevel);
						}
					}}
				/>
```

Gazook89:
I have a branch that is basically this, below, but I think I created it from the main repo rather than your fork/branch.  So I need to rebase it onto yours, if you wanted to merge it into your branch.  I can do that later, but gotta start making dinner.



It also simplifies a lot of the CSS, which isn't really visible here.  

My branch, for now, is here on my fork but with no deployment if you just want to see what the code changes look like: https://github.com/Gazook89/homebrewery/commits/Simplify-Zoom-Slider/

Gazook89:
Everything functions now.  I think there is a little more to do (I think the scrollToPage method can be just inside the toolbar component rather than brewRenderer.jsx) but since everything is working again wanted to drop an update:

gif

(if someone isn't following along, this is from my branch which is a branch of this PR, [can be found here](https://github.com/naturalcrit/homebrewery/compare/master...Gazook89:homebrewery:Simplify-Zoom-Slider))

Gazook89:
I mistakenly merged my branch on top of this branch (5e's branch).  Victor seems fine with the result of that merge, though, despite my git buffoonery.  So I think it's ready to be picked apart.  

I *suspect* it could still be improved, such as by eliminating DOM manipulation (`.getElementByID`) in favor of using references (`useRef()`), and by pulling more logic into the Toolbar component rather than having it inside brewRenderer.jsx utilizing `useContext`.  But I think my brain reached its limit for that.  

One curiosity in the implementation is that if you are zoomed out, where you can see 4+ pages, the "current page number" that is displayed is a little unclear on how it's being calculated.  When scrolled all the way to the top, it shows "1", and when scrolled all the way to the bottom, it shows the number of the last page--- so I'm not sure where the "dividing line" is between different page numbers, if that makes any sense.   For comparison, in Mac Preview app, if you are zoomed out where you can see multiple pages, the page that is in the *middle* of the viewer is the number that is displayed as "current page".....and it's not possible for the "current page" to show the first or last pages  (when you are zoomed out far).    This isn't a blocker, and I don't think it's really even a problem, but it was a curious thing I spent a little time looking at.

5e-Cleric:
>I suspect it could still be improved, such as by eliminating DOM manipulation (.getElementByID) in favor of using references (useRef()), and by pulling more logic into the Toolbar component rather than having it inside brewRenderer.jsx utilizing useContext. But I think my brain reached its limit for that.

You're right, i should change the ref, but the props are okay, we don't need to change that to context.
I prefer to use context for larger stuff.

5e-Cleric:
>One curiosity in the implementation is that if you are zoomed out, where you can see 4+ pages, the "current page number" that is displayed is a little unclear on how it's being calculated. When scrolled all the way to the top, it shows "1", and when scrolled all the way to the bottom, it shows the number of the last page--- so I'm not sure where the "dividing line" is between different page numbers, if that makes any sense. For comparison, in Mac Preview app, if you are zoomed out where you can see multiple pages, the page that is in the middle of the viewer is the number that is displayed as "current page".....and it's not possible for the "current page" to show the first or last pages (when you are zoomed out far). This isn't a blocker, and I don't think it's really even a problem, but it was a curious thing I spent a little time looking at.

Yes, this and how the zoom works is something we can't really fix if we are working with CSS properties.

G-Ambatte:
- [ ] Clicking Zoom % text entry shows cursor, but you cannot change the first two digits of the zoom level via keyboard entry
  - [ ] tabbing out seems to break a lot of things
- [ ] Clicking increase/decrease arrows in text boxes does not result in immediate change, update only occurs on exit
  - [ ] throttled updates?
- [ ] "Match Preview to Source" does not move to correct page at zoom levels other than 100%
  - [ ] reset zoom level when function activated?
- [ ] (optional) smooth transition between pages/zoom levels

Gazook89:
Not going to add it in this PR, but noting it so I don’t forget: an additional button(s) to set the zoom so the page width fills the preview pane, and another to fit the whole page in the preview pane. Possibly a reset button (to 100%), but I’m less convinced that is necessary

calculuschild:
Just one thing noticed immediately: popup dialog appears twice. Haven't looked at the code yet but are we duplicating an element somewhere?

image

calculuschild:
Otherwise function looks good. 👍 Just want to check for code quality before merging.

Gazook89:
Actually I bet it is just rendering twice.  The first time is rendering without the Toolbar component, then second time it is after, and the position is changed.  

I believe this gets into the my fear about manipulating the DOM directly (getElementById) rather than using references.  So I will take a further dive on that.

Gazook89:
I should also probably fix this: 

image

calculuschild:
> I believe this gets into the my fear about manipulating the DOM directly (getElementById) rather than using references.

Ah, yep. React does not like it when you do that. It puts its internal shadow DOM out of sync with the browser contents.

calculuschild:
Printing is messed up. Print output should not be affected by viewer zoom:

image

Even at 100% zoom, print appears to be offset vertically:

image

Gazook89:
@ericscheid and I were just chatting about using `@print` query to basically override any css zoom property when sent to print (and with PR's in the pipeline, unset any CSS Grid/Flex layouts).  So I'm hopeful that solution works, and I'll get it added.

shared/helpers.js
```diff
@@ -28,7 +28,6 @@ const printCurrentBrew = ()=>{
 		window.frames['BrewRenderer'].contentWindow.print();
 		//Force DOM reflow; Print dialog causes a repaint, and @media print CSS somehow makes out-of-view pages disappear
 		const node = window.frames['BrewRenderer'].contentDocument.getElementsByClassName('brewRenderer').item(0);
-		node.style.display='none';
```
calculuschild, line 31: Any particular reason this was removed? This, plus the two lines below, were needed to prevent out-of-view pages from disappearing when the print dialog is opened. https://github.com/naturalcrit/homebrewery/pull/3491#issuecomment-2127685865

It's possible that bug no longer exists with newer Chrome versions, but still not sure what removing it here is needed for.

client/homebrew/brewRenderer/toolBar/toolBar.less
```diff
@@ -0,0 +1,104 @@
+.toolBar {
```
calculuschild, line 1: Go ahead and lint this

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -91,6 +96,21 @@ const BrewRenderer = (props)=>{
 			...prevState,
 			viewablePageNumber : Math.floor(target.scrollTop / target.scrollHeight * rawPages.length)
 		}));
+
+		getCurrentPage(e);
+	};
+
+	const getCurrentPage = (e) => {
```
calculuschild, line 103: Note, `viewablePageNumber` should already contain the current page in view, and it's already calculated during `handleScroll()`.

Feel free to rename `viewablePageNumber` to `currentPageNumber` if that makes more sense. But we don't need to calculate it twice.

shared/helpers.js
Gazook89: ah I'm not sure, I think that is from one of @5e-Cleric earlier commits on the branch and I haven't noticed it.  Victor, do you know if that was needed?   I'll revert that and see what happens.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -214,12 +246,19 @@ const BrewRenderer = (props)=>{
 					onKeyDown={handleControlKeys}
 					tabIndex={-1}
 					style={{ height: state.height }}>
+
+					<ErrorBar errors={props.errors} />
+					<div className='popups'>
+						<RenderWarnings />
+						<NotificationPopup />
+					</div>
```
calculuschild, line 254: Here it is. Errorbar and Popups have been added here but they are already present above, outside the iFrame.

client/homebrew/brewRenderer/brewRenderer.less
```diff
@@ -1,10 +1,13 @@
 @import (multiple, less) 'shared/naturalcrit/styles/reset.less';
 
 .brewRenderer {
-	will-change : transform;
-	overflow-y  : scroll;
+	will-change   : transform;
+	overflow-y    : scroll;
+	padding-block : 30px;
```
calculuschild, line 6: Print offset is caused by this `padding-block`.

shared/helpers.js
5e-Cleric: I think i removed that because at the time the toolbar was the first child of the component and this line completely disabled it? don't recall, but we should be fine with removing it.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: Probably a conflict badly resolved, we will delete this.

client/homebrew/brewRenderer/brewRenderer.less
5e-Cleric: @Gazook89 see, i knew this padding would come back to bite us in the ass.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -0,0 +1,124 @@
...
+					<input
+						id='page-input'
+						class='text-input'
+						type='text'
+						name='page'
+						inputMode='numeric'
+						pattern='[0-9]'
+						value={pageInput}
+						onChange={(e)=>{
+							handlePageChange(e.target.value == false ? e.target.value : parseInt(e.target.value));}}
```
calculuschild, line 101: For testing purposes, in what case would e.target.value == false, and what is the expected behavior of setting `pageInput = false`?

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -0,0 +1,124 @@
...
+const { useState, useEffect } = React;
+const _ = require('lodash')
+
+const MAX_ZOOM = 300;
+const MIN_ZOOM = 10;
+
+const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
+
+	const [zoomLevel, setZoomLevel] = useState(100);
+	const [pageInput, setPageInput] = useState(currentPage);
```
calculuschild, line 12: `zoomLevel` is not named `zoomInput`, even though it is tracking a similar value in the toolbar. Can we rename `pageInput` to something else more fitting?

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -0,0 +1,124 @@
+require('./toolBar.less');
+const React = require('react');
+const { useState, useEffect } = React;
+const _ = require('lodash')
+
+const MAX_ZOOM = 300;
+const MIN_ZOOM = 10;
+
+const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
```
calculuschild, line 9: `onPageChange` is not used

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+	useEffect(()=>{
+		onZoomChange(zoomLevel);
+	}, [zoomLevel]);
+
+	useEffect(()=>{
+		setPageInput(currentPage);
+	}, [currentPage])
+
+	const handleZoomChange = (delta)=>{
```
calculuschild, line 22: Can we rename 'handleZoomIncrement` to make clear this only adds/subtracts a delta from the zoom?

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -0,0 +1,124 @@
...
+		const zoomChange = _.clamp(zoomLevel + delta, MIN_ZOOM, MAX_ZOOM);
```
calculuschild, line 23: `newZoomLevel`

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+		setZoomLevel(zoomChange);
+	};
+
+	const handlePageChange = (page)=>{
+		setPageInput(page);
+	};
```
calculuschild, line 30: Can we just use `setPageInput` directly instead of having this separate handler? (similar to how we use `setZoomLevel` directly on the zoom range slider.

calculuschild:
Ok, I think that's all my comments for now. I can re-review once those are resolved, but I don't expect much else requiring changes.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: So rename `handleZoomChange`?   Would `handleZoomButton` work?

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: If the value entered is zero, it would return false, right?  And so if it's false, it will populate the field with it's value, which is the pageNum state which remains what it was (since it hasn't been updated yet).

shared/helpers.js
Gazook89: reverted.  cls

client/homebrew/brewRenderer/toolBar/toolBar.less
Gazook89: linted. cls

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: yup, done and done.  cls

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: done

client/homebrew/brewRenderer/toolBar/toolBar.jsx
calculuschild: Its saving the input value whether its 0 or not. The only difference is we use parseInt for numbers > 0 but not for 0?

client/homebrew/brewRenderer/brewRenderer.jsx
Gazook89: got rid of handleScroll() function entirely in favor of getCurrentPage().  handleScroll was giving me an issue, and getCurrentPage seems to function fine.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: removed

client/homebrew/brewRenderer/toolBar/toolBar.jsx
calculuschild: That's fine. Just to clarify its handling a specific type of zoom change, not all zoom change.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: okay yeah, I'm not really sure what i was defending.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: Regarding your edit:    since there is a controlled text input, I can't just set it's `value` to `currentPage`, right?  Because then I wouldn't be able to change the text inside the box?

calculuschild: Should be able to still edit the textbox. It would display the inherited `currentPage` to start, and as soon as you enter a new value, or scroll up and down, a new `currentPage` would filter down and show the updated number.

calculuschild:
- If you delete the number in the page selector, or enter anything besides a numeral character, you get NaN and can't interact with the input box anymore:

image

- If you enter '0' while you are already at page 1, it will scroll to page 1, but still display '0'. Similarly, if you are already at the last page, you can enter a number greater than the last page and it will just display that number.
  - Probably need to set `pageNum` to the clamped number here, so it updates the inputbox even if the page isn't actually moved.

- is it possible to auto-select the text in the input box on click, so I can directly start typing the page number instead of deleting the old one, or trying to drag-select the text contents?

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: Actually, now addressing the input of non-integer values going to NaN, i think this handler method is an old vestige of preventing that issue that at some point was mistakenly stripped down.

So i'm likely going to bring this handler back, to check the input value before running parseInt on the input value.

Gazook89:
> you get NaN and can't interact with the input box anymore:

> If you enter '0' while you are already at page 1, it will scroll to page 1, but still display '0'. 

> is it possible to auto-select the text in the input box on clic

@calculuschild I think I have this all addressed in my last few commits.

Gazook89:
For some reason the navbar dropdown menus lose focus now that the toolbar is there.  It drops the menu as soon you move your cursor lower than the toolbar, even the though toolbar is *under* the menu (in the layer stack).  

So i'll try to fix that.

calculuschild:
Okey Dokey. Merging now. Good work guys. 👍

Issue #2278, opened 2022-08-14
Feature: Save Brew to Image


G-Ambatte:
The Region Capture feature of Chrome (Desktop v104 and up) allows a MediaStream to be cropped to a target element. It should be possible to use this to capture the browser's rendered output to either the clipboard (blob) or to a file (PNG).

[This article](https://developer.chrome.com/docs/web-platform/region-capture) has details about the feature.

---

[This paragraph](https://developer.chrome.com/docs/web-platform/region-capture/#occluding-and-occluded-content) mentions a potential future Element-level Capture API, which seems like it would be exactly what is needed for this functionality.

5e-Cleric:
Genuine doubt, what would be the difference between this and exporting a pdf and transforming that into a jpeg or png?

And between those two and Window's Snipping tool?

5e-Cleric:
Answering myself from the future: absolutely no difference.

G-Ambatte:
>what would be the difference between this and exporting a pdf and transforming that into a jpeg or png?

User convenience: ideally a one click process resulting in a high resolution image of the full content.

>And between those two and Window's Snipping tool?

Snipping Tool can result in lower resolution images, and while it is less complex than generating a PDF and converting that to an image, it is still a more complex process than direct image generation.

dbolack-ab:
I have a working draft for this issue not based on the chrome region grabbing but it has a few issues likely relating to CORS that I have not yet worked out, thus _**DRAFT**_.  https://github.com/naturalcrit/homebrewery/pull/3219

5e-Cleric:
@dbolack-ab i believe this issue refers to making the ENTIRE BREW to an image, not just 1 page.

dbolack-ab:
> > @dbolack-ab i believe this issue refers to making the ENTIRE BREW to an image, not just 1 page.
> 
> you would still do it page by page (capturing the page element).

Why would you do that? To make a CBR?

dbolack-ab:
Is this still desired?

G-Ambatte:
I believe so, I'm pretty sure it's one of the more desired features - people make a one page magic item, then have to jump through some hoops to share it.

Personally, I've put together something on the fly during the session and then had to quickly use the Snipping Tool to paste the page to Discord. So I would find the ability to save as or copy to the clipboard would be a very useful feature.

dbolack-ab:
I think we should consider closing this.

PR #3219, opened 2023-12-31
Automatic Brew thumbnail generation option

dbolack-ab:
This is a feature addition that adds the option of creating a thumbnail directly from the first page of the brew.

MagicImage

PR #3501, opened 2024-05-29
Improved Automatic Brew thumbnail generation option

dbolack-ab:
This PR replaces #3219.

This is a feature addition that adds the option of creating a thumbnail directly from the first page of the brew.

MagicImage

This feature had issues with XSS support/non-support at various offsite resources. This PR resolves that issue by creating a proxy endpoint that loads for those resources.

An additional field, thumbnailSm, has been added to the brewStub schema to hold this data, if used.


5e-Cleric:
Closes #2278 to some degree.

5e-Cleric:
Okay, first thoughts, the thumbnail doesn't show in the userpage brew card, as do normal images. Or on links, as far as i can see (checked in discord).The functionality seems to work, but if the image doesn't show, the PR is not a success. What could be the issue here?

Gazook89:
Where is the thumbnail image stored? Right now we just take hotlinks to other image hosts.  Does this PR make it so we are storing images or base64 somewhere?

dbolack-ab:
> Where is the thumbnail image stored? Right now we just take hotlinks to other image hosts. Does this PR make it so we are storing images or base64 somewhere?

If the `.thumbnail` value is 'Page 1' then it is in `.thumbnailSm`

dbolack-ab:
I have updated Vault to look at thumbnailSm.

I know that the UI solution here is a *little* clumsy but it was the lightest touch solution. If we need something a little more elegant then we should kick this back to draft.

Gazook89:
Just for the record, this was closed at least partially because of storage concerns, for either images or base64.  As mentioned in Gitter, the "demon cover page" in the snippets generates a 9kb thumbnail and the current average brew size is 6kb.  Additionally, some concern about starting to store images in mongo and then the possible future ability to store images in an actual image hosting service-- having split locations etc etc.  

A possible option is to preserve this PR but make the generator only offer the generated image to the user at the moment it is generated, and let the user download/copy-paste to imgur or other host and then drop their link into the normal thumbnail field.

PR #3502, opened 2024-05-30
[DRAFT] Homebrewery as an Electron App

G-Ambatte:
This draft PR is some initial work towards enabling the use of Homebrewery locally as a standalone Electron app.

While this PR produces the base functionality, the following issues have been identified:

- [ ] `prompt`, `alert`, `confirm` all do not work in ElectronJS
  - [ ] login system does not work - uses `prompt`
- [ ] switching to a new page often opens a new window

If you experiment with this and strike any issues, please record them here.

### DIFF
--- a/electron-main.js
+++ b/electron-main.js
@@ -1,0 +1,65 @@
+// Modules to control application life and create native browser window
+const { app, BrowserWindow } = require('electron');
+const path = require('path');
+// const express = require('express');
+// const cors = require('cors');
+
+const DB = require('./server/db.js');
+const server = require('./server/app.js');
+const config = require('./server/config.js');
+let PORT;
+
+const startLocalServer = (done)=>{
+	DB.connect(config).then(()=>{
+		// Ensure that we have successfully connected to the database
+		// before launching server
+		PORT = process.env.PORT || config.get('web_port') || 8000;
+		server.app.listen(PORT, ()=>{
+			console.log(`\n\tServer started at: ${new Date().toLocaleString()}`);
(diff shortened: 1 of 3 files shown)


G-Ambatte:
> we need to change how links work in local

Or just specify a new value for `NODE_ENV` : `electron` or something similar.

G-Ambatte:
> login system does not work - uses prompt

This makes me think of `<dialog>` element `showModal()` method. It's not blocking, but it does present in the `top-layer` so cannot be behind anything, and all elements outside of the `<dialog>` are rendered inert while it is present, so the user cannot perform any other actions while it is displayed - but crucially, server side functions can continue as normal.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal

Gazook89:
Why would there be an account system at all?

5e-Cleric:
> Why would there be an account system at all?

For the same reason windows offers different accounts, multiple people might want multiple accounts, i say we keep them for the local version.

G-Ambatte:
> Why would there be an account system at all?

If we were able to link to file system and thus have a users in their own folders, there might be _less_ need for an account system.

That said, I still expect that some users will want to lock their documents so that it can't be viewed accidentally - and thus if they are exposed in the filesystem, they will not be protected.
I'm not sure which option is the best way forward; at the moment, this is just a small experimental side branch.

dbolack-ab:
I'm gonna go out on a limb and suggest this move to its own project.

Sparks01:
Did this ever go anywhere?

Issue #3505, opened 2024-05-31
Provide a Linkable URL to ID Heading


mouse0270:
### Your idea:

So I am sorry for posting this as I am pretty sure this is a duplicate of issue #3272 however, the linked issues saying this is being resolved are closed #1923 and #3262

But I am trying to find a way to link to either a specific page or a specific heading within a shareable link. For example, using this Brew: [Fallout TTRPG by The Arcane Arcade](https://homebrewery.naturalcrit.com/share/5BLGepL_5gWo) I thought it would be useful to be able to provide links to users such as
- `https://homebrewery.naturalcrit.com/share/5BLGepL_5gWo#p82`
- `https://homebrewery.naturalcrit.com/share/5BLGepL_5gWo#WEAPON_MOD_PROP_POWERFUL`

so that users can click these links from outside sources and go to that specific page or that specific ID. However, even though linking to `#WEAPON_MOD_PROP_POWERFUL` appears to work while the document is loaded, it does not work when the user clicks a url and comes to the brew. Instead, it just keeps the user on the first page.

My assumption is because when the page loads, the pages have not been loaded yet, so the browser does not know where to direct the user to as that heading does not exist.

My request is, would it be possible to link players to an ID from the URL after the Global Headings have been registered. So, after a page loads and all headings are registered you check the URL and if there is a `#ID` at the end of the URL, you jump to that Heading if it exists?

5e-Cleric:
actually, this is a duplicate of #2820 which is not possible at this time, we'd need to create a function to check the url for the specific id handler and then scroll the iframe that contains the document to the desired heading... Will take a look at this, but i can't promise anything at this time.

5e-Cleric:
Just pushed a PR that should enable this. Closing as duplicate, all further discussion will be held at #2820

PR #3506, opened 2024-05-31
Get external link and scroll to element

### LINKED ISSUE #2820: External Direct link
I understand you can link to specific id attributes by doing the following:

`<a href="http://www.external-website.com/page#some-id">Link</a>`

but using that with a homebrewery share link just doesn't work, i get that is because the document is inside an iframe or because the page is not rendered before following the link, but is there a way around this?

5e-Cleric:
This does work, not sure it is perfect.

Instead of having to catch the exact time of load finish of the pages, the function waits 100ms before scrolling.
for this instance, i think this is appropiate, and easier to deal with.

Tested this with 90+ page doc, works fine. Unsetting smooth scroll as it triggers lots of page loads.

Closes #2820 and duplicate #3505

to test, try any brew url + `#id`


5e-Cleric:
Now also works with page ids

dbolack-ab:
How should this be tested?

5e-Cleric:
I'm guessing you will have a lot to say about the scrolling methods, cannot remember why i did it this way.

dbolack-ab (approved):
Works like a charm.  This is a brilliant add.

G-Ambatte:
With the removal of PPR and globally unique IDs, could we skip scrolling to page and go directly to the element instead?

dbolack-ab:
> With the removal of PPR and globally unique IDs, could we skip scrolling to page and go directly to the element instead?

We could, but I think scrolling to the page is the better idea and is consistent with the various link jumps we do

G-Ambatte:
> the various link jumps we do

...which is also behaviour that can now be reconsidered in the light of full document rendering and globally unique IDs.

That's not to say that the current behaviour is not necessarily desirous, but rather that until now, we haven't actually had the option to consider anything else.
It might be out of scope for this PR, but it's probably worth considering at some stage.

dbolack-ab:
> Do you mean rendering the iframe directly at that page?
> 
.....
> This PR is supposed to handle only url navigation into the page so that would make sense, specially with smooth scrolling disabled to prevent unnecessary renders. If you can point in the right direction i can try.

I firmly believe the current way is the right way. Stay page oriented.

5e-Cleric:
This could probably use the newest scrolling functions

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -76,9 +75,80 @@ const BrewRenderer = (props)=>{
...
 
-	useEffect(()=>{ // Unmounting steps
-		return ()=>{window.removeEventListener('resize', updateSize);};
+	useEffect(() => {
+		const locationHash = window.location.hash;
+		const iframe = document.getElementById('BrewRenderer');
+		
+		const pageIdRegex = /^#p\d+$/;
+		
+		const handleIframeLoad = () => {
```
calculuschild, line 84: We already have a function that fire when the iFrame loads: `frameDidMount`

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -76,9 +75,80 @@ const BrewRenderer = (props)=>{
...
+				if (pageIdRegex.test(locationHash)) {
+					const pageNumber = parseInt(locationHash.slice(2));
+					scrollToPage(pageNumber - 1);
+				} else {
+					const elementId = locationHash.slice(1);
+					if (elementId) {
+						getPageContainingElement(elementId)
+							.then((pageNumber) => {
+								if (pageNumber !== -1) {
+									scrollToPage(pageNumber);
```
calculuschild, line 95: Do we not want to just scroll to the element directly instead of the containing page? Would simplify logic.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -76,9 +75,80 @@ const BrewRenderer = (props)=>{
...
+                if (pageNumber > pages.length) {
+                    console.log('page not found');
+                } else {
+                    pages[pageNumber].scrollIntoView({ block: 'start' });
+                }
+            }
+        }
+    };
+	
+	const getPageContainingElement = (elementId) => {
```
calculuschild, line 135: If we really do want to get the parent page of the element, I believe we can just use:

`document.getElementById("theId").closest(".page")`

`.closest` grabs the nearest ancestor that matches a certain query.

calculuschild:
Blocked by #3484 

There is some rearranging of the scrolling logic there that will interfere with this. When I get that one merged it should be more clear what changes might be needed for this PR.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: Myself i prefer scrolling to the container page, much more clean experience. But i am not 100% settled on that

dbolack-ab: I have found scrolling to the page is REALLY annoying on big documents.

5e-Cleric: my experience is different, but this change is not worth fighting over, will do this change

calculuschild:
Studying this one a bit, `isMounted` is set to true the same time the initial page rendering is started, which means the iFrame is mounted but the pages inside might not actually be done rendering. I wonder if we can add a new state for "firstRenderComplete" or something, that gets set when `renderPages()` completes for the first time, and use that as the check to wait for scrolling.

5e-Cleric:
>  it seems rendering the pages is also waiting for isMounted to be true

Aaah that explains it then...

I'll try, not sure

ericscheid:
Maybe set up a `MutationObserver`?

```js
    const hash = window.location.hash;
    if (hash && iframeDoc) {
      const anchor = iframeDoc.querySelector(hash);

      if (anchor) {
        anchor.scrollIntoView();
      } else {
        // Use MutationObserver to wait for the element if it's not immediately available
        const observer = new MutationObserver((mutations, obs) => {
          const anchorElement = iframeDoc.querySelector(hash);
          if (anchorElement) {
            anchorElement.scrollIntoView();
            obs.disconnect(); // Stop observing once the element is found
          }
        });

        observer.observe(iframeDoc, {
          childList: true,
          subtree: true,
        });
```

5e-Cleric:
> Maybe set up a `MutationObserver`?
> 
> ```js
>     const hash = window.location.hash;
>     if (hash && iframeDoc) {
>       const anchor = iframeDoc.querySelector(hash);
> 
>       if (anchor) {
>         anchor.scrollIntoView();
>       } else {
>         // Use MutationObserver to wait for the element if it's not immediately available
>         const observer = new MutationObserver((mutations, obs) => {
>           const anchorElement = iframeDoc.querySelector(hash);
>           if (anchorElement) {
>             anchorElement.scrollIntoView();
>             obs.disconnect(); // Stop observing once the element is found
>           }
>         });
> 
>         observer.observe(iframeDoc, {
>           childList: true,
>           subtree: true,
>         });
> ```

Won't this trigger for every rerender of the preview?

ericscheid:
> Won't this trigger for every rerender of the preview?

ahem .. 
> `            obs.disconnect(); // Stop observing once the element is found`

So .. on page load it looks for the element corresponding to `window.location.hash`. If that target does not yet exist, then the MutationObserver is created. When it gets triggered it scrolls to it, and then deletes the observer. Subsequent rerenders are not observed, and so no further scrolling occurs.

5e-Cleric:
This should be ready for merge

dbolack-ab (changes requested):
When I select an out of bounds page, this sends me to Vault instead of the Brew.

I think this should instead go to the first or last page and provide an error message.

dbolack-ab (approved):
Previously reported error was in the testing, not the results.

calculuschild:
Ok, just cleaned up a bit. Merging now!

Thanks @5e-Cleric ! 💯 ⭐ 🎉

PR #3507, opened 2024-06-02
Add hooks to save and partially clean the target href tag from dompurify

dbolack-ab:
This adds a pre and post hook to preserve `target` attributes on `A` tags. 

Additionally, this adds `rel="noopener"` if the `target` is set to `_blank`.

Based on recommended solutions in the DOMPurify git issues..


5e-Cleric (approved):
We broke this, if it works i say we go with it.

G-Ambatte:
This PR will be obsolete if PR #3548 is merged.

dbolack-ab:
No longer needed

Issue #2688, opened 2023-02-23
`.wide+*` overriden by `.descriptive+*`


ericscheid:
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

Wide descriptive blocks produce weird margins for following sibling elements.

There are two CSS rules that are in conflict

```
.page .wide + * { margin-top: 0; }
.page .descriptive + * { margin-top: 0.55cm; }
```

Since the `.descriptive` rule has the same specificity level as the `.wide` rule, and so it is the last occuring declaration that takes precedence. 

## Solution:

1. swap the two rules' position in the stylesheet so the `margin-top:0` takes precedence, or
2. write a new rule for `.page .descriptive.wide { margin-top: 0; }`

 

### Code

```gfm
# Some Title
{{descriptive,wide
Here is a cute sub-heading
}}

## a heading in the left column

You will see it appears lower than the right column heading. This is because this heading has a top-margin.

\column

## a heading in the right column

You will see it appears higher than the left column heading.
```

PR #3517, opened 2024-06-05
Solve for .descriptive and .wide not getting along

dbolack-ab:
Uses suggested fix from Issue

Uses more more direct target rather than rearrange the file.

Solves #2688


5e-Cleric (approved):

G-Ambatte (changes requested):
This does NOT fix the reported issue; checking the reproduction text in the deployment shows that the issue still exists.
I believe the issue reporter made a typo when suggesting the correction, instead of `.page .descriptive.wide {`, they should have entered `.page .descriptive.wide + * {`.

Current behaviour in Production: 
image

Current behaviour in Test Deployment:
image

Behaviour with suggested change:
image

dbolack-ab:
> This does NOT fix the reported issue; checking the reproduction text in the deployment shows that the issue still exists. I believe the issue reporter made a typo when suggesting the correction, instead of `.page .descriptive.wide {`, they should have entered `.page .descriptive.wide + * {`.
> 
> Current behaviour in Production: image
> 
> Current behaviour in Test Deployment: image
> 
> Behaviour with suggested change: image

Fixed. Note sure why it looked right before but it definitely did not on reload today.

calculuschild:
Looks good @dbolack-ab ! Merging!

PR #3522, opened 2024-06-06
Tweak CM page line brightness on dark themes

G-Ambatte:
As per Reddit report (https://redd.it/1d8opte), on dark CodeMirror themes, it can be difficult to read the folded text between the bright backgrounds of `\page` lines. This PR tweaks the brightness down slightly by reducing opacity of the background from 75% to 50%.


G-Ambatte:
Visual preview of the change:

image

5e-Cleric:
Talk about a minor change...

when you're showing the change, it'd help more to see a comparison from old to new.

Also, this is an unnecessary change that depends too much on personal taste.

dbolack-ab:
> Also, this is an unnecessary change that depends too much on personal taste.

If so, should we then fork this particular theme for both tastes or is that too far?

5e-Cleric:
> > Also, this is an unnecessary change that depends too much on personal taste.
> 
> If so, should we then fork this particular theme for both tastes or is that too far?

i don't think anyone would notice the change.

Gazook89:
> fork the theme

I wonder if it's best to start making modifications to individual themes.  Not necessarily solely for the reason raised in this PR, but it might be something we have more and more reasons for over a long enough time.  And getting used to that *now*, rather than realizing we missed the opportunity far in the future, might be good.

We will have more and more custom syntax, widgets (*some day*), and other editor enhancements that could all benefit from thematic style customizations.  For example, right now we are only modifying the codemirror instance-- but I could see a future where the Snippet Bar and Menus are matched to the codemirror theme.  Are we offering the opportunity to customize the "Editor" (which to me includes the Editor Tools aka the snippets) or just the writing space?   I'd say we allow themes to change the whole Editor.

That's a whole other thing, though.  For *this* PR, I think the change is minor, it's specific to dark themes, doesn't make things worse, and easily reversible.  

Just for reference, here is the original reddit users screenshot of the existing styling:
dark theme

Gazook89 (approved):
Looks good to me

5e-Cleric (approved):

Gazook89 (approved):
try this again....

looks good to me!

dbolack-ab (approved):
Looks good.

dbolack-ab (approved):

5e-Cleric (approved):
looks good to me too

dbolack-ab (approved):

PR #3526, opened 2024-06-08
Fixing Monster Frame background in Safari for iphone and ipad

5e-Cleric:
Safari for mobile screens does not support background-blend-mode and background-repeat:repeat; in the same element, it deactivates the overlay, which makes our statblocks look grey:

image

A fix is set here, i manually overlayed the color with the image to make the background, and i'm changing it for the original.

This fix works, and other users will not notice the difference.


calculuschild:
> other users will not notice the difference.

The use of a grayscale image and blending was a deliberate choice for customizability of the monster stat block color. Changing this will break brews that have used this to adjust the color of the monster stat block.

dbolack-ab:
Is this a can't fix at Safari?

5e-Cleric:
> Is this a can't fix at Safari?
> 

https://bugs.webkit.org/show_bug.cgi?id=259081

I'm not sure webkit bugs are a thing after chromium was released, but this us the best i could find right away.

Gazook89:
iPhone 17.5.1

image

Gazook89:
Wait hang on

Gazook89:
Sorry I just opened the non deployment link without thinking, that’s the previous screenshot. 

Here is from the PR deployment, but without the green bit because it’s pretty difficult to open the css editor on my tiny screen (iPhone 12 mini):

image


Note, the top and bottom bars are cut off but a different issue

5e-Cleric:
The border image is not slicing as it should, but that doesn't surprise me, i did not provoke it, and it is out of the scope of this PR.

calculuschild:
Can you try opening this on deployment on iphone?

https://homebrewery-pr-3526.herokuapp.com/share/_h7ThoQlrJVc

I assume this will just turn into the yellow parchment color.

themes/assets/assets.less
```diff
@@ -7,6 +7,7 @@
 @noteBorderImage          : url('/assets/noteBorder.png');
 @descriptiveBoxImage      : url('/assets/descriptiveBorder.png');
 @monsterBlockBackground   : url('/assets/parchmentBackgroundGrayscale.jpg');
+@monsterBlockOverlay   : url('/assets/parchmentBackgroundOverlayed.jpg');
```
calculuschild, line 10: This graphic isn't used anymore, right?

Gazook89:
Thanks for the link, here it is:

image

calculuschild:
Ok, this seems to be working without messing up custom coloring. Thanks @5e-Cleric ! And thanks @Gazook89 for testing. Will merge now.

PR #3532, opened 2024-06-13
Add environment option for secure local website

### LINKED ISSUE #3900: Simplify using locally installed Homebrewery in HTTPS mode
See issue raised on Reddit: https://redd.it/1gv7fr1

Currently, on starting Homebrewery locally, the local URL will be generated. However on clicking this link, many default browsers will change the `http://` link to `https://`. It would greatly simplify their experience if we can allow Homebrewery to be used in HTTPS mode.

G-Ambatte:
This PR adds support for the NODE_ENV environment variable to be set to `local.secure`. This option, when combined with a `hostname` configuration item, creates a webserver that serves pages secured with certificates configured for the `hostname`.

At this point in time, this is completely experimental and only of limited use in local installations. It has NOT been thoroughly tested, so some Homebrewery features may not work correctly with this environment setting.

---

This PR resolves #3900.


dbolack-ab:
What is the purpose here?

calculuschild:
@G-Ambatte Is it ok if we close this one for now and bring it back to its own issue where the need for this can be discussed?

G-Ambatte:
> @G-Ambatte Is it ok if we close this one for now and bring it back to its own issue where the need for this can be discussed?

That's a good idea, I don't think the implementation is novel enough to warrant holding on to the PR.

dbolack-ab:
Closing per last exchange.

server.js
```diff
@@ -1,18 +1,34 @@
...
 const server = require('./server/app.js');
 const config = require('./server/config.js');
+const fs = require('fs');
+const http = require('http');
+const https = require('https');
+
+let webServer = http.createServer(server.app);
+
+const NODE_ENV = config.get('node_env');
+if(NODE_ENV === 'local.secure') {
```
G-Ambatte, line 11: If we change this line to a file existence check, rather than checking for a specific NODE_ENV type, then that should simplify this change.

G-Ambatte:
I think this PR was unintentionally re-opened. Closing again.

PR #3548, opened 2024-06-29
Improve HTML sanitization

### LINKED ISSUE #3547: Links in ToC broken with 3.13.0 release
## Links in ToC broken with 3.13.0 release

**Share Link** : https://homebrewery.naturalcrit.com/share/fFPgV7Rj1YC2

 Clicking a link in the generated ToC opens a new tab instead of jumping to the location.

Worked an hour or so ago, before the new client was pushed out.

Chrome is up to date
Version 126.0.6478.127 (Official Build) (64-bit)

### LINKED ISSUE #3549: DOMpurify breaks custom html elements
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

I had some quick headers was used inside a common line of text as global styles that now no longer work. For example, a tag like this

    A black text, <h7>a red text</h7> and a black text again
with a simple CSS like

    h7 {color: red}
It used to work fine, but now it seems that they are no longer recognized as headers and don't appear colored.

I don't know if there is a way to revert this change if this is a bug, or if there is a better way to define quick styles for texts inside a paragraph without requiring a lot of code.

### Code

_No response_

G-Ambatte:
This PR resolves #3547, and resolves #3549. It also allows us to close #3507 as that PR will be obsolete once this is merged.

~~This PR changes the DOMPurify configuration to allow the `target` attribute in the output. It also adds the `DOMPurify.addHook` to allow all custom elements in the output, while still stripping potentially harmful attributes.~~

This PR is based on the `vue-html-secure` package, using a similar method to parse the HTML elements prior to injection to the browser.
Currently, certain tags are completely forbidden (e.g. `<script>`), and some attributes (anything starting with `on`). Blacklisted tags are identified by simple string comparison, while attributes are passed through a series of test functions. Any tag that matches one in the `blacklistTags` array is removed, as is any attribute that causes a function in the `blacklistAttribs` array to return `true`.

Testing is achieved via the `jsdom` and `jsdom-global` packages, to allow implementation of HTML specifications outside of a browser.


5e-Cleric (approved):

G-Ambatte:
Working through the v3.13 checklist to make sure all the DOMPurify issues are resolved in this branch:

- [x] [Font imports broken (Reddit)](https://www.reddit.com/r/homebrewery/comments/1dr2qpz/v3130_breaks_importing_multiple_fonts/ ) - confirmed working, source copied from example in Reddit post produces this: 
image
- [x] [Custom HTML elements broken (Reddit)](https://www.reddit.com/r/homebrewery/comments/1dr21ri/v1130_broke_some_styles/) - Confirmed: 
image
- [ ] [ToC Links opening in separate tab (GitHub)](https://github.com/naturalcrit/homebrewery/issues/3547) - *Confirmed, DOMPurify is mangling "target" value of links.* Fixed in https://github.com/naturalcrit/homebrewery/pull/3548

calculuschild:
#3555 was linking to this as the resolved issue. Reopening.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -126,9 +125,9 @@ const BrewRenderer = (props)=>{
 
 	const renderStyle = ()=>{
 		if(!props.style) return;
-		const cleanStyle = props.style; //DOMPurify.sanitize(props.style, purifyConfig);
+		const cleanStyle = safeHTML(`<style>${props.style}</style>`);
```
dbolack-ab, line 128: This region of code will need a small rework for User Brews.

G-Ambatte: I believe my thinking here was the `<style>` is a safe tag, so there's no reason to exclude it from the sanitization process. That said, I have no objection to returning to the original configuration of sanitizing `props.style` and then wrapping the string in `<style>` tags at injection time - would that be less of a conflict with the User Brews PR?

dbolack-ab: It's not so much that as it is the User brews rearranged some things. This function is removed.

Whichever applied first, some tweaks will be needed. :)

G-Ambatte: I've just pushed a merge conflict resolution that should solve this now.

G-Ambatte: Oh hey it would probably help if I actually parsed the text through the function that is the whole point of this PR, too...

dbolack-ab:
Is anything blocking this moving forward?

5e-Cleric:
> Is anything blocking this moving forward?

Issues that broke last time seem to be fixed, @calculuschild this should move up the queue

client/homebrew/brewRenderer/safeHTML.js
```diff
@@ -0,0 +1,47 @@
...
+	const blacklistAttrs = [
+		(test)=>{return test.localName.indexOf('on') == 0;},
+		(test)=>{return test.value.replace(/[\u0000-\u0020\u00A0\u1680\u180E\u2000-\u2029\u205f\u3000]/g, '').toLowerCase().trim().indexOf('javascript:') == 0;}
+	];
+
+
+	elements.forEach((element)=>{
+		// Check each element for blacklisted type
+		if(blacklistTags.includes(element?.localName?.toLowerCase())) {
+			element.parentNode.removeChild(element);
```
Gazook89, line 28: Why not just `element.remove()`, rather than going up to the parent, and back down to the child?

client/homebrew/brewRenderer/safeHTML.js
```diff
@@ -0,0 +1,47 @@
...
+		if(element.hasAttributes()){
+			for (const attribute of element.attributes){
+				for (const test of blacklistAttrs) {
+					if(test(attribute)){
+						element.removeAttribute(attribute.localName);
+						break;
+					};
+				};
+			};
+		};
```
Gazook89, line 41: I'm not super bothered by it, but this stands out as a lot of nested if's and for's.

Can you use `[...element.attributes].forEach()` rather than the first `for(const attribute....`?

And then for the second `for` loop, use `if(blacklistAttrs.some(test=>test(attribute))){ remove.... }`

Gazook89 (approved):
I don't have much of substance to add with this review, it looks pretty good to me but I reckon I can't think of every test (particularly outside of the tests you already have written). I tried some stuff in the deployment and nothing seemed to work (in a good way).  

I wrote some line comments but neither one of them would hold this thing up.

client/homebrew/brewRenderer/safeHTML.js
G-Ambatte: > `[...element.attributes].forEach()`

I don't believe you can `forEach` works on a NodeList, but this would convert it to an array first, so it might work.

> `if(blacklistAttrs.some(test=>test(attribute))){ remove.... }`

That will probably work; I keep forgetting `some()` exists and finding harder ways to achieve the same result.

I'll experiment when I have some time; might be a few days though.

client/homebrew/brewRenderer/safeHTML.js
G-Ambatte: `remove` is a newer function that achieves the same result. Some "wisdom of the ancients" shows wider support for the `removeChild` process, which boils down to being supported by a couple of ancient versions of IE.
Ultimately, `remove` is about as widely supported as `removeAttribute` which is used later in the function, so I have no particular issue with changing to use `remove`.

5e-Cleric:
@G-Ambatte if you don't mind making a final pass, this already has 2 approving reviews

5e-Cleric:
I believe this was lost because it was placed in the "done" category in the backlog, fixed that.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -28,14 +27,16 @@ const INITIAL_CONTENT = dedent`
 	<base target=_blank>
 	</head><body style='overflow: hidden'><div></div></body></html>`;
 
+import { safeHTML } from './safeHTML.js';
```
calculuschild, line 30: Let's move the `import` up next to the other `require` lines.

calculuschild (changes requested):

tests/html/safeHTML.test.js
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable max-lines */
+
+require('jsdom-global')();
+
+import { safeHTML } from '../../client/homebrew/brewRenderer/safeHTML';
```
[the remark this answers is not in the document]
5e-Cleric: We could handle css in another following PR, keep this small,  i must say, there are css vulnerabilities, but are pretty specific and most are caught by the UA itself today.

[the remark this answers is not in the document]
calculuschild: That's fair.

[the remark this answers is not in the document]
G-Ambatte: There are no CSS specific tests. However, there is also nothing to stop a user breaking out of the `<style>` blocks in the Style Editor and inserting HTML, which is why the style text is also parsed by the `safeHTML` function.

client/homebrew/brewRenderer/brewRenderer.jsx
G-Ambatte: This is relocated in the current commit.

package.json
```diff
@@ -101,6 +102,8 @@
     "express-static-gzip": "2.1.7",
     "fs-extra": "11.2.0",
     "js-yaml": "^4.1.0",
+    "jsdom": "^24.1.0",
+    "jsdom-global": "^3.0.2",
```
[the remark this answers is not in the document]
G-Ambatte: These should now be shifted to Dev dependencies.

package.json
```diff
@@ -134,6 +135,8 @@
     "globals": "^15.11.0",
     "jest": "^29.7.0",
     "jest-expect-message": "^1.1.3",
+    "jsdom": "^24.1.0",
```
calculuschild, line 138: Is jsdom still needed? I don't see it imported anywhere.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -15,8 +15,7 @@ const Frame = require('react-frame-component').default;
 const dedent = require('dedent-tabs').default;
 const { printCurrentBrew } = require('../../../shared/helpers.js');
 
-const DOMPurify = require('dompurify');
-const purifyConfig = { FORCE_BODY: true, SANITIZE_DOM: false };
+const Themes = require('themes/themes.json');
```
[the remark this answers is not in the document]
G-Ambatte: I'm not actually sure how the Themes line got in there. Removed.

package.json
G-Ambatte: Doesn't look like it, I've removed it now.

G-Ambatte:
I believe everything is addressed now.

calculuschild:
Thanks @G-Ambatte !

PR #3552, opened 2024-07-01
Allow `=` in attributes

calculuschild:
Attribute parsing in curly injectors was truncating urls with `=` (actually, any attribute). This patch fixes the truncation, and also allows `?` and `=` as valid contents of an attribute inside "quotes". Same tweak has been applied to curly Divs and curly Spans.

Adds 2 tests to confirm urls are no longer truncated, and attributes can be added with `=` and `?` in the contents.


5e-Cleric (approved):
Can confirm image links are correctly output, looks ok to me

G-Ambatte (approved):

PR #3556, opened 2024-07-03
Export Brew as a CBZ file

dbolack-ab:
This PR exports the Brew as a CBZ archive readable by various "comic" document readers.

The export isn't quite perfect, there appear to be some issues with some borders ( order of operations, maybe? ) within the library being used. Given this is a minor issue with a subset of divs, I'm pushing this now for evaluation while I look for alternatives and/or fix the library.

To prevent menubar clutter, this PR moves the PDF printing item to a new Export menu and adds the CBZ export below it.


5e-Cleric:
I'm not sure any user would notice, but i like the idea of exporting to other file formats. We likely want to mark this as `export to comic file format(CBZ)`.

Also, if we are exporting to .CBZ, might as well export as .ZIP if users want too, right?

server/middleware/content-negotiation.js
```diff
@@ -1,12 +1,13 @@
 module.exports = (req, res, next)=>{
-	const isImageRequest = req.get('Accept')?.split(',')
-        ?.filter((h)=>!h.includes('q='))
-        ?.every((h)=>/image\/.*/.test(h));
-	if(isImageRequest) {
-		return res.status(406).send({
-			message : 'Request for image at this URL is not supported'
-		});
+	if(! req.url.startsWith('/xssp/')) {
```
G-Ambatte, line 2: This line is causing test failures, as it doesn't handle the null case for `req.url`.
It will probably work just fine if it's changed to `req.url?.startsWith(...)`.

G-Ambatte (commented):
While I don't have a use-case for this feature, I am not specifically against it's inclusion. 
At this point, I have only investigated the tests failures; I would suggest it should probably have it's own test suite prior to final merging.

dbolack-ab:
> Also, if we are exporting to .CBZ, might as well export as .ZIP if users want too, right?

Given a CBZ is just a zip file of images, it's just an extension.

ericscheid:
> Also, if we are exporting to .CBZ, might as well export as .ZIP if users want too, right?

A `.zip` of what, exactly? What is the underlying document format that is being zipped (i.e. other than CBZ) ?

dbolack-ab:
> A `.zip` of what, exactly? What is the underlying document format that is being zipped (i.e. other than CBZ) ?

A set of JPGs/PNGs. So far as I have been able to discern, there is no true format for CBR/CBZ files. It simply walks through the archive in alphabetical order. If there are folders, they are walked as well.

server/middleware/content-negotiation.js
dbolack-ab: I remember that fix not working for an inexplicable reason, but it is now.

5e-Cleric:
> > Also, if we are exporting to .CBZ, might as well export as .ZIP if users want too, right?
> 
> A `.zip` of what, exactly? What is the underlying document format that is being zipped (i.e. other than CBZ) ?

Well, let's cite wikipedia on this one:

>Comic book archive is not a distinct file format. It is a [filename extension](https://en.wikipedia.org/wiki/Filename_extension) naming convention.
>
> The filename extension indicates the archive type used:
>
> - .cb7 → [7z](https://en.wikipedia.org/wiki/7z)
> - .cba → [ACE](https://en.wikipedia.org/wiki/ACE_(compression_file_format))
> - .cbr → [RAR](https://en.wikipedia.org/wiki/RAR_(file_format))(https://en.wikipedia.org/wiki/Comic_book_archive#cite_note-2)
> - .cbt → [TAR](https://en.wikipedia.org/wiki/Tar_(file_format))
> - .cbz → [ZIP](https://en.wikipedia.org/wiki/ZIP_(file_format))
>
>Comic book archive files mainly consist of a series of [image files](https://en.wikipedia.org/wiki/Image_file) with specific naming, typically [PNG](https://en.wikipedia.org/wiki/Portable_Network_Graphics) or [JPEG](https://en.wikipedia.org/wiki/JPEG) files, stored as a single [archive file](https://en.wikipedia.org/wiki/Archive_file). Occasionally [GIF](https://en.wikipedia.org/wiki/GIF), [BMP](https://en.wikipedia.org/wiki/Windows_and_OS/2_bitmap), and [TIFF](https://en.wikipedia.org/wiki/Tag_Image_File_Format) files are seen. Folders may be used to group images in a more logical layout within the archive, like book chapters.

So it is a compressed file of images of every page, to make a comic book.
They are used by software like Calibre or CDisplay

ericscheid:
Yes, a `.cbz` is a file which is container of components, bundled into a zip. The components of the Comic Book Archive  are a bunch of basic files (mostly image files), organised according to particular fules. It's a format, or at least a meta-format.

What I was getting at with my question is are there other end-format constructs that we might want to possibly include as export formats? Things like, oh, ebook, or maybe even a SPA of html/image components for static hosting of brews elsewhere.

5e-Cleric:
I'd be against the website approach, we could export to html, but i would not go pass that.

But an `.epub` export sounds like it would be a great idea, specially as epub files are html files under the hood, should be pretty simple to make.

dbolack-ab:
> I'd be against the website approach, we could export to html, but i would not go pass that.
> 
> But an `.epub` export sounds like it would be a great idea, specially as epub files are html files under the hood, should be pretty simple to make. (more like a zip of html, css and xml files, but you get what i mean) There must be some library out there.
> 
> https://www.npmjs.com/package/@lesjoursfr/html-to-epub

I'll look at this in another PR in the future. We're off-track.

5e-Cleric:
So, don't we need https://github.com/naturalcrit/homebrewery/issues/2278 to make this feature? i thought we were stuck with that, or does https://github.com/naturalcrit/homebrewery/pull/3501 work?

dbolack-ab:
> So, don't we need #2278 to make this feature? i thought we were stuck with that, or does #3501 work?

#3501 works. Two bugs with it were fixed with this PR and backported into it. 

#2278 is functionally implemented with this PR.

5e-Cleric:
> > So, don't we need #2278 to make this feature? i thought we were stuck with that, or does #3501 work?
> 
> #3501 works. Two bugs with it were fixed with this PR and backported into it.
> 
> #2278 is functionally implemented with this PR.

Before actually linking this PR into the corresponding issues, a recommended reading for you:
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

This closes #2278

dbolack-ab:
Weird big regression. Kicking back to draft until I find it

dbolack-ab:
I'm going to withdraw this PR for now. I mostly did it to prove it could be done, but it seems like a feature that wouldn't be used so there's no point in fighting the regression.

I will reopen if it turns our I read the room wrong.

PR #3560, opened 2024-07-06
Limit htmlString to the first element ONLY

### LINKED ISSUE #3559: IDs and Styles leak from internal elements when Mustache Style Injectors are used
### Your idea:

When using an inline or block style injector, attributes like style or ID from internal elements leak to their parents.

Reproduction:
```
{{test
# H1
}}
{}
```

Even though there is nothing to inject into the wrapping block, we receive the following HTML output:
```
<div class="block test" id="h1">
<h1 id="h1">h1</h1>
</div>
```

This is most noticeable when using the image masks; some users have been adding styling like so:
```
{{imageMaskEdge2,--offset:0%,--rotation:180
  ![](https://i.imgur.com/GZfjDWV.png){height:100%}
  {--test:true}
}}
```

which is resulting in:
```
<div class="block imageMaskEdge2" style="--offset:0%; --rotation:180;">
<p style="height:100%; --test:true;" src="https://i.imgur.com/GZfjDWV.png" alt="">
  <img style="height:100%;" src="https://i.imgur.com/GZfjDWV.png" alt="">
</p>
</div>
```

The `--test:true` is correctly applied to the `<p>`, but the `<height:100%>` has been incorrectly applied from the internal `` styling.

G-Ambatte:
This PR resolves #3559.

This PR limits the `htmlString` to be processed by the `extractHTMLStyleTags` function in `markdown.js` to only the content of the first element - that is, from the beginning of the string to the location of the first `>`. This prevents any attributes from elements deeper in the DOM tree from incorrectly being applied to the parent element.

Reproduction brew: https://homebrewery.naturalcrit.com/share/fodf7wAEOrbC

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -787,8 +787,9 @@ const processStyleTags = (string)=>{
 	};
 };
 
+//Given a string representing an HTML element, extract all of its properties (id, class, style, and other attributes)
 const extractHTMLStyleTags = (htmlString)=>{
-	const firstElementOnly = htmlString.indexOf('>') > 0 ? htmlString.substring(0, htmlString.indexOf('>')) : htmlString;
+	const firstElementOnly = htmlString.split('>')[0];
 	const id         = firstElementOnly.match(/id="([^"]*)"/)?.[1]    || null;
 	const classes    = firstElementOnly.match(/class="([^"]*)"/)?.[1] || null;
 	const styles     = firstElementOnly.match(/style="([^"]*)"/)?.[1] || null;
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -788,10 +788,11 @@ const processStyleTags = (string)=>{
 };
 
 const extractHTMLStyleTags = (htmlString)=>{
-	const id         = htmlString.match(/id="([^"]*)"/)?.[1]    || null;
-	const classes    = htmlString.match(/class="([^"]*)"/)?.[1] || null;
-	const styles     = htmlString.match(/style="([^"]*)"/)?.[1] || null;
-	const attributes = htmlString.match(/[a-zA-Z]+="[^"]*"/g)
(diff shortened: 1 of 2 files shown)


G-Ambatte:
Reproduction code, as per Issue #3559 :
```
{{test
# H1
}}
{}

{{imageMaskEdge2,--offset:0%,--rotation:180
  ![](https://i.imgur.com/GZfjDWV.png){height:100%}
  {--test:true}
}}
```

Output:
```
<div class="block test">
<h1 id="h1">H1</h1>
</div>
<div class="block imageMaskEdge2" style="--offset:0%; --rotation:180;">
<p style="--test:true;">
  <img style="height:100%;" src="https://i.imgur.com/GZfjDWV.png" alt="">
</p>
</div>
```

G-Ambatte:
This more closely matches the output of v3.12.0, where the reproduction text produces the following:
```
<div class="">
<h1 id="h1">H1</h1>
</div>
<div class="block imageMaskEdge2" style="--offset:0%; --rotation:180;">
<p class="" style="--test:true;">
  <img class="" style="height:100%;" src="https://i.imgur.com/GZfjDWV.png" alt="">
</p>
</div>
```

G-Ambatte:
There may be a way to fix this instead by limiting the input to the `this.parser.parse([token])` step.

calculuschild:
> There may be a way to fix this instead by limiting the input to the `this.parser.parse([token])` step.

I think the way you did it is best. Some tokens require knowledge of their child tokens to parse properly, so stripping away the children for this step might complicate things more than it helps, and then we would have to re-parse with the children anyway.

G-Ambatte:
> Some tokens require knowledge of their child tokens to parse properly, so stripping away the children for this step might complicate things more than it helps

I actually had the exact same thought about an hour before I saw this.

calculuschild:
Added a test case and simplified the string splitting code. Looks good!

calculuschild (approved):

PR #3564, opened 2024-07-08
Stop Ink Friendly snippet making text unselectable in output PDF

### LINKED ISSUE #3563: Ink Friendly snippet causes PDF text to be unselectable
### Your idea:

The Ink Friendly snippet currently uses `filter : drop-shadow(0px 0px 3px #888) !important;`, but when exported to PDF, this makes the text in the output unable to be selected and copied as text; instead it can only be manipulated as an image.

Replacing the `filter` line with `box-shadow: 0px 0px 3px #888 !important;` should achieve the same visual result while still eliminating this undesirable behaviour.

G-Ambatte:
This PR resolves #3563.

This PR changes the Ink Friendly snippet to use `box-shadow` instead of `filter`. This should achieve the same visual effect while not rendering the output PDF as an image.

### DIFF
--- a/themes/V3/5ePHB/snippets.js
+++ b/themes/V3/5ePHB/snippets.js
@@ -349,7 +349,7 @@ module.exports = [
 					/* Ink Friendly */
 					*:is(.page,.monster,.note,.descriptive) {
 						background : white !important;
-						filter : drop-shadow(0px 0px 3px #888) !important;
+						box-shadow : 0px 0px 3px #888 !important;
 					}
 
 					.page img {
--- a/themes/V3/5ePHB/snippets.js
+++ b/themes/V3/5ePHB/snippets.js
@@ -349,7 +349,7 @@ module.exports = [
 					/* Ink Friendly */
 					*:is(.page,.monster,.note,.descriptive) {
 						background : white !important;
-						box-shadow : 0px 0px 3px #888 !important;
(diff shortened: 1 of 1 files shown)


dbolack-ab:
Is there a Brew we can compare before and after results with?

G-Ambatte:
> Is there a Brew we can compare before and after results with?

Example 1: Current
HB link: https://homebrewery.naturalcrit.com/share/wjdLwCIMiGf7
PDF: [Ink Friendly Example 1_ Current - The Homebrewery.pdf](https://github.com/user-attachments/files/16258822/Ink.Friendly.Example.1_.Current.-.The.Homebrewery.pdf)

Example 2: PR Changes
HB link: https://homebrewery.naturalcrit.com/share/oU5XmiPzlhMg
PDF: [Ink Friendly Example 2_ Proposed - The Homebrewery.pdf](https://github.com/user-attachments/files/16258830/Ink.Friendly.Example.2_.Proposed.-.The.Homebrewery.pdf)

You should see that the "current" brew contains only a single framed monster block and the Ink Friendly snippet. However, the PDF output is that of an image; the text is not able to be selected or interacted with.
However, you should also see that the PR changes contains the exact same framed monster block, and the only difference is in the style, where the Ink Friendly snippet has been replaced with the version from this PR. The resultant PDF has text that can be highlighted and interacted with as expected for text.

PR #3566, opened 2024-07-09
Vitreum-less build scripts

### LINKED ISSUE #3480: Migrate from Vitreum to Webpack AND/OR other services
[Vitreum](https://github.com/stolksdorf/vitreum) Has not been maintained for the last 4 years. Since then we have been running using [a fork by @calculuschild](https://github.com/calculuschild/vitreum), but this might prove like an inefficient idea in the long run.

Conversation has started talking about moving from Vitreum to Webpack or another bundling system, this issue is here to log that project's ideas and conversation.

It uses common tools: React, Browserify, and LiveReload.  so we might need to install other  tools as well.

G-Ambatte:
This PR seeks to resolve #3480.

The initial step has been to extract all the useful parts of Vitreum to the main Homebrewery project, so those build tools are not obfuscated in an additional project.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,7 @@
         "@babel/preset-react": "^7.26.3",
         "@googleapis/drive": "^8.14.0",
         "body-parser": "^1.20.2",
+        "browserify": "^16.5.0",
         "classnames": "^2.5.1",
         "codemirror": "^5.65.6",
         "cookie-parser": "^1.4.7",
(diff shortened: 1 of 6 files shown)


5e-Cleric:
I believe this is a change important enough that we should roll it on its own, to make sure we know where any error comes from.

calculuschild:
Interesting. Do you think this can turn into using webpack or something instead of Browserify in a future PR?

5e-Cleric:
Right, cannot review this if it has conflicts and deployment doesn't work

G-Ambatte:
I've taken a stab at bringing this up to current this morning, there are multiple issues with the change from `require` to `import`.

5e-Cleric:
Happy to report htis can be closed!

PR #3569, opened 2024-07-11
Unset text-decoration on Brew Item links

### LINKED ISSUE #3568: BrewItem links are underlined
### Your idea:

The items in the BrewItem card that have a `href` attribute are displaying an underline. This is caused by the following styling:

```
a:-webkit-any-link {
  text-decoration: underline;
}
```

G-Ambatte:
The styling has changed, adding a `text-decoration: underline;` to `a` with links (that is, that have a `href` set).

This PR resolves #3568.

This PR adds styling to `.BrewItem .links a`, adding `text-decoration: unset;` which prevents the errant underline from appearing.

### DIFF
--- a/client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
+++ b/client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
@@ -119,11 +119,12 @@
 		text-align       : center;
 		a{
 			.animate(opacity);
-			display   : block;
-			margin    : 8px 0px;
-			opacity   : 0.6;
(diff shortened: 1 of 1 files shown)


calculuschild:
This looks innocent enough. Where did the change come from? Is this from a Browser change?

G-Ambatte:
> This looks innocent enough. Where did the change come from? Is this from a Browser change?

I believe this is a browser change, I haven't tracked down the exact source of it, the styling is targeted at a much higher level than we would ever use.

PR #3580, opened 2024-07-17
Fix Table of Contents uppercasing

### LINKED ISSUE #3572: Table of contents UPPERCASE titles un-uppercaseable!
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

When generating the table of contents (Text editor -> Table of contents (beta) ->  Table of contents (beta)) 
H3 titles are "uppercased", while before they where respecting the formatting of the actual title in the editor and it's capitalization.

Maybe this is something that have to do with making things 5e like, but I selected Pathfinder in the brew information, I'm using custom CSS to make things similar to pf2e but in this case is not a css issue, the renderer automatically uppercase the h3 in my brew.

TOC BEFORE:
image

TOC AFTER:
image

text is UPPERCASE in html, not via css, cannot easily override...
image

### Code

_No response_

G-Ambatte:
This PR resolves #3572.

This PR changes the Table of Contents generation snippet to use the `textContent` property of the HTML element, rather than the `innerText` property, as `innerText` has been found to incorrectly include styling from `text-transform`, which is undesirable.

### DIFF
--- a/themes/V3/5ePHB/snippets/tableOfContents.gen.js
+++ b/themes/V3/5ePHB/snippets/tableOfContents.gen.js
@@ -35,7 +35,7 @@ const getTOC = (pages)=>{
 		const ToCExclude = getComputedStyle(heading).getPropertyValue('--TOC');
 
 		if(ToCExclude != 'exclude') {
-			recursiveAdd(heading.innerText.trim(), onPage, headerDepth.indexOf(heading.tagName), res);
+			recursiveAdd(heading.textContent.trim(), onPage, headerDepth.indexOf(heading.tagName), res);
 		}
 	});
 	return res;


G-Ambatte:
As this bug only affects headings that have been styled with the `text-transform` property, I've flagged this as a `P2: Minor Bug`.

G-Ambatte:
From the example reproduction as detailed in #3572:

image

As seen in the above image, the output ToC now has the lowercase `h3`.

dbolack-ab:
What is the difference between innerText and textContext?

dbolack-ab (approved):
Seems sensible.

dbolack-ab:
> What is the difference between innerText and textContext?

To answer my own question, innerText observes the text being hidden by styling where textContext does not. This seems like an unimportant distinction in our use case.

G-Ambatte:
Correct, innerText also reflects text-transform styling, which appears yo be the root cause of the original issue.

dbolack-ab:
> Correct, innerText also reflects text-transform styling, which appears to be the root cause of the original issue.

Clearly, I wasn't parsing correctly because . Right. :)

G-Ambatte:
I also knocked out this PR _very_ quickly, so I may not have been as clear or as careful with my wording as I would have been otherwise, in the interest of getting it done before my lunch break finished.

5e-Cleric:
Flagging this as P2 may be too exaggerate, i think it fits perfectly into P3, obscure tweak or fix for a single user. But no point in changing that now, flagging as solution found.

5e-Cleric (approved):

calculuschild:
Thanks @G-Ambatte ! Loving these small PRs.

PR #3584, opened 2024-07-19
Disable linewrapping in the style editor

dbolack-ab:
This will make using font and image imports manageable.

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -354,6 +354,7 @@ const Editor = createClass({
 					language='gfm'
 					view={this.state.view}
 					value={this.props.brew.text}
+					wrap={true}
 					onChange={this.props.onTextChange}
 					editorTheme={this.state.editorTheme}
 					rerenderParent={this.rerenderParent} />
@@ -367,7 +368,8 @@ const Editor = createClass({
 					view={this.state.view}
 					value={this.props.brew.style ?? DEFAULT_STYLE_TEXT}
 					onChange={this.props.onStyleChange}
-					enableFolding={false}
+					enableFolding={true}
+					wrap={false}
 					editorTheme={this.state.editorTheme}
 					rerenderParent={this.rerenderParent} />
 			</>;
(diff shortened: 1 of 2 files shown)


calculuschild:
Is there an existing issue that this should be linked to? I'm not sure I fully understand what this is trying to solve. 

The main thing I notice is that removing the wrapping requires the user to scroll around to access the full text, making it harder to deal with long imports. Do you perhaps have an example brew or screenshots comparing before and after that would clarify what I should be testing?

Gazook89:
This is spawned by a complaint I had around long text strings being visually bothersome.  In particular, base645 encoded strings for fonts or images.  People wanting to use a different font can basically either import from Github via Githack, or base64 encode the font directly in the editor.  The former requires several not-particularly-convenient steps and even requires a Github account, and the latter results in a super long nonsense string that can dominate the editor.  

I was hoping for some way to truncate long strings, particularly for these encoded strings and maybe even long URL's (for those images that don't use imgur).  

My initial thought was to basically set a max character count inside parentheses, like:

```
@import url('fkjiwjfkneidnelsoiicndk9387fo8r...')

or

![alt text](https://www.booger.com/9fjk4908fj3...)

or, to keep the last few characters visible for id'ing the extension...

![alt text](https://www.booger.com/9fjk...z73.png)
```

But in gitter I think we basically realized that an easy half-win in the Style Editor is to just not wrap lines.  This would reduce the vertical space needed for these long lines, and not much would be lost since most CSS doesn't take much horizontal space.  *And*, you likely wouldn't want your CSS code wrapping anyway-- particularly as we get to CSS Nesting, lines can have more depth and are more likely to wrap.

I think my more complicated proposal above is still valid, but regardless the Style Editor, which only accepts CSS, doesn't seem like a good place for text wrapping.

calculuschild:
Ok, so the concern is saving *vertical* space. That makes more sense. I was under the impression the goal was to make long strings easier to view and handle, which this PR does the opposite.

Since we also have a PR for Style Tab code folding https://github.com/naturalcrit/homebrewery/pull/3585 , might I propose that PR instead as a compromise? Users can selectively minimize long blocks at will, but they can also be expanded and fully visible without scrolling.

(For what it's worth, we should definitely be discouraging base64 encoding anyway; it makes brews extremely heavy and, makes Marked slower.)

Gazook89:
Does the code folding allow folding a single line, like `@import ...`?

calculuschild:
> Does the code folding allow folding a single line, like `@import ...`?

Yes. It folds `{}` blocks, and `@import` lines. We can add other cases if needed, but it currently handles just those.

image

image

calculuschild:
Going to close this in favor of #3585 .

PR #3585, opened 2024-07-19
Add Code folding on CSS style tab

dbolack-ab:
A simple addition for CSS folding in the style tab.


shared/naturalcrit/codeEditor/fold-css.js
```diff
@@ -0,0 +1,44 @@
...
+
+				return {
+					from : CodeMirror.Pos(start.line, 0),
+					to   : CodeMirror.Pos(end, cm.getLine(end).length)
+				};
+			}
+
+			// IMPORT FOLDING
+
+			const importMatcher = /^@import.*?[;]/;
```
G-Ambatte, line 32: I know I wrote this, but after further testing, I believe the square brackets in this regex are unnecessary. That said, it still works fine as is.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -431,7 +433,7 @@ const CodeEditor = createClass({
 				}
 				text = foldPreviewText || `Lines ${from.line+1}-${to.line+1}`;
 
-				text = text.trim();
+				text = text.trim().replace('{', '').trim();
```
G-Ambatte, line 436: Is this functionally different from `text = text.replace('{','').trim();`?

shared/naturalcrit/codeEditor/fold-css.js
```diff
@@ -0,0 +1,44 @@
+module.exports = {
+	registerHomebreweryHelper : function(CodeMirror) {
+		CodeMirror.registerHelper('fold', 'homebrewerycss', function(cm, start) {
+
+			// BRACE FOLDING
+			const startMatcher = /\{[ \t]*$/;
+			const endMatcher = /\}[ \t]*$/;
+			const prevLine = cm.getLine(start.line);
```
G-Ambatte, line 8: The `prevLine` naming is somewhat confusing; I believe that this came from the `fold-code.js`, but the folding blocks in that module are quite different to these CSS blocks. This could be renamed to something clearer: `startLine` or something similar.

G-Ambatte (changes requested):
These changes are functional; my comments are more about stylistic concerns than actual code issues.

shared/naturalcrit/codeEditor/codeEditor.jsx
dbolack-ab: In this context? Probably not. 

Didn't seem to be in my tests so I consolidated.

shared/naturalcrit/codeEditor/fold-css.js
dbolack-ab: I know I see these without the training semicolon frequently.

Shouldn't this be `/^@import.*/` ?

shared/naturalcrit/codeEditor/fold-css.js
dbolack-ab: It did indeed.

shared/naturalcrit/codeEditor/fold-css.js
G-Ambatte: At that point, it could just be `^@import`... I was originally working towards the semicolon as a mandatory terminator; every example at https://developer.mozilla.org/en-US/docs/Web/CSS/@import terminates with one. It's far less of an issue if `@import` is always a single line.

dbolack-ab: That does seem to be the case

calculuschild: Let's remove the square brackets because they do nothing.

```suggestion
			const importMatcher = /^@import.*?;/;
```

Looks like you need the semicolon though or anything after is not parsed.

shared/naturalcrit/codeEditor/codeEditor.jsx
```diff
@@ -431,9 +433,21 @@ const CodeEditor = createClass({
...
+				text = text.replace('{', '').trim();
+
+				// Extra data url chomping
+				// Try to make it pretty...
+				const startOfData = text.indexOf('data:');
+				if(startOfData) {
+					text = (startOfData > maxLength) ?
+						`${text.slice(0, text.indexOf(':') + 1)} ... ${text.slice(startOfData, startOfData + 5)} ...` :
+						`${text.slice(0, text.indexOf('data:') + 5)} ...`;
+				}
```
calculuschild, line 446: This section might need a rework. `text.indexOf` will return -1 if there is no match, which is not a falsy value. So the if statement will still run even if `data:` is not present.

This is causing non-`:data:` strings to get truncated aggressively:

##### Unfolded:
image

##### Folded - Before
image

##### Folded - After
image

calculuschild (changes requested):

shared/naturalcrit/codeEditor/codeEditor.jsx
5e-Cleric: Wow, instead of using actual css elipsis we just input the ... huh? Interesting choice.

calculuschild: I don't think CSS ellipses supports specifying a character limit, or does it?

ericscheid: IIRC, it's works on a linear width specified in px, pt, em, etc.
We could thus use the `ch` unit (which is the width of the `0` character).

Our editor uses a monospaced font.

5e-Cleric: There we go, monospaced font + width:5ch =  specified width of 4 characters (because the elipsis will take the space for 1 character..

calculuschild: Though we want a width of 50 characters. This PR is truncating to 4 by mistake.

dbolack-ab: I am resolving this solely against the truncation issue as the CSS based ellipses is a larger scoped change.

calculuschild:
Thanks @dbolack-ab ! Merging this now.

PR #3587, opened 2024-07-20
Add new var math functions

### LINKED ISSUE #3537: Enable `sign` and `abs` math functions
### Your idea:

A discord user would like to have the `sign` and `abs` functions enabled in our Math parser.  Because I'm not super familiar with that part of the repo, I'll keep it simple:

From user:

> Is there a way to include a "sign" function? That is, something like $[sign(3)] that would return "+"

> If they need a reason for both abs and sign, it comes from computing any modifier. The abs is needed to get the base number and the sign is needed to apply the sign.
```
$[StrSign]$[StrAbs]
```
> If the modifier is negative and the abs function isn't available, this would return "--2".

Relevant code:  https://github.com/naturalcrit/homebrewery/blob/master/shared/naturalcrit/markdown.js

Relevant Discord conversation: https://discord.com/channels/113801089362558983/950919905266507806/1252624469114097756

G-Ambatte:
This PR resolves #3537.

This PR adds the following new variable math functions:

- `abs(n)` : returns the absolute value of the content of  `n`;
- `sign(n)` : returns the sign of `n`: `+` for positive numbers, `-` for negative numbers, and `''` (empty string) for 0; and
- `signed(n)` : returns the value of `n` with its sign. In this function, `0` is treated as a positive number, returning `+0`.


G-Ambatte:
image

```
**SIGNED NEGATIVE**
:
$[a](-4)
:
$[signed(a)]

::


**SIGNED POSITIVE**
:
$[b](3)
:
$[signed(b)]

::

**SIGNED POSITIVE**
:
$[c](0)
:
$[signed(c)]


::


**SIGN NEGATIVE**
:
$[d](-4)
:
$[sign(d)]

::

**SIGN POSITIVE**
:
$[e](3)
:
$[sign(e)]


::


**ABSOLUTE NEGATIVE**
:
$[f](-4)
:
$[abs(f)]

::

**ABSOLUTE POSITIVE**
:
$[g](3)
:
$[abs(g)]
```

dbolack-ab (approved):
Looks good.

shared/naturalcrit/markdown.js
```diff
@@ -46,6 +46,27 @@ const mathParser = new MathParser({
 		array       : false, fndef       : false
 	}
 });
+// Add absolute function
+mathParser.functions.abs = function (a) {
+	return Math.abs(a);
+};
```
calculuschild, line 52: Don't need to define `abs`. It already exists in our `mathParser`, but has just been disabled (set to `false` above). Instead just set to `true`, and move it up with the other `true` functions. Might have to rearrange the block of disabled functions here to make it look nice again.

shared/naturalcrit/markdown.js
```diff
@@ -46,6 +46,27 @@ const mathParser = new MathParser({
...
+// Add sign function
+mathParser.functions.sign = function (a) {
+	if(a == 0) return '';
+	if(a > 0) {
+		return '\\+';
+	} else {
+		return '\\-';
+	}
+};
```
calculuschild, line 61:
```suggestion
mathParser.functions.sign = function (a) {
	if(a == 0) return '';
	if(a > 0) return '\\+';
	return '\\-';
};
```

shared/naturalcrit/markdown.js
```diff
@@ -46,6 +46,27 @@ const mathParser = new MathParser({
...
+// Add signed function
+mathParser.functions.signed = function (a) {
+	if(a >= 0) {
+		return `\\+${a}`;
+	} else {
+		return `\\${a}`;
+	}
+};
```
calculuschild, line 69:
```suggestion
mathParser.functions.signed = function (a) {
	if(a >= 0) return `\\+${a}`;
	return `\\${a}`;
};
```

shared/naturalcrit/markdown.js
```diff
@@ -46,6 +46,27 @@ const mathParser = new MathParser({
...
 });
+// Add absolute function
+mathParser.functions.abs = function (a) {
+	return Math.abs(a);
+};
+// Add sign function
+mathParser.functions.sign = function (a) {
+	if(a == 0) return '';
+	if(a > 0) {
+		return '\\+';
```
calculuschild, line 57: Are the escapes needed?

calculuschild (changes requested):
Seems functional. Just some small cleanup.

shared/naturalcrit/markdown.js
G-Ambatte: Yes and/or no

The reproduction text puts the `sign` and `signed` functions at the start of a line. However, the Markdown parser running after the variables and math have been resolved will quite happily accept these as list indicators, or... whatever the issue was for `+`, it escapes me right now.  
I can see it being argued either way: that these functions should output the raw text and nothing else, OR that they should output in a way that does not activate these Markdown functions.

I suspect that it comes down to "which is the better user experience?" and "which functionality is MORE unexpected?"

I started this PR leaning towards escaped output, but now I'm leaning more towards non-escaped.

shared/naturalcrit/markdown.js
G-Ambatte: I believe I created the function specifically because I was having difficulty getting the built in version to function. I'll take another swing at it; if there IS an issue with activating the functions, then it's probably better to find and resolve it now.

shared/naturalcrit/markdown.js
calculuschild: Both + and - are list markers. Based on your explanation I think escaped is better. The goal of `signed` is to get a mathematical symbol, not generate a list, so escaping would make sense then. Would be a confusing result otherwise I think.

shared/naturalcrit/markdown.js
G-Ambatte: Oh! I figured it out. It was the variable name! Once I changed the variable name to a letter that did not appear in the function name, the issue disappeared.

image

This might be an issue deserving of further investigation, but it falls outside the realm of this PR.

shared/naturalcrit/markdown.js
G-Ambatte: I've changed this PR to output the non-escaped versions now.

G-Ambatte:
> Seems functional. Just some small cleanup.

Cleanup should now be complete. Issue #3613 raised for the function/variable name clash discovered here.

calculuschild:
🤘 :finnadie: :metal:

PR #3589, opened 2024-07-21
Experimental Header Navigation

G-Ambatte:
This PR adds a button to the BrewRenderer. When clicked, this exposes the header navigation window, which lists all of the elements with IDs in the document.
It generates this information by calling `querySelectorAll('[id]')` to acquire a list of all elements that have an `id` attribute, then parsing these for the following conditions:

- has the class `page` - depth: 0; text: 'Page' + page number
- is `h1` through `h6` - depth: heading value, i.e. h1 => 1, h6 => 6; text: `innerText` value
- all other elements - depth: 7; text: `innerText` value

From one of the documents in my local install DB:

**CLOSED:**
image

**OPEN:**
image

**HOVERING OVER "PAGE 2:**
image

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -65,9 +65,11 @@ const BrewRenderer = (props)=>{
 		height             : PAGE_HEIGHT,
 		isMounted          : false,
 		visibility         : 'hidden',
+		showHeaderNav      : false
 	});
 
 	const mainRef  = useRef(null);
+	const pagesRef = useRef(null);
 
 	if(props.renderer == 'legacy') {
 		rawPages = props.text.split('\\page');
@@ -188,6 +190,60 @@ const BrewRenderer = (props)=>{
 		document.dispatchEvent(new MouseEvent('click'));
 	};
 
+	const toggleHeaderNav = ()=>{
+		setState((prevState)=>({
+			...prevState,
+			showHeaderNav : !prevState.showHeaderNav
+		}));
+	};
+
+	const renderHeaderNav = ()=>{
+		return <>
+			<i
+				className={`navIcon ${state.showHeaderNav ? 'fa-solid' : 'fa-regular'} fa-rectangle-list`}
+				onClick={toggleHeaderNav}
+			></i>
+			{state.showHeaderNav && renderHeaderLinks()}
+		</>;
+	};
+
+	const renderHeaderLinks = ()=>{
+		if(!pagesRef.current) return;
+		const elements = pagesRef.current.querySelectorAll('[id]');
+		if(!elements) return;
+		const navList = [];
+
+		elements.forEach((el)=>{
+			if(el.className.match(/\bpage\b/)) {
+				navList.push({
+					depth : 0,
+					text  : `Page ${el.id.slice(1)}`,
+					link  : el.id
+				});
+				return;
+			}
+			if(el.localName.match(/^h[1-6]/)){
+				navList.push({
+					depth : el.localName[1],
+					text  : el.innerText,
+					link  : el.id
+				});
+				return;
+			}
+			navList.push({
+				depth : 7,
+				text  : el.innerText,
+				link  : el.id
+			});
+		});
+
+		return _.map(navList, (navItem)=>{
+			return <p>
+				<a href={`#${navItem.link}`} target='_self'>{`${'-'.repeat(navItem.depth)}${navItem.text}`}</a>
+			</p>;
+		});
+	};
+
 	const rendererPath  = props.renderer == 'V3' ? 'V3' : 'Legacy';
 	const themePath     = props.theme ?? '5ePHB';
 	const baseThemePath = Themes[rendererPath][themePath].baseTheme;
@@ -232,11 +288,14 @@ const BrewRenderer = (props)=>{
 						&&
 						<>
 							{renderStyle()}
-							<div className='pages' lang={`${props.lang || 'en'}`}>
+							<div className='pages' lang={`${props.lang || 'en'}`} ref={pagesRef}>
 								{renderPages()}
 							</div>
 						</>
 					}
+				</div>
+				<div className='headerNav'>
+					{renderHeaderNav()}
 				</div>
 			</Frame>
 			{renderPageInfo()}
--- a/client/homebrew/brewRenderer/brewRenderer.less
+++ b/client/homebrew/brewRenderer/brewRenderer.less
@@ -31,12 +31,8 @@
 			visibility: hidden;
 		}
 	}
-	
-
-
-
-
 }
+
 .pane { position : relative; }
 .pageInfo {
 	position         : absolute;
@@ -65,6 +61,32 @@
 	background-color : #333333;
 }
 
+.headerNav {
+	position: fixed;
+	top: 0px;
+	left: 0px;
+	padding: 5px 10px;
+	background-color: #ccc;
+	border-radius: 5px;
+	max-height: 100vh;
+	overflow-y: auto;	
+	.navIcon {
+		cursor: pointer;
+	}
+	p {
+		padding: 2px;
+		a {
+			color: inherit;
+			text-decoration: none;
+			cursor: pointer;
+			&:hover {
+				text-decoration: underline;
+			}
+		}
+	}
+}
+
+
 @media print {
 	.brewRenderer {
 		height: 100%;
@@ -75,5 +97,8 @@
 				box-shadow: unset;
 			}
 		}
+	}
+	.headerNav {
+		visibility: hidden;
 	}
 }


G-Ambatte:
@calculuschild We might want a deployment on this one for testing

G-Ambatte:
Example of a custom element with custom ID:
image

G-Ambatte:
Some further tweaks:
image

5e-Cleric:
Once a deployment is ready, i will suggest some style options, maybe we could even pull something from the codemirror theme chosen.

Gazook89:
My thought is that this is another reason for a unified template for Share and Edit pages— share pages could even have a split panel as well. The left panel would have tabs including the table of contents/outline.

G-Ambatte:
In case it's not clear: this navigation lives in the BrewRenderer, so exists on both Edit and Share pages. I just realized that I've only posted screen shots from Edit pages, and that might cause some confusion.

Gazook89:
Yeah I know where it is currently in the PR. I am suggesting an alternate future, whether it’s done in this PR or not. It just feels like we already have a hideable sidebar/drawer interface available to us, and we might not need to invent another expanding component. Just my two cents.

Gazook89:
More thoughts on this in this one-person discussion: https://github.com/naturalcrit/homebrewery/discussions/2951

5e-Cleric:
Is this PR's popup supposed to take the entire space?

client/homebrew/brewRenderer/headerNav/headerNav.less
5e-Cleric, line 1: Recommended addition/change:

```
.headerNav {
    padding: 25px 30px 5px 10px;
    
    .navIcon {
          position:fixed;
          top:0px;
          padding-top:5px;
          
        &.active {
          width:470px;
          background:#dadada;
          z-index:10;
        }
    }
    &.active {
        padding: 25px 10px 5px 10px;
        height:100vh;
        width:500px;
        
        background:#dadada;
        
        p:has(.pageLink) {
          margin-block:10px 5px;
          &:after {
            content:'';
            height:2px;
            background:#929292;
            display:block;
            position:relative;
            top:4px;
          }
        }
        
        &::-webkit-scrollbar {
          width: 10px;
        }
        &::-webkit-scrollbar-corner {
          visibility: hidden;
        }
        &::-webkit-scrollbar-thumb {
            background: #929292;
            border-radius:5px;
        }
    }
}
```

to achieve:

<details><summary>Show image</summary>

image

</details>

client/homebrew/brewRenderer/headerNav/headerNav.jsx
```diff
@@ -0,0 +1,76 @@
...
+
+const HeaderNavItem = ({ link, text, depth, className })=>{
+	return <p>
+		<a href={`#${link}`} target='_self' className={className}>
+			{`${'-'.repeat(depth)}${text}`}
+		</a>
+	</p>;
+};
+
+export default HeaderNav;
```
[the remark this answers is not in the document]
5e-Cleric: Perhaps add a border- after pages, or an HR element too.

Also please add a title to this navigation, a simple `## Header Navigation` will suffice, like the title attrb on the button.

5e-Cleric (changes requested):
Overall i would make the requested changes, and following style changes.

client/homebrew/brewRenderer/headerNav/headerNav.jsx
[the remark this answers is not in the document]
5e-Cleric: Every link should have a class to indicate depth to be able to style this properly, not just the dash option (which i don't particularly like the look of)

G-Ambatte:
> Is this PR's popup supposed to take the entire space?
> 
> image
> 
> And please do include some way to close it at all times, this is a long document and the navigation is at least 10 pages long, and the only way to close is with a small button at the top?

I suspect it's a case of a really long title is causing the element to cover 100% of the viewable width. Some styling like `max-width: 10vw;` might be all that's required to stop it covering the whole BrewRenderer.

dbolack-ab:
I am very concerned that this pulls too many headers.  Would adding a pulldown to set the "Max" Header level be practical?

This PR has a natural extension to do the same navigation to the editing window.

ericscheid:
> I am very concerned that this pulls too many headers. Would adding a pulldown to set the "Max" Header level be practical?
> 
> This PR has a natural extension to do the same navigation to the editing window.

I can imagine a future PR could address this, perhaps by providing little `h1` → `h5` buttons across the top that auto-expand/collapse all headings to that level. (And have each heading that is a parent also be expando endabled).

Lets git this simpler version deployed first though.

Gazook89:
I think you can just put the button in the new toolbar, on the far left.  Rather than floating as it's own button.

5e-Cleric:
Further testing reveals somehow table headers are included in the previous h5 header text:

image
For reference:

image


Also, i am not sure about the width being proportional to window size (iframe size in this case), but if we can get to close it whenever we want it is not a big issue.

dbolack-ab:
The error looks to be from picking up the div containing the table's full content.

client/homebrew/brewRenderer/headerNav/headerNav.jsx
```diff
@@ -0,0 +1,76 @@
...
+	const toggleHeaderNav = ()=>{
+		setState((prevState)=>({
+			...prevState,
+			showHeaderNav : !prevState.showHeaderNav
+		}));
+	};
+
+	const renderHeaderLinks = ()=>{
+		if(!pagesRef.current) return;
+		const elements = pagesRef.current.querySelectorAll('[id]');
```
5e-Cleric, line 22: Right, we can let CSS do the heavy lifting here:

```suggestion
		const elements = pagesRef.current.querySelectorAll('*:is(h1,h2,h3,h4,h5,h6)[id]');
```

We could even use this to leverage the depth level later.

5e-Cleric: This should fix the aforementioned issue.

G-Ambatte: But `id` can exist on more than just headers.

5e-Cleric: Right, i thought we only wanted headings, but.. right, so the issue here is that any block with id gets caught, with all contents too, including in this case the table

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -424,6 +424,7 @@ const EditPage = createClass({
 						lang={this.state.brew.lang}
 						currentEditorPage={this.state.currentEditorPage}
 						allowPrint={true}
+						showHeaderNav={true}
```
Gazook89, line 427: Does this component really need to stretch all the way to up to editPage.jsx and sharePage.jsx files?  It seems like we could just assume that the headerNav is always going to present in both pages (not always visible, of course), so we don't need to pass it in as a boolean prop on BrewRenderer-- it could just be included in BrewRenderer unconditionally.

G-Ambatte: It could, if we also want it to appear on every other page, like New, Home, Change Log, Welcome...

Gazook89: I don't really see a problem with that.  For something like the changelog, especially, it'd be nice to just have a list of dates/versions.  Same deal with FAQ (though it's obviously not very long at this point).  The point of the site is to create documents, and I'm not sure we need to say "some documents have these tools, others don't".  

If there are pages where navigation by header *isn't* desired, then we should maybe consider stop making everything out of special-case brews and just make normal pages.

5e-Cleric: I agree with G-Ambatte here, FAQ perhaps, but the rest don't need navigation at all, 2 page documents....

Gazook89: I’m not sure why there needs to be a distinction between short and long documents here. For all documents are we going to make the button disappear if it’s shorter than 2 pages? What if it is 10 pages but has no headers? 

The goal should be to have a consistent, unified UI as much as possible. If a button exists for one brew, it should be available on all brews. If, for whatever reason, a button really surely doesn’t do anything on a particular brew, then it should still be there but be disabled. That isn’t the case here, though.

G-Ambatte:
Interesting interaction here:

Current state of the PR:
image

Current state of the PR with `content-visibility: auto` rule disabled:
image

G-Ambatte:
Confirmed, fixed by changing to `textContent` instead of using `innerText`.

G-Ambatte:
State of the PR:

image

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: By moving the activation button into the toolbar, this no longer requires changes to Edit or Share page, so those changes have been reverted.

client/homebrew/brewRenderer/headerNav/headerNav.jsx
G-Ambatte: I've improved the text handling, which should prevent the entire content of a table with an ID being inserted into the header navigation.

G-Ambatte:
I believe that I've addressed the review comments; if I've missed anything, please let me know.

client/homebrew/brewRenderer/headerNav/headerNav.jsx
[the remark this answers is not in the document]
G-Ambatte: This is now the case - the head navigation is now inside `nav` > `ul` > `li` > `a`, instead of just being inside a `div` > `p` > `a`. Each `a` now has a `depth-N` class assigned, which sets the amount of `padding-left`; typically, `(N - 1) * 0.5em`.

5e-Cleric (changes requested):
Left 1 change requested, and a question, otherwise pretty good.

G-Ambatte:
> @G-Ambatte Do you think there is worth in highlighting the selected element for a second?

I can see some value in it; although it seems like it might be best left for a future PR.

5e-Cleric:
As you wish, this is ready, i'd wait for more visual reviews, seeing if someone has anything to say about its styling.

5e-Cleric (approved):

calculuschild:
I haven't really looked at this one yet, but if I can at least do a visual check tomorrow before merging.

5e-Cleric:
Luckily we had not merged yet:

```
bundle.js:84613 Uncaught ReferenceError: updateState is not defined
    at Object.onPageChange (bundle.js:84613:5)
    at Object.handlePageVisibilityChange [as onVisibilityChange] (bundle.js:79678:25)
    at bundle.js:79591:41
    at Array.forEach (<anonymous>)
    at IntersectionObserver.threshold (bundle.js:79590:15)
```

when checking a brew in sharepage

G-Ambatte:
image

Initial glance at the error seems to indicate that it has nothing to do with the changes in this PR,

image

The issue is in SharePage.jsx, line 26, which this PR does not touch.

G-Ambatte:
Discovered a couple of minor things this morning - an empty header (e.g. `###`) results in an empty line in the nav, as the element has the `id` attribute but no attribute content.

Also, headers in the ToC appear in the navigation, meaning that about 30% of the nav points to the ToC rather than the actual content.

G-Ambatte:
I think the selector is in a good place now; any future improvements can probably have their own separate PR.

Example brew: https://homebrewery-pr-3589.herokuapp.com/share/Ieoj7n9yKS_m

image

calculuschild:
This looks functional enough to merge. I do think at some point it might be worth comparing this to the existing logic from the TOC generator snippet and try to make them share one common piece of code.

Gazook89:
I don't think it needs to happen here, but this reinforces my desire to utilize the Edit page component/layout for the Share page, and have this Header Navigation occur in the left pane.    So `/edit` has Brew Editor, Style Editor, Properties Editor and the `/share/` route has just Header Navigation and Properties Viewer.      The properties viewer in the title bar would be replaced by the Properties Viewer.

client/homebrew/brewRenderer/headerNav/headerNav.jsx
```diff
@@ -0,0 +1,94 @@
...
+				text  : el.textContent,
+				link  : el.id
+			});
+		});
+
+		return _.map(navList, (navItem, index)=>{
+			return <HeaderNavItem {...navItem} key={index} />;
+		});
+
+	};
```
5e-Cleric, line 57: Would you mind commenting this for future reference?

5e-Cleric (approved):
Left a comment but this is ready for merge.

client/homebrew/brewRenderer/headerNav/headerNav.jsx
G-Ambatte: Completed in latest commit.

PR #3615, opened 2024-08-03
Prevent self-selection of User Theme

### LINKED ISSUE #3614: Theme: Themes can select themselves as Theme Source
### Your idea:

As shown in the below image, it is possible for a brew tagged with `meta:theme` to select itself from the dropdown list. This seems incorrect.

image

G-Ambatte:
This PR resolves #3614.

This PR adds a check to stop a brew from adding itself to the dropdown list of available user themes.


calculuschild:
👉😎👉

PR #3624, opened 2024-08-09
Fix var name substring clash #3613

### LINKED ISSUE #3613: Markdown: Functions break if a variable name is a substring of the function names
### Your idea:

During development of #3587, I discovered that `abs(a)` did not work, but `abs(c)` did.

Exhaustive testing for `abs()`:
image

Initial testing for `floor()`:
image

As can be seen, in every case where a variable name used in that calculation is a substring of the function name, the function stops working.

My suspicion is that the issue lies here, `shared/naturalcrit/markdown.js:line 465`:
image

By replacing _every_ instance of the variable name, we are catching unintended characters. Instead, we need to only replace the instances identified in the original regex match.

G-Ambatte:
This PR resolves #3613.

This PR changes the regex of the variable name replacement, to ensure that names that are substrings of math functions or other variables are not unintentionally replaced.

CURRENT:
```
[a]:-4

$[abs(a)]
```
**OUTPUT:** `$[abs(a)]`

image

PR:
```
[a]:-4

$[abs(a)]
```
**OUTPUT:** `4`

image

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -462,7 +462,7 @@ const replaceVar = function(input, hoist=false, allowUnresolved=false) {
 		mathVars?.forEach((variable)=>{
 			const foundVar = lookupVar(variable, globalPageNumber, hoist);
 			if(foundVar && foundVar.resolved && foundVar.content && !isNaN(foundVar.content)) // Only subsitute math values if fully resolved, not empty strings, and numbers
-				replacedLabel = replacedLabel.replaceAll(variable, foundVar.content);
+				replacedLabel = replacedLabel.replaceAll(new RegExp(`/(?<!\\w)(${variable})(?!\\w)`, 'g'), foundVar.content);
 		});
 
 		try {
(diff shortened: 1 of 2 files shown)


5e-Cleric (commented):
Would need to test this before approving, can we get a deployment?

calculuschild:
> Would need to test this before approving, can we get a deployment?

Deployment live

5e-Cleric:
Can confirm this works, tested with `min()` because `abs()` is not live in the actual site.

`$[min()]` returns `<p>Infinity</p>`, is this expected behaviour?

G-Ambatte:
> Can confirm this works, tested with `min()` because `abs()` is not live in the actual site.

`abs()` is not live in Production, but IS live in master, so should work in the deployment.

> `$[min()]` returns `<p>Infinity</p>`, is this expected behaviour?

This matches the function in the currently live Production, so I'm going to say "yes", but hesitantly.

---

REPRODUCTION BREW IN DEPLOYMENT: https://homebrewery-pr-3624.herokuapp.com/share/wSI_KF8CbN6N

calculuschild:
Very nice. Good debugging on this one!

PR #3625, opened 2024-08-09
Tweak variable math split regex

### LINKED ISSUE #3622: Math functions with multiple parameters cannot take variable names
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

The expectation is the reproduction code will return `2`; instead it returns `$[min(2,DexMod)]`

See reproduction brew here - https://homebrewery.naturalcrit.com/share/VE5qZsAsuAfk

### Code

```gfm
[DexMod]: 4

$[min(2,DexMod)]
```

G-Ambatte:
This PR resolves #3622.

This PR adds the comma character to the regex used to split the variable string in order to identify variable names. 

CURRENT:
RegEx: `/[a-z]+\(|[+\-*/^()]/g`

`min(2,var)` => `min(` : matches `[a-z]+\(`; `)` : matches `[+\-*/^()]` => OUTPUT: `2,var`
As variable names cannot contain commas, this never matches a variable name and thus the replacement function does not run.
**NB:** changing the input string to `min(2,(var))` **does work** because the brackets cause the replacement function to correctly identify the variable name.

Without brackets:
image
With brackets:
image

PR:
RegEx: `/[a-z]+\(|[+\-*/^(),]/g`

`min(2,var)` => `min(` : matches `[a-z]+\(`; `,` : matches `[+\-*/^()]`; `)` : matches `[+\-*/^()]` => OUTPUT: `2`,`var`

Without brackets:
image
With brackets:
image


G-Ambatte (commented):
This is a comment review on my own PR to test GitHub integrations for Gitter.

shared/naturalcrit/markdown.js
5e-Cleric, line 1: I am unsure if a change this big and of this importance should be done in a single PR, are we sure we won't angry the programmer gods?

5e-Cleric (approved):

shared/naturalcrit/markdown.js
G-Ambatte: > Nor did I deem that thou, a mortal man,
> Could’st by a breath annul and override
> The immutable unwritten laws of Heaven.
> They were not born today nor yesterday;
> They die not; and none knoweth whence they sprang.

\- _Antigone_, by Sophocles

Ancient Greeks aside, it IS a change to the regex string, and while I do believe that my RegEx-ing has significantly improved over the last few years, I still do not trust that there will not be unintended consequences.

Issue #3629, opened 2024-08-12
Non-used URLs open bugged new page


5e-Cleric:
Any URL that is not defined in our router by default opens a `/new` page. However after the custom themes update (v3.14.0), this opens with errors in saving and no snippet menus:

![image](https://github.com/user-attachments/assets/9eaaf73b-c083-488f-9354-12b8e35a1ed3)

With the next error in console:

```
Error: Not Found
    at Request.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:78344:21)
    at Emitter.emit (https://homebrewery.naturalcrit.com/homebrew/bundle.js:1954:20)
    at XMLHttpRequest.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:78638:10)
{
  "status": 404,
  "method": "GET",
  "url": "/api/theme/undefined/undefined"
}
```

This should be fixed.

5e-Cleric:
Hmm, yeah, alright then.

PR #3630, opened 2024-08-13
Prevent loading of undefined renderer or theme theme bundle.

dbolack-ab:
This should solve for #3629 when an invalid path load in turn tries to load "undefined" themes.


G-Ambatte:
If we instead put a parameter sanity check as a guard clause in the function in `helpers.js`, then we should only have to update a single piece of code in the future, right?

Is there a reason that something like this wouldn't work?

```
const fetchThemeBundle = async (obj, renderer, theme)=>{
	if(!renderer || !theme) return;
	const res = await request
			.get(`/api/theme/${renderer}/${theme}`)
			.catch((err)=>{
				obj.setState({ error: err });
			});
	if(!res) return;
[...]
```

dbolack-ab:
> If we instead put a parameter sanity check as a guard clause in the function in `helpers.js`, then we should only have to update a single piece of code in the future, right?
> 
> Is there a reason that something like this wouldn't work?
> 
> ```
> const fetchThemeBundle = async (obj, renderer, theme)=>{
> 	if(!renderer || !theme) return;
> 	const res = await request
> 			.get(`/api/theme/${renderer}/${theme}`)
> 			.catch((err)=>{
> 				obj.setState({ error: err });
> 			});
> 	if(!res) return;
> [...]
> ```

I can give that a whirl. I tried catching this in a different place at it was uglier. This seems to work better and simpler.

calculuschild:
This doesn't actually address the root issue of #3629. The issue there, is invalid URLs are emitting garbage data. Instead, we unrecognized paths to be rerouted to `/`.

So this PR is probably a good safeguard, but we also need to solve the root issue.

calculuschild:
Seems to work as expected. Thanks!

PR #3634, opened 2024-08-14
Fix Nav flow

### LINKED ISSUE #2071: Awkward flow in navbar when brew title is long

5e-Cleric:
Fixes #2071


5e-Cleric:
Still would need to work out something for actual flow of the nav to a sidebar or something like that, for mobile or screens smaller than 1200px

5e-Cleric:
Hmm, tests are failing because less version was updated, but not on vitreum.git

Gazook89:
image

this is what the deployment looks like on my phone. It’s /new, and there is no title, so it seems to be shrinking that piece to nearly nothing

calculuschild:
Seems fine to me. Merging!

Thanks @5e-Cleric ! 💯 ⭐ 🎉

Issue #2076, opened 2022-03-14
Snippet bar items overlap when width reduced


Gazook89:
Another cosmetic thing I've noticed is that the Editor buttons in the snippet bar overlap the snippet menus when the width is reduced beyond a certain point.  The main issue is that it just looks a little sloppy, and could be more responsive.

<img width="488" alt="image" src="https://user-images.githubusercontent.com/58999374/158198477-83937b14-b013-4c88-93ad-e1ba0cb4672b.png">

[Here is a link to a branch on my fork](https://github.com/Gazook89/homebrewery/tree/flexbox-snippets-bar) with some adjustments, including:
- add a wrapper div `.snippets` around the Snippet menus, similar to how the 'undo/redo' various editors are contained by `.editors`
- add flexbox properties to the outer snippetbar div and the elements within them.
- Still need to figure out some stuff like preventing text/element wrapping at a certain point, and what to do with the tooltip on the editor icons.

<img width="488" alt="image" src="https://user-images.githubusercontent.com/58999374/158198729-4541e748-bff9-4c52-9f1a-2ecc5608d653.png">

<img width="488" alt="image" src="https://user-images.githubusercontent.com/58999374/158198903-6eb27702-fc33-4566-a919-f4ec1d95a6eb.png">

------

This could be made simpler, and likely better, if the Editors portion didn't need to be right-aligned in the bar.  And to that point, it might make more sense to move the Editors/Undo-Redo portion to the *left side* of the snippet bar, since they don't change based on context, and then followed by the snippet menus.

Gazook89:
I created another branch to try a new configuration of the snippet bar.  IMHO, it looks nice, and while I think it will feel odd to long time users at first, will make sense over time.  And if Swappable Themes is pushed through I think it will work better with custom snippet groups; the persistent editor icons are always in the same place and snippet menus can be added into infinitum.  But I may be grasping at straws to justify the time I spent on this.

Single line tool bar:
<img width="579" alt="image" src="https://user-images.githubusercontent.com/58999374/158214322-d7ad85ad-64a5-4d60-bf3d-b0f9d3be63ae.png">

Squished to two lines:
<img width="579" alt="image" src="https://user-images.githubusercontent.com/58999374/158214500-3b2ea983-6612-4bf6-98ab-2ebef6eac9cf.png">

An action gif:
<img width="579" alt="a gif of the snippetbar" src="http://www.giphy.com/gifs/GuPjXU5tGDmCSAgle9">

[...]

PR #3635, opened 2024-08-14
Snippet bar wrapping

5e-Cleric:
### Before

image

### After

image

No animations, no nothin, just wrapping when there is no space.


Gazook89:
The deployment isn't showing any change for me.

5e-Cleric:
> The deployment isn't showing any change for me.

huh

5e-Cleric:
Ok can confirm works perfectly fine in local, and there are no changes to push....

5e-Cleric:
Fixed it, for some reason less doesn't like container queries nested in deployment.

I have reason to believe this would work if our version of less wasn't 1 year behind.

Gazook89:
Okay, confirmed what I suspected might happen (I've made pretty much this exact PR before):   If the editor buttons wrap below the dropdown menus, you can't use the dropdown menus.  Via JS, the dropdown menus are placed below the snippet bar total height.  

If you hover over the menu button, then move your mouse down to select something from the menu, your mouse will move over the Editor buttons and thus off your hover-activated menu and close it.

I think an appropriate fix, without going overboard on UI changes or pulling in UI libraries, is to rearrange the entire snippet bar-- it would fix this probably *and* I believe make more logical sense.  Basically, flip everything around to reverse order.

Move the Editor buttons (Brew, Style, Properties) furthest to left.  These are the main categories of the editor and completely change the content.

Move the Undo/Redo buttons and codemirror theme selector to the space after the editors.  These are applicable to multiple editors (Brew and Style), and do not change.

Move the snippet menus to the end of the snippet bar.  These are dependent on the active editor, and are further dependent on the active Theme-- some themes may have more or fewer menus.  Placing them at the end allows that number to grow without causing as much issue (though the dropdown menu/hover issue will persist if we continue to break the snippet menus across multiple lines).    

Then allow wrapping at the same spot-- just before the snippet menus.

Gazook89:
I realize now that you already linked to the Issue where I show the re-ordering idea: https://github.com/naturalcrit/homebrewery/issues/2076#issuecomment-1067022259

So yeah....I don't know if my old branch even still exists on my computer or if it can be brought up to Master easily.  But that's my idea in a giffy nutshell.

5e-Cleric:
Damn, yeah, i should have given this PR more than 5 minutes... hmm i personally don't like the idea of swapping them, visually, but i guess it is the best we have... Could use the team's opinion on this, @calculuschild @G-Ambatte @ericscheid @dbolack-ab

dbolack-ab:
> Fixed it, for some reason less doesn't like container queries nested in deployment.
> 
> I have reason to believe this would work if our version of less wasn't 1 year behind.

Should we fix that first?

dbolack-ab:
> Damn, yeah, i should have given this PR more than 5 minutes... hmm i personally don't like the idea of swapping them, visually, but i guess it is the best we have... Could use the team's opinion on this, @calculuschild @G-Ambatte @ericscheid @dbolack-ab

I dunno. I kinda like the controls up the "centerline" of the window, particularly if the scroll-lock/jump buttons are added.

This is a somewhat radical suggestion. What about moving the Snippets to a single parent on the left and changing it to a vertical menu?

calculuschild:
Is CSS smart enough that we could leave the order the same, but have the snippets take the bottom row when needed?

Gazook89:
> Is CSS smart enough that we could leave the order the same, but have the snippets take the bottom row when wrapping?

Possibly with CSS Grid and grid areas, and @media breakpoints. I think that’s a weird way to fix it, compared to just reordering. 

@dbolack-ab  I’m not sure I have a complete understanding of your first proposal: where would are you suggesting the scroll related buttons go?

dbolack-ab:
> 
> **EDIT** Okay, I think i understand the second suggestion of a single parent button to trigger a "base" menu for the snippets, from which the rest of the snippet menus can sprout from. Correct me if I got that wrong. Basically it would add another click/hover to all snippets and increase the depth of each item. I do think that _would_ be a radical change in terms of what users will notice and react to (more so than just a simple reordering).

Yes, that covers it. And yeah, it is a non-trivial change, but it keeps the top bar clean and in place. The Snippets menu would be better served by being click to open on subs, however many menus have actions on click as well as having subs which means a possibly non-trivial reworking.

With the potential of incoming user snippets, it seems like our best bad choice. Also, rotating the menu to being vertical rather than horizontal might help with the mouseover problem.

Gazook89:
> With the potential of incoming user snippets, it seems like our best bad choice. Also, rotating the menu to being vertical rather than horizontal might help with the mouseover problem.

So would the menu be indicated with an icon only?  I assume you mean something like this (from my vs code): 
image

It may be helpful if you can throw a quick sketch of placement.  Just to be sure we aren't talking past each other.

> however many menus have actions on click as well as having subs

I didn't even realize this. I suppose it just takes the top snippet in the submenu?  I think this is odd-- I wouldn't miss it. 


----

Then there is this mockup I did for the ToC stuff (I see now i missed some comments/questions from you).  It moves the editor buttons to the bottom as tabs.  This particular screenshot is just of the "ToC" tab, but you can imagine on the Editor tabs that the snippet bar would remain near the top much as it is today.

----

One thing to keep in mind is that we shouldn't fret *too much* about narrow widths....at some point, the person still has to type in the editor (at least if they are inserting snippets), and the snippet names have a minimum width as well.  So we don't need to worry about how things will responsively adjust to super narrow widths.  At some point, it should just nicely start overflowing it's container and be hidden.

5e-Cleric:
> > Is CSS smart enough that we could leave the order the same, but have the snippets take the bottom row when wrapping?
> 
> Possibly with CSS Grid and grid areas, and @media breakpoints. I think that’s a weird way to fix it, compared to just reordering.
> 
> @dbolack-ab I’m not sure I have a complete understanding of your first proposal: where would are you suggesting the scroll related buttons go?
> 
> **EDIT** Okay, I think i understand the second suggestion of a single parent button to trigger a "base" menu for the snippets, from which the rest of the snippet menus can sprout from. Correct me if I got that wrong. Basically it would add another click/hover to all snippets and increase the depth of each item. I do think that _would_ be a radical change in terms of what users will notice and react to (more so than just a simple reordering).
> 
> I don't think it's a bad idea, except with our current components it adds another point of failure when trying to get through menus. Because our menus are on 'hover' only, it's easy to lose your progress through a menu each time there is another level of depth. I think this approach of a deeper menu tree would be improved with an actual UI library that can provide dropdown menus that stay open on click (and close "on click outside") and can be navigated entirely with keyboard arrows/tabs. One more point, that is more general, is that another level of nesting will decrease discoverability of all snippets-- they'll be further hidden, and it'll be "one click extra" more difficult to just try a bunch.

Okay, having read it all, can we merge this PR as is, or with items reversed, and leave that to another PR?

calculuschild:
> Okay, having read it all, can we merge this PR as is, or with items reversed, and leave that to another PR?

Can't merge as-is because the snippets are still unusable when wrapping happens.

5e-Cleric:
> Okay, confirmed what I suspected might happen (I've made pretty much this exact PR before): If the editor buttons wrap below the dropdown menus, you can't use the dropdown menus. Via JS, the dropdown menus are placed below the snippet bar total height.
> 
> If you hover over the menu button, then move your mouse down to select something from the menu, your mouse will move over the Editor buttons and thus off your hover-activated menu and close it.
> 
> I think an appropriate fix, without going overboard on UI changes or pulling in UI libraries, is to rearrange the entire snippet bar-- it would fix this probably _and_ I believe make more logical sense. Basically, flip everything around to reverse order.
> 
> Move the Editor buttons (Brew, Style, Properties) furthest to left. These are the main categories of the editor and completely change the content.
> 
> Move the Undo/Redo buttons and codemirror theme selector to the space after the editors. These are applicable to multiple editors (Brew and Style), and do not change.
> 
> Move the snippet menus to the end of the snippet bar. These are dependent on the active editor, and are further dependent on the active Theme-- some themes may have more or fewer menus. Placing them at the end allows that number to grow without causing as much issue (though the dropdown menu/hover issue will persist if we continue to break the snippet menus across multiple lines).
> 
> Then allow wrapping at the same spot-- just before the snippet menus.

### Uncollapsed
image

### Collapsed

image

5e-Cleric:
Gotta be honest, i like the unwrapped, but the second line being wider i don't particularly enjoy

Also, i am considering addnig names to the tabs, and separating them from the rest of the UI with some sort of colors or moving them to a separate place.

Gazook89:
> wider second line vs first line

That's why i stretched the editor buttons in my original take on this, again in this comment: [2076](https://github.com/naturalcrit/homebrewery/issues/2076#issuecomment-1067022259).  It is significantly less jarring.  

I also think that the editor buttons should be labelled, because I hate having to tell new users "Go to the Style Editor (the paint brush icon by the snippets)...".  For a critical piece of the editor, it shouldn't only be an icon.  

To that effect, I had been working on proper looking tabs, separate from the snippets.  I have (or had) a branch that widely adopted Radix UI library but I'm not sure I ever published it (there is a closed one, but that was an older attempt) which had editor tabs...I can't find it now.  But basically it looked like what is shown in the ToC discussion, [here](https://github.com/naturalcrit/homebrewery/discussions/2951#discussioncomment-6523588).  I still think it is an effective way to communicate that the buttons are not part of the snippets, that they are important, and help users find them. 

But without straying too far from where this PR is currently at, you could add text labels to the current icons-- and then drop them when their containers get too small, now that CSS Container Queries are a thing.

calculuschild (changes requested):
A few observations with the current behavior:

- Snippets do not wrap on the **style** tab. They jump over to the left but do not move into their own row. All tabs should probably have the same behavior
- Editor theme dropdown is cut off on the left
image
- I would personally prefer snippets stay on the left when unfolded if it can be done.
  - Our snippet subgroups unfold toward the right, which feels more natural than unfolding to the left or unfolding out on top of the BrewRenderer IMO
  - Snippets on the right means the dropdowns go off the edge
image
  - When folded, the snippets become left-aligned anyway. Reworking everything to switch between supporting left and right-alignment depending on the fold state seems more confusion than it's worth.
- I agree with Gazook's comments about making the top buttons resize to fill the top row


Agree/disagree on these?

5e-Cleric:
Right, i'm more looking right now for what should have been the full scope of the design, and probably a final decision on it, whether it is a poll or whatever, more than an actual review of the current state.

calculuschild:
> Right, i'm more looking right now for what should have been the full scope of the design, and probably a final decision on it, whether it is a poll or whatever, more than an actual review of the current state.

So, here is what I think the scope should be:

- Do not change the snippet bar to vertical
- Snippets should drop below the other editor tools
- Snippets should remain on the left side both when folded and unfolded
- Tools should expand to fill the top row when folded, and float right when unfolded
- Folding should happen in both the text and style editors

i.e., unfolded:
image


folded:
image

ericscheid:
The simplest thing we can start with is applying a css rule to `.snippetBar .groupName` to hide the `groupName` if the available space is small.

snippet group names hidden in narrow snippetbar

We can do this with [container queries](https://caniuse.com/?search=container), like this:

```css
/* Enable container queries for .snippetBar */
.snippetBar {
    container-type: inline-size;
}

/* Hide .groupName when .snippetBar is less than 540px wide */
@container (max-width: 540px) {
    .groupName {
        display: none;
    }
}
```

calculuschild:
I think the label hiding can go along with the wrapping if needed. We can do both.

Given how much work has gone into this PR already and how long it has been sitting, I would say just adjust this snippet wrapping PR with the behavior I mentioned above, and we can add a new issue with the idea for hiding the labels.

Unless anyone has a strong objection, I don't think further discussion is needed here to finish this.

client/homebrew/editor/snippetbar/snippetbar.less
```diff
@@ -2,99 +2,116 @@
...
 			}
-			&.history {
-				.tooltipLeft('History');
-				font-size : 0.75em;
-				color     : grey;
-				position  : relative;
-				&.active {
-					color : inherit;
+
+			& > div {
```
Gazook89, line 33: I think here you need to have `position: relative;`, so that the hover tooltips display next to the button.  Couple of alternatives, though, if you want to consider them:

- They tooltips could be displayed next to their *group* rather than each button, in which case just set `position: relative;` one level up.
- They tooltips currently can go outside the container and are not visible if the snippetbar is wrapped.  I would actually create two sets of rules, one for tooltip below and one for tooltip on the left.  Set the "left" rule when not wrapped, and the "below" rule for when wrapped.  
- Or, just get rid of the fancy tooltips entirely and use the `title` attribute, which is more accessible, and has built-in collision detection so it will always be visible.  

Ultimately, CSS Anchor Positioning would fix this very very easily, but it is not available in FF yet.  If you wanted to go ahead with it anyway, you could do a `@supports` rule and have one set for Chrome, one for FF.  In fact, the entire system of snippet menus could use it for better results and less code.  Probably it's own PR though.

Gazook89:
I echo @calculuschild's plan above, with the editor tools expanding to fill the space when the bar is wrapped.  I understand that right after it wraps the width of the buttons is a little silly, but ultimately i think it works out better.  

*However*, if going with the option where buttons are not stretched and they just stick to the left side (when wrapped), I think it's worth doing a slight color change.  Change the snippetBar background to a darker gray, and assign the lighter `#DDD` to the buttons themselves.  Remove the `border-left` properties off everything.  Then set the `gap` property on the `.snippetBar`, `.editors`, and `.snippets` elements to `1px` to replace those border-lefts.  This will get you something like this:

image

and it occurs to me now that your PR adds some container query that changes the width of the editor buttons, so without that it would look like this:

image

I think this background color change makes it look a little neater because I can better understand why there is a big empty gap:  the *editor tools are one group, and the snippets are another*.  This might be entirely self-evident to someone who hasn't looked at the HTML structure of the site if everything just has the same BG color.  Also, creating the space between the buttons with `gap` is a shorter way of getting the same effect with less chances for issues.

dbolack-ab:
> **EDIT** Okay, I think i understand the second suggestion of a single parent button to trigger a "base" menu for the snippets, from which the rest of the snippet menus can sprout from. Correct me if I got that wrong. Basically it would add another click/hover to all snippets and increase the depth of each item. I do think that _would_ be a radical change in terms of what users will notice and react to (more so than just a simple reordering).

That's exactly it. And yes, it's awkward in exactly that way, but frankly this seems like a best-effort failure mode management solution.

client/homebrew/editor/snippetbar/snippetbar.less
5e-Cleric: restored position

5e-Cleric:
## Update

The last iteration of this PR, after lots of discussion, is this:

Captura de pantalla 2024-10-20 002337

dbolack-ab:
That's solid.

calculuschild:
Ok, I think this is fine for now. There are other issues that probably need to be resolved, but they aren't new from this PR, so I will merge this and save those for a later PR so we can be working from a clean slate.

Specifically, this PR makes it more obvious that snippet dropdowns are getting clipped at the divider bar:

image

Issue #513, opened 2017-10-09
[FEATURE]: PageNumber Auto extension: Exclude tagged pages from count


DoctorDan1986:
I like to make the first page a cover page for my own stylistic purposes.  It also helps when exporting as a PDF.  However, the pageNumber Auto feature naturally counts this first page as page 1, which throws off the auto count for what I want.

I was thinking that you might include a method to exclude certain pages from the output, something like a 

`<div class='pageNumber null'></div>`
or 
`<div class='pageNumber skip'></div>`

The effect I'm looking for is that any page with this class will have a page counter value of 0, so that if one, say, put this on the first page of the brew, and used the pageNumber Auto on page 2, it would exclude page 1 from the count and start the count at 1 on page 2.

### Additional Details

**Share Link** :

or

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

PASTE BREW CODE HERE

</pre></code></details>

Gazook89:
I couldn't figure out how to make a snippet that you could place on a particular page to make the count be skipped or reset, but here is a CSS method for resetting the page counter.  You have to target the page you want by page number, so the downside is that any rearranging of your text means you will need to double check the reset page is still targeted....Not likely that big of deal in this case, because it is unlikely the first page or two are going to get jostled around much.

CSS:
```
.page#p1 {
	counter-reset:phb-page-numbers;	
}
```
This could be made into a snippet if approved by the powers that be.

Gazook89:
@calculuschild --- I think this goes one of two ways:  a snippet with the above is added to the Style Editor menus (which I can add very quickly), or just close it.    

Besides for the css snippet above, I don't see how this would be possible without big changes, such as custom markdown tokens(?).  I imagine that it is vanishingly rare that users would want to skip a page in their count, except in the first few pages in which case `counter-reset` will work.

5e-Cleric:
wouldn't this be solved with a snippet and `.page:has(.counterReset)`?

I know this is a 6yr old issue

G-Ambatte:
> wouldn't this be solved with a snippet and `.page:has(.counterReset)`?
> 
> I know this is a 6yr old issue

Something like this:

```
.page:has(.pageCounterSkip) {
    counter-increment: phb-page-numbers 0;
}
```

See example brew: https://homebrewery.naturalcrit.com/share/Q6xWK06zBUa_

A far more difficult question is then about how to handle the footer mirroring to match the new page numbering.

[...]

PR #3639, opened 2024-08-15
Rework page counters for skipping and resets.

dbolack-ab:
Solves #513

This adds the .skipCounting and .resetCounting classes for causing a page number to not be incremented or to reset the number at 1.

The ToC Snippet is corrected to match the displayed page numbers while correctly tracking the page ids.


themes/V3/Blank/style.less
```diff
@@ -12,7 +12,7 @@
 }
 
 @page { margin : 0; }
-body { counter-reset : page-numbers; }
+body { counter-reset : page-numbers 0; }
```
5e-Cleric, line 15: Default is 0 anyway, consider this change has no effect.

themes/V3/Blank/style.less
```diff
@@ -481,4 +480,13 @@ body { counter-reset : page-numbers; }
 	&:nth-child(even) {
 		.pageNumber { left : 30px; }
 	}
-}
+
+	.resetCounting {
+		counter-set : page-numbers 1;
+	}
```
5e-Cleric, line 486: wouldn't we want the next page to be 1? if so, we would set this one to 0. Not sure if i missed something.

dbolack-ab: No. Reset applies to the current page.

themes/V3/Blank/style.less
dbolack-ab: Preference for explicitness.

themes/V3/Blank/style.less
5e-Cleric: Sorry, i was wrongly assuming the resseting page would also be skipped, don't know where i got that assumption, you are right.

5e-Cleric:
This is working fine, i wonder if we should add a snippet for it.

dbolack-ab:
> This is working fine, i wonder if we should add a snippet for it.
> 
> I swear, at some point we will get a user saying: the skipcounting snippet does not work! while using manual numbering.

I will create a separate pr for that.

dbolack-ab:
Solves #2826

calculuschild (commented):
The new `.skipCounting` and `.resetCounting` CSS classes seem to work as expected.

TOC generation seems to point to the correct page, while displaying a page number matching the current count, so that seems ok.

Test document here:

https://homebrewery-pr-3639.herokuapp.com/new/8k6xsvPelZGj

I still want to dig into the code for cleanup, but it appears functional.

Questions before merging:
- ~Do we need to add snippets for these?~ Edit: I see this is linked to the other PR
- Should pages with `.skipCounting` be ignored in the TOC? I think the consensus from #3679 is "yes"
image
- What should happen to pages with both `.skipCounting` and `.resetCounting`?

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -1,41 +1,65 @@
...
+	const walkPages = ()=>{
+		let current = 0;
+		let skip = 0;
+		let reset = 0;
+		const pages = iframeDocument.querySelectorAll('.page');
+		_.each(pages, (page)=>{
+			current++;
+			if(page.querySelectorAll('.skipCounting').length > 0) {
+				skip += 1;
+			} else if(page.querySelectorAll('.resetCounting').length > 0) {
```
calculuschild, line 24: I don't think this accounts for the case where a page has both `.resetCounting` and `.skipCounting`. Might need to swap the order of the if statements here or remove the else? Not thinking straight what that logic needs to be right now.

You can see in a test document that the TOC will skip the reset and get out of sync.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -1,41 +1,65 @@
...
+	const pageMap = [];
+
+	const walkPages = ()=>{
+		let current = 0;
+		let skip = 0;
+		let reset = 0;
+		const pages = iframeDocument.querySelectorAll('.page');
+		_.each(pages, (page)=>{
+			current++;
+			if(page.querySelectorAll('.skipCounting').length > 0) {
```
calculuschild, line 22: Can probably just use `querySelector` for locating the CSS classes. Should be simpler and faster because we only need to know if one exists.

calculuschild (changes requested):

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -1,41 +1,65 @@
...
-
-	const iframe = document.getElementById('BrewRenderer');
-	const iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
-	const headings = iframeDocument.querySelectorAll('h1, h2, h3, h4, h5, h6');
-	const headerDepth = ['H1', 'H2', 'H3', 'H4', 'H5', 'H6'];
+	walkPages();
 
 	_.each(headings, (heading)=>{
+		const pageAnchor = heading.closest('.page').id;
 		const onPage = parseInt(heading.closest('.page').id?.replace(/^p/, ''));
```
calculuschild, line 58: So `pageAnchor` is always identical to `onPage` except a `p` is in front of the number, right?.

I wonder if we can make the logic easier to parse if we just use `onPage` throughout (they are the same number), and then just add the `p` back at the end when emitting the markdown.

calculuschild:
@Gazook89 Can you confirm this fixes https://github.com/naturalcrit/homebrewery/issues/2826 to your expectation?

themes/V3/5ePHB/snippets/tableOfContents.gen.js
```diff
@@ -1,41 +1,65 @@
...
+	const iframe = document.getElementById('BrewRenderer');
+	const iframeDocument = iframe.contentDocument || iframe.contentWindow.document;
+	const headings = iframeDocument.querySelectorAll('h1, h2, h3, h4, h5, h6');
+	const headerDepth = ['H1', 'H2', 'H3', 'H4', 'H5', 'H6'];
+
+	const res = [];
+
+	const pageMap = [];
+
+	const walkPages = ()=>{
```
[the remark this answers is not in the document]
dbolack-ab: Too much time in other scripting languages

themes/V3/5ePHB/snippets/tableOfContents.gen.js
dbolack-ab: Yeah, I did not think about that scenario.

themes/V3/5ePHB/snippets/tableOfContents.gen.js
dbolack-ab: It's not. But it's close enough I can fake it

calculuschild:
Something strange is now happening; If you generate a ToC and then delete it, and generate it again, each new ToC will include the contents of the previous ToC. It just gets longer and longer 🤣 

After 3 times:

image

dbolack-ab:
> Something strange is now happening; If you generate a ToC and then delete it, and generate it again, each new ToC will include the contents of the previous ToC. It just gets longer and longer 🤣

When I pulled the child functions out to global I pulled a couple of arrays out without verifying they needed to be.
This is now fixed.

calculuschild:
Perfect. Gonna just do one more check over and then merge.

calculuschild:
Seems to be working. Thank! 💯 🎉 ⭐

PR #3643, opened 2024-08-16
Change order of operations in mask img transform

### LINKED ISSUE #3636: Using scaleX or scaleY parameters causes skewing in rotated masks
### Your idea:

Consider the following:

```
{{imageMaskEdge5,--offset:0%,--rotation:235,--scaleX:1.0
  ![](https://i.imgur.com/GZfjDWV.png){height:100%}
}}
```

which results in the following:
image

Adjusting the `scaleX` parameter should increase the horizontal width of the mask:

```
{{imageMaskEdge5,--offset:0%,--rotation:235,--scaleX:1.5
  ![](https://i.imgur.com/GZfjDWV.png){height:100%}
}}
```

Instead, we see that the image inside the mask is skewed:
image

Altering `scaleY` results in a similar skewed result:
image

I believe that the issue stems from this block of styling:

```
.page [class*='imageMask']>p:has(img) {
    position: absolute;
    bottom: 50%;
    left: 50%;
    width: 50%;
    height: 50%;
    transform: translateX(-50%) translateY(50%) rotate(calc(-1deg * var(--rotation))) scaleX(calc(1 / var(--scaleX))) scaleY(calc(1 / var(--scaleY)));
}
```

and in particular, this line: `transform: translateX(-50%) translateY(50%) rotate(calc(-1deg * var(--rotation))) scaleX(calc(1 / var(--scaleX))) scaleY(calc(1 / var(--scaleY)));`.
Modifying this line to instead read `transform: translateX(-50%) translateY(50%) scaleX(calc(1 / var(--scaleX))) scaleY(calc(1 / var(--scaleY))) rotate(calc(-1deg * var(--rotation)));` results in the expected output, regardless of modifications to `scaleX` or `scaleY`:
image

G-Ambatte:
This PR resolves #3636.

This PR changes the order of operations in the image transform, correctly reversing the parent transformation when both rotation and scaling is applied.


calculuschild:
Very simple fix. Thanks!

PR #3644, opened 2024-08-17
Prevent styling of lines inside folded sections

### LINKED ISSUE #3604: Emojis don't like to be hidden
| Operating System | Browser | Renderer |
|:----------------:|:-------:|:--------:|
|      Windows     | Chrome  |    v3    |

### Description of the issue

Any page that uses the new emoji syntax will not stay folded when using the gutter arrows to fold. When using the Fold All editor tool or keyboard shortcut, the pages fold correctly, but unfolding any individual page will also unfold any page containing the new emoji syntax.

**Reproduction steps:**

1. Put the source code provided into a new brew
2. Use the Fold All function (`CMD/CTRL+[`) or click the Fold All toolbar icon.
3. Click the gutter arrow to unfold page 1.
**RESULT:** Page 1 AND Page 2 unfolds.
**EXPECTATION:** Only page 1 unfolds.

4. Repeat step 2.
5. Click the gutter arrow to unfold page 3.
**RESULT:** ONLY Page 2 unfolds.
**EXPECTATION:** Only page 3 unfolds.

### Code

```gfm
\page

:df_d2:

\page
```

G-Ambatte:
This PR resolves #3604.

This PR adds information about folded lines of the source, and prevents any styling from being applied to those lines.

REPRO BREW CODE:
```
\page

:df_d2:

\page
```


5e-Cleric:
Right now there is a nifty error:

### Expected behavior of `foldAll`:

image

### Actual behaviour:

image


This issue is not caused by this PR, but what the hell?

The emoji stuff seems to be solved.

G-Ambatte:
CURRENT PROD:

Source;
image

FoldAll:
image

DEPLOYMENT:

FoldAll:

5e-Cleric (approved):
Does work as expected, and changes are relatively small

calculuschild:
Before merging.... do we know why emojis specifically are causing the issue, and not any of the many other custom styling we use? In case there is a more appropriate solution of just fixing whatever the emojis are doing wrong.

G-Ambatte:
> Before merging.... do we know why emojis specifically are causing the issue, and not any of the many other custom styling we use? It's probably fine to just exclude folded regions from being processed like this PR does, but I would want to know what the emojis are doing wrong so we aren't accidentally burying a deeper issue that will come back and bite us.
> 
> I _suspect_ it's something with this section inside the emojis?
> 
> ```
> // Iterate over conflicting marks and clear them
> var marks = codeMirror.findMarks(startPos, endPos);
> marks.forEach(function(marker) {
>   marker.clear();
> });
> ```

Confirmed, I've just confirmed that this code also works to stop the emoji syntax unfolding folded pages:

```
// Iterate over conflicting marks and clear them
var marks = codeMirror.findMarks(startPos, endPos);
marks.forEach(function(marker) {
    if(!marker.__isFold) marker.clear();
});
```

calculuschild:
Ok, thanks for confirming! I just added your tested change to the emojis with a small comment. Merging!

Issue #956, opened 2020-05-27
[FEATURE]: Support for Static Page generation


Deastrom:
## Description

I would love to be able to maintain a campaign site for my adventures and love the formatting you are maintaining here.  When I found this tool I was looking for ways to do static page generation and hosting on github and/or gitlab.  Your markdown support and pdf generation is amazing and is exactly what I'm looking for.

Could you add support for generating static sites/pages for such campaign sites?  i.e. Hugo or Jekyll

paige404:
Being able to generate that static page from a git repository would have a lot of potential for community development. 

For instance, if I have different chapters of my homebrewery spread across markdown files in a git repository, it would allow people to submit pull requests to suggest edits, add polish, what have you. Having an option to use homebrewery as a build tool to generate the resulting page/pdf would be _very_ powerful.

paige404:
This may actually be similar functionality to #802

calculuschild:
Closing, as this is covered by two other issues.

- Downloadable HTML + CSS bundle: #661 
- Auto-building from CLI: #802

Issue #661, opened 2018-04-16
Embed Instructions and Generated Markup


netpoetica:
## Description

@stolksdorf Firstly, just FYI, this tool is amazeballs - great work.

In addition to [this single CSS file](https://github.com/stolksdorf/homebrewery/issues/10), it would be great if there was an "Embed" button on Homebrewery that gave a popup including:

- instructions to download that CSS file
- a style tag w/ that file and instructions to put it in your pages head
- an HTML version of what was created that can easily be copied to clipboard

What do you think?

The use case is, an author wants to host the content on their own site but doesn't want it to be a PDF. Very likely, in my case, I would still generate a PDF w/ a download like (and usually a link to Homebrewery next to it, in case people want the errata), but it would be great to show something inline on my own page.

I understand that having the CSS file is enough, but it would be useful to add a quick "EMBED </>" link to the Homebrewery content so the user could host it themselves.

ericscheid:
See also #956

5e-Cleric:
Would it be agreeable to give an option to download as a zip file containing the HTML and CSS files? or embedding the styles into the HTML then export as that? I can work on this feature but we should add this as a toggle in accountpage, or a pick a brew to download like this, as i don't think this is going to be a popular feature.

dbolack-ab:
This and the related issues should be closed and reopened with a cosolidated feature plan.

5e-Cleric:
Agreed, we should close this and put an issue with all the desired output formats or really consider if this is something we want, which i think not, honestly

PR #3647, opened 2024-08-19
Html endpoint

dbolack-ab:
A possible solution for the HTML-only output using the existing sharePage logic rather than restoring the old printPage.

Solves #956
Possibly solves #661

- [ ] Test on phones to see if solves margin-*: auto readability issues, sort out how to remove margin styling
- [ ] Add to menu bar
- [ ] Discuss if a better path name  ( embed? export? ) should be used.
- [ ] Verify works for embedding.

test using `.../html/shareId`


calculuschild:
This puts new stuff into the `sharePage`, but then ignores almost all of the `sharePage` logic.

I think this might be easier to maintain and make more sense just moving the new portions to a separate dedicated page, and just display the `brewRenderer` without having to work around or hide everything else. It shouldn't be as complex as the old print page (doesn't need to deal with loading values from `/new`, etc)

calculuschild:
I don't think this solves #661 though. My understanding from reading that issue is that they want a bundle of CSS + HTML they can copy and host on their own website, so more a download link than an HTML view. I would want to discourage users embedding our site anyway; we are not a CDN to host files for other people's web pages.

Issue #1469, opened 2021-07-28
Colon "vertical spacing" syntax does not appear in <pre> blocks


Gazook89:
Was writing a guide and tried to show the new colon syntax for vertical spacing using an example in a `<pre>` code block, but the colon is instead rendered as `<div class='blank'></div>`.

<img width="1208" alt="image" src="https://user-images.githubusercontent.com/58999374/127264018-a536ecaa-e9f4-44c0-a46f-0d46928c30f3.png">

This is in the staging site for v3.

calculuschild:
Ah. Ok. The way I implemented it is probably not a good way. I literally just search the whole document for any `:` alone on a line and then replace it with the div. What I need to do is make it into an actual Marked.js extension so it works properly inside of code blocks, etc. I will work on this shortly.

Gazook89:
I'm not sure if it's exactly the same issue, but there is another weird thing about pre blocks.  The below image might have some code that is hard to understand at first glance, but essentially the issue is that inside the pre block on the right, moustache divs are getting extra line breaks inside them.  So in my editor I have:
```
{{pageNumber
1
}}
```
But that renders in the pre block as:
```
{{pageNumber

1

}}
```

<img width="1209" alt="image" src="https://user-images.githubusercontent.com/58999374/127266748-5c2d24fd-c8d0-422c-8267-eecb875375a2.png">

Everything still *shows*, so it's not horrible, but just takes more space in the pre block.

calculuschild:
> I'm not sure if it's exactly the same issue, but there is another weird thing about pre blocks. 

That was a separate issue, and I have fixed that one now. Go check it out.

The original issue is still being worked on, although a temporary workaround would be to use `<pre></code>` and replace colons with `&#58;`

```
<pre><code>Here is a colon
&#58;
Tada!
</code></pre>
```

should appear as 

```
Here is a colon
:
Tada!
```

PR #3648, opened 2024-08-19
Fix for the colons in codeblocks issue.

dbolack-ab:
This was the lightest-weight solution I could find without potentially breaking definition lists for the `:` in codeblocks concern.

It may be a better fix to promote this to a proper token, I am unsure.

- [x] Catches `:` series
- [x] Does not interfere with DF Lists
- [x] Does not translate inside of PRE blocks.

Solves #1469


calculuschild:
Rather than modifying the `code` renderer, this should be done by just adding a Marked extension that finds `:::` tokens. `code` tokens should already be smart enough to ignore Markdown syntax inside if that syntax is properly registered as a Marked extension.

dbolack-ab:
> Rather than modifying the `code` renderer, this should be done by just adding a Marked extension that finds `:::` tokens. `code` tokens should already be smart enough to ignore Markdown syntax inside if that syntax is properly registered as a Marked extension.

Hence my second paragraph. However, getting it to not battle it out with the definition lists has proven a larger challenge than expected.

dbolack-ab:
Found my problem with making it a token, however I do not know if we have any weird user cases this may break.

calculuschild:
I just fiddled around a bunch looking for more edge cases. Found a few and added them as tests, then made small tweaks to get those tests passing.

I can't seem to find any more undesired behaviors so I think this is good to merge.

PR #3654, opened 2024-08-20
Add catch-all for invalid paths

### LINKED ISSUE #3629: Non-used URLs open bugged new page

calculuschild:
Fixes #3629 

`res.route` is the currently-matched route. If nothing has been matched by this point (route = undefined), we have an invalid route. In that case we need to redirect to the homepage at `/`. Eventually this could be a dedicated 404 page, but that can come later.

Would appreciate a second look just to verify this isn't breaking our normal routes, including API requests (saving/loading/etc.)

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -420,6 +420,14 @@ if(isLocalEnvironment){
 	});
 }
 
+// Catch-all route for invalid routes
+app.use((req, res, next) => {
+	if (!req.route)
+		return res.redirect('/');
+
+	return next();
+});
+
 //Render the page
 const templateFn = require('./../client/template.js');
 const renderPage = async (req, res)=>{
--- a/server/app.js
+++ b/server/app.js
@@ -14,6 +14,7 @@ const GoogleActions = require('./googleActions.js');
 const serveCompressedStaticAssets = require('./static-assets.mv.js');
 const sanitizeFilename = require('sanitize-filename');
 const asyncHandler = require('express-async-handler');
+const templateFn = require('./../client/template.js');
 
 const { DEFAULT_BREW } = require('./brewDefaults.js');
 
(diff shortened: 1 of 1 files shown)


server/app.js
```diff
@@ -420,6 +420,14 @@ if(isLocalEnvironment){
...
 }
 
+// Catch-all route for invalid routes
+app.use((req, res, next) => {
+	if (!req.route)
+		return res.redirect('/');
+
+	return next();
+});
+
```
5e-Cleric, line 430: Can this be merged into a check in the other function in line 462?

5e-Cleric:
In terms of functionality, and routes, everything seems to be working correctly, tested all pages, and saving routes, deleting, changing brew from drive to hb and back... No issue seen yet.

server/app.js
calculuschild: The rerouting needs to happen before rendering the page; otherwise we spend resources rendering a blank page that never gets sent.

5e-Cleric: Right, so:

```
app.use(asyncHandler(async (req, res, next)=>{
	if (!req.route)
		return res.redirect('/');

        const page = await renderPage(req, res);
	if(!page) return;
	res.send(page);
});
```

Or am i missing something? ( i haven't tested if this would work)

calculuschild: Rendering the page happens starting at line 432. Rerouting needs to happen before that. Line 462 is too late.

calculuschild: Ah wait, 432 is just defining the render function, but it doesn't get called until 462. I suppose we could merge them, but I would then want to move the `renderPage()` definition down. I.e., I want all "route/reroute" handling to be together without `renderPage()` definition in between

PR #3660, opened 2024-08-22
Image wrapping

### LINKED ISSUE #380: Wrapping text around an image

calculuschild:
Fixes #380, supersedes #3126

Deployed at https://homebrewery-pr-3660.herokuapp.com/

Adds `.wrapLeft` and `.wrapRight` CSS classes that can be applied to any image to cause image wrapping. This is done modifying the Image renderer of Marked, to emit the image `src` attribute as a CSS variable via `style="--HB_src: url(src)"` so the URL can be accessed from our stylesheets  (HTML attributes are not normally accessible from CSS).

Note, floating, and hence wrapping does not work for images with `position:absolute`. Not sure if we should have these classes automatically set `position:unset` at risk of confusing users because the image moves around, or not, at risk of confusing users because wrapping isn't working.

Eventually, if browsers implement `attr()`, we could undo the variable exposure workaround and just access `src` directly; users would not see any change in their brews.

image

I have verified all of this working myself, but want at least one more eye to confirm

#### Reviewers, please copy this checklist and mark what you have verified:
*feel free to add sub-bullets for specific cases; I will copy them up here*
- [ ] New features work as expected
  - [x] `wrapLeft` wraps text on left side
  - [x] `wrapRight` wraps text on right side
  - [ ] Wrapping works when injected into images
  - [ ] Wrapping works when set via the `style` tab (e.g., use `--HB_src` to make *all* images wrap)
  - [ ] Images can be positioned via `margin` on the image
  - [x] New snippets are functional
  - [x] Snippet icons are clear
- [x] Old features still work
  - [x] Not breaking existing curly injection on images
  - [x] Not breaking normal image display
- [ ] Edge case / stress testing
- [x] Opportunities for simplification / refactoring
- [x] Legibility and appropriate comments

```
- [ ] New features work as expected
  - [ ] `wrapLeft` wraps text on left side
  - [ ] `wrapRight` wraps text on right side
  - [ ] Works for injected images
  - [ ] Works for images via `style` tab
  - [ ] Images can be positioned via `margin`
  - [ ] New snippets are functional
  - [ ] Snippet icons are clear
- [ ] Old features still work
  - [ ] Not breaking existing curly injection on images
  - [ ] Not breaking normal image display
- [ ] Edge case / stress testing
- [ ] Opportunities for simplification / refactoring
- [ ] Legibility and appropriate comments
```
</details>


Gazook89:
> Not sure if we should have these classes automatically set position:unset at risk of confusing users because the image moves around, or not, at risk of confusing users because wrapping isn't working.

Does it make sense to create a new snippet for each 'wrapRight' and 'wrapLeft', which drops in the related classname, but also drops in `position:relative` right after it?  That way we have 'example' snippets that show it working, but without as much subtle background surprises. 

That may be unnecessary snippet bloat, though.  And there is a part of me that thinks users at some point just sort of need to know some CSS.....(but the rest of me kind of cringes at the suggestion that our homebrew tool requires technical CSS knowledge).

calculuschild:
If a snippet, it might make sense to insert a full new image, rather than try to modify an existing one. Then we don't have to deal with whatever properties the user has set, and they can adapt it as needed.

However, this kind of "modifier" class would be a good candidate for widgets.

Note, there are also issues with `position:relative`. If the image has any positioning coordinates (`left:100px`), they will shift the image but not the text flow, so it will appear that the text is wrapping around a phantom space separate from the graphic.

Gazook89:
I think `position: relative` is my preference:

```
![homebrew mug](http://i.imgur.com/hMna6G0.png) {wrapRight,top:50px,right:130px,width:280px,position:relative}
```
yields:
image

Without relative positioning, the image will only ever be squarely inside the column.


-----

I also liked @dbolack-ab syntax which included an easy `shape-image-threshold` for setting the threshold of what level of alpha (or luminance? not sure what it is based) creates the shape.  For instance, the difference between `shape-image-threshold: 0.1` vs `0.5`:

imageimage

And the syntax included an easy option for setting the `shape-margin`.  `0px` and `30px`:

imageimage

So without the neat and tidy syntax, i think these are further arguments for custom snippets for Left and Right, making these additional properties visible.

calculuschild:
> I think `position: relative` is my preference:
> 
> ```
> homebrew mug {wrapRight,top:50px,right:130px,width:280px,position:relative}
> ```
> 
> yields: image
> 
> Without relative positioning, the image will only ever be squarely inside the column.

What browser are you using? On my end I get this (the "phantom wrapping" I mentioned)

image

Gazook89:
Ah, Firefox.  So firefox does this *better* than chrome (IMHO).  

Well shoot.  So an image can only be wrapped if the entire width of the image fits within the column.  that's a big bummer.

calculuschild:
You can use negative margins to move things around, but yeah. This would seem to be the case.

Gazook89:
Negative margins is a good idea, and seems to work.  If it hadn't worked, I was pretty much ready to throw up my hands and say "whats the point of text wrapping at all". 

Okay.  So the `position:unset` seems fairly settled to me.  I do support new snippets that showcase `shape-margin`, `shape-image-threshold` and possibly now even `margin-left: -200px` or whatever.

calculuschild:
Are those features critical to a snippet, or would they fit better as a note in the FAQ, for example? Or as a comment under the snippet?

Mainly I don't want to bloat the snippet and make it look scary for the majority of users who probably won't touch it.

Gazook89:
I think that having a small amount of default `shape-margin` is ideal, whether that is included in the inline styling via the snippet or baked into the class.  `shape-image-threshold` isn't as important, leaving it at the default `0` is likely what most people expect.  `margin-[left/right]` is highly context dependent, so that doesn't need to be included--- but despite being context dependent, I think it'll be commonly wanted.  Maybe an html comment that says "Image can be shifted relative to column with `margin-left or -right`"

calculuschild:
> I think that having a small amount of default `shape-margin` is ideal, whether that is included in the inline styling via the snippet or baked into the class. `shape-image-threshold` isn't as important, leaving it at the default `0` is likely what most people expect. `margin-[left/right]` is highly context dependent, so that doesn't need to be included--- but despite being context dependent, I think it'll be commonly wanted. Maybe an html comment that says "Image can be shifted relative to column with `margin-left or -right`"

Ok. We do have a default `shape-margin` in the class. I will leave out `shape-image-threshold`, and I will add an HTML comment for margins.

5e-Cleric:
## Full test by 5e-Cleric

- [ ] New features work as expected
  - [ ] `wrapLeft` wraps text on left side
  - [ ] `wrapRight` wraps text on right side
  - [ ] Works for injected images
  - [ ] Works for images via `style` tab
  - [ ] Images can be positioned via `margin`
  - [ ] New snippets are functional
  - [ ] Snippet icons are clear
- [ ] Old features still work
  - [ ] Not breaking existing curly injection on images
  - [ ] Not breaking normal image display
- [ ] Edge case / stress testing
- [ ] Opportunities for simplification / refactoring
- [ ] Legibility and appropriate comments

5e-Cleric (approved):
Don't see much space for optimization code wise, and the code is pretty straightforward.

Still needs further testing but this is okay.

calculuschild:
> Works for injected images (what does this mean?)

Text wrapping works when injected into an image. `{wrapLeft}`

> Works for images via style tab (what does this mean?)
 
 (updated this one above) Text wrapping using the `--HB_src` works when added via the style tab. e.g.:
 
 ```
img {
		shape-outside : var(--HB_src);
		float         : left;
		shape-margin  : 0.2cm;
}
 ```
 
> Images can be positioned via margin sometimes only the text moves

Do you have an example where adding a margin to the image moves the text instead?

5e-Cleric:
> Do you have an example where adding a margin to the image moves the text instead?

Well, this is my fault, but i think users will make this mistake too, i swapped one wrap for the other, but did not change the margin side, and tried to make the margin do something.

calculuschild:
> > Do you have an example where adding a margin to the image moves the text instead?
> 
> Well, this is my fault, but i think users will make this mistake too, i swapped one wrap for the other, but did not change the margin side, and tried to make the margin do something.

@5e-Cleric Since then, we have added the appropriate left/right margin to the snippet by default. Does that help any?

5e-Cleric:
Does work better out of the box, i think

PR #3665, opened 2024-08-24
VIEWER TOOLS PART 2 - Zoom to Fit, Fit to Width

Gazook89:
*This PR builds on the work done in #3499 (it is branched from it).  Either that PR can be merged first and then this one (if approved) can be merged later....or, this can be merged by itself and everything in 3499 will be included.*

This PR adds  "Zoom to Fit" and a "Fit to Width" buttons to the viewer toolbar in the Preview pane.  It looks like this:

https://github.com/user-attachments/assets/e9d797b0-6080-49f5-82db-ec4602e9ef39

Zoom to Fit:   gets the largest single dimension (of either width or height) from all pages, and then adjusts the zoom so that it will entirely within the viewport.  It *does not* fit the "current page" to the viewport, a choice i made.

Fit to Width: gets the largest *width* dimension of all the pages and sets the zoom so that it fills the viewport based on that.  Same deal regarding "current page" as above.


client/icons/icon-components/zoomIcons.jsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
```
calculuschild, line 1: Could we just put the icons as plain SVGs along with all our other custom icons? I don't think we need to devote a whole separate component for icons when we already have icons being used in plenty of other spaces. Titles, etc. can be just applied on the button itself instead of embedded in the SVG.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -37,10 +39,67 @@ const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
 		setPageNum(pageNumber);
 	};
 
+
+	const calculateZoom = (mode)=>{
```
calculuschild, line 43: I think this should be renamed to `handleZoomFit` or `handleZoomFitButtons` or something.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -20,7 +22,7 @@ const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
 	}, [currentPage]);
 
 	const handleZoomButton = (delta)=>{
```
calculuschild, line 24: So when I proposed naming this `handleZoomIncrement` before, I assumed this would be just for the +/- buttons. Now it is being used also for the "fit" buttons, which means `calculateZoom` has to awkwardly calculate a delta instead of being able to more straightforwardly just set the zoom level it calculated, so I propose one of two refactors:

1) Have `calculateZoom` just `setZoomLevel(desiredZoom - margin)` at the end, and rename this to `handleZoomIncrement` to distinguish it from the `fit` buttons.
2) Rename this back to `handleZoomChange`, but have the input just be the zoom level you want to set instead of a delta, then update the +/- buttons so they send `zoomLevel +- 20`.

2 might be better so you can still have a centralized place to perform rounding and clamping.

client/icons/icon-components/zoomIcons.jsx
Gazook89: I'll mess with it again.  I was having trouble with getting the color of the icons to be adaptable to whatever color their "context" was (changing the `color` property on a container div should also change the color of the icon).  As a component, I can use `currentColor`.  

For example in the snippet bar, with some custom icons, you can see that the color of the icon remains 'black' even if i specifically set those icons to `color: red`:

image

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -37,10 +39,67 @@ const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
...
+		let desiredZoom = 0;
+
+		if(mode == 'fill'){
+			// find widest page, in case pages are different widths, so that the zoom is adapted to not cut the widest page off screen.
+			let widestPage = 0;
+			[...pages].forEach((page)=>{
+				const width = page.offsetWidth;
+				if(width > widestPage)
+					widestPage = width;
+			});
```
calculuschild, line 58: Seems like an opportunity for a `reduce`. What about something like this?

```suggestion
			const widestPage = [...pages].reduce((maxWidth, page) => Math.max(maxWidth, page.offsetWidth), 0);
```

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -37,10 +39,67 @@ const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
...
+
+			desiredZoom = (iframeWidth / widestPage) * 100;
+
+		} else if(mode == 'fit'){
+			// find the page with the largest single dim (height or width) so that zoom can be adapted to fit it.
+			let maxDimRatio = 0;
```
calculuschild, line 64: should this be `minDimRatio`?

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -37,10 +39,67 @@ const ToolBar = ({ onZoomChange, currentPage, onPageChange, totalPages })=>{
...
+
+			[...pages].forEach((page)=>{
+				const widthRatio = iframeWidth / page.offsetWidth;
+				const heightRatio = iframeHeight / page.offsetHeight;
+				const dimensionRatio = Math.min(widthRatio, heightRatio);
+				if(dimensionRatio < maxDimRatio || maxDimRatio == 0){
+					maxDimRatio = dimensionRatio;
+				}
+			});
```
calculuschild, line 73: The minimizing logic simplifies a lot if we start with `Infinity`

```suggestion
			let minDimRatio = Infinity;

			[...pages].forEach((page)=>{
				const widthRatio = iframeWidth / page.offsetWidth;
				const heightRatio = iframeHeight / page.offsetHeight;
				minDimRatio = Math.min(widthRatio, heightRatio, minDimRatio);
			});
```

or

```suggestion
			const minDimRatio = [...pages].reduce((minRatio, page) => 
				Math.min(minRatio, iframeWidth / page.offsetWidth, iframeHeight / page.offsetHeight), Infinity
			);

```

client/icons/icon-components/zoomIcons.jsx
calculuschild: Check out this reply. Apparently this is how FontAwesome gets its SVGs to use the `color`. We should be able to do this as well:

https://stackoverflow.com/a/65147574

Gazook89: Yeah, i think that is basically what I have done.  The method in that SO answer is to use `currentColor` in the SVG itself...but it requires you to inline the SVG in the HTML (you can't use ``), so rather than a big SVG string in the middle of the file I am dropping it into a component as a separate file (using SVGR to write the component, as noted in the answer comments).

calculuschild: Ok, this is gonna sound a little crazy. What if we make our custom icons masks, instead of displaying it as the `content` property. For example, replace the `customIcons.less` with this:

```
.fac {
  display          : inline-block;
  background-color : currentColor;
  mask-size        : contain;
  mask-repeat      : no-repeat;
  mask-position    : center;
}
.position-top-left {
  mask-image: url('../icons/position-top-left.svg');
}
.position-top-right {
  mask-image: url('../icons/position-top-right.svg');
}
.position-bottom-left {
  mask-image: url('../icons/position-bottom-left.svg');
}
.position-bottom-right {
  mask-image: url('../icons/position-bottom-right.svg');
}
.position-top {
  mask-image: url('../icons/position-top.svg');
}
.position-right {
  mask-image: url('../icons/position-right.svg');
}
.position-bottom {
  mask-image: url('../icons/position-bottom.svg');
}
.position-left {
  mask-image: url('../icons/position-left.svg');
}
.mask-edge {
  mask-image: url('../icons/mask-edge.svg');
}
.mask-corner {
  mask-image: url('../icons/mask-corner.svg');
}
.mask-center {
  mask-image: url('../icons/mask-center.svg');
}
.book-front-cover {
	mask-image: url('../icons/book-front-cover.svg');
}
.book-back-cover {
  mask-image: url('../icons/book-back-cover.svg');
}
.book-inside-cover {
  mask-image: url('../icons/book-inside-cover.svg');
}
.book-part-cover {
  mask-image: url('../icons/book-part-cover.svg');
}
.davek {
  mask-image: url('../icons/Davek.svg');
}
.rellanic {
  mask-image: url('../icons/Rellanic.svg');
}
.iokharic {
  mask-image: url('../icons/Iokharic.svg');
}
```

And voila! All our custom icons can be recolored:

image

Gazook89: Okay, done and done.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: I'm not sure I agree-- when it is called, it is `onClick={()=>handleZoomButton(calculateZoom('fit'))}`, so it seems weird there were be a handler inside a handler.  This function isn't changing anything on the page or specifically reacting to a button being pressed, it's just doing a calculation which is then used as an input to a handler function.  And it takes 'fit' or 'fill' as it's arguments, so it's pretty much right there.  

Maybe `calculateChange`?  

I'm going to let this simmer a bit, and maybe when addressing the other comments something better will come up.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: wow that maxBy is good.  done.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: i changed it to calculateChange, in combo with addressing another comment

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: done and done

PR #3671, opened 2024-08-27
Small Patch for Toolbar Positioning

Gazook89:
This PR is a two lines change to the LESS files for toolbar and brewRenderer.  It fixes the issue with the scrollbar going behind the toolbar, and it simplifies things in terms of flow (no absolute positioning).  

I believe absolute positioning was originally used because we were using JS to set the iframe height, (or some other element), and we stopped doing that.


calculuschild:
Thanks @Gazook89 !

client/homebrew/brewRenderer/toolBar/toolBar.less
```diff
@@ -1,5 +1,5 @@
 .toolBar {
-	position         : absolute;
+	position         : relative;
 	z-index          : 1;
 	box-sizing       : border-box;
 	display          : flex;
```
5e-Cleric, line 7: I believe the declaration of position can be removed entirely.

PR #3679, opened 2024-08-29
Add Page number alteration snippets

dbolack-ab:
This creates Text Editor Menu Snippets for the new SkipCounting and ResetCounting classes used by the auto page numbering system.

 - [x] The  Icons and labels might not be the best. Suggestions welcome
 - [x] It might be wise to move this to a Subsnippet menu for all the page numbering snippets. Discuss?
 - [x] Presently, the ToC snippet will generate entries on pages that have the skipCounting class. Should those entries be discarded as they will be inaccurate or should we consider that a user error?

This PR is a child of #3639.


5e-Cleric:
> The Icons and labels might not be the best. Suggestions welcome
Could we get a deployment for this?

> It might be wise to move this to a Subsnippet menu for all the page numbering snippets. Discuss?

I think i agree

Presently, the ToC snippet will generate entries on pages that have the skipCounting class. Should those entries be discarded as they will be inaccurate or should we consider that a user error?

ToC should be as accurate as we can, yes, jump entries in pages without numbering.

5e-Cleric:
blocked merging until #3639 , but we can review

calculuschild:
@dbolack-ab Thank you for splitting this up

- For icons, I might suggest `rotate-left` for the RESET, and `xmark` for SKIP
- I agree, we have multiple numbering-related snippets now. Let's combine them into a group
- `.skipCounting` entries should not be included in the TOC in my opinion (but this should be handled in #3639)

5e-Cleric (approved):
Functionality works, good icons, this looks good

Issue #2198, opened 2022-05-29
Deduplicate pagination functions in Editor and BrewRenderer


G-Ambatte:
As per [this comment](https://github.com/naturalcrit/homebrewery/pull/2162#discussion_r869850180) on PR #2162, with the merging of that PR, the pagination function in the BrewRenderer will be duplicated in the Editor.

It should be possible to use `React.cloneElement` to add props to the `this.props.children` (as identified in [this comment](https://github.com/naturalcrit/homebrewery/pull/2162#discussion_r884207224)) being handled in `SplitPane.jsx` to pass a unified pagination function as a prop to both child elements.

5e-Cleric:
I am not sure if the current toolbar even gives a third page navigation function. Should check.

dbolack-ab:
Is this still an issue with the Intersection Observer?

calculuschild:
I think this was handled somewhere around #3729 or #3484

PR #3682, opened 2024-08-29
Use IntersectionObserver to detect current page

G-Ambatte:
This PR uses the IntersectionObserver to detect which page elements are in the viewport and thus visible to the end user. This data is updated as the user scrolls, and is passed to a new function `updatePageArray` in `helpers.js`.

This PR also adds two further new functions to `helpers.js`:
-  `getPageNumber` : returns an integer, the average value of the content of the `pageArray` which should be the most central page in the range of visible pages
- `getVisiblePageRange` : returns a string, if the array length is 3 or less, joined by commas (e.g. `1, 2, 3`) and for lengths greater than 3, the first and last page separated by a dash (e.g. `1 - 5`).

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -13,7 +13,7 @@ const RenderWarnings = require('homebrewery/renderWarnings/renderWarnings.jsx');
 const NotificationPopup = require('./notificationPopup/notificationPopup.jsx');
 const Frame = require('react-frame-component').default;
 const dedent = require('dedent-tabs').default;
-const { printCurrentBrew } = require('../../../shared/helpers.js');
+const { printCurrentBrew, updatePageArray, getPageNumber } = require('../../../shared/helpers.js');
 
 const DOMPurify = require('dompurify');
 const purifyConfig = { FORCE_BODY: true, SANITIZE_DOM: false };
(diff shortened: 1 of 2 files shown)


Gazook89:
What problem is this fixing?

calculuschild:
@Gazook89 I believe this is working toward https://github.com/naturalcrit/homebrewery/issues/2198

5e-Cleric:
If this works, it should solve an issue at #3694

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -35,8 +35,28 @@ const BrewPage = (props)=>{
 		index    : 0,
 		...props
 	};
+
+	const pageRef = useRef();
+
+	const rootMargin = '-50% 0px';
```
calculuschild, line 41: Any reason for using `rootMargin` intead of `threshold`?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -35,8 +35,28 @@ const BrewPage = (props)=>{
...
+
+	useEffect(()=>{
+		const observer = new IntersectionObserver(
+			([entry])=>{
+				const pageNo = parseInt(entry.target.getAttribute('id').slice(1));
```
calculuschild, line 46: I wonder if we can just use the component's `index` prop directly instead of reading it from the DOM target?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -35,8 +35,28 @@ const BrewPage = (props)=>{
...
+				updatePageArray(entry.isIntersecting, pageNo);
+			},
+			{ rootMargin }
+		);
+		observer.observe(pageRef.current);
+		return ()=>{
+			if(pageRef.current == null) return;
+			observer.unobserve(pageRef.current);
+		};
+	}, [pageRef.current, rootMargin]);
```
calculuschild, line 56: Neither of these values trigger a redraw when changed, so they won't have any effect here in the `useEffect` dependency array. The effect will just fire once when the page component is mounted and then never again. Might as well just have an empty array here, no? `[]`

shared/helpers.js
```diff
@@ -50,8 +50,42 @@ const fetchThemeBundle = async (obj, renderer, theme)=>{
 	}));
 };
 
+let pageArray = [];
```
calculuschild, line 53: What is the purpose of the `pageArray` over simply setting the current page directly when the `intersectionObserver` fires?

G-Ambatte: My thinking was that - particularly with the new BrewRenderer toolbar - that there can be multiple pages visible in the view port, so an array of page numbers is more appropriate than a single value.

G-Ambatte:
This should be superseded by PR #3845.

5e-Cleric:
> This should be superseded by PR #3845.

Any reason you didn't close it then?

PR #3688, opened 2024-08-29
Change BrewRenderer background in Heroku test deployments

G-Ambatte:
This PR modifies the BrewRenderer to change the default background when the `HEROKU_APP_NAME` variable exists. This variable only exists when created during the PR deployment process in Heroku apps, or when specifically set on a Development machine.

The main purpose of this PR is to create a visual indication that someone is working on a deployment, NOT the main website. This is primarily intended to be a Dev Experience QoL improvement, which is why I've marked it as a low priority change.


calculuschild:
What issue is this trying to solve?

5e-Cleric:
> What issue is this trying to solve?

This is trying to help us in the development, so we don't confuse local vs deployment vs live. Minor inconveniences.

dbolack-ab:
Can we get a deployment on this one to validate?

dbolack-ab (approved):
Looks good. Need to see the deployment.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -188,6 +188,12 @@ const BrewRenderer = (props)=>{
 		}));
 	};
 
+	const styleObject = { height: state.height };
```
5e-Cleric, line 191: height:100vh works fine here, i tested in another PR. This works too, but is less readable.

dbolack-ab (approved):
Looks good, deployment works great for the intended use.

calculuschild:
@G-Ambatte Can you address 5e-Cleric's open comment and then I can merge this.

G-Ambatte:
> @G-Ambatte Can you address 5e-Cleric's open comment and resolve conflicts, and then I can merge this.

I've implemented the suggested change... which led me to investigate if `state.height` was actually being used anywhere after the change. As best I can determine it was not, so I have also removed the now obsolete state and event listeners from BrewRenderer.

calculuschild (approved):

calculuschild:
Nice added cleanup. Merging!

PR #3689, opened 2024-08-29
Add Style Tab Snippets for Globally toggling additional header inclusion

dbolack-ab:
This add a style tab submenu with entries to add 
 - H3
 - H3, H4
 - H3, H4, H5
 - H3, H4, H5, H6

to the Table of contents when generated. 

The menu default if to add all headers.


5e-Cleric (approved):

calculuschild (commented):
Some observations / questions for discussion:

- The default is already to generate H1-H3. Do we need a separate button for that?
- This overrides any specifically-excluded elements, for example Monster Stat Blocks which are written to exclude the headers inside
  - If that is intended, it should apply to all headers. Monsters have H2 and H3, but this PR only targets H3 +.
- Would this fit better in the Blank snippets?

dbolack-ab:
> * The default is already to generate H1-H3. Do we need a separate button for that?

No, I suppose not.

> * This overrides any specifically-excluded elements, for example Monster Stat Blocks which are written to exclude the headers inside

  Does the monster block specificity not trump this? The code *should* be looking at the value of --TOC from the header Element when determining to include it or not.



> * Would this fit better in the Blank snippets?

Yes. It's in here at the moment because that is where the ToC snippet is at this moment.

calculuschild:
> No, I suppose not.

Ok, lets remove that snippet then.

> Does the monster block specificity not trump this? The code should be looking at the value of --TOC from the header Element when determining to include it or not.

Looks like it doesn't. Monster block rule is:

```
h4,h5,h6,.page:has(.frontCover),.page:has(.backCover),.page:has(.insideCover),.monster,.noToC,.toc {
    --TOC: exclude
}
```

And this PR has:

```
.page :where(h3, h4, h5, h6) {
    --TOC: include;
}
```

I believe the `.monster` and `.page:where` have the same specificity, so the contents of the STYLE tab being the last style applied, will override equal-level specificity.

> Yes. It's in here at the moment because that is where the ToC snippet is at this moment.

OK. We can move it later along with the TOC generator stuff.

dbolack-ab:
> Ok, lets remove that snippet then.

Done.

> Looks like it doesn't. Monster block rule is:

Okay, that tells me that the Monster block needs to be adjusted as it is ( just by a nose ) an override of the default behavior. I will see if I can figure it out but that may be a scooch beyond my CSS skills/

ericscheid:
> Okay, that tells me that the Monster block needs to be adjusted as it is ( just by a nose ) an override of the default behavior. I will see if I can figure it out but that may be a scooch beyond my CSS skills/

Change the `.monster` selector to `.page .monster` and it'd have higher specificity than `.page where(...)`.

calculuschild:
Also make sure any change still works with the ToCDepthHx snippets. Those ones *should* override the exclusions in a monster block.

dbolack-ab:
> Also make sure any change still works with the ToCDepthHx snippets. Those ones _should_ override the exclusions in a monster block.

I believe that the Depth snippets set the default document behavior, meaning that classes that contradict that rule should trump.

ericscheid:
> > Also make sure any change still works with the ToCDepthHx snippets. Those ones _should_ override the exclusions in a monster block.
> 
> I believe that the Depth snippets set the default document behavior, meaning that classes that contradict that rule should trump.

👍 

Agreed. The more specific rules override the default rules. Part of the reason for going this _interrogate-css_ strategy was so that by deferring precedence to CSS logic we could avoid having to code up all the logic ourselves.

calculuschild:
> Agreed. The more specific rules override the default rules. Part of the reason for going this interrogate-css strategy was so that by deferring precedence to CSS logic we could avoid having to code up all the logic ourselves.

Also agreed. If the Monster class by default excludes its headers from the ToC, and we are now explicitly adding `ToCDepthH3` onto that DIV, `ToCDepthH3` should now override the Monster default via CSS logic.

dbolack-ab:
> The `ToCDepthHx` snippets set the behavior for the one DIV they are applied to, not the full document. So their individual DIV should trump the default document behavior.

If that is the case, I have constructed the classes incorrectly and these are not Global Toggles. 

> 
> > Agreed. The more specific rules override the default rules. Part of the reason for going this interrogate-css strategy was so that by deferring precedence to CSS logic we could avoid having to code up all the logic ourselves.
> 
> Also agreed. If the Monster class by default excludes its headers from the ToC, and someone adds `ToCDepthH3` onto that DIV specifically to _include_ those values, `ToCDepthH3` should be the more specific rule to override the Monster default via CSS logic. This is in line with the current deployed behavior. I'm just saying we just don't want to break it with this PR, since this touches some of those CSS rules. If we raise the `monster` specificity, we should ensure the `ToCDepthHx` specificity is still high enough to be the "trump-all" setting it was designed to be.

calculuschild:
Ok. I *think* we are on the same page but just talking past each other.

Just to make it explicit, does this match what you intend the "trump" order to be?

1) Default page headers
2) Monster Block divs (specially excluded) 
3) ToCDepthHx divs (specially *included*)
4) Global toggles (this PR)

ericscheid:
> Ok. I _think_ we are on the same page but just talking past each other.
> 
> Just to make it explicit, does this match what you intend the "trump" order to be?
> 
> 1. Default ToC settings
> 2. Global Toggles
> 3. Monster Block divs (specially excluded, trumps global toggles)
> 4. ToCDepthHx divs (specially _included_, trumps the div it is on)

The precedence should be determined by the css rules (speificity, but also the cascade), so what are the use cases for those css rules? Then we can double-check we have the right rules.

For example:

```
// this h3 should appear in the ToC
{{monster,frame
{{ToCDepthH3
### Actions
...
}}
}}

// but this one shouldn't
{{ToCDepthH3
...
{{monster,frame
### Actions
}}
}}

// while with these two I have questions..
{{monster,frame,ToCDepthH3
...
}}
{{ToCDepthH3,monster,frame
...
}}


```

calculuschild:
> The precedence should be determined by the css rules

Again, this is already agreed. The question is what specificities do we want to set for each rule to achieve the behavior we want, because we are free to do that.

If we want `ToCDepthHx` to maintain its original and current behavior of "trump everything", then for all three examples you listed, the H3 should appear in the ToC, because it is designed to override all contained headers with `--TOC: include;`.

If we want to *change* that, to put `ToCDepthHx` on even specificity with the Monster stat block so that only cascading order makes the difference, then we can do so. But I would suggest moving that to a different PR so as to not muddy up the intent of this PR.

So, to be clear, my understanding of the use cases for the CSS rules:

|# | Rule  | Use Case |
|--|------|-------|
| 1 | Top-level Header Settings | Exclude H4-H6 from ToC |
| 2 | Global Toggles (this PR) | Add H4-H6 to ToC, as long as it is not specifically excluded elsewhere |
| 3 | Block-level Settings (ToC, Monster, Cover Pages) | Exclude internal H's from ToC regardless of the global toggles (Monster headers should still be excluded, ToC itself should still be excluded, etc. |
| 4 | `ToCDepthHx` | Override specific exclusion rules. E.g., force a Monster Block, Cover Page, the ToC itself to be included in the ToC |

Hence, these rules should be designed on a different specificity level such that 2 *always* overrides 1. 3 *always* overrides 2. 4 *always* overrides 3. Yes, this behavior is achieved by CSS (specificity). The more specific rules override the default rules. Then, by deferring precedence to this CSS logic we avoid having to code up all the logic ourselves.  

If two rules of the same specificity interact (e.g., `ToCDepthH3` and `ToCDepthH6` together), then the cascade will manage them. Easy peasy.

ericscheid:
> 2 always overrides 1. 3 always overrides 2. 4 always overrides 3

- Nesting `Do-This` div inside a `Dont-Do-This` div for rules of the same specificity should have the `Do-This` rule applied.

- Reversing the nesting to `Dont-Do-This` _inside_ a `Do-This` div for rules of the same specificity should have the `Dont-Do-This` rule applied. This is the _cascade_ part of `CSS`.

It would be weird to have the `Do-This` rule apply in both cases (because we finagled the specificity to achieve that).

calculuschild:
> This is the cascade part of CSS.

Yes, you seem to understand the fundamentals of CSS. So if we don't want them to do this, we _increase the specificity_ of the rule we want to trump everything. I.e. Rule 4 should have the highest specificity of all (perhaps with `!important`) so that no matter how it is nested, it trumps and overrides the contents it encloses. The desired behavior.

themes/V3/5ePHB/snippets.js
```diff
@@ -27,35 +27,166 @@ module.exports = [
...
+											}\n\n`
+							},
+							{
+								name : 'Enable H1-H6 all pages',
+								icon : 'fas fa-dice-six',
+								gen  : dedent`/* Add ToC include for H3, H4, H5, and H6 level Headers */
+											.page h3, .page h4, .page h5, .page h6 {
+												--TOC: include;
+											}\n\n`
+							},
```
calculuschild, line 188: We want to replace these with the `.tocGlobal` classes then, right? Or remove `.tocGlobal` classes and leave these? They are identical behavior.

dbolack-ab: replace.  That was sloppy of me.

calculuschild:
Alright. Thanks for the fixes. Merging this now. 💯 ⭐ 🎉

PR #3692, opened 2024-08-30
Style snippets, more page sizes, A3, A5, Card

5e-Cleric:
THe imgur images used as backgrounds come from https://imgur.com/a/IWQRo8a, i am unsure of the origin or license of these images, and therefore discussion is needed, or new, free backgrounds might be needed.


5e-Cleric:
> This is my album. The first image is just the normal HB background I believe. The rest are either adjustments of that, or made entirely by me (either scans, digitally drawn, or some combination). I’m okay with whatever permissive license is wanted.

Best case scenario, CCO or CC BY. Not sure, entirely up to you, people will add their background most times anyway.

calculuschild:
Two comments:

- When we link to imgur, those are always "example" images that we intend the user to change to their own link. If we are supplying the graphics ourselves, they should be hosted on our own server.

- Backgrounds belong within Themes, no? Snippets are meant more to provide functional example pieces of code, not as an artwork source.

5e-Cleric:
> Two comments:
> 
> * When we link to imgur, those are always "example" images that we intend the user to change to their own link. If we are supplying the graphics ourselves, they should be hosted on our own server.

Indeed, i did not add them as such in case we didn't have rights to them.

> * Backgrounds belong within Themes, no? Snippets are meant more to provide functional example pieces of code, not as an artwork source.

We should at least provide one extra background option so users can learn how to change the background.

The page sizes have been requested in multiple occasions, and we have space for them. The background is my idea to make users want to use alternative backgrounds.

calculuschild:
> We should at least provide one extra background option so users can learn how to change the background.

Then let's just make it a "learning example" snippet and have just one image.

Though it does bring up a question of where do we stop in providing this type of snippet that does not insert a component but just overrides the theme. I think it would be untenable to make a snippet to override *every* possible CSS rule in our themes (thinking back to the "let's make a hotkey shortcut for every possible snippet" phase). But if we start adding them now where is the line?

> The background is my idea to make users want to use alternative backgrounds.

In an attempt to move them away from the PHB theme? Again, that seems like the job of Themes.

> The page sizes have been requested in multiple occasion

The page sizes I agree with

5e-Cleric:
I will think about this, meanwhile, is it ok if i remove the background snippet and we merge the page sizes? Should i add more? like Legal, or A2? I dont think there is any need, but i guess i could add a card size one...

calculuschild:
Yes, we can make this just page sizes for now. Card size would be nice (like Magic cards or something. 2.5 x 3.5 inches)

5e-Cleric:
> Yes, we can make this just page sizes for now. Card size would be nice (like Magic cards or something. 2.5 x 3.5 inches)

Right, what is that in american, a 1,5 sardines x 5 9mm bullets?

Sorry, bad joke, don't like inches, will add this.

calculuschild:
You can put in mm if you want. I just happened to know the 2.5 x 3.5 off the top of my head.

calculuschild (dismissed):
Just need to :

- [x] Remove background image snippets
- [x] Add any other page sizes, in mm (cards?)

5e-Cleric:
for the card snippet also unset columns and declared font-size at 12px, smaller, which should be more appropiate.

themes/V3/Blank/snippets.js
```diff
@@ -437,11 +437,29 @@ module.exports = [
...
-					.page{
+					.page {
 						width  : 210mm;
 						height : 296.8mm;
 					}\n\n`
 			},
+			{
+                name: 'A3 Page Size',
+                icon: 'far fa-file',
+                gen: dedent`/* A3 Page Size */
```
calculuschild, line 448: Please check tabs vs spaces as well

themes/V3/Blank/snippets.js
```diff
@@ -453,6 +471,27 @@ module.exports = [
 						columns : unset;
 					}\n\n`
 			},
+			{
+				name: 'Card Page Size',
+				icon: 'far fa-file',
+				gen  : dedent`/* A4 Page Size */
```
calculuschild, line 477: This comment should say "card size" or something

themes/V3/Blank/snippets.js
5e-Cleric: linted

calculuschild (approved):

PR #3694, opened 2024-08-30
Fix toolbar issues

5e-Cleric:
Changes flow of brewRenderer and splitpane, adds classes to splitpanes, left pane is named `editorPane`, and right pane is named `previewPane`. Grid is used in the pane to establish a defined height for brewRenderer and toolbar.

This way, the toolbar can be positioned statically, and the scrollbar does not hide behind it.

Also changed some margins and paddings to make spaces match at the beggining and end of a brew preview.

To anyone reviewing, this PR doesn't make any prominent visual change, other than the scrollbar not going behind the element.


client/homebrew/brewRenderer/brewRenderer.less
```diff
@@ -31,6 +34,12 @@
 
 .pane { position : relative; }
 
+.pane.previewPane {
+	display: grid;
+	grid-template-columns: 1fr;
+	grid-template-rows: 32px 1fr;
```
Gazook89, line 40: can this just be 'auto' or something, for the height, so that it just accommodates whatever height the toolbar is (as set by toolbar.less)?  Rather than setting it here at a magic number like 32px?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -217,7 +217,7 @@ const BrewRenderer = (props)=>{
 					onScroll={getCurrentPage}
 					onKeyDown={handleControlKeys}
 					tabIndex={-1}
-					style={{ height: state.height }}>
+					style={{ height: "100vh" }}>
```
Gazook89, line 220: If a js variable isn't needed here, does the height need to be set in the jsx file at all?  Why not just set the `height` property in the brewRenderer.less file?

client/homebrew/brewRenderer/brewRenderer.less
```diff
@@ -3,16 +3,19 @@
 .brewRenderer {
 	overflow-y    : scroll;
 	will-change   : transform;
-	padding-top   : 30px;
+	padding-block : 30px;
```
Gazook89, line 6: I haven't opened this up locally yet, so maybe this is a stupid question, but if the toolbar is no longer absolutely positioned and overlapping the brewRenderer, we don't need 30px of padding on top at all, right?  We only had that extra padding because we wanted to be sure nothing was hidden under the toolbar.

Gazook89 (changes requested):
I did this review before cloning the branch to my local machine, so maybe some of the questions are bad, but they are the first things I would check anyway.  

Before I left on my trip I was thinking that we were having issues with the toolbar disappearing behind the navbar because of extra re-renders between the toolbar and Brew Renderer components.  This PR seems to "take care of it", but doesn't really solve that problem (if it is indeed a problem).   And I don't mean to minimize the improvements in this PR-- the `absolute` positioning *is* a problem that also needed to be fixed.  

I'll check it out tomorrow and see if i can poke it harder

client/homebrew/brewRenderer/brewRenderer.less
5e-Cleric: grid is not my forte, but it should be possible.

I must say, our structure is set to work like this in other places, like the navbar or the snippetbar, both heights are hardcoded everywhere, instead of allowing for any flow.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: The height seems to be used all throughout the component, for different methods, including scroll functions,  i am not familiar with the internal workings of this file, and i preferred not to "fix" something that wasn't broken.

I changed this instance because a hardcoded value in pixels was affecting the layout greatly.

client/homebrew/brewRenderer/brewRenderer.less
5e-Cleric: Because of changes in flow, this padding is all the vertical space between the brew and the top of its parent, whenever you can check out the PR, you will see this space is there for a reason, I changed it for the weird combination of margins in pages and paddings from outside that caused that chaos.

5e-Cleric:
Yes, I don't know for sure what causes the extra renderings, but I can take a look, and I should fix this, if  I shall find it.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: removed height state, unnecessary, you were right

5e-Cleric:
Cleaned rendering of brewRenderer, cleaned unused CSS from toolbar, blocked on #3682 because page count depends right now on a faulty calculation function.

Gazook89:
I pulled this onto my local branch and the changes seem fine.  I don't really think it is 'blocked' by 3682-- the changes here really don't have anything to do with that, unless I'm misunderstanding something.  This is mostly just some light CSS and structure changes, and that one is changing how pages are calculated.

5e-Cleric:
This PR is causing a bug on deployment and local for me, but not for @Gazook89, this small bug requires a bit of testing, to see if it is only on my machine.

### Check the page number
https://github.com/user-attachments/assets/cb60d259-3966-4226-a183-24c490777f5c

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -139,7 +139,7 @@ const SplitPane = createClass({
 
 	render : function(){
 		return <div className='splitPane' onPointerMove={this.handleMove} onPointerUp={this.handleUp}>
-			<Pane
+			<Pane className="editorPane"
```
calculuschild, line 142: Should these be named `leftPane` and `rightPane`, since they are now being used for `vault`, etc.

calculuschild (approved):
I like the cleanup here. Thanks!

shared/naturalcrit/splitPane/splitPane.jsx
5e-Cleric: probably

5e-Cleric:
> I like the cleanup here. Thanks!
> 
> > This PR is causing a bug on deployment and local for me, but not for @Gazook89, this small bug requires a bit of testing, to see if it is only on my machine.
> 
> I see this on my end too. Will that be resolved by #3682 or do you need to dig in more deeply here?

I can't be sure without testing. It makes sense that the error derives from where we calculate the current page, but i need to test that as fix.

5e-Cleric:
with #3705 this PR is obsolete, closing

calculuschild:
Are these fixes still useful for when the toolbar is currently visisble?

5e-Cleric:
> Are these fixes still useful for when the toolbar is currently visisble?

Not really.

PR #3696, opened 2024-08-31
All Snippet entries that have subsnippets but not generators.

dbolack-ab:
The Snippet menu builder excludes any entries without `gen` functions/strings.

This change allows snippet to be included if they lack and `gen` but contain `subsnippets`

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -70,7 +70,9 @@ const Snippetbar = createClass({
 	mergeCustomizer : function(oldValue, newValue, key) {
 		if(key == 'snippets') {
 			const result = _.reverse(_.unionBy(_.reverse(newValue), _.reverse(oldValue), 'name')); // Join snippets together, with preference for the child theme over the parent theme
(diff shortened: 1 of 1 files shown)


client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -70,7 +70,9 @@ const Snippetbar = createClass({
 	mergeCustomizer : function(oldValue, newValue, key) {
 		if(key == 'snippets') {
 			const result = _.reverse(_.unionBy(_.reverse(newValue), _.reverse(oldValue), 'name')); // Join snippets together, with preference for the child theme over the parent theme
-			return _.filter(result, 'gen'); //Only keep snippets with a 'gen' property.
+			return _.filter(result, function(snip) {
+				return(snip.hasOwnProperty('gen') || snip.hasOwnProperty('subsnippets'));
+			});
```
calculuschild, line 75: Any downside if we just simplify this to:

```suggestion
			return result.filter(snip.gen || snip.subsnippets);
```
?

.gen and .subsnippets have to exist *and* be truthy (no empty strings for .gen, etc.)

calculuschild:
Merged!

PR #3699, opened 2024-09-03
Set Focus on Editor after Tab Switches

Gazook89:
This sets the focus on the editor (and thus reveals cursor position) between editor changes (style to text to meta and back).

It *does not* store the cursor position between page reloads.  All it does is set focus.

One side-effect is that this is *setting focus whenever the page loads*.  Some people have expressed fondness for the editor having focus right away on page load, but I'm sure some will be surprised since it is a change from the existing load behavior (where no focus is set at all).

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -82,6 +82,7 @@ const Editor = createClass({
 		if(prevProps.moveSource !== this.props.moveSource) {
 			this.sourceJump();
 		};
+		this.codeEditor.current?.codeMirror.focus();
 	},
 
 	handleControlKeys : function(e){
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -82,7 +82,6 @@ const Editor = createClass({
 		if(prevProps.moveSource !== this.props.moveSource) {
 			this.sourceJump();
 		};
-		this.codeEditor.current?.codeMirror.focus();
 	},
 
 	handleControlKeys : function(e){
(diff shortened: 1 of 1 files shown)


5e-Cleric:
i do prefer focus right away but i will need to scroll anyway to reach the work area

calculuschild:
Having this in the `componentDidUpdate`, this will focus the editor any time the textEditor props or state changes. Is there any edge case where the editor receives some update and we do *not* want to focus on it?

For example, saving causes an update, so if you are typing a page number into the viewer tools and an auto-save kicks in, you will lose focus of the viewer tool. I don't know how likely this is but it could be annoying.

Another difference, is now if you click in the brewRenderer, the focus will stay on the text editor because the two are updating each other. I don't see this as an issue yet but it could be something to look out for.

Not any strong feelings against this, but just wanted to bring it up if someone can think of a weird case where this could be a problem.

> It does not store the cursor position between page reloads. All it does is set focus.

Related to this, I didn't realize before but cursor position *is* stored switching between Text and Style editor. Not page reloads, but at least within the same page.

calculuschild (approved):

Gazook89:
What about moving it here:

```jsx
	handleViewChange : function(newView){
		this.props.setMoveArrows(newView === 'text');
		this.setState({
			view : newView
		}, ()=>{
			this.codeEditor.current?.codeMirror.focus();
			this.updateEditorSize();
		});	//TODO: not sure if updateeditorsize needed
	};
```

calculuschild:
> What about moving it here:

If this maintains the behavior you want I agree this is probably a more appropriate place. This is good for me.

5e-Cleric (approved):
Minimal changes, works as promised nothing to complain about

PR #3704, opened 2024-09-04
Update gitattributes to stop JSON from constantly appearing as modified

### LINKED ISSUE #3703: [DX] Stop editorThemes.json and themes.json from constantly showing as modified
### Your idea:

Every time the build process runs, the files `themes.json` and `editorThemes.json` always show as modified, despite file checks showing that the content is unchanged.

Testing today shows that the line ending encoding is being changed - the build process creates files with `LF` line end encoding, while the pulled file shows `CRLF` line end encoding.

Further testing shows that modifying the `.gitattributes` file to include the line `*.json text eol=lf` prevents this issue from occurring.

G-Ambatte:
This PR resolves #3703.

This PR modifies the `.gitattributes` file to add the line `*.json text eol=lf`, prevent line end encoding in JSON files from being changed on commit or pull.
Initial testing shows that this change prevents the JSON files from constantly appearing in the modified files list.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 package-lock.json binary
+
+*.json text eol=lf


calculuschild:
Seems to work. Thanks @G-Ambatte !

PR #3705, opened 2024-09-04
Add button to toggle Preview tools

Gazook89:
Toggles a state variable to either visible or hidden which is used to set a related class on the toolbar.  The hiding is done with CSS, just reducing the width of the toolbar and the opacity of the tools.  A [few](https://discord.com/channels/113801089362558983/1078734540085932052/1278034176662310944) [complaints](https://discord.com/channels/113801089362558983/1078734540085932052/1278095710679466064) about the always-visible toolbar came in on discord when it was first rolled out.  When shown the work in this PR, at least one of the users said this fix looked good.

https://github.com/user-attachments/assets/a400853a-0bab-4ff8-980b-45e5026280a2

Soon, we will need to figure out how this works with the Header navigation PR #3589.  Possibly this toolbar and the toggle for it will need to change.


calculuschild (approved):
Looks good for now 👍

PR #3711, opened 2024-09-05
Store limited Brew History in Local Storage

### LINKED ISSUE #3070: LocalStorage backups of brews
It happens often enough that a user will mistakenly overwrite one of their brews or delete something they didn't mean to. Without Google Drive, the chances of getting the file back are slim, and require a lot of crossing fingers that a good enough version is in our database backups (not to mention how slow and obnoxious searching the backup system is in MongoDB Atlas).

As mentioned in #2975 , it might be a good idea to save a local copy of the work in the browser, similar to what we do in `/new`.

Given the size of brews being fairly small, I don't think we need to do anything as complex as diffs. In fact, it may be helpful to record multiple copies of each brew based on a logarithmic dropoff time or something.

Say we store 5 versions of each brew. Upon each autosave trigger, whether autosave is enabled or not, we check if any backup slots are available due to becoming expired, according to the following schedule. When a snapshot lower on the list expires, the others move down to take its place.

| Backup Slot | Time before expiry |
| :-----------------:|:-------------------------:|
| 1  | 2 minutes |
| 2 | 10 minutes |
| 3  | 1 hour |
| 4  | 12 hours |
| 5  | 2 days |

To explain the sequence, imagine this:

A user opens the edit page of a brew. We mark the current time.

The user is typing away, and then stops for a moment, triggering an autosave. It hasn't yet been 2 minutes, so no additional Snapshot is taken at this time.

A little while later, the user stops again, and another autosave is triggered. We see that at least 2 minutes have passed since our marked time, so Snapshot 1 will be taken. Since there are no older snapshots in slots 4-5, it can move down to fill the lowest possible space, landing in slot 5, where it will wait for 2 days. We again mark the current time.

Now that slot 1 is open again, if another autosave occurs after 2 more minutes, we take another snapshot and move it down to slot 4. A few minutes later, and all 5 slots are now filled.

Now, when the next 2 minutes are up,  slot 1 expires, and becomes available for a new snapshot.
When 10 minutes are up, Slot 2 will expire, and the current slot 1 will take its place.
When 1 hour is up, Slot 3 will expire, and the current slot 2 will take its place, and the current slot 1 will take slot 2.
etc.

The benefits of this are that we only have to store 5 brews at a time, rather than hundreds of brews tracking every possible change, and the user doesn't lose large blocks of work if they only need to go back a few minutes. We also don't have to deal with complex diff tracking.

Note that each slot duration is only a maximum. I.e., after 2 days when Slot 5 expires, the 2-day backup slot is really only going back 12 hours to the previous Slot 4. But this could be fixed by just adding a 6th or more backup.

In implementing this we should be sure to only expire the oldest item if multiple expiry times happen to align at the same moment (i.e., after revisiting a brew a few days later) so all backups are not lost at the same time. Other items that have been moved to fill the gap would have their expiry timer reset again.

G-Ambatte:
This experimental PR saves a limited brew history in local storage, under the key `HOMEBREWERY-HISTORY-{shareId}-{n}`, where `n` is in the range 1-5.

This will resolve #3070.

Delay between history updates is controlled by the following array:
```
const HISTORY_SAVE_DELAYS = [
	1,			// 1 minute
	30,			// 30 minutes
	60,			// 60 minutes
	24 * 60,		// 24 hours
	5 * 24 * 60		// 5 days
];
```

Every time the Edit Page autosave function runs, the following takes place:
- If the key does not exist, the brew's title, text, style, and the current time are combined into an object, JSON.stringified, and saved in the key.
- If the key does exist, the encoded object is parsed and the current time is checked against the `savedAt` value + the relevant delay value. If the delay has expired, the stored key is updated with the current values.

---

### To Dos

- [x] UI interface
  - [ ] view stored data
  - [x] overwrite current values with stored data
- [x] garbage collection/expiry
- [x] consider adding a minimum version differential, instead of or in conjunction with time


G-Ambatte:
Because it seemed like a good idea, I have also added a `version` differential requirement to the update function. In the previous iteration, returning to the document that you were working on yesterday could immediately wipe out 80% of all stored data. By adding a minimum number of version changes as well, the stored data will not update unless the version has also changed by a minimum number, so returning to edit yesterday's document will only overwrite the historical versions once the user actually starts making changes.

G-Ambatte:
Sample Local Storage keys:
image

calculuschild:
> returning to the document that you were working on yesterday could immediately wipe out 80% of all stored data.

Ideally, if the check finds that any entries are expired, only the oldest entry will be cleared, and all others will move up to the next time delay slot so only one entry expires at a time.

So in your example, if you return the next day, item in slot 4 (24 hours) would expire, but slot 3 would take its place; you would still have 3 entries from the previous day.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -233,6 +251,41 @@ const EditPage = createClass({
 		}));
 	},
 
+	updateHistory : function(){
+		const historyKeys = [];
+		[1,2,3,4,5].forEach((i)=>{
+			historyKeys.push(`${HISTORY_PREFIX}-${this.props.brew.shareId}-${i}`);
+		});
+		historyKeys.forEach((key, i)=>{
```
calculuschild, line 259: I know this is a work in progress, but I have some suggestions on the logic sequence.

Pseudocode:
1) Check if any slot has expired:
  a) Check from oldest to newest (i.e. from slot 5 to 1) for expiry time vs current time
  b) If the slot is expired, remove it, and stop checking for expiry of any other slots (expire only one slot at a time)
2) Check if there is any empty slot and fill it:
  a) Check from oldest to newest (i.e., from slot 5 to 1) for an empty slot.
  b) If an empty slot is found:
    i)  If any newer slot above is already filled, move the next newest slot to fill the space
    ii) If If no newer slot above is already filled, create a new entry from the current brew
    iii) However the slot was filled, reset the expiry date for that entry based on its new slot position, even if it was already expired
  c) continue checking for the next empty slot and repeat (i.e. cascade all slots down as needed until no empty slot remains, resetting their expiry timers for any entry that was moved)
  
Thus, there will always be 1 entry at least 24 hours old (sitting in the 5-day slot), 1 entry 1 hour old (sitting in the 24 hour slot), etc, and we aren't expiring 80% of the entries at once. They all "graduate" to the next level when an older one dies; they don't all die together.

G-Ambatte: I realized the logic was critically flawed last night - for example, a slot with a 5 day timer, edited continuously for 5 days and 5 minutes, would contain data that is 5 minutes old. That's clearly not the intended functionality!

Thanks for digging out the pseudocode, I was about to go look for it.

G-Ambatte:
- [ ] Check if any slot has expired:
  - [ ] Check from oldest to newest (i.e. from slot 5 to 1) for expiry time vs current time
  - [ ] If the slot is expired, remove it, and stop checking for expiry of any other slots (expire only one slot at a time)
- [ ] Check if there is any empty slot and fill it:
  - [ ] Check from oldest to newest (i.e., from slot 5 to 1) for an empty slot.
  - [ ] If an empty slot is found:
    - [ ]  If any newer slot above is already filled, graduate the next newest slot in line to fill the empty slot
    - [ ] If no newer slot above is already filled, create a new entry from the current brew in the empty slot
    - [ ] However the slot was filled, reset the expiry date for that entry based on its new slot position, even if it was already expired
  - [ ] continue checking for the next empty slot and repeat (i.e. cascade all slots down as needed until no empty slot remains, resetting their expiry timers for any entry that was moved)

G-Ambatte:
I have added a garbage collection function; when the function is called and a set period of time has passed without the data in local storage being updated, the history data will be removed.

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,120 @@
...
+            return key.startsWith(HISTORY_PREFIX);
+        })
+        .forEach((key)=>{
+            const collectAt = new Date(JSON.parse(localStorage.getItem(key)).expireAt);
+            collectAt.setMinutes(collectAt.getMinutes() + GARBAGE_COLLECT_DELAY);
+            if(new Date() > collectAt){
+                localStorage.removeItem(key);
+            }
+        });
+};
```
calculuschild, line 120: I'm not sure I understand the purpose of the garbage collection. Are we just arbitrarily wiping peoples' backups if it's been too long?

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,120 @@
...
+
+    // Data from brew to be stored
+    archiveBrew.title = brew.title;
+    archiveBrew.text = brew.text;
+    archiveBrew.style = brew.style;
+    archiveBrew.version = brew.version;
+    archiveBrew.shareId = brew.shareId;
+
+    // Calculated values
+    archiveBrew.savedAt = new Date();
```
calculuschild, line 61: It looks like `savedAt` isn't used anymore.

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,120 @@
...
+    archiveBrew.expireAt = new Date();
+    archiveBrew.expireAt.setMinutes(archiveBrew.expireAt.getMinutes() + HISTORY_SAVE_DELAYS[slot]);
+    
+    // Store data
+    const key = getKeyBySlot(brew, slot);
+    localStorage.setItem(key, JSON.stringify(archiveBrew));
+    return;
+};
+
+export function updateHistory(brew) {
```
calculuschild, line 71: I think ultimately the logic is correct here, but there's got to be a more straightforward way to do what is effectively just iterating over your `history` array and moving entries around.

Rather than carrying around this `numbers` array, I think things will become a lot simpler if you just use normal for loops and have a `maxSlot` number.

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,120 @@
...
+    numbers.toReversed().every((slot)=>{
+        const storedVersion = history[slot];
+
+        // If slot has expired, update all lower slots and break
+        if(new Date() >= new Date(storedVersion.expireAt)){
+            const slots = Array.from(Array(slot - 1).keys());
+            slots.toReversed().every((slot)=>{
+                const actualSlot = slot + 1;
+                // Move data from actualSlot to actualSlot + 1
+                updateStoredBrew({ ...history[actualSlot], shareId: brew.shareId }, actualSlot + 1);
```
calculuschild, line 90: Do you need to re-set `shareId` here? Isn't it already included in `history[actualSlot]`?

client/homebrew/utils/versionHistory.js
calculuschild: Not tested, but I think this might be equivalent:

```
export function updateHistory(brew) {
    const maxSlot = 5;
    const history = [];

    // Load data from local storage to History object
    for (let i = 1; i <= maxSlot; i++) {
        history[i] = getVersionBySlot(brew, i);
    }

    // Walk each version oldest to newest
    for (let slot = maxSlot; slot >= 1; slot--) {

        // If slot has expired, update all lower slots and break
        if (new Date() >= new Date(history[slot].expireAt)) {
            for (let i = slot - 1; i >= 1; i--) {
                updateStoredBrew(history[i], i + 1);	// Move data from slot i to slot i + 1
            }
            updateStoredBrew(brew, 1);	// Fill the first slot with current data
            break;					// Break out of the loop as we found an expired value
        }
    }
}
```

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,120 @@
...
+                // If first slot, fill with current data
+                if(actualSlot == 1) {
+                    updateStoredBrew(brew, 1);
+                    // Break after updating first slot
+                    return false;
+                }
```
calculuschild, line 96: Thinking through what happens here if this is the very first entry.... Let me know if this is correct:

 `history` will be full of `DEFAULT_STORED_BREW`. The one in slot 5 will expire, then all the other `DEFAULT_STORED_BREW` in slots 1-4 will slide up, and the first entry will be in slot 1, set to expire in 1 minute.
 
 Next save, another `DEFAULT_STORED_BREW` expires, everything slides up, and a new entry in slot 1.
 
 So it will take 5 saves before any actual save data reaches slot 5. I think that's fine; shouldn't cause any user data to be expired prematurely.

calculuschild:
If you can't tell, I'm very excited about this one.

client/homebrew/utils/versionHistory.js
G-Ambatte: Correct, I was using it for garbage collection and then right before commit, I changed GC to use `expireAt` instead. Still not sure which will end up being the correct approach.

client/homebrew/utils/versionHistory.js
G-Ambatte: At this stage, the process only update the entries for the document currently being edited, so once a user completes their brew and never edits it again, the local storage entries persist forever. I'm not sure what the actual limit on local storage is - I seem to recall reading 5MB somewhere - it seemed prudent to have some function that checks ALL entries in local storage and cleans out the ones that have not been altered in a significant period.

Currently I have the non-testing GC expiration value set at 28 days.

client/homebrew/utils/versionHistory.js
G-Ambatte: As I recall, that was to cover the instance where the local storage for that slot did not exist, so the history slot is filled with default values - and because this data was then being passed back to the update function to write it into an adjacent slot, it was being written to a different key.

I have since changed the default object to just use the current brew values, so it not actually be necessary any longer.

client/homebrew/utils/versionHistory.js
G-Ambatte: The way it's currently functioning, if the local storage key is empty, the current brew data is loaded instead with `expireAt` set in the past.
So the first time the function runs, it sees slot 5 is expired, overwrites it with the content of slot 4 - which is current brew values. But the act of writing it to slot 5 sets the `expireAt` to 5 days from current.
THEN the process of updating the lower slots does the same for slots 1 - 3, shuffling them down a slot (into 2 - 4) with their respective `expireAt` times - which, again, having been empty/no , are current brew values.
FINALLY, slot 1 is written with the current brew values.

Hopefully I've managed to explain it well enough: in short, the first time the function runs, we end up with the current data in every slot from 1 through 5, albeit with different values set for `expireAt`.

client/homebrew/utils/versionHistory.js
G-Ambatte: Digging into the documentation at [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API/Storage_quotas_and_eviction_criteria#web_storage), they list the local storage limit as 5MB.

However, there are other options if this is insufficient.

calculuschild: Ok, this is fine then. I just wasn't understanding the motive at first.

G-Ambatte: IndexedDB looks like a really good option - although far more complex than LocalStorage.

client/homebrew/utils/versionHistory.js
G-Ambatte: I suspect it will still be useful to keep the `savedAt` data for the UI, although given the current functionality, it might be more useful if it is not overwritten during the storage slot movement process - so the UI can indicate the age of the data be restored, e.g. "Saved data from 19:00 August 10, 2024"

client/homebrew/utils/versionHistory.js
calculuschild: Yes it does. I say we just do localStorage for now as a minimum viable product and we can switch to indexedDB as its own project.

client/homebrew/utils/versionHistory.js
G-Ambatte: I've changed `savedAt` to no longer be overwritten during the updating process, so the time that the stored data was current will be available to the UI.

client/homebrew/utils/versionHistory.js
G-Ambatte: I'll mark this as resolved; currently the garbage collection is only running during the save process of Edit Page, but the intent is that it will be called often.

client/homebrew/pages/editPage/editPage.jsx
G-Ambatte: This should also now be resolved with the latest commits. There's still some way to go on actual UI, but the underlying data is there and updating as expected.

G-Ambatte:
Data, propagating through local storage:
image

Current state of the UI:
image
Currently the UI re-uses the `snippetGroup` and `snippet` styling to dropdown menus.
Hovering over a History item shows the timestamp (UTC) of when that item was saved.

Currently clicking the menu items does not have an effect.

client/homebrew/utils/versionHistory.js
G-Ambatte: This piece of code has been changed.

G-Ambatte:
This is currently functional.

At this point, there's no method of viewing the stored data before overwriting the document in the editor, BUT the change is stored in the Undo history, so users can CTRL+Z (or whatever their preferred Undo method is) to get back to their current document.

5e-Cleric:
Okay, so this still saves the document in mongo/Drive, but keeping older versions in localStorage, right?

)

can we process the time data to be showed in an appropiate unit? Minutes, days, months...

5e-Cleric:
image

It goes from v1 to v6? what is the number related to?

client/homebrew/editor/editor.jsx
```diff
@@ -1,4 +1,4 @@
-/*eslint max-lines: ["warn", {"max": 300, "skipBlankLines": true, "skipComments": true}]*/
+/*eslint max-lines: ["warn", {"max": 500, "skipBlankLines": true, "skipComments": true}]*/
```
5e-Cleric, line 1: this can be easily checked in eslint.config.mjs, not sure why is this here, also, the max is currently 200...

client/homebrew/editor/editor.jsx
5e-Cleric, line 1: are these changes actually related to the brew history stuff? looks like a small cleaning of the component, which could be its own PR

client/homebrew/editor/editor.jsx
G-Ambatte: `editor.jsx` currently has ~475 lines; including this line changes the limit for this file - as you can see, it was previously set at 300 but I have increased the limit to 500.
The next option would be to review the module and break it out into smaller chunks if suitable.

client/homebrew/editor/editor.jsx
G-Ambatte: There are some changes that have been included when I linted the files I was working on; I tried to keep the Linter fixes in their own commits.

G-Ambatte:
> Okay, so this still saves the document in mongo/Drive, but keeping older versions in localStorage, right?

Correct, this does not change the process of saving to Mongo or Google Drive; this is in addition to, not instead of.

> can we process the time data to be showed in an appropiate unit? Minutes, days, months...

image
`snippetbar.jsx:L169-171` does exactly this; if it's more than 60 minutes, changes to hours; if more than 24 hours, changes to days.
Currently, garbage collection is set at 28 days, so it should clean out old entries before they get too far into "months" territory.

G-Ambatte:
> If no version is saved, should all those appear?
> 
> How many changes or time does it need to create a new version? Like, does it not create a new one until 30 minuts have passed from the previous?

DEFAULTS:
image

Currently, slot 0 is not used.

After the first time a save process runs - autosave or otherwise - the local history slots are checked, from slot 5 (with the longest period before expiry) to slot 1 (with the shortest period to expiry). In this scenario, there is no stored data, so they are all considered expired.
The content of slot 5 is discarded, and the content of slot 4 moved into 5; then 3 into 4; 2 into 3; and then 1 into 2. Finally, the current data is written into slot 1. In this scenario, there is no data in slots 1 through 4, so current data is used instead, resulting in the current data occupying all five available history slots.

client/homebrew/editor/editor.jsx
5e-Cleric: wait, is that comment actually changing ESlint rules for that specific document?

G-Ambatte: Yes.

calculuschild:
> How many changes or time does it need to create a new version? Like, does it not create a new one until 30 minuts have passed from the previous?

@5e-Cleric Let me try to explain as well: A new version is saved every 2 minutes (the shortest delay value). However, we only hold on to a total of 5 versions, so if we expired the oldest one every two minutes, we would only have a total of 10 minutes of history. Instead, version we store has a different delay before it expires. The oldest one won't be thrown out until 2 days have passed. The next oldest will stay for 12 hours, etc. So every two minutes when a new version comes, we expire the oldest one we can that is past its threshold, but all the other ones survive. This means most of the time the 2-minute slot will be just overwritten again and again until one of the older spaces below opens up a space and all the others can slide down to fill the space.

In this way, after 12 hours, we can guarantee we have a backup somewhere between 12 hours and 2 days in slot 5, another backup between 1 hour and 12 hours in slot 4, etc.

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -58,14 +62,21 @@ const Snippetbar = createClass({
 	},
 
 	componentDidUpdate : async function(prevProps) {
+		const update = {};
+		let newData = false;
```
calculuschild, line 66: There's no need to try to bundle the setState into one update object like this. React does this automatically. Just use `this.setState({snippets : snippets})` and then `this.setState((prevState)=>{thistoryExists : !prevState.historyExists});` or whatever so the two update steps don't have to be intertwined.

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -138,6 +149,35 @@ const Snippetbar = createClass({
...
+
+		return <div className='dropdown'>
+			{_.map(historyItems, (item, index)=>{
+				if(!item.savedAt) return;
+
+				const saveTime = new Date(item.savedAt);
+				const diffMs = new Date() - saveTime;
+				const diffMins = Math.floor(diffMs / (60 * 1000));
+
+				let diffString = `about ${diffMins} minutes ago`;
```
calculuschild, line 167: Can we have seconds too? It was throwing me off seeing "0 minutes" the whole time.

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -138,6 +149,35 @@ const Snippetbar = createClass({
...
+	replaceContent : function(item){
+		return this.props.updateBrew(item);
+	},
+
+	renderHistoryItems : function() {
+		const historyItems = getHistoryItems(this.props.brew);
+
+		return <div className='dropdown'>
+			{_.map(historyItems, (item, index)=>{
+				if(!item.savedAt) return;
```
calculuschild, line 161: There should be a way we can hide the duplicate entries. I think it will be a source of confusion for a lot of people.

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,123 @@
...
+export const HISTORY_SLOTS = 5;
+
+const DEFAULT_HISTORY_SAVE_DELAYS = {
+ 	'0' : 0,           // 0 minutes (if not specified)
+	'1' : 2,			// 2 minutes
+	'2' : 10,			// 10 minutes
+	'3' : 60,			// 60 minutes
+	'4' : 12 * 60,     // 12 hours
+	'5' : 2 * 24 * 60  // 2 days
+};
```
calculuschild, line 11: Use consistent tabs/ spaces between rows

calculuschild, line 11: Use consistent tabs/ spaces between rows

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,123 @@
...
+	// Calculated values
+	archiveBrew.savedAt = brew?.savedAt || new Date();
+	archiveBrew.expireAt = new Date();
+	archiveBrew.expireAt.setMinutes(archiveBrew.expireAt.getMinutes() + HISTORY_SAVE_DELAYS[slot]);
+
+	// Store data
+	const key = getKeyBySlot(brew, slot);
+	localStorage.setItem(key, JSON.stringify(archiveBrew));
+	return;
+};
```
calculuschild, line 58: Not critical, but would you be opposed to just setting the object directly?

```suggestion
function updateStoredBrew(brew, slot = 0) {
  const archiveBrew = {
    title    : brew.title,
    text     : brew.text,
    style    : brew.style,
    version  : brew.version,
    shareId  : brew.shareId,
    savedAt  : brew?.savedAt || new Date(),
    expireAt : new Date()
  };

  archiveBrew.expireAt.setMinutes(archiveBrew.expireAt.getMinutes() + HISTORY_SAVE_DELAYS[slot]);

  const key = getKeyBySlot(brew, slot);
  localStorage.setItem(key, JSON.stringify(archiveBrew));
}

```

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,123 @@
...
+        .filter((key)=>{
+        	return key.startsWith(HISTORY_PREFIX);
+        })
+        .forEach((key)=>{
+        	const collectAt = new Date(JSON.parse(localStorage.getItem(key)).savedAt);
+        	collectAt.setMinutes(collectAt.getMinutes() + GARBAGE_COLLECT_DELAY);
+        	if(new Date() > collectAt){
+        		localStorage.removeItem(key);
+        	}
+        });
```
calculuschild, line 122: Double-check for tabs vs spaces on this file

calculuschild (changes requested):
Seems functional. Just few small formatting/display things.

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: I could change the functionality slightly: instead of using the current values instead of an empty slot, if the function instead skips the update for that slot, still cycling through the remaining slots. I expect this to result in the following functionality:
1. On first run, slot 1 gets filled (because it always gets filled with current data), so the menu would only show slot 1, version 1 data (current).
2. On second run, slot 2 is filled with slot 1's old data. Slots 3-5 remain unfilled. So slot 1, v2 data (current); slot 2, v1 data
3. Third run, slot 2 data progresses to 3, 1 to 2. So slot 1, v3 data (current); slot 2, v2 data; slot 3, v1 data.

...and so on until slot 1 has v5 data and slot 5 has v1 data.

As there will never be a point where multiple slots have the same data, it will (or at least, _should_) be clearer to the casual observer which slots contain the oldest data.

client/homebrew/utils/versionHistory.js
G-Ambatte: I did, then commented it, then uncommented it... I don't think the Linter catches spacing before comments.

G-Ambatte: Actually, I'll just take those comments out and add one that specifies that the values are in minutes. I think `2 * 24 * 60` makes it clear that the value is 2 days; it would might require clarification if the value was just `2880`.

client/homebrew/utils/versionHistory.js
G-Ambatte: Not sure what happened there. Should be corrected now.

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: Should be added now.

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: It seems to be working perfectly now:

image

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: I've disentangled them now; it all seems happy. Pushing the commit shortly.

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -59,13 +63,17 @@ const Snippetbar = createClass({
...
 			});
-		}
-	},
+		};
 
+		if(historyExists(this.props.brew) != this.state.historyExists){
+			this.setState({
+				historyExists : !this.state.historyExists
+			});
+		};
```
calculuschild, line 75: React is smart enough to not re-render if the new state value is the same as the old value. We don't have to do our own check if the value is different.
```suggestion
		this.setState({
			historyExists : historyExists(this.props.brew)
		});
```

client/homebrew/utils/versionHistory.js
```diff
@@ -0,0 +1,121 @@
...
+	'4' : 12 * 60,
+	'5' : 2 * 24 * 60
+};
+
+const DEFAULT_GARBAGE_COLLECT_DELAY = 28 * 24 * 60;
+
+const DEFAULT_STORED_BREW = {
+	shareId  : 'default_brew',
+	expireAt : '2000-01-01T00:00:00.000Z'
+};
```
calculuschild, line 19: Is DEFAULT_STORED_BREW used anywhere now with your logic change?

G-Ambatte: No, it is not - I thought about using `DEFAULT_STORED_BREW.expireAt` in the last update to `getVersionBySlot()` but immediately realized that it would always be a static value, so I could just set it inside the function.
I have removed `DEFAULT_STORED_BREW` in the latest commit, I'll push it shortly.

calculuschild (approved):

calculuschild:
Before merging I just want to double-check the behavior. Just testing from the deployment I'm not sure the slots are expiring as I expect:

Slot 3 is over 3 hours old, but it's not expiring when I make a new save (should expire when 1 hour old); it just overwrites slot 1. Similarly slot 2 is over 40 minutes old but it's also not expiring (should expire after 10).

image

When home I will double-check with custom config values to make sure I'm not seeing things.

G-Ambatte:
> Before merging I just want to double-check the behavior. Just testing from the deployment I'm not sure the slots are expiring as I expect:
> 
> Slot 3 is over 3 hours old, but it's not expiring when I make a new save (should expire when 1 hour old); it just overwrites slot 1. Similarly slot 2 is over 40 minutes old but it's also not expiring (should expire after 10).
> 
> image
> 
> When home I will double-check with custom config values to make sure I'm not seeing things.

Expiry is set from the moment that the slot is updated, while the saved time is always the time that the data was first saved. Also, the update function only runs when the save function is called; if the document is not being updated, then the function is not being called, so infrequent document updates may cause the data to persist much longer in it's slot than otherwise expected.

As a debugging feature, we could add the slot expiry time to the UI, so we can see exactly when the slot will expire and ensure it matches expectations.

calculuschild:
Ok, my concerns have been assuaged. Merging now.

Thanks @G-Ambatte ! 💯 ⭐ 🎉

PR #3723, opened 2024-09-10
Add sorting options to Vault Page

5e-Cleric:
### Sorting options to Vault

Hello everyone, here i am again, with the next vault update, sorting!

Before anyone asks, i don't think sorting by author will be feasible, but i could try.

When reviewing, keep in mind, because of pagination, every click in a filter is a new search.

This is implemented using the same filter bar from the user page, with different functionality, same css but simplified when possible.

image


dbolack-ab:
One small UX critique. My expectation was that clicking on the "arrow box" ( rough space where either arrow would appear) it would toggle direct, rather than the word.

5e-Cleric:
> One small UX critique. My expectation was that clicking on the "arrow box" ( rough space where either arrow would appear) it would toggle direct, rather than the word.

Feel obligated to point out, this is how it currently works in the userpage, literally using the same code. I do agree with you, but i'd rather get the functionality in place, then fixing both in another PR

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -34,7 +37,10 @@ const VaultPage = (props)=>{
 
 	useEffect(()=>{
 		disableSubmitIfFormInvalid();
-		loadPage(pageState, true);
+		const initialSort = props.query.sort || 'title';
+		const initialDir = props.query.dir || 'asc';
+		const initialSorting = `${initialSort}/${initialDir}`;
+		loadPage(pageState, true, initialSorting);
```
calculuschild, line 43: Any special reason `initialSorting` needs to be a string that you join and split up later? Seems it would be more straightforward to just pass the sort category and direction as separate parameters? Or if the must be combined, as an object?

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -248,6 +263,43 @@ const VaultPage = (props)=>{
 		</div>
 	);
 
+	const renderSortOption = ( optionTitle, optionValue)=>{
+		const oppositeDir = dirState === 'asc' ? 'desc' : 'asc';
+
+		return (
+			<div
+			className={`sort-option ${
+                sortState === optionValue ? `active ${oppositeDir}` : ''
```
calculuschild, line 272: I don't see that `acc` or `desc` in the className here has any effect. It's not mentioned in the `.less` file at all.
```suggestion
                sortState === optionValue ? `active` : ''
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -248,6 +263,43 @@ const VaultPage = (props)=>{
...
+					loadPage(1, false, `${optionValue}/${oppositeDir}`)
+					}}>
+                {optionTitle}
+            </button>
+            {sortState === optionValue && (
+                <i
+                    className={`sortDir fas ${
+                        dirState === 'asc' ? 'fa-sort-up' : 'fa-sort-down'
+                    }`}
+                ></i>
```
calculuschild, line 285:
```suggestion
                <i className={`sortDir fas ${dirState === 'asc' ? 'fa-sort-up' : 'fa-sort-down'}`} />
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -248,6 +263,43 @@ const VaultPage = (props)=>{
...
+			<div
+			className={`sort-option ${
+                sortState === optionValue ? `active ${oppositeDir}` : ''
+            }`}
+			>
+				<button onClick={() => {
+					loadPage(1, false, `${optionValue}/${oppositeDir}`)
+					}}>
+                {optionTitle}
+            </button>
```
calculuschild, line 279:
```suggestion
				<button onClick={() => loadPage(1, false, `${optionValue}/${oppositeDir}`)}>
					{optionTitle}
				</button>
```

client/homebrew/pages/vaultPage/vaultPage.jsx
```diff
@@ -248,6 +263,43 @@ const VaultPage = (props)=>{
 		</div>
 	);
 
+	const renderSortOption = ( optionTitle, optionValue)=>{
```
calculuschild, line 266: Check tabs vs spaces on this file

server/vault.api.js
```diff
@@ -55,6 +63,7 @@ const findBrews = async (req, res)=>{
 
 	await HomebrewModel.find(combinedQuery, projection)
 		.skip(skip)
+		.sort(sortConditions(sort, dir))
```
calculuschild, line 66: Sorting must be performed before any `limit` or `skip` step. Otherwise it is only sorting on the current page, and later pages will have results that should have gone before.

client/homebrew/pages/vaultPage/vaultPage.jsx
5e-Cleric: there is no need for separate parameters, it is overall less words, i think. don't really care

calculuschild: But it adds an unnecessary step of combining/splitting the string, which is a very awkward way of doing things. If there is no special reason let's just keep them separate.

5e-Cleric: solved

server/vault.api.js
5e-Cleric: Does not

5e-Cleric: > When you chain [skip()](https://www.mongodb.com/docs/manual/reference/method/cursor.skip/#mongodb-method-cursor.skip) and [limit()](https://www.mongodb.com/docs/manual/reference/method/cursor.limit/#mongodb-method-cursor.limit), the method chaining order does not affect the results. The server always applies the skip operation based on the sort order before it applies the limit on how many documents to return.

Best i could find in the official docs to support that statement.

client/homebrew/pages/vaultPage/vaultPage.jsx
5e-Cleric: In my machine, there are no spaces, all tabs, not sure if you have the latest changes or if there is a display difference.

5e-Cleric:
Everything should be ready

server/vault.api.js
```diff
@@ -29,12 +29,20 @@ const rendererConditions = (legacy, v3)=>{
 	return {}; // If all renderers selected, renderer field not needed in query for speed
 };
 
+const sortConditions = (sort, dir) => {
+	return {
+		[sort]: dir === 'asc' ? 1 : -1
+	};
```
calculuschild, line 35:
```suggestion
	return { [sort]: dir === 'asc' ? 1 : -1 };
```

server/vault.api.js
```diff
@@ -29,12 +29,18 @@ const rendererConditions = (legacy, v3)=>{
 	return {}; // If all renderers selected, renderer field not needed in query for speed
 };
 
+const sortConditions = (sort, dir) => {
+	return { [sort]: dir === 'asc' ? 1 : -1 };
+}
```
calculuschild, line 34:
```suggestion
};
```

calculuschild:
Ok, seems good to me. Nice work @5e-Cleric ! ⭐ 🎉 💯

5e-Cleric:
Yay!

PR #3724, opened 2024-09-11
Add Pagella Font family to Blank

dbolack-ab:
## Description

    Add Pagella face to Blank Template.
    
    Pagella is a TeX update of the URW Palladio L face that is a good substitute for the commonly used Book Antiqua and Palatino faces.
    
    https://en.wikipedia.org/wiki/Palatino#:~:text=The%20first%20legal%20free%20version,on%20URW%20Palladio%20L%20font.

### Reviewer Checklist

- [x] Font family works
  - [x] Regular
  - [x] Bold
  - [x] Italic
  - [x] Bold Italic
- [x] Snippet Works


5e-Cleric:
Could you add a sample of the font to the PR?

dbolack-ab:
image
image
image
image

Gazook89 (approved):
From what I can tell from a [little search on the license](https://www.latex-project.org/lppl/), it seems fine by itself.  

The only other comment I have (beyond my question on "purpose" in the regular conversation thread) is that I believe many of our other fonts are `.woff2` format.  I don't know if anything is lost/gained between `woff` and `woff2`, but just calling that out.

calculuschild:
Woff2 just has better compression, so we should prefer woff2 here

Gazook89:
Also as I am looking at the theme I am working on now, I can likely swap out the base text font I was going to use ("de Aetna") for Pagella if that means fewer fonts are added to the repo.  I don't see much reason to have many different serif fonts for the standard base text, and Pagella seems like a good choice due to the variety of glyphs.

I know we have Booksanity already, but I think for original themes not based on some other source we can move towards more 'open' fonts.

5e-Cleric (approved):
Looks good to me, thanks @dbolack-ab !

5e-Cleric:
Everything works, tested and reviewed, merging!

PR #3733, opened 2024-09-17
fix

5e-Cleric:
Please do test this before releasing. fixes all errors reported in stage so far.


calculuschild:
Yep. Seems to fix it and the transition from enabling/disabling the history dropdown works as expected.

PR #3736, opened 2024-09-19
Properties Editor Structure and Style

Gazook89:
This is mostly a small refactoring of CSS and some slight changes to HTML structure in advance of the new TagInput component.  Things I noticed while working on that PR are the many places that font sizes were set, and the variety of font-size values used in different places.  And this also updates the "publish" button description, some box sizing and borders, etc.

No methods/functions were harmed in the production of this PR.  

I'd like to get this pushed through before doing much styling on the TagInput component update.

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -311,7 +311,7 @@ const MetadataEditor = createClass({
 
 	render : function(){
 		return <div className='metadataEditor'>
-			<h1 className='sectionHead'>Properties Editor</h1>
+			<h1>Properties Editor</h1>
 
 			<div className='field title'>
 				<label>title</label>
@@ -359,9 +359,7 @@ const MetadataEditor = createClass({
 
 			{this.renderRenderOptions()}
 
-			<hr/>
-
-			<h1 className='sectionHead'>Authors</h1>
+			<h2>Authors</h2>
 
 			{this.renderAuthors()}
 
@@ -372,9 +370,7 @@ const MetadataEditor = createClass({
 				notes={['Invited author usernames are case sensitive.', 'After adding an invited author, send them the edit link. There, they can choose to accept or decline the invitation.']}
 				onChange={(e)=>this.handleFieldChange('invitedAuthors', e)}/>
 
-			<hr/>
-
-			<h1 className='sectionHead'>Privacy</h1>
+			<h2>Privacy</h2>
 
 			<div className='field publish'>
 				<label>publish</label>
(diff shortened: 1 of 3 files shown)


client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -304,17 +304,14 @@ const MetadataEditor = createClass({
...
-				</a>
+				<small><a href='/legacy' target='_blank' rel='noopener noreferrer'>Click here to see the demo page for the old Legacy renderer!</a></small>
 			</div>
 		</div>;
 	},
 
 	render : function(){
 		return <div className='metadataEditor'>
-			<h1 className='sectionHead'>Brew</h1>
+			<h1 className='sectionHead'>Properties Editor</h1>
```
calculuschild, line 314: This is the title of the section, not the whole tab (i.e along with "authors" and "privacy")

calculuschild:
Is this ready for merging or do you want to tinker on it any more?

client/homebrew/editor/metadataEditor/metadataEditor.jsx
Gazook89: I believe it’s fine the way I have it. Many things in life have an H1, then some content, then are followed with h2 and more content. 

Nothing is really lost by removing “brew” from the top here. Everything in that panel is about the brew. We don’t alter anything else with the units like the account. 

But, we do frequently reference the Properties Editor when helping users, and it’d be nice to have it labeled (beyond the hover state on the bar). 

Just my two cents

calculuschild: If you want a top-level title that says "Properties Editor", that is fine. But I meant that having this element on the same level as "Authors" and "Privacy" implies that only part of this panel is the Properties Editor, when "Properties Editor" should apply to the whole page. We are missing one higher-level header where "Properties Editor" belongs.

An equivalent menu at a restaurant:

### Sandwiches (shouldn't this be "Breads", and "Sandwiches" at a higher level at the top?)

- Whole Wheat
- Pumpernickle
- White

### Vegatables

- Lettuce
- Tomato

### Meats

- Turkey
- Bacon

Gazook89: Okay, so just leave Properties Editor as H1, and demote "Privacy" etc to H2

Gazook89: What about this:

image

calculuschild: That's fine, yes. Thanks!

PR #3765, opened 2024-09-27
Change default value of currentEditorCursorPageNum

### LINKED ISSUE #3764: Legacy Brews cannot be viewed
A recent change to the BrewRenderer is causing Legacy brews to be unable to render at all on the SharePage. The issue appears to be from a recent change to line 144 of BrewRenderer.jsx, changing `props.currentEditorPage` to `props.currentEditorPage - 1`.

REPRODUCTION STEPS:
1. Attempt to open the Share page of any Legacy brew, new or old, with or without actual content.

EXPECTED RESULT:
Brew is rendered

CURRENT RESULT:
Loading icon sits on page forever. Page background image does not load.

G-Ambatte:
This PR resolves #3764.

This PR changes the default value of `currentEditorCursorPageNum`, which causes the BrewRenderer to correctly render Legacy Brews again on the Share Page.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -55,7 +55,7 @@ const BrewRenderer = (props)=>{
 		theme                      : '5ePHB',
 		lang                       : '',
 		errors                     : [],
-		currentEditorCursorPageNum : 0,
+		currentEditorCursorPageNum : 1,
 		currentEditorViewPageNum   : 0,
 		currentBrewRendererPageNum : 0,
 		themeBundle                : {},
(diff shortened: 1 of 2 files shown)


calculuschild:
So for future reference, what happened here is #3729 moved all the tracking of page numbers up into the `/edit`, `/new`, and `/` home pages, to be passed down to the `editor.jsx` and `brewRenderer.jsx` so they can be in sync for page jumping purposes. The default initial value for both the editor and brewrenderer page is 1, but since the `/share` page does not have this page syncing feature, the correct initial value was not being passed down from anywhere.

This also means the page viewer toolbar is not being updated on the share page. I will push a fix to this PR to handle that as well.

PR #3767, opened 2024-09-27
Add zoom property to BrewRenderer print styling

### LINKED ISSUE #3744: BrewRenderer toolbar zoom level affects Print sizing
When using the zoom function of the new BrewRenderer toolbar, attempting to print the brew results in pages at the zoomed size, not the intended print size. Possibly could be fixed by adding an `@media print` rule of `zoom: 100%` to override the toolbar's `zoom` setting (or whatever property it been uses to achieve this effect).

G-Ambatte:
This PR fixes #3744.

This PR adds the `zoom` property to the `@media print` rules for the BrewRenderer, preventing the `zoom` property set in inline styles (by the BrewRenderer toolbar) from affecting the print output.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.less
+++ b/client/homebrew/brewRenderer/brewRenderer.less
@@ -39,6 +39,7 @@
 		overflow-y  : unset;
 		.pages {
 			margin : 0px;
+			zoom: 100% !important;
 			& > .page { box-shadow : unset; }
 		}
 	}


Gazook89 (approved):
Does it have to be `!important`?  I don't know, just asking.

Otherwise, i approve of this single line change.

G-Ambatte:
> Does it have to be `!important`?  I don't know, just asking.
> 
> Otherwise, i approve of this single line change.

It does, or else it won't override the inline style.

PR #3786, opened 2024-10-03
Enable --experimental-require-module flag on node build in order to support ESM module only modules and update Node

dbolack-ab:
Created to support #3401.

Tests pass. Just not sure if this is a bad idea for reasons I am unaware of.


Gazook89:
I believe i am a fan of this, but i'm not terribly educated on it.  It's also been awhile since I bumped into whatever limitation it was imposing-- I think maybe when I was working with Radix UI, or maybe it was Shoelace UI (now part of Web Awesome).  It's not a pressing need for me personally, now, but I do remember being frustrated that I couldn't use some ESM modules in certain places.  

It might not be terribly relevant to getting this flag enabled, but are you running into issues with a library only providing ESM modules, or just modules that you are writing in ESM yourself?

dbolack-ab:
> I believe i am a fan of this, but i'm not terribly educated on it. It's also been awhile since I bumped into whatever limitation it was imposing-- I think maybe when I was working with Radix UI, or maybe it was Shoelace UI (now part of Web Awesome). It's not a pressing need for me personally, now, but I do remember being frustrated that I couldn't use some ESM modules in certain places.
> 
> It might not be terribly relevant to getting this flag enabled, but are you running into issues with a library only providing ESM modules, or just modules that you are writing in ESM yourself?

Yes. nanoid.

calculuschild:
This seems to work... It feels hacky though.

One alternative is to just set it as another environment variable `NODE_OPTIONS=--experimental-require-module` similar to what we do with `NODE_ENV="local"`. Then it would apply to any `node/npm` call, not just the specific scripts in our package.json. 

This is specifically just an issue on the server code; our client code gets bundled and passed through Babel, which already takes care of this. Our server code is run raw without any kind of processing, but I'm not sure how to make that work with Babel easily.

I think the actual recommended solution is to turn our project into a module, and use `import` everywhere. But that seems like a lot of effort for something that might not quite work.

I will merge this for now, and then the nanoid one. We can look at alternatives later with some more digging.

PR #3022, opened 2023-09-03
[NEW THEME]:  Unearthed Arcana

G-Ambatte:
This PR adds a new theme: Unearthed Arcana. It mimics the very basic styling of those documents, with monochrome text and no visible background texture.

PR #3792, opened 2024-10-04
Add missing stubs for brew snippets and master pages (Master Pages Pt 1)

dbolack-ab:
This adds some missing bits for snippets stubs and adds the same for master pages.

I chose to name the field `templates` as it may be a better term for the user base than master pages when it comes to intent of the feature. 

Updates tests to account for the helper function returning the template stub.

In addition to providing stubs, this provides direct support for #2804 and #3022.


calculuschild:
Just FYI on this, I am going to put this on lower priority until we decide if Master pages are a feature we actually want to pursue yet. Just given the size of this feature, I don't want to get myself too involved in reviews until the discussion in #3783 comes to a conclusion on both need and approach.

PR #3799, opened 2024-10-07
View Modes - Single / Facing Pages / "Flow" arrangements

### LINKED ISSUE #1379: [FEATURE]: Multiple pages preview  /  popout preview
Would be nice to quickly toggle a Fit to Screen option for easy screen shots of whole single pages. As is, typically it requires using the browser zoom.

To an extreme, it could be possible to fit multiple pages to a screen using either float, css flex box, or css grid. Zoom buttons within the preview pane would set to "full size", fit page to screen, 2 pages per screen, 4 pages to screen, or 9 pages to screen (which could be used to display "trading cards") and this would be responsive sizing, not necessarily strict dimensions.

Gazook89:
## Description
This PR adds a toggle for changing the page arrangement in the Preview pane, and is the last piece of the new Toolbar that I had in mind for the Toolbar.  Using the 3-way toggle, users can choose different visual arrangements for viewing brews.  This setting is set in the *toolbar*, so it is not part of the brew metadata and neither does it carry over between brews/page reloads.

The three settings:

- Single: the same view we've always had-- one page at a time, in a vertical scroll.
- Facing: sets the pages to be side by side in pairs, as if displayed in a book.
- Flow: sets the pages to fill the horizontal space and wrap to a new row when full-- works well for 'card' layouts.

In addition to these three layouts, another "options" button is added next to that toggle.  The options button opens as a little 'popup' much like the "saving error" notifications, and contains some inputs:

- Row Gap/Column Gap: these sliders alter the spacing between pages when using the "Facing" and "Flow" arrangements.  This likely isn't a critical thing to control, but I added it in case folks have different ideas about perfect spacing (for screenshots maybe).  
- Start on right:  This input is disabled unless you are on the Facing mode, and changes where the first page is situated.  If a user has a Cover Page on page 1, likely it makes the most sense to have "Start on right" set to true.

Additional options could be added, but i honestly couldn't think of anything else.  They should be strictly for modifying the *view*, and nothing about the actual brew, I think-- though it is tempting to put a "print friendly" button somewhere on the toolbar, I don't think that'll happen.

- Closes #1379 

### Toolbar:

image

image

### Facing Pages:

image

### Flow:

image

### Possible future changes
<del>I am considering changing the "Flow" mode to using Grid rather than Flex.  Grid can still handle what Flex does, but gives more options such as setting a max number of columns/pages in a row, which could then be a setting in the Options box.</del>  I'm not sure this is possible any longer after a quick check.  If someone knows otherwise, let me know.

Another thing:  possibly an additional view mode is the "page flip" from [pageflip.js](https://github.com/Nodlik/StPageFlip) or [turn.js](http://www.turnjs.com)

### Reviewer Checklist

- [x] Verify new features are functional
  - [x] Can toggle between "single", "facing", and "flow" layouts and is visually correct
  - [x] Adjustments to Options settings carry through to iframe (Gap settings, and "Start on right")
  - [x] Style is consistent with the rest of the UI
- [x] Verify old features have not broken
  - [x] Toolbar can still be hidden
  - [x] Test on at least a few existing brews
- [x] Test for edge cases / try to break things
  - [x] Displays well enough (i.e. nothing hidden) in Firefox
  - [x] Clicks outside of Options box closes the box
- [x] Identify opportunities for simplification and refactoring

```
  - [ ] Can toggle between "single", "facing", and "flow" layouts and is visually correct
  - [ ] Adjustments to Options settings carry through to iframe (Gap settings, and "Start on right")
  - [ ] Style is consistent with the rest of the UI
  - [ ] Toolbar can still be hidden
  - [ ] Test on at least a few existing brews
  - [ ] Displays well enough (i.e. nothing hidden) in Firefox
  - [ ] Clicks outside of Options box closes the box
```

</details>


ericscheid:
The `Start on right` option could instead be implemented as a fourth view setting option ("book"?). Just an idea.

5e-Cleric:
> The `Start on right` option could instead be implemented as a fourth view setting option ("book"?). Just an idea.

I'd write it as "separate cover" or "start by second page" or something like that.

Gazook89:
> instead be implemented as a fourth view setting option ("book"?)

That is basically what "Facing" is-- it would make more sense I think to just rename "Facing" to "Book", which is fine with me.  I had steered away from "Book" at the time because I thought it might have certain connotations to it that might set expectations for something else, but I'm not sure that was well-founded now.   As a fourth option, I'm not sure it makes any sense.


> I'd write it as "separate cover" or "start by second page" or something like that.

I just went with the phrasing from affinity publisher:
image

I'm not sure  either of those provides any additional clarity over "start on right".  I do think that there will be users wondering "why would I want my pages unbalanced and have the first page all by itself", but I also think that it will be easy enough to click the button to set it to the left.  Or maybe, this particular setting (left or right start) *should* be a brew metadata property, so authors can decide what works best for their brew.  Or perhaps the author sets the "default"/initial value, and then viewers can still have this option to change it.

5e-Cleric:
> I'm not sure either of those provides any additional clarity over "start on right". I do think that there will be users wondering "why would I want my pages unbalanced and have the first page all by itself", but I also think that it will be easy enough to click the button to set it to the left. Or maybe, this particular setting (left or right start) _should_ be a brew metadata property, so authors can decide what works best for their brew. Or perhaps the author sets the "default"/initial value, and then viewers can still have this option to change it.

Start on right or left sounds good, we could also go for start on odd page or start from even. Same thing.

I think this should be in the viewing stuff, not the metadata panel. Definitely.

About that, is this CSS going to override brews' own code to change layout?

Gazook89:
It applies inline styles to the `.pages` element. So if someone is mucking about outside of `.page` then yes, they’ll have to `!important` their rules

5e-Cleric:
## Review

- [x] Verify new features are functional
  - [x] Can toggle between "single", "facing", and "flow" layouts and is visually correct
  - [x] Adjustments to Options settings carry through to iframe (Gap settings, and "Start on right")
  - [x] Style is consistent with the rest of the UI
- [x] Verify old features have not broken
  - [x] Toolbar can still be hidden
  - [x] Test on at least a few existing brews
  - [ ] Displays well enough (i.e. nothing hidden) in Firefox
  - [x] Clicks outside of Options box closes the box

### Comments

- 2 page brews look ridiculous with the facing layout
- The config wheel should not be visible in single view mode
- The layout toggle should have some indicator to understand it is a toggle (maye that is just me)
- Focus in the checkbox looks terrible, maybe we can adjust it



- Should the snap zoom to width snap to two pages if that mode is selected?


Will keep reviewing

Gazook89:
> 2 page brews look ridiculous with the facing layout

Yup, that's just the way they are.  If someone is making a 2 page brew, hopefully one of them isn't a cover page.  But if it is, the viewer can switch it to start on the left, or they can use two other viewing modes.

> The config wheel should not be visible in single view mode

Perhaps.  I left it as-is because I was contemplating an option on the single view to have horizontal scrolling rather than vertical.  What do you think about that?  For now, i'll make the view settings disabled on Single, though.

> The layout toggle should have some indicator to understand it is a toggle (maye that is just me)

I would agree, I think I intended to do that, put it off, and then just got so used to how the toggle worked that I forgot that it might not be natural to others.  I'll think about how to manage that visually.  Perhaps I will put three little dots under it, and have the index of the current mode colored; a quick picture works as better example:

image

Or, if there were good icons, I could do a set of three buttons (only one active at a time, like a radio input group).  I'm not sure this works well with just text, though, since it would take so much space:

image

> Focus in the checkbox looks terrible, maybe we can adjust it

Agree, I think the checkboxes need a little makeover.  I don't think *focus states* has really had any...erm...*focus* in terms of styling up until recently.  I will take a crack at improving it in this particular toolbar while keeping in mind that a future PR may do a more comprehensive restyling of various inputs and their different states.

>  Should the snap zoom to width snap to two pages if [facing pages] is selected?

I'll let someone else decide this or follow whatever the consensus is.  I'll leave it as the last thing I do, if at all.

> Initial value of the sliders does not match the actual value, at all

I'll have to come back to this when i have more time to really think about it. You may ask "what is there to think about", and honestly i'm just not sure it really matters if their initial values line up with their actual values.

Gazook89:
Regarding a better way to indicate that there are three options, rather than clicking on a single button to reveal them, what about this?

image

dbolack-ab:
> > instead be implemented as a fourth view setting option ("book"?)
> 
> That is basically what "Facing" is-- it would make more sense I think to just rename "Facing" to "Book", which is fine with me. I had steered away from "Book" at the time because I thought it might have certain connotations to it that might set expectations for something else, but I'm not sure that was well-founded now. As a fourth option, I'm not sure it makes any sense.

When I was doing more formal layout the term of the day was "spreads"

Also, for future expansion, you need to consider "back to front" books.

Gazook89:
In my latest commits I just switched to icons, though I suppose the `title` of each can be updated to whatever.  But I also didn't think 'spreads' was wholly appropriate either, because I think of spreads as more of a generic term, more akin to what I have been calling 'modes'.   Correct me if I'm wrong, but I thought a 'spread' basically meant 'arrangement', and that a spread could be displaying 3 pages side-by-side, for example a tri-fold brochure.  In which case, I wouldn't want to say that a two-page 'book' layout is the "spread" view, because that is only *one type of spread*.  

I think more generally I can stop saying "View Modes", and instead go with "View Spreads" or just "Spreads" and update my variable names to that convention.

> Also, for future expansion, you need to consider "back to front" books.

Eh, i suppose maybe one day, but I haven't ever seen a single demonstrated need for it (aka no user has ever mentioned it), and until then it's not something I am going to think about.  There would need to be like dozens of users suddenly saying they want it before I personally would go down that path.

Gazook89:
At this point, with the latest commit, I only have these concerns from @5e-Cleric to address:

- [ ] Focus in the checkbox looks terrible, maybe we can adjust it
- [ ] Should the snap zoom to width snap to two pages if [facing pages] is selected?

And I'm contemplating whether some of these things should have default values that are set by brew props, so that authors can say "by default, display this as a book spread".   But I'd be happy to leave that as a new PR.

Gazook89:
I don't know how I'm how i am going to get around this issue with the "current page number" field showing `NaN` when really zoomed out.  I might just set a condition that if the value is not a number, then return `1` and just cheese it like that.

image

5e-Cleric:
Looks good, page navigation is wonky with multiple pages, but that is to be expected, can we make the next page button move 2 if the pages mode is "facing"?

Gazook89:
> can we make the next page button move 2 if the pages mode is "facing"?

I think I might leave this for a future thing, because the whole page navigation thing as you mention is wonky.  But I expect that the basic implementation would be easy, possibly one line:
```
<button
					id='next-page'
					className='tool'
					onClick={()=>scrollToPage(pageNum + (spread === 'facing' ? 2 : 1))}
					disabled={pageNum >= totalPages}
				>
```
(more a note to myself than anything here).

5e-Cleric (approved):
Looks good overall, i am not sure what are the advantages of the anchored component over the dialog, but i trust there is a noticeable difference.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -3,23 +3,49 @@ const React = require('react');
...
+			spreads.forEach((spread)=>pagesContainer.classList.remove(spread));
+			pagesContainer.classList.add(spread);
+			['recto', 'verso'].forEach((leaf)=>pagesContainer.classList.remove(leaf));
+			pagesContainer.classList.add(startOnRight ? 'recto' : 'verso');
+		}
+	}, [spread, startOnRight]);
+
+	useEffect(()=>{
+		onStyleChange({ '.page': pageShadows ? {} : { boxShadow: 'none' } });
+	}, [pageShadows]);
```
calculuschild, line 47: Generally, state is not intended to be used just as a trigger to put all the side effects in a `useEffect()` like this. Instead, any `onClick()` functions should call an actual handler function which applies the side effects *and* sets the new state. Then you don't need a useEffect at all.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -173,6 +174,27 @@ const BrewRenderer = (props)=>{
 		}));
 	};
 
+	const handleStyle = (newStyle)=>{
```
calculuschild, line 177: This step seems way too complex. We should be able to reuse a similar process to what we did with setting the zoom. I.e.:

1) User clicks a button in the toolbar
2) Handler function from the parent gets called with the new setting
3) Renderer in the parent uses the new setting to conditionally change its style

If we need a whole custom processor for styles like this something has gone wrong

Gazook89:
I think this is pretty much ready to be reviewed again.  Thanks @calculuschild for the react changes.

Gazook89:
Actually, I think before merging this in, #3845 should get a second look and be merged before/at the same time.  It handles a lot of the "current page" questions that having "facing" and "flow" spreads will raise.

calculuschild:
#3845 relies on some functions that will no longer exist as a result of the refactoring in this PR. I will merge this one now since it is reviewed and complete, then we can resolve the conflicts in 3845 to relocate each piece as needed.

PR #3800, opened 2024-10-07
Change local version history to use Indexed DB

### LINKED ISSUE #3763: Move localstorage of history versions to an alternative
Currently, we have enabled a limited brew history feature stored in localstorage (see #3711).

However, we have also realised this has the inevitable conclusion of filling up the localStorage and collapsing some features that depend on it.

That's why the idea has risen, about moving such feature (already merged) to IndexedDB or minimongo.

this should be fixed before the next version (current version 3.15.0).

G-Ambatte:
This PR resolves #3763.

This PR changes the history version to use the IndexedDB system rather than LocalStorage, as the maximum storage of IndexedDB is much larger than LocalStorage.

This PR is still currently a work in progress, as the IndexedDB interface is quite different to the LocalStorage interface.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -5,7 +5,7 @@ const createClass = require('create-react-class');
 const _     = require('lodash');
 const cx    = require('classnames');
 
-import { getHistoryItems, historyExists } from '../../utils/versionHistory.js';
+import { historyCheck, loadHistory } from '../../utils/versionHistory.js';
 
 //Import all themes
 const ThemeSnippets = {};
@@ -50,30 +50,31 @@ const Snippetbar = createClass({
 			renderer      : this.props.renderer,
 			themeSelector : false,
 			snippets      : [],
-			historyExists : false
+			historyExists : false,
+			showHistory   : false
 		};
 	},
 
-	componentDidMount : async function() {
+	componentDidMount : async function(prevState) {
 		const snippets = this.compileSnippets();
 		this.setState({
 			snippets : snippets
 		});
 	},
 
-	componentDidUpdate : async function(prevProps) {
+	componentDidUpdate : async function(prevProps, prevState) {
 		if(prevProps.renderer != this.props.renderer || prevProps.theme != this.props.theme || prevProps.snippetBundle != this.props.snippetBundle) {
 			this.setState({
 				snippets : this.compileSnippets()
 			});
 		};
 
-		if(historyExists(this.props.brew) != this.state.historyExists){
+		const checkHistoryExists = (await historyCheck(this.props.brew)) ?? false;
+		if(checkHistoryExists != prevState.historyExists){
 			this.setState({
-				historyExists : !this.state.historyExists
+				historyExists : checkHistoryExists
 			});
-		};
-
+		}
 	},
 
 	mergeCustomizer : function(oldValue, newValue, key) {
@@ -151,30 +152,34 @@ const Snippetbar = createClass({
 		return this.props.updateBrew(item);
 	},
 
-	renderHistoryItems : function() {
-		const historyItems = getHistoryItems(this.props.brew);
+	renderHistoryItems : async function() {
+		if(!this.state.historyExists || !this.state.showHistory) return;
 
-		return <div className='dropdown'>
-			{_.map(historyItems, (item, index)=>{
-				if(!item.savedAt) return;
+		const historyItems = await loadHistory(this.props.brew);
+		console.log('renderHistoryItems', historyItems);
+		return;
 
-				const saveTime = new Date(item.savedAt);
-				const diffMs = new Date() - saveTime;
-				const diffSecs = Math.floor(diffMs / 1000);
+		// return <div className='dropdown'>
+		// 	{_.map(historyItems, (item, index)=>{
+		// 		if(!item[1].savedAt) return;
 
-				let diffString = `about ${diffSecs} seconds ago`;
+		// 		const saveTime = new Date(item[1].savedAt);
+		// 		const diffMs = new Date() - saveTime;
+		// 		const diffSecs = Math.floor(diffMs / 1000);
 
-				if(diffSecs > 60) diffString = `about ${Math.floor(diffSecs / 60)} minutes ago`;
-				if(diffSecs > (60 * 60)) diffString = `about ${Math.floor(diffSecs / (60 * 60))} hours ago`;
-				if(diffSecs > (24 * 60 * 60)) diffString = `about ${Math.floor(diffSecs / (24 * 60 * 60))} days ago`;
-				if(diffSecs > (7 * 24 * 60 * 60)) diffString = `about ${Math.floor(diffSecs / (7 * 24 * 60 * 60))} weeks ago`;
+		// 		let diffString = `about ${diffSecs} seconds ago`;
 
-				return <div className='snippet' key={index} onClick={()=>{this.replaceContent(item);}} >
-					<i className={`fas fa-${index+1}`} />
-					<span className='name' title={saveTime.toISOString()}>v{item.version} : {diffString}</span>
-				</div>;
-			})}
-		</div>;
+		// 		if(diffSecs > 60) diffString = `about ${Math.floor(diffSecs / 60)} minutes ago`;
+		// 		if(diffSecs > (60 * 60)) diffString = `about ${Math.floor(diffSecs / (60 * 60))} hours ago`;
+		// 		if(diffSecs > (24 * 60 * 60)) diffString = `about ${Math.floor(diffSecs / (24 * 60 * 60))} days ago`;
+		// 		if(diffSecs > (7 * 24 * 60 * 60)) diffString = `about ${Math.floor(diffSecs / (7 * 24 * 60 * 60))} weeks ago`;
+
+		// 		return <div className='snippet' key={index} onClick={()=>{this.replaceContent(item);}} >
+		// 			<i className={`fas fa-${index+1}`} />
+		// 			<span className='name' title={saveTime.toISOString()}>v{item.version} : {diffString}</span>
+		// 		</div>;
+		// 	})}
+		// </div>;
 	},
 
 	renderEditorButtons : function(){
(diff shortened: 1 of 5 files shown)


G-Ambatte:
For reference, I have set the following data in my `config/local.json` file as the custom test values:

```
{
"historyConfig" : {
		"HISTORY_SAVE_DELAYS" : {
			"0" : 0,
			"1" : 1,
			"2" : 2,
			"3" : 3,
			"4" : 4,
			"5" : 5
		},
		"GARBAGE_COLLECT_DELAY" : 10
	}
}
```

client/homebrew/utils/versionHistory.js
```diff
@@ -11,28 +13,33 @@ const DEFAULT_HISTORY_SAVE_DELAYS = {
 	'5' : 2 * 24 * 60
 };
 
-const DEFAULT_GARBAGE_COLLECT_DELAY = 28 * 24 * 60;
+const HB_DB = 'HOMEBREWERY-DB';
+const HB_STORE = 'HISTORY';
 
-const HISTORY_SAVE_DELAYS = global.config?.historyData?.HISTORY_SAVE_DELAYS ?? DEFAULT_HISTORY_SAVE_DELAYS;
-const GARBAGE_COLLECT_DELAY = global.config?.historyData?.GARBAGE_COLLECT_DELAY ?? DEFAULT_GARBAGE_COLLECT_DELAY;
```
calculuschild, line 17: I think the way you had this before was simpler; just set the HISTORY_SAVE_DELAYS and GARBAGE_COLLECT_DELAY once at this point, instead of partially defining it over multiple locations.

client/homebrew/utils/versionHistory.js
```diff
@@ -11,28 +13,33 @@ const DEFAULT_HISTORY_SAVE_DELAYS = {
...
-function updateStoredBrew(brew, slot = 0) {
+// function getVersionBySlot(brew, slot){
+// 	// Read stored brew data
+// 	// - If it exists, parse data to object
+// 	// - If it doesn't exist, pass default object
+// 	const key = getKeyBySlot(brew, slot);
+// 	const storedVersion = IDB.get(key);
+// 	const output = storedVersion ? storedVersion : { expireAt: '2000-01-01T00:00:00.000Z', shareId: brew.shareId, noData: true };
+// 	return output;
+// };
```
calculuschild, line 38: Looks like we can delete this block

client/homebrew/utils/versionHistory.js
```diff
@@ -43,74 +50,110 @@ function updateStoredBrew(brew, slot = 0) {
...
+	if(!IDB) return false;
+	const historyExists = await IDB.keys(await createHBStore())
+		.then((keys)=>{
+			return [...keys].some((key)=>{
+				return key.startsWith(`${HISTORY_PREFIX}-${brew.shareId}`);
+			});
+		})
+		.catch(()=>{return false;});
+
+	return historyExists;
```
calculuschild, line 79:
```suggestion
	const keys = await IDB.keys(await createHBStore())
		.catch(() => []);
	return keys.some(key => key.startsWith(`${HISTORY_PREFIX}-${brew.shareId}`));

```

client/homebrew/utils/versionHistory.js
```diff
@@ -43,74 +50,110 @@ function updateStoredBrew(brew, slot = 0) {
...
+				if(!history[updateSlot - 1]?.noData) {
+					historyUpdate.push(parseBrewForStorage(history[updateSlot - 1], updateSlot + 1));
+				}
 			};
 
 			// Update the most recent brew
-			updateStoredBrew(brew, 1);
+			historyUpdate.push(parseBrewForStorage(brew, 1));
+
+			IDB.setMany(historyUpdate, await createHBStore());
```
calculuschild, line 133: Does this need an await? I guess nothing else relies on the result here so it can be done asynchronously,

client/homebrew/utils/versionHistory.js
```diff
@@ -43,74 +50,110 @@ function updateStoredBrew(brew, slot = 0) {
...
+
+				// if(key.startsWith(`${HISTORY_PREFIX}`)) {	// This check was to make sure we were only selecting the history keys, should be unnecessary
+				const collectAt = new Date(value.savedAt);
+				collectAt.setMinutes(collectAt.getMinutes() + GARBAGE_COLLECT_DELAY);
+				if(new Date() > collectAt){
+					IDB.del(key, await createHBStore());
+				};
+				// }
+			});
 		});
```
calculuschild, line 158: The beauty of `await` is we can avoid `.then` and read things in a more traditional sequence. Bonus, is it tends to be quite a bit shorter too.

```suggestion
	const entries = await IDB.entries(await createHBStore());

	for (const [key, value] of entries) {
		const collectAt = new Date(value.savedAt);
		collectAt.setMinutes(collectAt.getMinutes() + GARBAGE_COLLECT_DELAY);

		if (new Date() > collectAt)
			await IDB.del(key, await createHBStore());
	}
```

client/homebrew/editor/snippetbar/snippetbar.jsx
```diff
@@ -151,12 +164,18 @@ const Snippetbar = createClass({
...
 
+	toggleHistoryMenu : function(){
+		this.setState({
+			showHistory : !this.state.showHistory
+		});
+	},
+
 	renderHistoryItems : function() {
-		const historyItems = getHistoryItems(this.props.brew);
+		if(this.state.historyItems.length == 0) return;
```
calculuschild, line 174: is `historyItems` redundant with `historyExists`? HistoryExists also requires at least one entry is in `historyItems`. Seems `historyExists` could be removed altogether and just use `historyItems` like you do here?

calculuschild (changes requested):
Functionality seems good. I couldn't find anything breaking.

Feedback is then just standard code cleanup stuff.

client/homebrew/utils/useIDB.js
```diff
@@ -0,0 +1,80 @@
+import {
```
calculuschild, line 1: This whole file can be deleted

client/homebrew/utils/versionHistory.js
G-Ambatte: Yes, I commented it out in case I broke something by no longer having it available, but it should be fine to just delete now.
It will be saved in a reversible commit anyway.

client/homebrew/utils/versionHistory.js
G-Ambatte: Reverted to the local consts, rather than trying to use config values.

client/homebrew/utils/versionHistory.js
```diff
@@ -43,74 +50,110 @@ function updateStoredBrew(brew, slot = 0) {
...
-	return Object.keys(localStorage)
-		.some((key)=>{
-			return key.startsWith(`${HISTORY_PREFIX}-${brew.shareId}`);
-		});
+async function createHBStore(){
+	return await IDB.createStore(HB_DB, HB_STORE);
+}
+
+export async function historyCheck(brew){
+	if(!IDB) return false;
```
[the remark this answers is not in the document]
G-Ambatte: This was originally a longer check to ensure `window.IndexedDB` existed, but eventually it got changed to the current form - which, as you correctly point out, can never be false.
Removed in the latest commit.

client/homebrew/utils/versionHistory.js
calculuschild: Config values are fine if you really want them, but my point was you can assign these values here instead of setting it to the default here and then loading the config later. Just do the default and config handling all in one place.

client/homebrew/utils/versionHistory.js
G-Ambatte: This is implemented in the latest commit, with a minor tweak or two to get it working properly.

client/homebrew/utils/versionHistory.js
G-Ambatte: This is implemented in the latest commit.

client/homebrew/utils/versionHistory.js
```diff
@@ -43,74 +50,110 @@ function updateStoredBrew(brew, slot = 0) {
...
+					return dataArray.forEach((data)=>{
+						history.push(data ?? DEFAULT_HISTORY_ITEM);
+					});
+				})
+				.catch(()=>{
+					historyKeys.forEach(()=>{
+						history.push(DEFAULT_HISTORY_ITEM);
+					});
+				});
+
```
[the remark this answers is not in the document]
G-Ambatte: In theory, no error should be possible: at one stage I was getting errors that I thought were related to `window` or `window.IndexedDB` not be instantiated yet, but I'm fairly certain now that I was wrong and that was not actually the cause.

client/homebrew/editor/snippetbar/snippetbar.jsx
G-Ambatte: Correct, I started out just having `historyExists` is a boolean function, but I had been considering using a `historyItems.length` declaration instead. However, as `loadHistory` should return a non-zero length Array every time, it should be slightly more complicated than just a `length` check.

client/homebrew/utils/versionHistory.js
G-Ambatte: Technically, no it doesn't.

Possibly it should have one, just for consistency.

client/homebrew/editor/snippetbar/snippetbar.jsx
calculuschild: If `loadHistory` returns a non-zero length Array everytime, is this heck for `historyItems.length == 0` ever going to be true?

G-Ambatte: Only if it hasn't populated it yet... in which case the renderHistoryItems function should not be accessible, as it is locked behind the historyExists function.
This is implicit in the way the render function is called, but I think I'll add it as an explicit guard clause inside the renderHistoryItems function.

client/homebrew/utils/useIDB.js
G-Ambatte: Removed in the latest commit

client/homebrew/utils/versionHistory.js
G-Ambatte: I've added await to this call and also the Garbage Collection IDB.delete function, so that the IDB calls are all consistently called with await.

calculuschild:
Okay here goes..... Merging! Thanks @G-Ambatte for figuring IndexedDB out!

💯 ⭐ 🎉 🌟 🎆 💥 🍾

Issue #2585, opened 2023-01-03
[SERVER] Improved Messaging System


G-Ambatte:
### Your idea:

Recording Gitter discussion for future feature development:

> **erics**: [...] maybe we should put in place a generic handler for “message from server”. That way any page, and I mean any page, could pop a UI thing to the user with a custom message (e.g. “new version, please refresh”, or “going down for maintenance in 10 mins”, or “happy new year!”). The UI thing could be a banner.

> **jeddai**: [...] it's certainly reasonable imo to have a mongo collection of messages with a start/stop ISO date, an endpoint to fetch them, and a repeating timeout on a minute to fetch them from the server. And if it's in mongo we can add a feature to the admin panel to CRUD the messages

G-Ambatte:
In terms of fetching messages from the server, is a repeated fetch from every user the best way to handle this? Wouldn't it create less load on the server to use WebSockets* instead, broadcasting the message event repeatedly (with the client set up to ignore repeated messages) rather than having every user make an additional browser request of the server every sixty seconds?

---

* bearing in mind that I don't have any direct experience with WebSockets and/or knowledge of their best practices. Just that this seems like a perfect use-case for them.

jeddai:
That's a good option as well, I don't know whether it would put more strain on the server or not, but I can for sure say it'll put less strain on the database.

A workflow could look like this:

1. Establish the MongoDB collection
2. On initial app load, send the current messages
3. Open a websocket connection, and use it to send messages to the active clients

When a new message is created, it would broadcast the message. If a message is updated, that would be broadcast as well and the clients could handle that by matching the ObjectID+version.

G-Ambatte:
I think I did some work along these lines a year or two back, I'll see what I can dig out and throw at a Draft PR as a proof of concept.

calculuschild:
Backend is implemented, but not the actual message delivery system.

5e-Cleric:
Other elements should use the same components
- [ ] Render Warnings
- [ ] HTML error warnings

PR #3805, opened 2024-10-10
Add notification client side

5e-Cleric:
This creates the client implementation of notifications, should solve #2585 

Gets all notifications from the database, then filters out any already dismissed, expired, or inactive.

Renders the dialog appropiately.

also made some minor changes to the admin page, as noticed minor issues with layout and style there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: cimg/node:20.8.0
+      - image: cimg/node:20.17.0
       - image: mongo:4.4
 
     working_directory: ~/homebrewery
@@ -27,7 +27,7 @@ jobs:
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
 
-      - run: sudo npm install -g npm@10.2.0
+      - run: sudo npm install -g npm@10.8.2
       - node/install-packages:
           app-dir: ~/homebrewery
           cache-path: node_modules
@@ -45,7 +45,7 @@ jobs:
 
   test:
     docker:
-      - image: cimg/node:20.8.0
+      - image: cimg/node:20.17.0
 
     working_directory: ~/homebrewery
     parallelism: 1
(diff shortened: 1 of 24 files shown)


Gazook89:
Please be sure some screen shots are included when you are ready

5e-Cleric:
> Please be sure some screen shots are included when you are ready

no UI has changed, not sure what screenshots do you want, this substitutes the current hardcoded notifications for any notification in the database as described in the original comment.

5e-Cleric:
In order to test this i recommend loading up the deployment, all three current notifications should show up, then go to the localstorage section of the dev tools, and delete some of the notifications dismiss keys to test if it partially renders (which it does in my machine).

Further features will come in next PRs, could include:

- Decide color of the notification
- Reopen all active notifs from navigation

5e-Cleric:
Adding image as per @Gazook89 's request:

image
As we can see, even if some notifications are dismissed, the ones not dismissed still render.

If one dismisses the dialog, any notification is dismissed.
I manually deleted that entry in localstorage to show this.

dbolack-ab:
This worked for me as described.

dbolack-ab (approved):
This worked as expected.

I am a little concerned about clutter on the user side.

Gazook89:
> I am a little concerned about clutter on the user side.

For this PR, I don't think that is a concern I would have.  This PR is basically just changing how the notifications are managed by admins with the slight change that not every active notification is shown again when a new one is added (so it should actually be less clutter than current).  

I do think we could do something like smaller notification notices, possibly that expand (an HTML `details/summary` element) to show the details.

5e-Cleric:
> I do think we could do something like smaller notification notices, possibly that expand (an HTML `details/summary` element) to show the details.

I think if we did that, users would not ever expand them. Although, if we only dismissed them if they had opened them...

calculuschild:
HTML is not being parsed correctly:

image

calculuschild:
Ok, this looks good to me. Merging!

PR #3809, opened 2024-10-10
Adress small accessibility concerns

5e-Cleric:
- Adds pointer cursor icon to the entire foldgutter element
- Adds highlight to the foldgutter element
- Adds not-allowed cursor icon to snippetbar functions if they are not available (brew history stuff)

- Adds ew-resize cursor to toolbar zoom function (doesn't work in chrome, for some reason)
- Changes width of page number input to be dynamic
- Adds not-allowed cursor icon to disabled buttons


calculuschild (approved):
All new functionality looks fine, and this is mostly just CSS so I don't see any bugs.

Just need to fix the merge conflicts and this can merge.

PR #3810, opened 2024-10-10
Updates to correct for issues missed in enabling --experimental-require-module flag.

dbolack-ab:
It was determined this works fine with 20.18 and corresponding NPM. Unsure if I typoed and that's why it worked locally or not


Gazook89:
I suggest just re titling this PR

PR #3813, opened 2024-10-11
Test: Convert /server and depends to ESM style code calling

dbolack-ab:
I've having a little distrust of my local environment with jest. creating a PR to compare against circleci results


dbolack-ab:
Closing in favor of #3902

PR #3816, opened 2024-10-11
Refactor version history to use custom store wrapper

G-Ambatte:
This PR refactors the IndexedDB usage in the new version history functions, pulling it out to it's own wrapper module.

As best I can determine, this has not affected the functionality of the code at all, but provides a useful interface layer for any future use of IndexedDB.


5e-Cleric (approved):
History function works just fine, approving this change

calculuschild:
To clarify... the only thing this does is wrap all the functions so they call `await createCustomStore()` internally? If yes, does idb-keyval not have a way we can just set the default database or something?

I agree this makes indexedDB usage a little cleaner, but tons of wrappers feels like a code smell that should have a less "boilerplatey" solution.

calculuschild:
Just for fun, maybe a Proxy would be appropriate here? I haven't used these before but I think something like this might work:

```
import * as IDB from 'idb-keyval/dist/index.js';

const IDBProxy = new Proxy(IDB, IDBWrapper) ;

const IDBWrapper = {
	get: (target, prop) => {		// target is IDB, prop is IDB.prop (so `entries`, `setMany`, etc.)
		return (...args) => {			// `get:` here is overwriting how we handle "getting" `IDB.prop`
			return target[prop]( ...args, await createCustomStore() );
		}
	}
};

//Use the proxied version of the IDB package
IDBProxy.set('key', 'value');
```

G-Ambatte:
At this point, I've left both options in the `customIDBStore.js` file... They both work, as far as I have been able to tell so far.

5e-Cleric:
Toaster suggests the following change to the `initCustomStore` function, which is shorter, and to my utrained eye, seems very close:

```
export function initCustomStore(db, store) {
	const storePromise = IDB.createStore(db, store);

	return {
		entries: async () => IDB.entries(await storePromise),
		keys: async () => IDB.keys(await storePromise),
		values: async () => IDB.values(await storePromise),
		clear: async () => IDB.clear(await storePromise),
		get: async (key) => IDB.get(key, await storePromise),
		getMany: async (keys) => IDB.getMany(keys, await storePromise),
		set: async (key, value) => IDB.set(key, value, await storePromise),
		setMany: async (entries) => IDB.setMany(entries, await storePromise),
		update: async (key, updateFn) => IDB.update(key, updateFn, await storePromise),
		del: async (key) => IDB.del(key, await storePromise),
		delMany: async (keys) => IDB.delMany(keys, await storePromise),
	};
}
```

G-Ambatte:
> ```
> export function initCustomStore(db, store) {
> 	const storePromise = IDB.createStore(db, store);
> 
> 	return {
> 		entries: async () => IDB.entries(await storePromise),
> 		keys: async () => IDB.keys(await storePromise),
> 		values: async () => IDB.values(await storePromise),
> 		clear: async () => IDB.clear(await storePromise),
> 		get: async (key) => IDB.get(key, await storePromise),
> 		getMany: async (keys) => IDB.getMany(keys, await storePromise),
> 		set: async (key, value) => IDB.set(key, value, await storePromise),
> 		setMany: async (entries) => IDB.setMany(entries, await storePromise),
> 		update: async (key, updateFn) => IDB.update(key, updateFn, await storePromise),
> 		del: async (key) => IDB.del(key, await storePromise),
> 		delMany: async (keys) => IDB.delMany(keys, await storePromise),
> 	};
> }
> ```

I've applied this change now; everything appears to work correctly. The main issue is the loss of explanatory comments for future devs, but hopefully the code is clean enough that additional comments are not necessary.

G-Ambatte:
> you can add the comments back, it should still be shorter. but i do think it is somewhat easy to understand, perhaps add a comment explaining what the proxy does?

I think it's clean enough now that the comments should be unnecessary. The result of initIDBProxy should be functionally identical to the result of initCustomStore; as mentioned in a previous comment, at this stage I've left both in place. I suspect initCustomStore is much easier to understand then initIDBProxy.

calculuschild:
Ok, at this point they both look about the same amount of "clean", so let's just go with one of the options since they are functionally identical. Keep whichever you prefer and remove the other @G-Ambatte. Then we can merge.

G-Ambatte:
> Keep whichever you prefer and remove the other @G-Ambatte.

As much as I like the Proxy version, I think it's harder to follow than the Wrapper, so in the interests of being readable for future maintainers, I suspect that will be the best route.

calculuschild:
Ok, merging!

PR #3817, opened 2024-10-12
Generate Templates Snippets menu (Master Pages Pt 2)

dbolack-ab:
This step adds the functionality to pull Brew Templates from the brew frontmatter and from the brew's theme chain and transform it into a snippets menu.

This is part 2 of master pages/templates sork. The first is in #3792.


dbolack-ab:
Pulling for lack of need.

PR #3818, opened 2024-10-12
Insert Templates/Master Pages into the Rendering pane. (Master Pages Part 3)

dbolack-ab:
This PR adds the functionality to insert the Template/Master Page Markdown+ into pages.

Usage:
`\page Source Document Name:Template Name`

example:

`\page Template Tester:Master Test 1`

Templates will be applied in the following manner.

1. If the `\page` declaration has a template label, it is used for the page.
2. If the `\page` declaration does not have a template label, the most recently used template is applied.
3. If the `\page` declaration does not have a template and there are no prior declarations, the `Blank` template in the active document will be used - if it exists.

(V3 only) rawPages are now constructed via a function that parses out to capture the template names and will drop the first page if it is empty in order to allow template setting on page 1.


dbolack-ab:
Pulling for lack of need.

PR #3820, opened 2024-10-13
Add script fix tool to Admin page #3801

### LINKED ISSUE #3801: [ADMIN] Add brew script cleaning function
### Your idea:

Currently, users are breaking their brews by saving scripts in them - Imgur embed scripts and the like. While an ideal solution would be to prevent these scripts from breaking the site, in the interim a simple tool which updates the brew content without needing to render it would allow any admin user to fix this issue.

Basically, make an Admin page function that takes in a brew's ShareID and then does a find and replace in that brew's text content to eliminate any script open or close tags.

G-Ambatte:
This PR resolves #3801.

This PR adds functionality to the existing BrewLookup tool UI, to call a new Admin API function that parses a brew's title, text, and description properties, and replaces any and all instances of `<script` or `</script` with `<s` or `</s` respectively.

NB: Because the replacement function runs on the text property WITHOUT parsing it through any `splitText` functions, it also checks style and metadata.


G-Ambatte:
Because the tool scans for `</script` without the end bracket, we can test with a brew set up like so:

image

Brew located on the Admin page:

image

Clicking **Scan for SCRIPTs** initiates a search, if a matching tag is found, **CLEAN BREW** appears:

image

After clicking **CLEAN BREW**, the button is removed:

image

Refreshing the original Edit Page, we can see the `script` tags have been replaced with `s` tags:

image

5e-Cleric:
trying to test this, i see no indication of whether the function is scanning or not, nothing happens.

[HB-ahref=javascriptalertundefinedundefinedThisisaJavaScriptinjectionviahrefattributeundefinedundefinedClickmea.txt](https://github.com/user-attachments/files/17497344/HB-ahref.javascriptalertundefinedundefinedThisisaJavaScriptinjectionviahrefattributeundefinedundefinedClickmea.txt)

G-Ambatte:
Testing with the HB Bomb:

1. **PRODUCTION :** Create **NEW FROM FILE** using HB_Bomb.txt
2. **PRODUCTION :** Confirmed brew is broken - image
3. **DEPLOYMENT :** Navigate to Admin page, enter brew ID into Brew Lookup - image
4. **DEPLOYMENT :** Run brew search - image
5. **DEPLOYMENT :** Run **Scan for SCRIPTs** - image
6. **DEPLOYMENT :**  Run **CLEAN BREW** - image
7. **PRODUCTION :** Refresh Edit Page, observe that brew loads correctly - image

G-Ambatte:
> trying to test this, i see no indication of whether the function is scanning or not, nothing happens.

I think what is needed is an indication that the scan is running/completed and what the result is, so that there is some reaction, even on a brew that has no `script` tags to find.

Gazook89:
Why not just automatically scan the brew, rather than having to press a button to do it.  After the search it would just do it, and tell you immediately "hey, there are some scripts in this".  It could have an alternate state "hey, no scripts in this brew".  

Then the option to "clean" it.

G-Ambatte:
> Why not just automatically scan the brew, rather than having to press a button to do it. After the search it would just do it, and tell you immediately "hey, there are some scripts in this". It could have an alternate state "hey, no scripts in this brew".
> 
> Then the option to "clean" it.

That's a good idea - I originally anticipated script detection being a larger load than it actually appears to be. I should be able to include this fairly quickly this evening.

G-Ambatte:
The most recent changes:

1. The number of script tags is displayed as part of the normal search for a brew
2. When the CLEAN BREW button is clicked, this number updates to zero and the CLEAN BREW button is removed.

Brew with detected SCRIPT tags:
image

Brew without detected SCRIPT tags/after "cleaning":
image

5e-Cleric:
image

When i run the file, i get an immediate 0 scripts detected... and there are scripts

G-Ambatte:
> When i run the file, i get an immediate 0 scripts detected... and there are scripts

How strange... It appears that https://homebrewery-pr-3820.herokuapp.com/source/yv7Y9xV-82o8 was identical to https://homebrewery-pr-3820.herokuapp.com/source/Tuvq-LKs-6tp, but the latter was the one I generated from the file linked earlier in the piece which detected and cleaned the script tags without issue. I have not yet determined why the former does not.

G-Ambatte:
Now working via the Homebrew API functions, and confirmed working with Google brews.

5e-Cleric:
it seems to work, even if other vulnerabilities are still around, this is only meant to be a tool for us to clean corrupt brews.

This does not remove other javscript sources, #3548 should take care of that. But this could clean the title element as well, if only because that change breaks the userpage too.

5e-Cleric (approved):
Overall, nice function, i hope we never need to use it.

Thank you for this.

server/homebrew.api.js
```diff
@@ -87,8 +87,18 @@ const api = {
 			// Get relevant IDs for the brew
 			const { id, googleId } = api.getId(req);
 
+			const accessMap = {
+				edit  : { editId: id },
+				share : { shareId: id },
+				admin : {
+					$or : [
+						{ editId: { $regex: id, $options: 'i' } },
+						{ shareId: { $regex: id, $options: 'i' } },
```
calculuschild, line 96: Why do we need regex for this?

server/admin.api.js
```diff
@@ -100,6 +100,27 @@ router.get('/admin/finduncompressed', mw.adminOnly, (req, res)=>{
...
+	function cleanText(text){return text.replaceAll(/(<\/?s)cript/gi, '');};
+
+	const brew = req.brew;
+
+	const properties = ['text', 'description', 'title'];
+	properties.forEach((property)=>{
+		brew[property] = cleanText(brew[property]);
+	});
+	// Tag cleaning is commented out as it is impossible to enter a script tag in tags
+	// brew.tags = cleanText(brew.tags.join('\n')).split('\n');
```
calculuschild, line 116: Do you expect this commented section to ever be needed in the future or should it just be removed?

server/admin.api.js
```diff
@@ -66,23 +70,19 @@ router.post('/admin/cleanup', mw.adminOnly, (req, res)=>{
...
-			{ editId: { $regex: req.params.id, $options: 'i' } },
-			{ shareId: { $regex: req.params.id, $options: 'i' } },
-		]
-	}).exec()
-	.then((brew)=>{
-		if(!brew)	// No document found
+router.get('/admin/lookup/:id', mw.adminOnly, asyncHandler(HomebrewAPI.getBrew('admin', false)), async (req, res, next)=>{
+	const brew = req?.brew ?? undefined;
+
+	try {
```
calculuschild, line 76: This `try` block should no longer be necessary since we have removed the database/Google action inside that could potentially fail. Instead we use `getBrew()` which has its own error handling inside.

If there is any issue getting a brew, `getBrew()` already generates and throws the appropriate error, and then `asyncHandler` passes the error to the central error handler in `app.js` which knows to send the 404 to the client.

So if we found the brew, we can just return the brew. If there was an error, we won't reach this line anyway.

```
router.get('/admin/lookup/:id', mw.adminOnly, asyncHandler(HomebrewAPI.getBrew('admin', false)), async (req, res, next)=>{
		return res.json(brew);
});
```

client/admin/brewUtils/brewLookup/brewLookup.jsx
```diff
@@ -12,22 +14,45 @@ const BrewLookup = createClass({
...
+			})
+			.catch((err)=>this.setState({ error: err }))
+			.finally(()=>{
+				this.setState({
+					searching : false
+				});
+			});
+	},
+
+	cleanScript(){
```
calculuschild, line 49: Let's pull this out so we aren't defining a function inside of another function.

client/admin/brewUtils/brewLookup/brewLookup.jsx
```diff
@@ -46,12 +71,23 @@ const BrewLookup = createClass({
...
 				<dt>Last Updated</dt>
 				<dd>{Moment(brew.updatedAt).fromNow()}</dd>
 
 				<dt>Num of Views</dt>
 				<dd>{brew.views}</dd>
+
+				<dt>Number of SCRIPT tags detected</dt>
+				<dd>{this.state.scriptCount}</dd>
 			</dl>
+			{this.state.checkForScript &&
```
calculuschild, line 86: Could we possibly remove `checkForScript` state altogether and just use `scriptCount > 0` (or `this.state.scriptCount &&`, since 0 is falsy) here since this is the only place it is used?

calculuschild (changes requested):
Functionality seems good. Just some cleanup requested.

server/homebrew.api.js
G-Ambatte: I was copying the existing functionality from the Brew Lookup function, which used regex for case-insensitive search... I think we can probably skip that; I suspect that anyone with Admin access is probably going to assume that the search is case-sensitive.

G-Ambatte: Regex searching removed in latest commit.

server/admin.api.js
G-Ambatte: Comment removed in latest commit.

client/admin/brewUtils/brewLookup/brewLookup.jsx
G-Ambatte: Removed checkForScript in latest commit.

server/admin.api.js
G-Ambatte: Applied in the latest commit.

server/admin.api.js
```diff
@@ -66,23 +70,9 @@ router.post('/admin/cleanup', mw.adminOnly, (req, res)=>{
...
-			return res.status(404).json({ error: 'Document not found' });
-		else
-			return res.json(brew);
-	})
-	.catch((err)=>{
-		console.error(err);
-		return res.status(500).json({ error: 'Internal Server Error' });
-	});
+router.get('/admin/lookup/:id', mw.adminOnly, asyncHandler(HomebrewAPI.getBrew('admin', false)), async (req, res, next)=>{
+	const brew = req?.brew ?? undefined;
```
calculuschild, line 74: I don't think even this line is needed. If brew is undefined, `getBrew` has already failed and we wouldn't get this far.

G-Ambatte: At this point, I think the whole function/endpoint is obsolete.

calculuschild: Hm. Do we already have an endpoint that just gets a brew and returns it?

client/admin/brewUtils/brewLookup/brewLookup.jsx
G-Ambatte: Which function are we pulling out here?

server/admin.api.js
5e-Cleric: Yes, getBrew in homebrew.api, but that returns the text and textbin as well. This function does not.

client/admin/brewUtils/brewLookup/brewLookup.jsx
G-Ambatte: I had to change it to `this.state.scriptCount > 0` as using just `this.state.scriptCount` would result in a `0` in the output when the button was not displayed.

client/admin/brewUtils/brewLookup/brewLookup.jsx
G-Ambatte: I've pulled the counting function out to a separate function call now, and it gets properly called after the brew text is updated. This required a minor update to the `updateBrew` function, specifically to the section that removes the `text` field after compressing to `textBin`.

server/admin.api.js
G-Ambatte: `getBrew` is a middleware, not an endpoint.

server/homebrew.api.js
```diff
@@ -365,7 +374,7 @@ const api = {
 			// Compress brew text to binary before saving
 			brew.textBin = zlib.deflateRawSync(brew.text);
 			// Delete the non-binary text field since it's not needed anymore
-			brew.text = undefined;
+			if(!req?.keepText) brew.text = undefined;
```
calculuschild, line 377: We should never want to keep the raw text. This defeats the purpose of the text compression.

client/admin/brewUtils/brewLookup/brewLookup.jsx
calculuschild: > Which function are we pulling out here?

This was referring to `cleanScript` which was defined inside of `lookup`

server/homebrew.api.js
G-Ambatte: We only want it returned for the brews which have just been cleaned - but we don't need it kept in the saved data; only in the returned data. I'll change that function so that the data that is actually saved is not changed.

G-Ambatte: This should now be resolved by the latest commit.

calculuschild: Sorry, I'm still not clear why this change was needed. Moving the function definition of `cleanscript` to a different place shouldn't have impacted this.

Can you help me understand what broke when cleanScript was moved out of `lookup`?

G-Ambatte: Not so much a thing that broke, but rather a change in process to remove an assumption.

Previously, the function would replace the tags in the brew, then set the number of detected tags to zero. However, it did this without actually running the function to check the text - and because when the brew is saved to Mongo, the text property is removed, running the check function again on the returned brew data would result in an error.

The current iteration adds a `returnText` property to the request, and when the save process detects the presence of this field, the text property is copied out of the data to be saved, and then added to the object to be returned post-save, so the data written to disk is unchanged, but the `text` property is returned to be checked again.

calculuschild: Ok, so this is just to facilitate re-scanning multiple times?

G-Ambatte: Essentially, yes - to allow the updated text to be checked to confirm that it has been cleaned correctly.

calculuschild: I'm going to fiddle with this code for a bit, if that's ok.

G-Ambatte: 100%, please do.

client/admin/brewUtils/brewLookup/brewLookup.jsx
```diff
@@ -46,12 +69,23 @@ const BrewLookup = createClass({
...
+				<dt>Created Time</dt>
+				<dd>{brew.createdAt ? Moment(brew.createdAt).toLocaleString() : 'No creation date'}</dd>
+
 				<dt>Last Updated</dt>
 				<dd>{Moment(brew.updatedAt).fromNow()}</dd>
 
 				<dt>Num of Views</dt>
 				<dd>{brew.views}</dd>
+
+				<dt>Number of SCRIPT tags detected</dt>
```
calculuschild, line 81:
```suggestion
				<dt>SCRIPT tags detected</dt>
```

server/homebrew.api.js
calculuschild: Just backed out the last two commits and made it automatically perform `lookup()` again after cleaning, so it just fetches the updated brew and scans it again.

`updateBrew()` is really only designed to return the brew stub, not the whole document, so I think adding the separate *keepText* flag might end up being an escape hatch into behavior it isn't designed to do. Instead, it might be worth a more direct refactor to `updateBrew()` if we decide we need to reconsider exactly what fields it returns, or if it should return the whole document.

Issue #3822, opened 2024-10-13
Delete unnecessary elements


5e-Cleric:
There are a number of unnecessary elements in our code that can be safely deleted for cleanup, mainly, wrapper divs that are duplicated, probably as a necessary step in development, but ultimately useless.

The main example on this is the navigation, which currently has the following structure:

```
<nav>
    <div class="navContent">
        {children}
    </div>
</nav>
```
As it should be obvious, that content div has no real role, and can be deleted safely, as long as:
1. it is not referenced by any function
2. proper styles are secured

5e-Cleric:
Similarly, i am not sure we need the div.content as parent of the splitpane, with good styling that one might not be needed either

5e-Cleric:
An update to that, the 2 initially reported elements are created by the iframe component, not by us.

5e-Cleric:
Can confirm the .content div can be removed from editpage, newpage, homepage... any page with splitpane

PR #3823, opened 2024-10-13
Clean up DOM

5e-Cleric:
This PR removes unnecessary wrapping divs following the criteria described in #3822, so far:

- Deletes div.navContent (child of nav)
- Deletes div.content (parent of splitPane)


calculuschild:
Looks good to me!

PR #3829, opened 2024-10-15
Memoize BrewRenderer

### LINKED ISSUE #3830: Lag when scrolling editor panel as of 3.16.0
## Description

See
https://reddit.com/r/homebrewery/comments/1g3pxfb/lagging_code_panel_after_3160/
and
https://www.reddit.com/r/homebrewery/comments/1g3p1ku/input_lag_after_the_recent_update/

calculuschild:
## Description
Memoizes the `brewRenderer.jsx` component. Normally, any time a parent component updates, it re-renders all it's child components as well because it re-sends their props. Scrolling sends a new "currentPage" to the editor/new/home page, which triggers a re-render.

This PR adds a `memo` marker to BrewRenderer, and an equality function to specify the condition when a render should be triggered. This prevents unnecessary markdown parsing, etc. when simply scrolling around; only props that actually affect the render output will trigger a render.

Greatly reduces lag during scrolling of the text editor panel, though further updates are possible (Editor.jsx also redraws the custom markdown highlighting, which we could similarly memoize or otherwise optimize).

**NOTE:** The equality function makes liberal use of `?.` optional chaining because `/new` seems to not have some of the initial props set. Might be a sign of something to check in `/new` but not of immediate urgency to get this fix out.

- Closes #3830 

## QA Instructions

Take a large brew in the Edit Page and compare lag during scrolling of the text editor on this branch and on live HB. Example from Reddit: https://homebrewery.naturalcrit.com/share/NJBuzODSboxi

Test with editor/renderer sync on and off.

Confirm all other changes that should trigger a re-render of BrewRenderer still fire (changing theme, renderer, style content, text content, etc.)

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -1,7 +1,7 @@
 /*eslint max-lines: ["warn", {"max": 300, "skipBlankLines": true, "skipComments": true}]*/
 require('./brewRenderer.less');
 const React = require('react');
-const { useState, useRef, useCallback } = React;
+const { useState, useRef, useCallback, memo } = React;
 const _ = require('lodash');
 
 const MarkdownLegacy = require('naturalcrit/markdownLegacy.js');
@@ -47,7 +47,7 @@ const BrewPage = (props)=>{
 const renderedPages = [];
 let rawPages      = [];
 
-const BrewRenderer = (props)=>{
+const BrewRenderer = memo((props)=>{
 	props = {
 		text                       : '',
 		style                      : '',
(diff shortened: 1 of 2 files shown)


5e-Cleric:
I offer https://homebrewery.naturalcrit.com/share/9EnXpdgPfRvI as a testing brew, because it is meant for it and not coming from a reddit user that might want to keep their work a little more secure.

dbolack-ab:
Can confirm, this is a significant improvement with a clone of Liorith Supplement: Casters ( https://homebrewery.naturalcrit.com/share/NJBuzODSboxi / ) and the Technomancers book I use as a big brew test. ( https://homebrewery.naturalcrit.com/share/b8XlqkNvF-vA is mu copy - I lost track of the original owner )

calculuschild:
Superceded by #3831

PR #3841, opened 2024-10-20
Refactor of ErrorBar component

5e-Cleric:
This PR refactors the ErrorBar component to be functional, and use the Dialog component:

image

As you can see the style remains basically the same as before, as does the rendering of text.


calculuschild (changes requested):
I have not been able to get this to actually display. The box is rendered and appears in the DOM, but does not actually become visible.

In the mean time I have some small tweaks to the JSX file, and the HTML output is unchanged, but still not visible. I'm assuming this is something with the CSS, possibly a conflict after updating to master.

G-Ambatte:
> I have not been able to get this to actually display. The box is rendered and is listed in the DOM, but does not actually become visible.

The Dialog components displays when there is at least one entry in the `dismissKeys` array - we'll need some changes to the Dialog component to use it in the manner intended by this PR.

5e-Cleric:
Superseded by #3972

PR #3845, opened 2024-10-21
Intersection Observers for getting "Current Page(s)"

### LINKED ISSUE #3824: Rethink how current page number is figured in Preview Pane
### Your idea:

I'm interested in how to better get the current page number "in view" in the Preview Pane.  This is because with the "View Modes / Spreads" PR, and zoom controls, it is possible to have multiple pages in full view and it becomes unclear what page should be the 'current' page.  And, there is some struggle with the current method calculating a page number that is higher than the total number of pages.

I know this has been discussed in the past, particularly Intersection Observer.  I just don't see it in any PR's or open/closed issues so bringing it up formally here.

From what I can tell, the basic method is to set up `useEffect` such that whenever the array of `.page`s changes, add that to the observer (with the iFrame or brewRenderer being the "observing" node) so that every page is watched.  Each page can be set up with a `threshold` of 50% or something, so any page more than halfway into the iFrame is counted as being 'in view' (the exact threshold can be debated).   Rather than make a choice about which page number to display in the "current page" indicator, I vote we just give a page range.  So if I'm zoomed out and I see 4 pages more than halfway in view, my "current page" would be "3-6".

I understand this likely needs to be able to return an *actual single* current page for things like "scrollTo", to which I don't really have a solid answer right now.  Maybe it's possible to return two results back from the observer callback:  the array of all pages in view, and just the page that is dead center of the viewport which can be found by setting the `rootMargin` to `50% 50% 50% 50%`.  

### Resources
[MDN - Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

Gazook89:
This PR is mostly motivated by the work on the #3799 PR which allows users to view many pages at once, but was causing issues with what is the "current page" and also issues of viewing "page 10 out of 9 total pages"-like issues in the Toolbar.

## Description

This sets up two Intersection Observers: the first captures every page that is at least 30% visible inside the `.pages` container, and the second captures every page that has at least one pixel on the horizontal center line of `.pages`.  Both can be arrays of integers (page index).

The `visiblePages` array is duplicated and formatted into a `formattedPages` state variable as strings, which gets displayed in the toolbar. `visiblePages` is kept so that it can be used in the "previous/next page" buttons of the toolbar later.

The toolbar displays the `formattedPages` strings, unless the user clicks into the page input and enters their own integer (only a single integer, no range), which can then jump the preview to that page on Enter or blur().

The Arrow 'change page' buttons jump the preview back and forth by a 'full set'.  Meaning that if you are viewing one page a time it will move through the previous/next pages one at a time, but if you are viewing 10 pages at once it will move you through 10 pages at a time.

- Closes #3824
- #3694 -- ultimately scuttled for another PR but raises issue of bad page count calculations

First, sorry for the bad single commit on this one, it doesn't really walk you through the development.

An example document (500 pages) with some additional styling to make it easy to see page numbers:

[HB-NecromanticArmorofSaladDressing.txt](https://github.com/user-attachments/files/17656392/HB-NecromanticArmorofSaladDressing.1.txt)

**or**, just put this CSS in any long brew:
```css
.page:has(.insideCover, .frontCover, .backCover)::after {
  display: grid;
}

.page::after {
  all:unset;
  content: counter(page-numbers);
  position: absolute;
  top: 0;
  left: 0;
  width: 100%;
  height: 100%;
  background: #AAAA;
  display: grid;
  place-items: center;
  font-size: 200px;
  }
  
.page:nth-child(even):after {
  transform: unset;
 }

```

To Test:

1. Load the page and it should load up with `1 / [total pages]` in the toolbar.
2. You should see it change when scrolling in the Preview pane, stopping at `1` or the total page count without going over.
3. Then, change the zoom so you see multiple pages and you should see *ranges* of pages in that input, like `10 - 12`.  You can also change the current "Spread" to facing or flow.  
4. Try the "previous/next" arrows. It should not overrun past the limits.
5. Try typing a single integer in the 'current page' input and hitting `enter` or clicking outside of it-- it should take you to that page.  Typing non-integer characters should be ignored, as well as ranges, and integers outside the total page count will bring you to the last page.
7. After entering your own page number, scrolling again continues to work.

Single page with >30% of area inside view:
image

Book spread, shows that only pages with >30% of area inside view are shown in the 'current pages' input:
image

Many pages, shows the "range":
image

### Reviewer Checklist

**No checklist yet. (I have to go to bed)**

  - [ ] Page ranges are shown when multiple pages in view (more than a sliver).
  - [ ] Prev/Next Page buttons in toolbar move through based on how many pages are in view.
  - [ ] Accurately describes pages in view when pages are in `flex` format (multiple pages in row)
  - [ ] Brew/Source Jump buttons still work
  - [ ] Prev/Next buttons in toolbar still work
  - [ ] Does not show current page higher than total number of pages, or lower than "1". 
  - [ ] Functions at any zoom level
  - [ ] Doesn't cause performance issues

```
  - [ ] Page ranges are shown when multiple pages in view (more than a sliver).
  - [ ] Prev/Next Page buttons in toolbar move through based on how many pages are in view.
  - [ ] Accurately describes pages in view when pages are in `flex` format (multiple pages in row)
  - [ ] Brew/Source Jump buttons still work
  - [ ] Prev/Next buttons in toolbar still work
  - [ ] Does not show current page higher than total number of pages, or lower than "1". 
  - [ ] Functions at any zoom level
  - [ ] Doesn't cause performance issues
```

</details>


G-Ambatte:
When the toolbar showing 100% zoom and page displaying `7 - 8`, the following occurs:
- clicking _Next Page_ results in the view moving to page 9 - Expectation: either end of visible range (8) or page immediately after visible range (9)
- clicking _Previous Page_ results in the view moving to page 5 - Expectation: either start of visible range (7) or page immediately before visible range (6)

Presumably this issue will also appear at other zoom levels.

Gazook89:
> clicking Next Page results in the view moving to page 9 - Expectation: either end of visible range (8) or page immediately after visible range (9)

To be clear, are you saying that this expectation is met?  Because when I do it, `7 - 8` and "next", it goes to `9`.  And then you are saying that the actual problem is this:

> clicking Previous Page results in the view moving to page 5 - Expectation: either start of visible range (7) or page immediately before visible range (6)

Just want to be sure you are listing *one* problem, not two.

G-Ambatte:
When the toolbar showing 40% zoom and page displaying `7 - 9`, the following occurs:
- clicking _Next Page_ results in the view moving to pages `10 - 11` - Expectation: either page range starting at the end of visible range (`9 - 10`) or page range starting immediately after visible range (`10 - 11`)
- clicking _Previous Page_ results in the view moving to pages `4 - 5` - Expectation: either page range ending at the start of visible range (`6 - 7`) or page range ending immediately before the start of the visible range (`5 - 6`)

So it appears that the _Previous Page_ option is consistently selecting one page earlier than expected.

G-Ambatte:
> Just want to be sure you are listing _one_ problem, not two.

Correct, _Next Page_ is working within my intuitive expectations, while _Previous Page_ does not.

Gazook89:
So it's taking the number of visible pages (3), walking back that number of pages from the "first in range" (or, subtracting 3 from 7), and then using that result to plug into `scrollToPage`.   

Because you are only viewing partials of the 'top' and 'bottom' pages, you have 3 visible pages.  But when `scrollToPage` is called, it places the first page right at the top of the page, and only 2 pages are visible in the pane.  So you are jumping back to page 4, but placing it right at the top so only it and page 5 are visible.

I'll have to think about how this should be handled

Gazook89:
Perhaps before calculating the next range to scroll to, I can call `scrollToPage` with the first page in the original array to move it fully into view which would minimize the length of the visiblePages array, and then go forward with the rest of the "previous" button logic.

Gazook89:
@G-Ambatte i pushed a change, see if it's better.  The behavior still isn't entirely consistent between the "Previous" and the "Next" buttons, but it should no longer skip a whole page entirely.     

When you are viewing pages `6 - 7` at 100% zoom, and
- you do "Previous": it will jump to `5 ` with page five aligned at top of pane.
- you do "Next": it will jump to `7`

When you are viewing pages `50 - 53` at 39% zoom, and
- you do "Previous": it will jump to `48 - 49 ` with page 48 aligned at top of pane.
- you do "Next": it will jump to `53 - 54` with page 53 aligned at top of pane.

So basically when you are advancing a "page", it will show you the last page of your previous state plus whatever else registers as 'visible'.

And when you are going back a "page", it will show you a completely new set of pages that immediately precede your previous set.

When only *one* page is registered as "visible" it just works as expected. 

Is this acceptable?

G-Ambatte:
> @G-Ambatte i pushed a change, see if it's better. The behavior still isn't entirely consistent between the "Previous" and the "Next" buttons, but it should no longer skip a whole page entirely.
> 
> When you are viewing pages `6 - 7` at 100% zoom, and
> 
>     * you do "Previous": it will jump to `5 ` with page five aligned at top of pane.
> 
>     * you do "Next": it will jump to `7`
> 
> 
> When you are viewing pages `50 - 53` at 39% zoom, and
> 
>     * you do "Previous": it will jump to `48 - 49 ` with page 48 aligned at top of pane.
> 
>     * you do "Next": it will jump to `53 - 54` with page 53 aligned at top of pane.
> 
> 
> So basically when you are advancing a "page", it will show you the last page of your previous state plus whatever else registers as 'visible'.
> 
> And when you are going back a "page", it will show you a completely new set of pages that immediately precede your previous set.
> 
> When only _one_ page is registered as "visible" it just works as expected.
> 
> Is this acceptable?

Sorry, just realized I didn't respond to this. From testing, it appears to operate as expected. Looks good!

Gazook89:
Some changes made and things fixed.  I think it's outlined well enough in the commit messages, but quick notes:

- The 'fit to view' zoom button, which normally sets the zoom so that a single full page fits just within the preview pane, now sets the zoom to fit two pages side-by-side *if the current spread is 'facing'*. 
- This logic is changed:

    > When you are viewing pages 50 - 53 at 39% zoom, and you do "Next": it will jump to 53 - 54 with page 53 aligned at top of pane.

    Now, it will jump you to page 54 at the start of the range and any other pages after it.

- did a small css change to `justify-content` properties that I didn't know existed until now:  they are now `safe center` in both *facing* and *flow* spreads.  Basically, they are centered unless there isn't space enough for one page, in which case it becomes basically `flex-start` (aligned to the left).

**Oh, and, `master` has been merged in**.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -185,7 +215,10 @@ const ToolBar = ({ displayOptions, currentPage, totalPages, onDisplayOptionsChan
 					className='previousPage tool'
 					type='button'
 					title='Previous Page(s)'
-					onClick={()=>scrollToPage(pageNum - 1)}
+					onClick={()=>{
+						const rangeOffset = visiblePages.length > 1 ? 1 : 0;
+						scrollToPage(_.min(visiblePages) - visiblePages.length + rangeOffset);
+					}}
 					disabled={pageNum <= 1}
```
5e-Cleric, line 222: This line should be updated following your new functionality, so it is disabled when the page range starts from 1, and the next page button when the page range ends in the last page. The CSS will provide the styling when the disabled attribute is set.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -214,7 +247,18 @@ const ToolBar = ({ displayOptions, currentPage, totalPages, onDisplayOptionsChan
...
+						// the other pages will always be the same (since the viewport doesn't change).
+						// So this needs to scroll to the 'max', then see what is newly in view,
+						// and if the same pages are visible, do it again but +1.
+						const start = _.max(visiblePages);
+						scrollToPage(start);
+						if(start === _.max(visiblePages)){
+							scrollToPage(start + 1);
+						};
+					}}
 					disabled={pageNum >= totalPages}
```
5e-Cleric, line 262: as previously mentioned, this needs updating.

5e-Cleric:
The following is more like a personal take on this, but i think it can be useful, in the jsx for the page numebr input add a manual style attribute like so:

`style={{width:`${pageNum.toString().length + 2}ch`}}`

and in the CSS changing the width to set a `max-width:10ch` or whatever looks reasonable.

This should keep a small input when a small value is set, and grow with it to a maximum we deem appropiate. It accounts for the number of characters in the pageNum string, plus 2 for good measure, acting like a faux padding.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: done and....

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: ...done

Gazook89:
> The following is more like a personal take on this, but i think it can be useful, in the jsx for the page numebr input add a manual style attribute like so:
> 
> `` style={{width:`${pageNum.toString().length + 2}ch`}} ``
> 
> and in the CSS changing the width to set a `max-width:10ch` or whatever looks reasonable.
> 
> This should keep a small input when a small value is set, and grow with it to a maximum we deem appropiate. It accounts for the number of characters in the pageNum string, plus 2 for good measure, acting like a faux padding.

The idea is fine, but I'm going to push through a slightly different implementation.   I'm not going to do `+2ch` just to get some padding-- i can just use `padding` for that.  Also, I'd rather have `min-width` setup than `max-width`, so that it starts with ~5 character width and then can grow to accommodate any length.   

*you might think this can get out of control if someone types in a super long number...but the input width won't expand until the toolbar is rerendered, which doesn't happen until the number is 'submitted', at which point it'll be knocked down to the 'max' page which will be a more reasonable character length, hopefully*

What about this?  Need more padding or is it fine?

image

image

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -32,16 +34,16 @@ const ToolBar = ({ displayOptions, currentPage, totalPages, onDisplayOptionsChan
 			setPageNum(parseInt(pageInput)); // input type is 'text', so `page` comes in as a string, not number.
 	};
 
+	// scroll to a page, used in the Prev/Next Page buttons.
```
5e-Cleric, line 37: Not sure if this comment actually adds anything

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -69,6 +77,28 @@ const ToolBar = ({ displayOptions, currentPage, totalPages, onDisplayOptionsChan
...
+		if(pages.length === 0) return '';
+
+		const sortedPages = [...pages].sort((a, b)=>a - b); // Copy and sort the array
+		const ranges = [];
+		let start = sortedPages[0];
+
+		for (let i = 1; i <= sortedPages.length; i++) {
+			// If the current page is not consecutive or it's the end of the list
+			if(i === sortedPages.length || sortedPages[i] !== sortedPages[i - 1] + 1) {
+				// Push the range to the list
```
5e-Cleric, line 91: Same as before, L91 comment doesn't add anything. Is this from the toaster?

5e-Cleric (approved):
Overall looks great!

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: I think as I wrote it i was just erring on the side of "too many comments" rather than too few.  Just before writing this reply i was reading a blog post that had this bit in it:

> You know how wired earbuds always, always get tangled when you place them in a drawer or your pocket or something for few seconds? Codebases do the exact same thing, when there are multiple people writing on things, fixing each others' bugs. Everyone has a different way of thinking, so it's only natural that things over time get a bit tangled up.
> ...
> When going through old code, especially some that has barely any comments, it can take a very long time to understand what is actually going on. I honestly spent most of my time trying to understand how the thing even works, what is called when, where the icons positions are updated, and so on.

I don't know if it's worth keeping this comment in or not, but right now, if it were me, I would.  I don't think it's bad to have a comment above each function describing what it does and where it's called even if it's pretty obvious to us right now what it does.

Someone else can remove it or resolve the conversation however they see fit.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: It probably is a remainder from cGPT, though I'm not sure it should be removed.  It's part of a series of comments on each step of the method, so it'd be odd to leave one step uncommented.  Someone else can remove it or do whatever to resolve this.

5e-Cleric: yes, but a comment that says `push the range to the list`, then the next line is `range.push(` is pretty redundant IMO

calculuschild:
One odd thing. Not necessarily a dealbreaker, but I'm noticing on the facing-page view, scrolling between pages will add a weird flicker where one of the other pages temporarily appears in the list. You can see i this video it's entering a state with 500, 502-503 visible. Looks like this is only if the panel is partially covering up the right-side pages so we hit a point where the left side is in view but not the right side. Could possibly be a small tweak in a later PR to only list sequential pages or something.

https://github.com/user-attachments/assets/13bbd470-0f7d-4452-bd47-24630255f7d3

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -214,8 +248,19 @@ const ToolBar = ({ displayOptions, currentPage, totalPages, onDisplayOptionsChan
...
+						// then the 'max'/last page in view will always be the same, and
+						// the other pages will always be the same (since the viewport doesn't change).
+						// So this needs to scroll to the 'max', then see what is newly in view,
+						// and if the same pages are visible, do it again but +1.
+						const start = _.max(visiblePages);
+						scrollToPage(start);
+						if(start === _.max(visiblePages)){
+							scrollToPage(start + 1);
+						};
+					}}
```
calculuschild, line 262: I'm not sure I understand this process. Don't we always want to jump to `_max(visiblePages) + 1` from the get-go?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -36,8 +36,46 @@ const BrewPage = (props)=>{
...
+		// Observer for tracking the page at the center of the iframe.
+		const centerObserver = new IntersectionObserver(
+			(entries)=>{
+				entries.forEach((entry)=>{
+					if(entry.isIntersecting)
+						props.onCenterPageChange(props.index + 1); // Set this page as the center page
+				});
+			},
+			{ threshold: 0, rootMargin: '-50% 0px -50% 0px' } // Detect when the page is at the center
+		);
```
calculuschild, line 67: Can you just clarify what is the purpose of the centerPage observer? Is it just to handle the edge case where no page is 30% in view (a very small window, zoomed in a lot, etc)?

calculuschild: *If* that's the case, I don't see any way to do this without two observers, but it feels kinda weird. However, I wonder if we can reduce some of the extra structure that has to be passed around just to get this to work.

Any chance we can have `centerObserver` also call `onVisibilityChange`, thus eliminating `onCenterPageChange()` and the associated `centerPage` state?

5e-Cleric:
I have said it before and say it again, in single page view, no two pages should be displayed in the current page input. Retain the current working  while adding new. Please.

Gazook89:
@calculuschild 

> I'm noticing on the facing-page view, scrolling between pages will add a weird flicker in the toolbar where one of the other pages temporarily appears in the list. 

So this because it is showing which pages have n% of their area inside the view.  If you've got the panel open so your pages are pushed beyond the right edge of the window, and then you scroll so that vertically you are moving more and more of the page out of view, suddenly there isn't enough of the page in view.  Basically it's saying "these pages are most predominantly in view" and ignores slivers.  What that percentage is can be adjusted if we don't think it's quite right.

I specifically allowed for disconnected multiple ranges because I wanted to focus on what the user is actually seeing in most cases, rather than saying "here are some page numbers, even ones not in view".   That said, my original purpose with that was because I was contemplating that "Flow" might actually allow a user to set *n* pages in a row, rather than use flexbox-- then users could zoom and pan around a grid of pages, and I'd want to show actual in-view pages rather than a single range.   I'm open to whatever in terms of this.

(Also, your mp4 video doesn't work for me, so I'm just kind of assuming we are talking about the same thing).

@5e-Cleric 

> I have said it before and say it again, in single page view, no two pages should be displayed in the current page input. Retain the current working while adding new. Please.

When single page view is zoomed out so that you can see multiple pages, is it better to only have one page number displayed, or all of them?  

And in my experience, if I have scrolled so that I'm working at the bottom of a page and so have 2 pages pretty much equally split in my Preview window, I like to be sure I know which page is which.  At a glance, `88-89` is easier for me to quickly parse as "the page at the top is `88` and the page on the bottom is `89`", rather than a single number and guessing which page it is referring to.

My *guess*, and it's just a guess, is that displaying 2 page numbers at once is mostly noticeable because you are *really* looking at it.  Testing the PR and scrolling around rapidly between pages while solely watching that counter is not really equivalent to just typing out or reading page after page.

I guess I would also be concerned about having different implementations for different views and not carrying over the same expectations.

client/homebrew/brewRenderer/brewRenderer.jsx
Gazook89: The centerPage is what is used for the jumpTo function, for jumping the editor, as well as what you said.  In terms of "weirdness", I had the same thought at the time and I asked around on some discord channel, blogs, and cGPT (i know i know...) and basically the consensus is that multiple observers isn't really a problem at all.  

I believe that for jumpTo function, the `centerPage` state is needed so that just one integer is being used, rather than some array.  

I have to go pickup my kid soon so I haven't actually pulled this back open before commenting now, so i can look again later and maybe have a different answer then.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
Gazook89: Intuitively, I think you are right.  But i just remember fussing with this so much and I'm not sure why, and why I left such a big comment on something seemingly pretty simple.  It could just be a remnant scrap from earlier iteration that didn't get cleaned up....or, it portends doom to anyone that might involve themselves with deeply cursed code.  

As I said, you are probably right, but I need to come back with more time and retry it on my local before saying "yes/no".

client/homebrew/brewRenderer/brewRenderer.jsx
calculuschild: For jumpto, I think we can just extract like the last or first or middle page from viewablePages.

I get the two observers needed to check two different types of "in view". That's fine. Mostly I don't like how centerPage feels like a subset of viewablePages and yet requires 2 functions and a state to keep track of. Their use overlaps so closely it feels like there should be a way to merge the two somehow.

calculuschild:
> So this because it is showing which pages have n% of their area inside the view.
> I'm open to whatever in terms of this.

Yeah this sounds like what I was talking about.

I think the way the viewing experience currently works, the grid view reflows so all pages in view are always contiguous (they don't flow outside the viewport). Which means the only time you can get the separate ranges is if you are doing something weird at a very large zoom level or very narrow panel looking at the corners of four pages. I think at that point it might make more sense to just display a single range as the min-to-max visible pages.

> If I have scrolled so that I'm working at the bottom of a page and so have 2 pages pretty much equally split in my Preview window, I like to be sure I know which page is which.  At a glance, `88-89` is easier for me to quickly parse as "the page at the top is `88` and the page on the bottom is `89`", rather than a single number and guessing which page it is referring to.

I think I agree with this in terms of unifying the experience across modes. Though if I understand the code right, it would be a simple check to just always use the centerPage when in single-page mode.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
calculuschild: Going to just simplify this. I don't see any behavior difference.

client/homebrew/brewRenderer/brewRenderer.jsx
calculuschild: I have left both observers as well as separate "visible" and "center" states, but merged some of the logic to reduce the number of functions passed around.

calculuschild:
Managed to trim this down to about 60% the original added length. Retested everything and all appears to be still in order.

I'm going to leave the listing of multiple pages in single-page mode, since you can easily see more than 1 page if you zoom out. Tested that it is a simple change to make, but I think I side with @Gazook89 on this for now. Can be debated in a later issue.

Going to merge this now. Good work @Gazook89 !

PR #3851, opened 2024-10-23
Refactor UserPage as a functional component

5e-Cleric:
This PR turns the UserPage Component into a functional component, reducing the overall code in around 20 lines of code.

Next step should be doing the same refactor to listPage.

This should be pretty easy to review.

### DIFF
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -1,12 +1,11 @@
 const React = require('react');
-const { useState, useEffect } = require('react');
+const { useState } = React;
 const _ = require('lodash');
 
 const ListPage = require('../basePages/listPage/listPage.jsx');
 
 const Nav = require('naturalcrit/nav/nav.jsx');
 const Navbar = require('../../navbar/navbar.jsx');
-
 const RecentNavItem = require('../../navbar/recent.navitem.jsx').both;
 const Account = require('../../navbar/account.navitem.jsx');
 const NewBrew = require('../../navbar/newbrew.navitem.jsx');
@@ -14,13 +13,12 @@ const HelpNavItem = require('../../navbar/help.navitem.jsx');
 const ErrorNavItem = require('../../navbar/error-navitem.jsx');
 const VaultNavitem = require('../../navbar/vault.navitem.jsx');
 
-const UserPage = (props) => {
-    const { username = '', brews = [], query = '', error = null } = props;
-    const usernameWithS = username + (username.endsWith('s') ? `’` : `’s`);
+const UserPage = ({ username = '', brews = [], query = '', error = null, ...props }) => {
 
-    const groupedBrews = _.groupBy(brews, (brew) => {
-        return brew.published ? 'published' : 'private';
-    });
+    const [currentError, setCurrentError] = useState(error);
+
+    const usernameWithS = username + (username.endsWith('s') ? `’` : `’s`);
+    const groupedBrews = _.groupBy(brews, brew => brew.published ? 'published' : 'private');
 
 	const brewCollection = [
 		{
@@ -34,12 +32,6 @@ const UserPage = (props) => {
 			brews: groupedBrews.private || []
 		}] : [])
 	];	
-
-    const [currentError, setCurrentError] = useState(error);
-
-    const errorReported = (error) => {
-        setCurrentError(error);
-    };
 
     const navItems = () => (
         <Navbar>
@@ -55,8 +47,7 @@ const UserPage = (props) => {
     );
 
     return (
-        <ListPage brewCollection={brewCollection}  navItems={navItems()} query={query} reportError={errorReported}
-        />
+        <ListPage brewCollection={brewCollection}  navItems={navItems()} query={query} reportError={(error) =>setCurrentError(error)} />
     );
 };
 
--- a/client/homebrew/pages/userPage/userPage.jsx
+++ b/client/homebrew/pages/userPage/userPage.jsx
@@ -13,42 +13,42 @@ const HelpNavItem = require('../../navbar/help.navitem.jsx');
 const ErrorNavItem = require('../../navbar/error-navitem.jsx');
 const VaultNavitem = require('../../navbar/vault.navitem.jsx');
 
-const UserPage = ({ username = '', brews = [], query = '', error = null, ...props }) => {
+const UserPage = ({ username = '', brews = [], query = '', error = null, ...props })=>{
 
-    const [currentError, setCurrentError] = useState(error);
+	const [currentError, setCurrentError] = useState(error);
 
-    const usernameWithS = username + (username.endsWith('s') ? `’` : `’s`);
-    const groupedBrews = _.groupBy(brews, brew => brew.published ? 'published' : 'private');
+	const usernameWithS = username + (username.endsWith('s') ? `’` : `’s`);
+	const groupedBrews = _.groupBy(brews, (brew)=>brew.published ? 'published' : 'private');
 
 	const brewCollection = [
 		{
-			title: `${usernameWithS} published brews`,
-			class: 'published',
-			brews: groupedBrews.published || []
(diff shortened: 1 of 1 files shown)


client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -14,69 +14,50 @@ const HelpNavItem = require('../../navbar/help.navitem.jsx');
 const ErrorNavItem = require('../../navbar/error-navitem.jsx');
 const VaultNavitem = require('../../navbar/vault.navitem.jsx');
 
-const UserPage = createClass({
-	displayName     : 'UserPage',
```
Gazook89, line 18: Just leaving a note that possibly we may want to add something like:

```
UserPage.displayName = 'UserPage'
```
at the end of the component (after the closing bracket).  My understanding right now is that the build process may automatically be doing this, inferring the name from the function name, so we probably *don't* need to do this.  So really, just a note that if we start to notice in error messages that the component name isn't showing, we can come back and add this ourselves.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -14,69 +14,50 @@ const HelpNavItem = require('../../navbar/help.navitem.jsx');
...
+			brews: groupedBrews.published || []
+		},
+		...(username === global.account?.username ? [{
+			title: `${usernameWithS} unpublished brews`,
+			class: 'unpublished',
+			brews: groupedBrews.private || []
+		}] : [])
+	];	
+
+    const [currentError, setCurrentError] = useState(error);
```
Gazook89, line 38: My preference would be to have the `state` stuff placed right at the top of the component below props, before any other functions since that is traditionally where we have placed it.  And it's nice to see what variables are *props* and which are *state* right at the beginning.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -14,69 +14,50 @@ const HelpNavItem = require('../../navbar/help.navitem.jsx');
...
+
+    const errorReported = (error) => {
```
Gazook89, line 40: Can this be renamed to `handleReportError`?  My gut feeling is that this would be easier to understand as a function handling a child component prop, following the same convention elsewhere.  As-is, it seems like a regular "local helper" function like `usernameWithS`.

In fact, we could likely get rid of the function entirely by just setting up `ListPage` like this:

```
<ListPage brewCollection={brewCollection}  navItems={navItems()} query={query} reportError={()=>setCurrentError(error)} />
```

*note, you can also throw the last `/>` on line 59 on the same line as the rest of it since it's not doing much on it's own, and reduce your lines of code by 1*

client/homebrew/pages/userPage/userPage.jsx
5e-Cleric: Hmm, good point, from the docs:

> The displayName string is used in debugging messages. Usually, you don’t need to set it explicitly because it’s inferred from the name of the function or class that defines the component. You might want to set it explicitly if you want to display a different name for debugging purposes or when you create a higher-order component.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -14,69 +14,50 @@ const HelpNavItem = require('../../navbar/help.navitem.jsx');
...
-			</Nav.section>
-		</Navbar>;
-	},
-
-	render : function(){
-		return <ListPage brewCollection={this.state.brewCollection} navItems={this.navItems()} query={this.props.query} reportError={this.errorReported}></ListPage>;
-	}
-});
+const UserPage = (props) => {
+    const { username = '', brews = [], query = '', error = null } = props;
```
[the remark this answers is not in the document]
5e-Cleric: In these cases, i'd recommend making a suggestion instead of a code block, that way we can change it with a click 😄

Gazook89 (changes requested):
I started by leaving some individual comments before thinking i would complete a review, apologies for that.

Overall looks pretty simple.  I would just move some things around and remove a function that I think only calls a setState function.

5e-Cleric:
All comments visited, and found some other opportunities to reduce number of lines without sacrificing legibility.

client/homebrew/pages/userPage/userPage.jsx
calculuschild: Yeah I don't think it's needed for functional components.

client/homebrew/pages/userPage/userPage.jsx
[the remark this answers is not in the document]
calculuschild: This is fine, but just my two cents, I prefer the `props` object for all the props. Makes it very clear which values are coming from the parent component.

My preference is this format:

```
const myComponent = (props) => {
    props = { username = '', brews = [], query = '', error = null, ...props };
```
or spread out
```
const myComponent = (props) => {
	props = {
	    username = '',
	    brews    = [],
	    query    = '',
	    error    = null,
	    ...props
	};
```

client/homebrew/pages/userPage/userPage.jsx
Gazook89: I read somewhere that it doesn't automatically grab the name if you are using arrow functions rather than named functions.  But then somewhere else said that Babel or other build tools take care of it.  So this was just sort of a "hey, in case things don't work later, park this is in the back of your brain".  I'm just going to mark this thread resolved.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -1,82 +1,54 @@
...
-	}
-});
+const UserPage = ({ username = '', brews = [], query = '', error = null, ...props }) => {
+
+    const [currentError, setCurrentError] = useState(error);
+
+    const usernameWithS = username + (username.endsWith('s') ? `’` : `’s`);
+    const groupedBrews = _.groupBy(brews, brew => brew.published ? 'published' : 'private');
+
+	const brewCollection = [
```
calculuschild, line 23: There are a couple of weird space/tab areas. Just run linting on this and I think it is good to merge.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -1,82 +1,54 @@
...
+            <Nav.section>
+                {currentError && (<ErrorNavItem error={currentError} parent={null}></ErrorNavItem>)}
+                <NewBrew />
+                <HelpNavItem />
+                <VaultNavitem />
+                <RecentNavItem />
+                <Account />
+            </Nav.section>
+        </Navbar>
+    );
```
calculuschild, line 47: I *think* this doesn't need to be a function:

```suggestion
    const navItems = (
        <Navbar>
            <Nav.section>
                {currentError && (<ErrorNavItem error={currentError} parent={null}></ErrorNavItem>)}
                <NewBrew />
                <HelpNavItem />
                <VaultNavitem />
                <RecentNavItem />
                <Account />
            </Nav.section>
        </Navbar>
    );
```

client/homebrew/pages/userPage/userPage.jsx
5e-Cleric: not at home atm, if you can run the linter and merge, nice, if not, i will be home in 2 hours.

calculuschild: I cannot either atm.

client/homebrew/pages/userPage/userPage.jsx
```diff
@@ -1,82 +1,54 @@
...
+
+    return (
+        <ListPage brewCollection={brewCollection}  navItems={navItems()} query={query} reportError={(error) =>setCurrentError(error)} />
```
calculuschild, line 50: With above change, will need to change navItems here too:

```suggestion
        <ListPage brewCollection={brewCollection}  navItems={navItems} query={query} reportError={(error) =>setCurrentError(error)} />
```

5e-Cleric:
Kinda forgot about this one, linted, merged and ready to go, also destructured the props separately as mentioned in other PRs, @calculuschild should be ready for merge

PR #3853, opened 2024-10-23
fix snippetbar weirdness

5e-Cleric:
adds another hardcoded breakpoint for style tab, removes the snippets element from the properties tab, and sets correct styles for all wrappings


calculuschild:
Maybe this is just me, but I think all tabs should have the same limit so the toolbar buttons aren't shifting around when you change tabs:

https://github.com/user-attachments/assets/2f2743c0-e873-410e-ae55-29211b6fd984

5e-Cleric:
> Maybe this is just me, but I think all tabs should start wrapping at the same width so the toolbar buttons aren't shifting around when you change tabs:

I wish i could satisfy this need, but the wrapping is automatic when the two items collide, i am not manually setting the wrap, just the expansion of the items.

https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap

calculuschild:
> I wish i could satisfy this need, but the wrapping is automatic when the two items collide, i am not manually setting the wrap, just the expansion of the items.

I used the wrong wording then. The expansion of the items is what I am talking about. I would like the toolbar buttons to expand at the same panel width on all three tabs.

Gazook89:
You mean actually add padding to just the `snippets` div that makes it the same width as the Brew Editor `snippets` div?  If so, I think that would be tough, because not every theme will have the same set of menus on Brew or Style tabs.

calculuschild:
> You mean actually add padding to just the `snippets` div that makes it the same width as the Brew Editor `snippets` div?  If so, I think that would be tough, because not every theme will have the same set of menus on Brew or Style tabs.

I was thinking there would be some way to `calc` it. Maybe not. 

I mostly just don't like the buttons changing places when I'm trying to switch between tabs.

calculuschild:
This looks great. Thank you. I will merge this now before anyone else comes in and objects. 😛

Gazook89:
Is this relevant to this discussion: https://shkspr.mobi/blog/2024/10/you-can-use-text-wrap-balance-on-icons/

5e-Cleric:
> Is this relevant to this discussion: https://shkspr.mobi/blog/2024/10/you-can-use-text-wrap-balance-on-icons/

Not really? we dont want buttons wrapping the text, i even set the text wrap to unset in the last PR.

Btw, you might want to check text-balance pretty, it is also pretty good

Gazook89:
From the article:

> But the name is, I think, slightly misleading. It doesn't only work on text. It will work on any content. For example - I have a row of icons at the bottom of this page. If the viewport is too narrow, a single icon might drop to the next line. That can look a bit weird.

So you can balance spans, divs, etc.


> Btw, you might want to check text-balance pretty, it is also pretty good

There isn't a `text-balance` property, so I assume you are talking about the exact same property `text-wrap: balance' that this article is talking about.

5e-Cleric:
> From the article:
> 
> > But the name is, I think, slightly misleading. It doesn't only work on text. It will work on any content. For example - I have a row of icons at the bottom of this page. If the viewport is too narrow, a single icon might drop to the next line. That can look a bit weird.
> 
> So you can balance spans, divs, etc.

any inline block is treated as a character, so yes. Still better used for text, in this PR, because it is already relying on flex, it is basically useless, flex-wrap works in an easier and more apt manner for the flow of elements.

> > Btw, you might want to check text-balance pretty, it is also pretty good
> 
> There isn't a `text-balance` property, so I assume you are talking about the exact same property `text-wrap: balance' that this article is talking about.

Yes, brain no braining, text-wrap pretty

PR #3735, opened 2024-09-19
TagInput - Functional component for tag-like inputs

Gazook89:
# Description
This supersedes PR #3469.   This PR is just the very basic functionality of tags:  add, remove, update.  Subsequent PRs will layer on additional features/functionality in discrete chunks and descriptions are below.  Each of the additional features is a branch on the previous PR-- they build on each other.  As such, they can merged in order, or the last PR can be merged in and it will contain everything prior to it.  Here is the breakdown:

<details><summary>#3735 - Part 1 - Basic Functions</summary>
The PR removes the current StringArrayEditor.jsx entirely, and replaces it with tagInput.jsx which creates a functional component called TagInput.  Obviously it doesn't have to be limited to only "tags" as we called them in HB-- it can be used for the Author list as well and where you might want a *string array editor*.  I just prefer the shorter name for brevity.

- [x] add tags
- [x] remove tags individually
- [x] edit existing tags
- [x] save tags to brew metadata
</details>

<details><summary>[No PR yet] - Part 2 - Keyboard Navigation</summary>

Part 2 focuses on adding keyboard navigation to the tag input.  Things like using `Enter`, `,`, or `Tab` to submit a new value, or tabbing/shift+tabbing to move around the tags.

[Here is the branch on my remote](https://github.com/Gazook89/homebrewery/tree/TagInput-Part-2-Keyboard-Nav).  

</details>

<details><summary>[No PR yet] - Part 3 - Validations</summary>

Part 3 adds methods to validate the tag inputs so that we can control what values are input.  It differs from the original StringArrayEditor by having the validator functions centralized with other inputs in `validations.js`, rather than declaring the functions directly in props or as `valuePatterns`.  

[Here is the branch on my remote](https://github.com/Gazook89/homebrewery/tree/TagInput-Part-3-Validations). 

</details>

<details><summary>[No PR yet] - Part 4 - Uniqueness</summary>

Part 4 is basically a supplement to Part 3-- it adds a check to see if the new value is unique from the existing tags before submitting it.  By default, it must be unique, but via props the component could accept duplicate values.

[Here is the branch on my remote](https://github.com/Gazook89/homebrewery/tree/TagInput-Part-4-Uniqueness). 

</details>

## Quick Overview of TagInput
<details><summary>Click to expand</summary>

[...]

PR #3854, opened 2024-10-24
Rework combobox.jsx to functional component

Gazook89:
## Description

This is a rework of the Combobox component.  It is now a Functional component.  It makes a few small changes to structure (the options are all buttons now).  It also adds better keyboard navigation (which is pretty much all of the "new lines of code").  Beyond that, it is mostly the same at this point.

Right now, only the Language picker uses this component.  I started to rework it because I want to use it in the TagInput component (aka StringArrayEditor).  

Beyond whatever is suggested in review, some additional things I want to do:

- [x] add a "expand"/"open" button to the right side of the input as a button to open it (in truth, the whole input toggles the menu so this will just be a little visual indicator).
- [ ] a quick pass on accessibility issues (roles, hidden navigation)
- [ ] get the validators working with it again.

Then, as a subsequent PR, I want to work out a good way to get submenus to work.

- Related #3735

Try out the Language picker, using your mouse and navigating it with the keyboard.  You should be able to:

- Tab into it and focus the input
- Space to open the menu with all options, Space again to close it
- Type any characters to open menu to filtered options
- Typing Space should still work to add whitespace, without opening/closing the menu, as long as the input isn't otherwise empty
- Esc to close the drop down
- Enter to select an option
- ArrowDown from input to open and navigate through menu, cycling to first option if reached the bottom
- ArrowUp to move through menu, focusing the input when you move past the first option
- when focus is on menu option, be able to use Backspace to move focus to input and delete characters.
- Tab out to next input closes the dropdown and moves focus.
- entering a value that filters to zero results displays a 'no matches' item.

Possibly some of these behaviors will change as i look at WCAG for comboboxes and I'll try to update this list.  

### Reviewer Checklist

```
	- [ ] Tab into it and focus the input
	- [ ] Space to open the menu with all options, Space again to close it
	- [ ] Type any characters to open menu to filtered options
	- [ ] Typing Space should still work to add whitespace, without opening/closing the menu, as long as the input isn't otherwise empty
	- [ ] Esc to close the drop down
	- [ ] Enter to select an option
	- [ ] ArrowDown from input to open and navigate through menu, cycling to first option if reached the bottom
	- [ ] ArrowUp to move through menu, focusing the input when you move past the first option
	- [ ] when focus is on menu option, be able to use Backspace to move focus to input and delete characters.
	- [ ] Tab out to next input closes the dropdown and moves focus.
	- [ ] entering a value that filters to zero results displays a 'no matches' item.
  - [ ] Still correctly sets the `lang` attribute and is preserved with the brew metadata
```

</details>


client/components/combobox.jsx
```diff
@@ -1,128 +1,166 @@
...
+
+	const handleInputChange = (evt)=>{
+		const newValue = evt.target.value;
+		setInputValue(newValue);
+		setCurrentOption(-1);
+
+		const filtered = React.Children.toArray(props.children).filter((option)=>{
+			return autoSuggest.filterOn.some((filterAttr)=>{
+				return option.props[filterAttr]?.toLowerCase().startsWith(newValue.toLowerCase());
+			});
```
5e-Cleric, line 45: I believe we can make this using `some`, making it more concise without losing readability, make the change if you like

```suggestion
const filtered = props.children.filter(({ props }) =>
  autoSuggest.filterOn.some(filterAttr =>
    props[filterAttr]?.toLowerCase().startsWith(newValue.toLowerCase())
  )
);

```

client/components/combobox.jsx
```diff
@@ -1,128 +1,166 @@
...
+			}
+
+			// Prevent modifier keys from triggering dropdown (example, shift+tab or tab to move focus)
+			else if(!modifiers.includes(evt.key)) {
+				setShowDropdown(true);
+			}
+		}
+
+
+	};
```
5e-Cleric, line 113: With that many else ifs, we might benefit from using the hated switch:

```suggestion
    const handleKeyDown = (evt) => {
    const modifiers = ['Meta', 'Shift', 'Alt', 'Control', 'Tab'];
    const isArrowDown = evt.key === 'ArrowDown';
    const isArrowUp = evt.key === 'ArrowUp';
    const optionsLength = filteredOptions.length;

    if (inputFocused || currentOption >= 0) {
        switch (evt.key) {
            case ' ':
                if (inputValue === '') {
                    evt.preventDefault();
                    setShowDropdown(prev => !prev);
                }
                break;
                
                // ArrowDown moves to the next option
            case 'ArrowDown':
                evt.preventDefault();
                setShowDropdown(true);
                setCurrentOption((currentOption + 1) % optionsLength);
                break;
                
                // ArrowUp moves to the previous option
            case 'ArrowUp':
                evt.preventDefault();
                const newIndex = currentOption - 1;
                setCurrentOption(newIndex >= 0 ? newIndex : -1);
                if (newIndex === -1) inputRef.current.focus();
                break;
                
                // Escape key closes the dropdown
            case 'Escape':
                setShowDropdown(false);
                setCurrentOption(-1);
                inputRef.current.focus();
                break;
                
                // Backspace key while menu is open still deletes characters in input
            case 'Backspace':
                if (showDropdown) inputRef.current.focus();
                break;
            case 'Tab':
                setShowDropdown(false);
                setCurrentOption(-1);
                break;
                
                // Prevent modifier keys from triggering dropdown (example, shift+tab or tab to move focus)
            default:
                if (!modifiers.includes(evt.key)) setShowDropdown(true);
                break;
        }
    }
};

```

Please double check it in case i made many errors.

client/components/combobox.less
```diff
@@ -23,16 +27,26 @@
...
+			all: unset;
+			display: block;
+			width: 100%;
+			position:relative;
             font-size: 11px;
             font-family: Open Sans;
             padding: 5px;
             cursor: default;
             margin: 0 3px;
+			box-sizing: border-box;
```
5e-Cleric, line 43: I believe our css applies this to all elements, consider changing `all:unset` if the approach is taking more lines than it should (you probably already have, but must suggest it)

client/components/combobox.less
```diff
@@ -23,16 +27,26 @@
...
+			margin: 0;
             //border-bottom: 1px solid darkgray;
             &:hover {
+                background-color: rgb(223, 221, 221);
+			}
+			&:focus {
                 filter: brightness(120%);
                 background-color: rgb(163, 163, 163);
             }
```
5e-Cleric, line 52: Overall, i'd say keep the colors in hex whenever possible, less characters, same legibility IMO

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -240,16 +240,15 @@ const MetadataEditor = createClass({
...
-				return <div className='item' title={`${englishName.of(code)}`} key={`${index}`} data-value={`${code}`} data-detail={`${localName.of(code)}`}>
-					{`${code}`}
-					<div className='detail'>{`${localName.of(code)}`}</div>
-				</div>;
-			});
-		};
+		const langOptions = _.map(langCodes.sort(), (code, index)=>{
+			const localName = new Intl.DisplayNames([code], { type: 'language' });
+			const englishName = new Intl.DisplayNames('en', { type: 'language' });
+			return <button className='item' title={`${englishName.of(code)}`} key={`${index}`} data-value={`${code}`} data-detail={`${localName.of(code)}`}>
```
5e-Cleric, line 246: can't all of these be written as `title={englishName.of(code)}`?

5e-Cleric (commented):
Overall looks fine, i made more suggestions than i thought, sorry for making them separate.

client/components/combobox.jsx
Gazook89: This is nearly the exact same as what I have unless I’m missing something or don’t know how to read a diff. 

The React.Children.toArray is important before there may be only a single child, not an array, so this would convert it to an array. Or, it could have multiple children but just not passed as an array. 

But regarding `some`, pretty sure I’m already using it the exact same way?

client/components/combobox.jsx
Gazook89: I considered it, but I opted for else if before I personally think it is more readable and understandable (don’t have to worry about “making many errors”. It doesn’t save many (any?) lives, either (though I can see I have some extra white space). 

I’ll reconsider if Switch/Case has a huge amount of support among reviewers

client/components/combobox.jsx
5e-Cleric: basically, alright, then this is resolved. Did not think of the single element possibility

client/components/combobox.jsx
Gazook89: I'm going to mark this as 'resolved' for right now mostly because it's enormous amount of text in the middle of the thread.  If anyone else comments that they want Switch/Case statements rather than else/if, I'll *switch* it.

client/components/combobox.less
Gazook89: Largely the `all: unset` is to unset everything we have currently setup for *every* `button` on the site:

```css
button {
	-webkit-transition: background-color .25s ease;
	-moz-transition: background-color .25s ease;
	-o-transition: background-color .25s ease;
	transition: background-color .25s ease;
	display: inline-block;
	padding: .6em 1.2em;
	cursor: pointer;
	background-color: #27AE60;
	font-family: 'Open Sans',sans-serif;
	font-size: .8em;
	font-weight: 800;
	color: white;
	text-decoration: none;
	text-transform: uppercase;
	border: none;
	outline: none;
}
```

Probably it'd be ideal to change this styling to be based on *some* class, rather than styling every button with a very heavy-handed style.  As we start to use actual button elements rather than just `div` elements with `onClick` attributes, we are going to want to style those buttons in certain contexts ("navbar", "snippetbar", "viewbar", etc).

If I made that change in this PR, though, the PR would quickly spread to cover changes in many more files.  So probably left as it's own thing for now.  Instead, I'll just do `all: unset` and then build the style from the ground up.

Without putting much thought on it, I think the current "default button" styling should apply to the class name `call-to-action`, though I bet there are better names.

client/components/combobox.less
Gazook89: Agreed, thanks for pointing that out.  I think it's a remnant from when i was trying to do tricky color manipulations here.  I went further and lightened the grays a bit more too.  This will be resolved when I push up my changes, so marking it now.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
Gazook89: yes, good point, thanks -- fixed

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
+	const [inputValue, setInputValue] = useState(props.value || '');
+	const [showDropdown, setShowDropdown] = useState(false);
+	const [inputFocused, setInputFocused] = useState(false);
+	const [currentOption, setCurrentOption] = useState(-1);
+	const [filteredOptions, setFilteredOptions] = useState(React.Children.toArray(props.children));
+	const inputRef = useRef(null);
+	const optionRefs = useRef([]);
+	const componentRef = useRef(null);
+
+	useEffect(()=>{
```
calculuschild, line 15: Here, you are re-defining `handleClickOutside()` every time `showDropdown` changes value. I don't think this is necessary; you can define the function outside of `useEffect`. Same for the other functions below.

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
-	componentWillUnmount : function() {
-		if(this.props.trigger == 'click')
-			document.removeEventListener('click', this.handleClickOutside);
-	},
-	handleClickOutside : function(e){
-		// Close dropdown when clicked outside
-		if(this.refs.dropdown && !this.refs.dropdown.contains(e.target)) {
-			this.handleDropdown(false);
+
+		document.addEventListener('pointerdown', handleClickOutside);
```
calculuschild, line 22: This is also adding a new event listener every time `showDropdown` changes.

If you want to add a listener once when the component first mounts, change the `[showDropdown]` to an empty array `[]`

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
+	// Handle keyboard navigation
+	const handleKeyDown = (evt)=>{
+		const modifiers = ['Meta', 'Shift', 'Alt', 'Control', 'Tab'];
+		if(inputFocused || (currentOption >= 0)) {
+			const optionsLength = filteredOptions.length;
+
+			if(evt.key === ' ' && (inputValue === '' || (inputRef.current.selectionStart === 0 && inputRef.current.selectionEnd === inputValue.length))){
+				evt.preventDefault();
+				setShowDropdown(!showDropdown);
+			}
```
calculuschild, line 61: What is the purpose of this one?

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
-			this.handleDropdown(false);
+
+		document.addEventListener('pointerdown', handleClickOutside);
+		return ()=>{document.removeEventListener('pointerdown', handleClickOutside);};
+	}, [showDropdown]);
+
+	useEffect(()=>{
+		onSelect(inputValue);
+		// handleInputChange({ target: { value: inputValue } });
+	}, [inputValue]);
```
calculuschild, line 29: `useEffect` is not usually not needed for this type of case. Here, we trigger two renders, first when `inputValue` changes, and then again when `onSelect()` changes something in the parent `metaDataEditor`.

Instead of setting `inputValue` and then handling side effects later in `useEffect`, both `setInputValue` and calling `onSelect()` should happen together in a handler function, probably in your `handleInputChange()` function.

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
-					}
+		setFilteredOptions(filtered);
+	};
+
+	/* eslint-disable brace-style */
+
+	// Handle keyboard navigation
+	const handleKeyDown = (evt)=>{
+		const modifiers = ['Meta', 'Shift', 'Alt', 'Control', 'Tab'];
+		if(inputFocused || (currentOption >= 0)) {
```
calculuschild, line 55: If we are already focused, why do we need to keep setting focus for the different keys below?

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
+		document.addEventListener('pointerdown', handleClickOutside);
+		return ()=>{document.removeEventListener('pointerdown', handleClickOutside);};
+	}, [showDropdown]);
+
+	useEffect(()=>{
+		onSelect(inputValue);
+		// handleInputChange({ target: { value: inputValue } });
+	}, [inputValue]);
+
+	useEffect(()=>{
```
calculuschild, line 31: Similar comment here. Changing `currentOption` *and* the side effect of focusing the element should happen inside of a handler function, not inside `useEffect`.

See https://react.dev/learn/you-might-not-need-an-effect

client/components/combobox.jsx
calculuschild: I'll add another vote for `switch`. This seems like the perfect use case for that.

client/components/combobox.jsx
```diff
@@ -1,128 +1,167 @@
...
-			},
-		};
-	},
-	getInitialState : function() {
-		return {
-			showDropdown : false,
-			value        : '',
-			options      : [...this.props.options],
-			inputFocused : false
+const Combobox = ({ onSelect, onEntry, autoSuggest = { filterOn: ['data-value'], suggestMethod: 'includes', clearAutoSuggestOnClick: false }, ...props })=>{
```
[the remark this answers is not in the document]
Gazook89: donezo.

client/components/combobox.jsx
calculuschild: Sorry, missed it in your original comment. Ignore this.

client/components/combobox.jsx
Gazook89: Got it, good point on this one.  done.

client/components/combobox.jsx
Gazook89: If just change this to onMount `[]`, it doesn't work.  But it works if i remove the `showDropdown` condition in the handler:

```
const handleClickOutside = (evt)=>{
		console.log(componentRef.current, showDropdown)
		if(componentRef.current && !componentRef.current.contains(evt.target)) {
			setShowDropdown(false);
		}
	};
```

So that's what i'll do.

Gazook89:
Thanks for reviewing.  I'm going to leave some comments unaddressed because I'm in the middle of refiguring out how focus is managed and a few small behavior changes as part of the "align with WCAG" portion of this.  I didn't think I'd get two code reviews so quick :)  but they were still valuable and I changed a few things as suggested.

client/components/combobox.jsx
calculuschild: Without any dependency array it now adds a new listener on *every* render, which is even worse.

Let's see if we can figure out why `[]` didn't work

5e-Cleric:
its been a year and a half, is it okay if we close this? it maybe should be done from scratch

Gazook89:
Going to close this and come back to it, maybe, with smaller & better commits and goals.

PR #3855, opened 2024-10-26
Refactor splitpane as a functional component

5e-Cleric:
The next functional refactor in this chaos of a run.

the component still works in the same way.

### DIFF
--- a/shared/naturalcrit/splitPane/splitPane.jsx
+++ b/shared/naturalcrit/splitPane/splitPane.jsx
@@ -3,112 +3,90 @@ const React = require('react');
 const { useState, useEffect, useRef } = React;
 const cx = require('classnames');
 
-const SplitPane = (props) => {
-	props = {
-		storageKey         : 'naturalcrit-pane-split',
-		onDragFinish       : function(){}, //fires when dragging
-		showDividerButtons : true,
-		...props
-	};
-	const pane1 = useRef(null);
-	const pane2 = useRef(null);
-	const [currentDividerPos, setCurrentDividerPos] = useState(null);
-	const [userSetDividerPos, setUserSetDividerPos] = useState(null);
-	const [windowWidth, setWindowWidth] = useState(null);
+const SplitPane = (props)=>{
+	const {
+		storageKey = 'naturalcrit-pane-split',
+		onDragFinish = ()=>{},
+		showDividerButtons = true
+	} = props;
+
 	const [isDragging, setIsDragging] = useState(false);
+	const [dividerPos, setDividerPos] = useState(null); // Initial divider position is set to `null`
 	const [moveSource, setMoveSource] = useState(false);
 	const [moveBrew, setMoveBrew] = useState(false);
 	const [showMoveArrows, setShowMoveArrows] = useState(true);
 	const [liveScroll, setLiveScroll] = useState(false);
 
-	const storageKey = props.storageKey || 'naturalcrit-pane-split';
-	const onDragFinish = props.onDragFinish || (() => {});
+	const dividerRef = useRef(null);
 
-	// Fetch saved divider position and scroll state on mount
-	useEffect(() => {
-		setWindowWidth(window.innerWidth);
-		const dividerPos = window.localStorage.getItem(storageKey);
-		const liveScrollSetting = window.localStorage.getItem('liveScroll') === 'true';
-		setLiveScroll(liveScrollSetting);
+	// Set initial divider position and liveScroll only after mounting
+	useEffect(()=>{
+		const savedPos = window.localStorage.getItem(storageKey);
+		setDividerPos(savedPos ? parseInt(savedPos, 10) : window.innerWidth / 2);
+		setLiveScroll(window.localStorage.getItem('liveScroll') === 'true');
 
-		if (dividerPos) {
-			const limitedPos = limitPosition(dividerPos, 0.1 * (window.innerWidth - 13), 0.9 * (window.innerWidth - 13));
-			setCurrentDividerPos(limitedPos);
-			setUserSetDividerPos(dividerPos);
-		} else {
-			setCurrentDividerPos(window.innerWidth / 2);
-			setUserSetDividerPos(window.innerWidth / 2);
-		}
-
-		const handleResize = () => {
-			const newPos = limitPosition(userSetDividerPos, 0.1 * (window.innerWidth - 13), 0.9 * (window.innerWidth - 13));
-			setCurrentDividerPos(newPos);
-			setWindowWidth(window.innerWidth);
+		const handleResize = ()=>{
+			setDividerPos((pos)=>limitPosition(pos,0.1 * (window.innerWidth - 13), 0.9 * (window.innerWidth - 13))
+			);
 		};
 		window.addEventListener('resize', handleResize);
+		return ()=>window.removeEventListener('resize', handleResize);
+	}, [storageKey]);
 
-		return () => {
-			window.removeEventListener('resize', handleResize);
-		};
-	}, []);
-
-	const limitPosition = (x, min = 1, max = window.innerWidth - 13) => {
+	const limitPosition = (x, min = 1, max = window.innerWidth - 13)=>{
 		return Math.round(Math.min(max, Math.max(min, x)));
 	};
 
-	const handleUp = (e) => {
+	const handleUp =(e)=>{
 		e.preventDefault();
-		if (isDragging) {
-			onDragFinish(currentDividerPos);
-			window.localStorage.setItem(storageKey, currentDividerPos);
+		if(isDragging) {
+			onDragFinish(dividerPos);
+			window.localStorage.setItem(storageKey, dividerPos);
 		}
 		setIsDragging(false);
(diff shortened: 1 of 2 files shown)


shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
-const SplitPane = createClass({
-	displayName     : 'SplitPane',
-	getDefaultProps : function() {
-		return {
-			storageKey         : 'naturalcrit-pane-split',
-			onDragFinish       : function(){}, //fires when dragging
-			showDividerButtons : true
+const SplitPane = (props)=>{
+	const {
+		storageKey = 'naturalcrit-pane-split',
```
calculuschild, line 8: storageKey doesn't need to be a prop. It is never passed in by any parent component.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
+
+	// Set initial divider position and liveScroll only after mounting
+	useEffect(()=>{
+		const savedPos = window.localStorage.getItem(storageKey);
+		setDividerPos(savedPos ? parseInt(savedPos, 10) : window.innerWidth / 2);
+		setLiveScroll(window.localStorage.getItem('liveScroll') === 'true');
+
+		const handleResize = ()=>{
+			setDividerPos((pos)=>limitPosition(pos,0.1 * (window.innerWidth - 13), 0.9 * (window.innerWidth - 13))
+			);
```
calculuschild, line 30: `handleResize` (and in general, most functions) don't need to be defined inside of `useEffect()`. You can move this out into the normal component flow.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
-
-	limitPosition : function(x, min = 1, max = window.innerWidth - 13) {
-		const result = Math.round(Math.min(max, Math.max(min, x)));
-		return result;
-	},
-
-	handleUp : function(e){
+		window.addEventListener('resize', handleResize);
+		return ()=>window.removeEventListener('resize', handleResize);
+	}, [storageKey]);
```
calculuschild, line 34: This should probably just be an empty `[]` so it only fires once, instead of any time `storageKey` updates.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
+	);
+};
+
+const Pane = ({ width, children, isDragging, className, moveBrew, moveSource, liveScroll, setMoveArrows })=>{
+	const styles = width
+		? { flex: 'none', width: `${width}px` }
+		: { pointerEvents: isDragging ? 'none' : 'auto' };
+
+	return (
+		<div className={cx('pane', className)} style={styles}>
```
calculuschild, line 115: Shouldn't need `cx`. Can just do ``className={`pane ${className}`}``

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
+			{renderDivider()}
+			<Pane isDragging={isDragging}>{props.children[1]}</Pane>
+		</div>
+	);
+};
+
+const Pane = ({ width, children, isDragging, className, moveBrew, moveSource, liveScroll, setMoveArrows })=>{
+	const styles = width
+		? { flex: 'none', width: `${width}px` }
+		: { pointerEvents: isDragging ? 'none' : 'auto' };
```
calculuschild, line 112: Let's keep this comment

```suggestion
		: { pointerEvents: isDragging ? 'none' : 'auto' }; //Disable mouse capture in the right pane; else dragging into the iframe drops the divider
```

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
+		const newSize = limitPosition(e.pageX);
+		setDividerPos(newSize);
+	};
+
+	const liveScrollToggle = ()=>{
+		window.localStorage.setItem('liveScroll', String(!liveScroll));
+		setLiveScroll(!liveScroll);
+	};
+
+	const moveArrows = showMoveArrows && (
```
calculuschild, line 66: Let's continue to call this `renderMoveArrows` for consistency with all the other functions that output JSX.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
+				id='scrollToggleDiv'
+				className={`arrow ${liveScroll ? 'lock' : 'unlock'}`}
+				onClick={liveScrollToggle}
+			>
+				<i id='scrollToggle' className={`fas fa-${liveScroll ? 'lock' : 'unlock'}`} />
+			</div>
+		</>
+	);
+
+	const renderDivider = ()=>(
```
calculuschild, line 87: `renderDivider` doesn't need to be a function either I guess.

shared/naturalcrit/splitPane/splitPane.jsx
calculuschild: Actually, I don't think `Pane` is passed in a className from anywhere. I think we can remove all references to it and just to `className="pane"`.

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
-	},
-
-	componentWillUnmount : function() {
-		window.removeEventListener('resize', this.handleWindowResize);
-	},
-
-	handleWindowResize : function() {
-		// Allow divider to increase in size to last user-set position
-		// Limit current position to between 10% and 90% of visible space
-		const newLoc = this.limitPosition(this.state.userSetDividerPos, 0.1*(window.innerWidth-13), 0.9*(window.innerWidth-13));
```
calculuschild, line 58: There is one behavior difference I note. Maybe it's not critical but wanted to point it out. In the original behavior, changing the window size will remember the last divider position set by the user and *grow back to that size* if the window becomes larger again. In this PR, shrinking the window and making it large again does not regain the divider position.

First is the original behavior. Then I switch tabs to the new behavior.
https://github.com/user-attachments/assets/c4f7b6a7-f833-40af-86e9-4e65e5c57dbb

shared/naturalcrit/splitPane/splitPane.jsx
```diff
@@ -1,200 +1,121 @@
...
+	} = props;
+
+	const [isDragging, setIsDragging] = useState(false);
+	const [dividerPos, setDividerPos] = useState(null); // Initial divider position is set to `null`
+	const [moveSource, setMoveSource] = useState(false);
+	const [moveBrew, setMoveBrew] = useState(false);
+	const [showMoveArrows, setShowMoveArrows] = useState(true);
+	const [liveScroll, setLiveScroll] = useState(false);
+
+	const dividerRef = useRef(null);
```
calculuschild, line 20: Looks like this dividerRef doesn't do anything. It's set and then never referenced anywherer.

shared/naturalcrit/splitPane/splitPane.jsx
5e-Cleric: Alright, this is left from the original comp. resolved

shared/naturalcrit/splitPane/splitPane.jsx
5e-Cleric: Very interesting, i'll have a look

5e-Cleric:
```
const newLoc = this.limitPosition(this.state.userSetDividerPos, 0.1*(window.innerWidth-13), 0.9*(window.innerWidth-13));

this.setState({
			currentDividerPos : newLoc,
			windowWidth       : window.innerWidth
		});
```

Seems identical to 

`setDividerPos((pos)=>limitPosition(pos, 0.1 * (window.innerWidth - 13), 0.9 * (window.innerWidth - 13)));`

So i don't see where this could come from

Note: i erased the windowWidth state because it was not being referenced, just set. Not even in the original code.

calculuschild: It's the `userSetDividerPos`, not `windowWidth`.

calculuschild: Do you think this is a behavior worth holding on to or no?

5e-Cleric: I think it is worth maintaining this behavior, if it was intentional.

5e-Cleric: Fixed it without the extra state, which was causing issues:


https://github.com/user-attachments/assets/fac00044-4eb6-4df1-9e56-e5431adba140

5e-Cleric:
also updated comments as i saw fit, and added a class to the divider when dragging, to prevent some distasteful flickering effect when dragging.

calculuschild:
Ok, I think this is good now. Merging!

PR #3861, opened 2024-10-29
Vault styles quickfix

5e-Cleric:
i did not check vault when testing #3823 , and removing one of those divs broke styling at vault:

image

this fixes that


5e-Cleric:
It is a 3 line change but linting makes it look long.

PR #3863, opened 2024-10-30
Fix crashes if title is invalid as URI

5e-Cleric:
This PR wraps reddit link generation encodeURI in a try catch block to prevent crashes if the title has characters that are not being able to be encoded.

to be more specific, the following string: `𓀔𓀇𓀅𓀋𓀡𓀡𓀕𓀠𓀧𓀨𓀣𓀷𓀷𓀿𓀿𓁀𓁶𓁰 𓁴𓁿𓂀𓁾𓁵𓁯𓂞𓂤𓂗𓃃𓂾𓂺𓂹 𓃞𓃙𓃖𓃓𓃕𓃓𓃜𓃘𓃙𓃟𓃛𓃞 𓂺𓃂𓂿𓂺𓃃�` breaks the reddit encoding url process, yes, i know this is a stupidly wrong title, but it has revealed a weird bug.

When that string is parsed to generate the reddit URL, and the function crashes, it breaks the whole app, giving the following error in heroku:

```
2024-12-15T12:04:50.412169+00:00 app[web.1]: URIError: URI malformed
2024-12-15T12:04:50.412184+00:00 app[web.1]:     at encodeURIComponent (<anonymous>)
2024-12-15T12:04:50.412185+00:00 app[web.1]:     at ReactClassComponent.getRedditLink (/app/build/homebrew/ssr.cjs:4682:76)
2024-12-15T12:04:50.412185+00:00 app[web.1]:     at ReactClassComponent.renderNavbar (/app/build/homebrew/ssr.cjs:4706:18)
2024-12-15T12:04:50.412186+00:00 app[web.1]:     at ReactClassComponent.render (/app/build/homebrew/ssr.cjs:4720:14)
2024-12-15T12:04:50.412186+00:00 app[web.1]:     at Gb (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.production.min.js:65:127)
2024-12-15T12:04:50.412187+00:00 app[web.1]:     at Ib (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.production.min.js:67:211)
2024-12-15T12:04:50.412197+00:00 app[web.1]:     at W (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.production.min.js:73:89)
2024-12-15T12:04:50.412197+00:00 app[web.1]:     at Ib (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.production.min.js:67:482)
2024-12-15T12:04:50.412198+00:00 app[web.1]:     at W (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.production.min.js:73:89)
2024-12-15T12:04:50.412198+00:00 app[web.1]:     at Ib (/app/node_modules/react-dom/cjs/react-dom-server-legacy.node.production.min.js:71:304)
```

Sorry for making another PR, this is low prio.


calculuschild:
Can you provide a string that actually fails here? I want to test this.  I can't seem to reproduce it with the string you shared on Gitter Oct 30:

> if you are curious, the string that showed this error was: 𓀔𓀇𓀅𓀋𓀡𓀡𓀕𓀠𓀧𓀨𓀣𓀷𓀷𓀿𓀿𓁀𓁶𓁰 𓁴𓁿𓂀𓁾𓁵𓁯𓂞𓂤𓂗𓃃𓂾𓂺𓂹 𓃞𓃙𓃖𓃓𓃕𓃓𓃜𓃘𓃙𓃟𓃛𓃞 𓂺𓃂𓂿𓂺𓃃�

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -380,7 +380,15 @@ const EditPage = createClass({
...
-		return `https://www.reddit.com/r/UnearthedArcana/submit?title=${encodeURIComponent(title)}&text=${encodeURIComponent(text)}`;
+		let encodedTitle;
+		try {
+			encodedTitle = encodeURIComponent(title);
+		} catch (error) {
+			console.error("Error encoding title for Reddit link:", error);
+			encodedTitle = encodeURIComponent("Check out my homebrew!"); // Fallback title
+		}
+	
+		return `https://www.reddit.com/r/UnearthedArcana/submit?title=${encodedTitle}&text=${encodeURIComponent(text)}`;
```
calculuschild, line 391: Looks like `toWellFormed()` exists, which can take a string and replace any invalid characters with a placeholder. Might be more direct to fix the string so the error doesn't happen in the first place.

```suggestion
return `https://www.reddit.com/r/UnearthedArcana/submit?title=${encodeURIComponent(title.toWellFormed())}&text=${encodeURIComponent(text)}`;
```

calculuschild (changes requested):

client/homebrew/pages/editPage/editPage.jsx
5e-Cleric: is this function used elsewhere in our code?

calculuschild: I tested this with your broken document and can confirm `toWellFormed()` fixes the issue, so I will replace with that.

There is only one other place where we URI encode user-made content, when generating the filename for download. The download function already runs through the `sanitize-filename` package, which I tested and does handle this malformed title, so we don't need to change anything there.

PR #3866, opened 2024-10-31
Fix the `.list` node for the StringArrayEditor

Gazook89:
fixes the icon buttons on the stringArrayEditor/tagInput `.list` element so that they are inline, given the correct background, and aligned correctly.  Restores pink validation styling.

## Description

Here it is broken, on Live:
image

Here it is fixed:
image

- [gitter discussion](https://matrix.to/#/!IuzKyDJxDuanSbtTQt:gitter.im/$_08cnMbqx2srlhedAmE0h_skmtSEPFWyPmhxmFVqccs?via=gitter.im&via=matrix.org&via=integrations.ems.host)

Try using the string array editor (aka "tags" and "authors" inputs) in the Properties Editor.  The three states should all display properly (non-editing 'badge', editing badge, and empty badge/input).

## Add'l comments

Likely the metadataEditor.less file could be overhauled more completely, but I'm just focused on fixing this one thing.  In my TagInput PR there are significant changes to the .less file so I don't want to muck about too much in there right now if i don't have to.

### DIFF
--- a/client/homebrew/editor/metadataEditor/metadataEditor.less
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.less
@@ -245,25 +245,24 @@
 		display   : flex;
 		flex      : 1 0;
 		flex-wrap : wrap;
+		align-items: center;
 
 		> * { flex : 0 0 auto; }
 
 		#groupedIcon {
 			#backgroundColors;
 			position   : relative;
-			top        : -0.3em;
-			right      : -0.3em;
-			display    : inline-block;
+			height: 100%;
+			display    : flex;
+			align-items: center;
+			justify-content: center;
 			min-width  : 20px;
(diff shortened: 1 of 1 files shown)


calculuschild:
Does this one conflict with #3894 at all? Do you have a preference of which gets merged first?

Gazook89:
> Does this one conflict with https://github.com/naturalcrit/homebrewery/pull/3894 at all? Do you have a preference of which gets merged first?

I would just go with 3894, and this can be closed, reopening if necessary.  The tagInput PRs have a few more in the pipe, and i can't remember right now which bits of styling happen where but I'm sure I didn't take anything from this PR and reuse it (intentionally).

5e-Cleric:
Then closing this for now, as #3894 should fix the issue.

PR #3868, opened 2024-11-01
Snippetbar Rework: Popovers, Anchor Positioning, Enhancements

Gazook89:
### Background
This PR is primarily an effort to bring forward the use of the HTML attribute `popover` and the new [CSS Anchor Positioning API (view in Chrome)](https://developer.chrome.com/blog/anchor-positioning-api).  Right off the bat I'll say I'm not sure we want to go this route *right away*, because we are still waiting on Firefox to rollout anchor positioning.  [I'm eternally optimistic that it will be soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1838746).  But hopefully this can at least be illustrative of how `popover` and anchoring can work together in a UI.  

image

Using the `popover` attribute in an element means that the element is rendered to a **top layer** root element, basically a sibling to the `<html>` element.  This means it is higher in the layers than any other item, and so can't be clipped/hidden behind anything except other elements that come after it in **top layer**.  In the DOM, the element is still wherever you put it in your code, and can be styled just as you would normally expect, with a couple of exceptions....it's just rendered above everything else.

Popovers do lose some ability to position them.  In Firefox, where Anchor Positioning isn't a thing yet, this makes it difficult to place things like the menus in the right spot because your only option is to use absolute positioning *in relation to the viewport*;  you cannot position it relative to a parent/ancestor.  In Chrome, you can use Anchor Positioning on popover elements, which pretty much covers all our bases.

image

Anchor positioning is for positioning an element in relation to another.  Notably, the two elements do not need to be related to each other in the DOM-- they don't need to be parent/child/sibling etc.  You just give an element an anchor name, and then reference that name in the other element's css anchor position property. To set where the anchored element is positioned in relation to the anchor, there are a few easy keyword-based properties that are basically "position the anchored element's left top corner against the anchor's right top corner".  You can do more with sizing, margins, etc.  Anchor positioning can dynamically keep an element within the viewport until the element would otherwise be totally out of view, or switch the positioning if the anchored element doesn't fit anymore, etc.  Anchored elements can even have different corners anchored to different elements.

### So what, ignore Firefox?
So the big downer is that because Firefox doesn't have Anchor Positioning yet, it's actually pretty difficult (as far as i can tell), to have a decent fallback.  Popover elements lack some positioning abilities that would help mitigate the issue.  As this PR currently stands at time of writing, the Firefox experience is pretty disappointing. Some options:

- Pretend Firefox doesn't exist and just fire this off now (this is the current state; i don't love it).
- Use a library (Floating-UI) as basically a polyfill for Firefox, and try to write the code such that when Firefox adds Anchor Positioning it is easy to remove the polyfill.  Floating-UI is a popular library for doing everything Anchor Positioning does, before AP was added to browsers and is the basis for other very popular UI libraries.
- Remove all the Anchor Positioning stuff and just use Floating-UI for everything.  
- Sit on this PR until Firefox catches up.

### This is a big PR, what's in here anyways?
This started off as a quick effort to show that Anchor Positioning and Popovers are the way to fix the "menus clipping behind the preview pane" problem.  But as I had to make a little adjustment here and there, things got out of hand.  Particularly with snippetBar.less, which has been significantly altered.

- snippetBar.jsx: Big changes to the html structure of snippets, menus, and the bar itself.  Lots of changing divs to buttons, setting style objects via JS (to match anchors to anchored elements).  Refactored the history items to use Moment.js which is already used in the repo.
- snippetBar.less:  Refactored a lot of duplicate or unneeded lines.  Widely implemented CSS Grid and subgrids in the snippet menus to make things line up better, such as when a submenu trigger doesn't have an associated left-hand icon.  Added a tinge of color to the editor tabs when they are hovered or active.
- reset.less, core.less: made some changes to the styling for buttons.  Update reset.less to include `button` so that UA styling is reset.  Update core.less so that not all buttons have the dramatic styling applied to them which later need to be overridden.
- Some other small changes to make everything work.  Should be decent-ish notes about those in the commit history.

Even if this PR is rejected, I think there is a lot of value in taking the changes to metadataEditor.less as their own thing and implementing them in another PR, which is what I should have done before I got carried away.

### What's left?
Fix the firefox issue.

And I still think there is opportunity for better display of the snippet bar, particularly around the editor tabs.  They should have full text labels since they are very important buttons and have the least intuitive icons (in my opinion).  But I think that can be tackled in a greater re-envisioning of the editor pane.

I may also take a crack at revising the snippet menu html structure again to handle better keyboard navigation.

- Closes #

Right now, I'm mostly interested in what people think of how it actually works in the deployment.  Specifically, look at it in Chrome.  Also look in Firefox to see how crappy it is.  

I will update the reviewer checklist later.

### Reviewer Checklist

  - [ ] Feature A does X
  - [ ] Feature B does Y

<details><summary>Copy this list</summary></details>

### DIFF
--- a/client/components/dialog.jsx
+++ b/client/components/dialog.jsx
@@ -2,7 +2,7 @@
 import React from "react";
 const { useRef, useEffect } = React;
 
-function Dialog({ dismisskeys, closeText = 'Close', blocking = false, ...rest }) {
+function Dialog({ dismisskeys = [], closeText = 'Close', blocking = false, ...rest }) {
 	const dialogRef = useRef(null);
 
 	useEffect(()=>{
--- a/client/homebrew/brewRenderer/toolBar/toolBar.less
+++ b/client/homebrew/brewRenderer/toolBar/toolBar.less
@@ -13,6 +13,7 @@
 	height           : auto;
 	padding          : 2px 0;
 	font-family      : 'Open Sans', sans-serif;
+	font-size        : 13px;
 	color            : #CCCCCC;
 	background-color : #555555;
 	& > *:not(.toggleButton) {
@@ -90,10 +91,6 @@
 		width            : auto;
 		min-width        : 46px;
 		height           : 100%;
-		padding          : 0 0px;
-		font-weight      : unset;
-		color            : inherit;
-		background-color : unset;
 		&:hover { background-color : #444444; }
 		&:focus { outline : 1px solid #D3D3D3; }
 		&:disabled {
--- a/client/homebrew/editor/metadataEditor/metadataEditor.less
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.less
@@ -79,6 +79,7 @@
 				text-overflow : ellipsis;
 			}
 			button {
+				.colorButton();
 				padding          : 0px 5px;
 				color            : white;
 				background-color : black;
@@ -138,16 +139,16 @@
 		margin-bottom : 15px;
 		button { width : 100%; }
 		button.publish {
-			.button(@blueLight);
+			.colorButton(@blueLight);
 		}
 		button.unpublish {
-			.button(@silver);
+			.colorButton(@silver);
 		}
 	}
 
 	.delete.field .value {
 		button {
-			.button(@red);
+			.colorButton(@red);
 		}
 	}
 	.authors.field .value {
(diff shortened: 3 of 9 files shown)


5e-Cleric:
Should we block this until firefox comes around then?

5e-Cleric:
> Do you mind also disabling the editor themes button in the properties panel? given that this PR touches snippetbar anyway.

This is not necessary anymore.

Related to this PR, in the PR to make the snippetbar wrap correctly, i may have introduced a visual issue we could fix here:

image
Meaning, the name popout we have for the tools show to the left always, and are now clipped behind the screen side, for the first item.

dbolack-ab:
What are the downsides if we add the polyfill - assuming the polyfill plays nice?

calculuschild:
Closing in favor of #1213

PR #3869, opened 2024-11-01
Fix-splitpane-overflowing-page

5e-Cleric:
reverts half the changes in #3823, basically brings back a deleted div that changed an overflow, which was hidden but caused visual issues.


Gazook89 (approved):
This all looks well and good to me.  I experience the issue on my local branches since merging `master` into them, and on the staging site, but issue seems to be resolved when i test this PR locally.  No notes.

PR #3870, opened 2024-11-03
Local snippet editor

dbolack-ab:
This adds brew-based user snippets.

This follows the document pattern using `snippet Snippet Name` instead of page to indicate the start of a snippet.

A new `Brew Snippets` menu is added at the end of the existing menu list with Snippets organized under their owning document names (themeBundles return their snippets as well)

To Do:
- [ ] Highlighting
- [ ] Determine what/if snippet should populate the snippet editor

To test:
- [ ] Add a Snippet to a theme brew
  - [ ] Ensure the snippet shows up on a brew using the theme.
  - [ ] Ensure the snippet inserts
- [ ] Add a snippet to the brew using the theme above
  - [ ] Ensure that both menu items appear
  - [ ] Ensure the snippet insterts

Example Snippet
```
\snippet Example Snippet
Hello!
I'll be your snippet for the day!
```

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -21,6 +21,7 @@ const DEFAULT_STYLE_TEXT = dedent`
 					color: black;
 				}`;
 
+const DEFAULT_SNIPPET_TEXT = ``;
 let isJumping = false;
 
 const Editor = createClass({
@@ -35,6 +36,7 @@ const Editor = createClass({
 			onTextChange  : ()=>{},
 			onStyleChange : ()=>{},
 			onMetaChange  : ()=>{},
+			onSnipChange  : ()=>{},
 			reportError   : ()=>{},
 
 			onCursorPageChange : ()=>{},
@@ -51,7 +53,7 @@ const Editor = createClass({
 	getInitialState : function() {
 		return {
 			editorTheme : this.props.editorTheme,
-			view        : 'text' //'text', 'style', 'meta'
+			view        : 'text' //'text', 'style', 'meta', 'snip'
 		};
 	},
 
@@ -61,6 +63,7 @@ const Editor = createClass({
 	isText  : function() {return this.state.view == 'text';},
 	isStyle : function() {return this.state.view == 'style';},
 	isMeta  : function() {return this.state.view == 'meta';},
+	isSnip  : function() {return this.state.view == 'snip';},
 
 	componentDidMount : function() {
 
@@ -457,6 +460,20 @@ const Editor = createClass({
 					onChange={this.props.onMetaChange}
 					reportError={this.props.reportError}
 					userThemes={this.props.userThemes}/>
+			</>;
+		}
+
+		if(this.isSnip()){
+			return <>
+				<CodeEditor key='codeEditor'
+					ref={this.codeEditor}
+					language='gfm'
+					view={this.state.view}
+					value={this.props.brew.snippets ?? DEFAULT_SNIPPET_TEXT}
+					onChange={this.props.onSnipChange}
+					enableFolding={true}
+					editorTheme={this.state.editorTheme}
+					rerenderParent={this.rerenderParent} />
 			</>;
 		}
 	},
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -158,7 +158,7 @@ const Editor = createClass({
 
 	highlightCustomMarkdown : function(){
 		if(!this.codeEditor.current) return;
-		if(this.state.view === 'text')  {
+		if(this.state.view === 'text') {
 			const codeMirror = this.codeEditor.current.codeMirror;
 
 			codeMirror.operation(()=>{ // Batch CodeMirror styling
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -499,7 +499,7 @@ const Editor = createClass({
 					historySize={this.historySize()}
 					currentEditorTheme={this.state.editorTheme}
 					updateEditorTheme={this.updateEditorTheme}
-					snippetBundle={this.props.snippetBundle}
+					themeBundle={this.props.themeBundle}
 					cursorPos={this.codeEditor.current?.getCursorPosition() || {}}
 					updateBrew={this.props.updateBrew}
 				/>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ orbs:
 jobs:
   build:
     docker:
-      - image: cimg/node:20.17.0
+      - image: cimg/node:20.18.0
       - image: mongo:4.4
 
     working_directory: ~/homebrewery
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -6,6 +6,7 @@ const _     = require('lodash');
 const cx    = require('classnames');
 
 import { loadHistory } from '../../utils/versionHistory.js';
+import { brewSnippetsToJSON } from '../../../../shared/helpers.js';
 
 //Import all themes
 const ThemeSnippets = {};
@@ -114,6 +115,9 @@ const Snippetbar = createClass({
 
 			oldSnippets = _.keyBy(compiledSnippets, 'groupName');
 		}
+		const userSnippetsasJSON = brewSnippetsToJSON(this.props.brew.title, this.props.brew.snippets, this.props.snippetsBundle);
+		compiledSnippets.push(userSnippetsasJSON);
+
 		return compiledSnippets;
 	},
 
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -41,6 +41,7 @@ const Snippetbar = createClass({
 			unfoldCode        : ()=>{},
 			updateEditorTheme : ()=>{},
 			cursorPos         : {},
+			themeBundle       : [],
 			updateBrew        : ()=>{}
 		};
 	},
@@ -64,7 +65,7 @@ const Snippetbar = createClass({
 	},
 
 	componentDidUpdate : async function(prevProps, prevState) {
-		if(prevProps.renderer != this.props.renderer || prevProps.theme != this.props.theme || prevProps.snippetBundle != this.props.snippetBundle) {
+		if(prevProps.renderer != this.props.renderer || prevProps.theme != this.props.theme || prevProps.themeBundle != this.props.themeBundle) {
 			this.setState({
 				snippets : this.compileSnippets()
 			});
@@ -101,15 +102,12 @@ const Snippetbar = createClass({
 	},
 
 	compileSnippets : function() {
-		console.log('compileSnippets');
 		let compiledSnippets = [];
 
 		let oldSnippets = _.keyBy(compiledSnippets, 'groupName');
 
-		console.log(this.props.themesBundle);
-
-		if( this.props.themesBundle) {
-			for (let snippets of this.props?.themesBundle?.snippets) {
+		if(this.props.themeBundle.snippets) {
+			for (let snippets of this.props.themeBundle.snippets) {
 				if(typeof(snippets) == 'string')	// load staticThemes as needed; they were sent as just a file name
 					snippets = ThemeSnippets[snippets];
 
@@ -119,8 +117,8 @@ const Snippetbar = createClass({
 				oldSnippets = _.keyBy(compiledSnippets, 'groupName');
 			}
 		}
-		console.log(this.props.themesBundle);
-		const userSnippetsasJSON = brewSnippetsToJSON(this.props.brew.title, this.props.brew.snippets, this.props?.themesBundle?.snippets);
+
+		const userSnippetsasJSON = brewSnippetsToJSON(this.props.brew.title, this.props.brew.snippets, this.props.themeBundle.snippets);
 		compiledSnippets.push(userSnippetsasJSON);
 
 		return compiledSnippets;
@@ -265,6 +263,10 @@ const Snippetbar = createClass({
 				<div className={cx('meta', { selected: this.props.view === 'meta' })}
 					onClick={()=>this.props.onViewChange('meta')}>
 					<i className='fas fa-info-circle' />
+				</div>
+				<div className={cx('snip', { selected: this.props.view === 'snip' })}
+					onClick={()=>this.props.onViewChange('snip')}>
+					<i className='fas fa-th-list' />
 				</div>
 			</div>
 
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -65,7 +65,10 @@ const Snippetbar = createClass({
 	},
 
 	componentDidUpdate : async function(prevProps, prevState) {
-		if(prevProps.renderer != this.props.renderer || prevProps.theme != this.props.theme || prevProps.themeBundle != this.props.themeBundle) {
+		if(prevProps.renderer != this.props.renderer ||
+			prevProps.theme != this.props.theme ||
+			prevProps.themeBundle != this.props.themeBundle ||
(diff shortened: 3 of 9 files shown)


Gazook89:
What marks the end of a snippet

ericscheid:
How is this different from using variables to insert blocks of text?

dbolack-ab:
> How is this different from using variables to insert blocks of text?
> 
> (aside from the editor UI — which could be done for variables too/instead)

Abstractly? Only that a menu exists. Practically? It is a better organizational scheme with semi-self-documentation via the menu.

When I suggested creating user snippets via variables it got an unfavorable reaction.

calculuschild:
> How is this different from using variables to insert blocks of text?

A variable could insert a block of text, but that block would not be editable.

Example, a snippet that inserts a custom NPC statblock. The user can then manually edit the NPC's name, stats, etc. to match their use.

If they just put a variable `$[customStatBlock]`, there isn't a straightforward way to now change the internal statblock text.

This reveals the key difference: a variable enables easy duplication of a fixed text. A snippet enabled easy insertion of a template text, intended to be filled out and edited.

5e-Cleric:
Can we get a deployment for this? @calculuschild

5e-Cleric:
I'm sure you are aware, but the deployment is not working atm

dbolack-ab:
> I'm sure you are aware, but the deployment is not working atm

Checking what I can

dbolack-ab:
> I'm sure you are aware, but the deployment is not working atm

I think  calc fixed it. It's working now.

5e-Cleric:
Okay, checking this, all base snippets are completely gone, is that intended? Shouldn't the base theme snippets appear with the custom ones?

5e-Cleric:
I also can't write at all in the snippet editor tab.

dbolack-ab:
> I also can't write at all in the snippet editor tab. No error in the console, or anything weird, it just doesn't work.

Test it on /edit.

The other paths aren't set up yet.

5e-Cleric:
Is this intended to make css snipets as well?

5e-Cleric:
> ~What marks the end of a snippet~
> 
> Edit: oh, I see there is a whole new editor tab in addition to Brew, Styles, Properties. So nothing needs to indicate the end of a snippet because presumably each snippet would start with `\snippet` and continue until the next `\snippet`. I was thinking this was something built in to the existing Brew editor.

no? i tried and this:

```
/snippet
{{color:red

my name

}}

/snippet

aaa
```

produces just one snippet

dbolack-ab:
> > ~What marks the end of a snippet~
> > Edit: oh, I see there is a whole new editor tab in addition to Brew, Styles, Properties. So nothing needs to indicate the end of a snippet because presumably each snippet would start with `\snippet` and continue until the next `\snippet`. I was thinking this was something built in to the existing Brew editor.
> 
> no? i tried and this:
> 
> ```
> /snippet
> {{color:red
> 
> my name
> 
> }}
> 
> /snippet
> 
> aaa
> ```
> 
> produces just one snippet

You need to name the Snippet.

```
/snippet My name in Red
{{color:red

my name

}}

/snippet aaa

aaa
```

5e-Cleric:
A few requests:

- [ ] disable the `brew snippets` button (not hide) in the snippetbar if there are no snippets to hide.
- [ ] move the properties tab to the end again, i think that should be last

dbolack-ab:
> A few requests:
> 
> * [ ]  disable the `brew snippets` button (not hide) in the snippetbar if there are no snippets to hide.
> * [ ]  move the properties tab to the end again, i think that should be last

Added

Gazook89:
It doesn't look like the emoji syntax works.  I think that will work once you have highlighting in (or at least, I think it's handled in the same markdown.js file for the regular editor).  

<del>I was wondering if a snippet could be given an icon using our standard set of font icon libraries.  As-is, I don't have any ideas how that icon would be loaded into the existing icon slot we use in other menus, but we could still allow people to write `\snippet :image: Inline Image` and get something closely approximating those menus.  But I suppose they won't render in the snippet menu anyway.....</del>

5e-Cleric:
> It doesn't look like the emoji syntax works. I think that will work once you have highlighting in (or at least, I think it's handled in the same markdown.js file for the regular editor).

unrelated error, fixed in #3911.

dbolack-ab:
> > It doesn't look like the emoji syntax works. I think that will work once you have highlighting in (or at least, I think it's handled in the same markdown.js file for the regular editor).
> 
> unrelated error, fixed in #3911.

Thanks! I was wondering about that!

dbolack-ab:
> A few requests:
> 
> * [x]  disable the `brew snippets` button (not hide) in the snippetbar if there are no snippets to show.
> * [x]  move the properties tab to the end again, i think that should be last

There's currently no styling for adding disabled to the class. What should it look like?

5e-Cleric:
> > A few requests:
> > 
> > * [x]   disable the `brew snippets` button (not hide) in the snippetbar if there are no snippets to show.
> > * [x]   move the properties tab to the end again, i think that should be last
> 
> There's currently no styling for adding disabled to the class. What should it look like?

I see two good options:

image

With `color:grey` and `cursor:not-allowed`

OR:

image

dbolack-ab:
> > > A few requests:
> > > 
> > > * [x]    disable the `brew snippets` button (not hide) in the snippetbar if there are no snippets to show.
> > > * [x]    move the properties tab to the end again, i think that should be last
> > 
> > 
> > There's currently no styling for adding disabled to the class. What should it look like?
> 
> I see two good options:
> 
> image
> 
> With `color:grey` and `cursor:not-allowed`
> 
> OR:
> 
> image

I lean towards the first one.

5e-Cleric:
image


deployment not working atm

5e-Cleric:
If you could fix these conflicts, i think this is closer to being merged.

5e-Cleric:
A little reminder, could you fix this?

5e-Cleric:
We could, in this PR or a next one, let users decide on an icon for the snippet, using our emoji syntax

Also, was there a possibility of making CSS snippets?

dbolack-ab:
> We could, in this PR or a next one, let users decide on an icon for the snippet, using our emoji syntax

Or not use one at all...

This has me wondering if we could/should allow users to set one as part of the naming of the snippet

`\snippet This is my snippet :someEmoticon:`

> Also, was there a possibility of making CSS snippets?

I hadn't really thought on it. It would need some additional logic and probably a distinct `\snippet` label if we want to segregate them. If we don't, then a little logic to duplicate it over to the Style tab.

Gazook89:
I don't think we need to have icons for every single snippet.  I've been thinking that we already have too many icons.  Few of the ones we have any clear meaning at all, and because every single option has one they don't really offer any visual benefit for easily finding a snippet.

5e-Cleric:
There are conflicts to fix, 

and 
image

If one uses the forward bar, the snippet is captured, but the name is wrong, same if no bar is set.

please make the regex look for the backwards bar and the word, not just the word.

dbolack-ab:
> There are conflicts to fix,
> 
> and image
> 
> If one uses the forward bar, the snippet is captured, but the name is wrong, same if no bar is set.
> 
> please make the regex look for the backwards bar and the word, not just the word.
> 
> As for the rest:
> 
> * [x]  Add a Snippet to a theme brew
>   
>   * [x]  Ensure the snippet shows up on a brew using the theme.
>   * [x]  Ensure the snippet inserts
> * [x]  Add a snippet to the brew using the theme above
>   
>   * [x]  Ensure that both menu items appear
>   * [x]  Ensure the snippet insterts

How did you generate this one?

5e-Cleric:
> > There are conflicts to fix,
> > and image
> > If one uses the forward bar, the snippet is captured, but the name is wrong, same if no bar is set.
> > please make the regex look for the backwards bar and the word, not just the word.
> > As for the rest:
> > 
> > * [x]   Add a Snippet to a theme brew
> >   
> >   * [x]   Ensure the snippet shows up on a brew using the theme.
> >   * [x]   Ensure the snippet inserts
> > * [x]   Add a snippet to the brew using the theme above
> >   
> >   * [x]   Ensure that both menu items appear
> >   * [x]   Ensure the snippet insterts
> 
> How did you generate this one?

As per my last comment:

`\snippet foo` correctly generates a snippet named `foo`
`/snippet foo` incorrectly generates a snippet named `/snipppet foo`
`snippet foo` incorrectly generates a snippet named `snippet foo`

dbolack-ab:
@5e-Cleric Should be a lot better now.

5e-Cleric:
So, now no snippets show in the homepage `\`

dbolack-ab:
> So, now no snippets show in the homepage `\`
> 
> Also, the syntax hightling should follow the correct syntax, if you can:

This was due to the highlighting not clearing...

5e-Cleric:
Right, that is great, the other issue however is still standing:

image

dbolack-ab:
> Right, that is great, the other issue however is still standing:
> 
> image

Another regression from merges. yay.

dbolack-ab:
Working now, but I don't know why.

5e-Cleric:
> Working now, but I don't know why.

my favourite bit.

5e-Cleric (approved):

calculuschild:
Does this offer a way to make nested snippet menus or apply an icon (perhaps via the `:icon:` syntax)?

In general, I think this PR works for now for merging, but I think as a next step (separate PR?) I would like to move away from the custom `\snippet` syntax.  YAML already is the standard "key:value metadata syntax" for Markdown and would allow for a nesting structure as well as secondary properties besides just snippet name and contents.

This would also greatly simplify things on our end because we can avoid parsing a custom `\snippet` syntax at all. Is YAML easy for users to write? Probably not at first glance, but the type of user who would make custom snippets is also probably the type to be willing to learn YAML.

But also, probably more user friendly: I think we could implement a fillable form with all the needed fields so people don't have to worry about the syntax at all.

But here's what YAML would look like for reference:
```
My Custom Snippet:
  content: |
    Custom Text Stuff
  icon: ':fas_dice:'
  subsnippets:
  
    - Subsnippet 1:
        content: |
          {{customThing1
            spread over
            multiple lines
          }}
        icon: ':fas_dice_one:'
        
    - Subsnippet 2:
        content: |
          {{customThing2}}
        icon: ':fas_dice_two'
        
My Other Snippet:
  content: |
    Other Text Stuff
  icon: ':fas_dice:'
  
My Snippet with No Icon:
  content: |
    More other Text Stuff
```

5e-Cleric:
I'd keep users away from yaml, unless we have some sort of intellisense for it, because indentation is harder to track than brackets IMO

calculuschild:
> Maybe we should move OUR snippets to YAML, from JSON, first if we go that route.
> 
> We’d be using what we preach, and we could also offer a help link directly to our repo snippets as a reference for new snippet creators.

So, I see what you mean here. But my first thought is that our internal code is Javascript, which is directly compatible with JSON. The Homebrewery interface is Markdown, which is directly compatible with YAML. There is not an inherent need to force them together.

However the main obstacle with doing it for our internal code, is we are not actually using pure JSON either. Pure JSON would not allow us to put function calls in the snippet generation. So we use `.js` files with objects. I don't think YAML allows embedding function calls in its properties either, so we are kind of stuck keeping it in Javascript.

> I'd keep users away from yaml, unless we have some sort of intellisense for it, because indentation is harder to track than brackets IMO

These are good points. Codemirror can be configured to display indents and syntax highlighting for YAML, so I don't think those alone are going to be blocking. But I also think a fillable form would again avoid these issues anyway.

client/homebrew/editor/editor.jsx
```diff
@@ -52,7 +60,7 @@ const Editor = createClass({
 	getInitialState : function() {
 		return {
 			editorTheme : this.props.editorTheme,
-			view        : 'text' //'text', 'style', 'meta'
+			view        : 'text' //'text', 'style', 'meta', 'snip'
```
Gazook89, line 63: personally i'd just name this tab `snippets` rather than `snip`-- it seems like an unnecessary abbreviation in something that'll get minified later.

5e-Cleric:
well, related to this, the default snippet is not loaded as a snippet (which i would expect it to) when loading new.

Also, changes in the snippet tab are not stored in localstorage, is this desired?

dbolack-ab:
1) I don't think there is any real need for nested user snippets other than compulsory over-organization. 
2) I've proposed dropping Icons on the user snippets menu - I don't think they provide enough value for the real estate
3) YAML is obtuse to IT people.  Exposing users to it seems mean.

With these first two things in mind, I see no benefit to moving users out of a pattern they already understand for documents. It is a little awkward, but short of translating it in a wizard and widget setup, I think it will do.

Best bet - wait on feedback. *I* expect this feature will only really be of use to power users who make templates and similar shared docs.

Gazook89:
Is YAML still bad if we don't allow subsnippets or icons at first, only adding those later if we so choose?  Without icons or subsnippets, this is what it would look like:

```
\snippets
Weapon:
  content: |
    {{weapon
    ## Weapon
    ...
    }}

Potion:
  content: |
    {{potion
    ## Potion
    ...
    }}
```

That doesn't look as bad as the examples with subsnippets and other properties.  But it still has that YAML form so it *could* be extended later.  I guess my knock against it is the aforementioned spaced indents, which isn't great.  But for something that is set *once*, not getting edited a lot, it might be worthwhile to maintain that YAML format and extensibility.

dbolack-ab:
I mean... I want this feature in the product and if I had to use YAML formatting for it I'd probably not use it. And I've been in IT 30 years. I used to hand-edit sendmail configs and I'd rather not touch YAML.  Meh.

calculuschild:
Regardless of the input method, as long as the backend stores it in a fixed structure, we can always change the input mode at a later date.

Given that, @dbolack-ab brings up a good point in https://github.com/naturalcrit/homebrewery/discussions/4142.  It might make better sense to store the snippets alongside the other metadata instead of in its own <code>```snipets</code> block. If we do that, that should be part of this PR, and then any changes to input syntax/fillable form can come later.

dbolack-ab:
> well, related to this, the default snippet is not loaded as a snippet (which i would expect it to) when loading new.

Looking at this now, I kinda like this behavior as it doesn't push a useless snippet into the stack. That said, I'm going to fix this unless others agree with me. :)

calculuschild:
For the CSS in the Style tab, if the user has never edited its contents, the default style text is not actually applied to the document. Its role is not to actually influence the document behavior but to instruct the user.

I would expect the Snippet default text to serve the same function and not actually push an empty Snippet into every brew.

dbolack-ab:
Need to update tests and find teh test timeout regression

client/homebrew/editor/editor.jsx
```diff
@@ -159,12 +171,18 @@ const Editor = createClass({
 
 				for (let i=customHighlights.length - 1;i>=0;i--) customHighlights[i].clear();
 
+				let userSnippetCount = 1; // start snippet count from page 2
```
5e-Cleric, line 174: what does the comment here mean? what page?

client/homebrew/editor/editor.less
```diff
@@ -11,10 +11,18 @@
 			background : #33333328;
 			border-top : #333399 solid 1px;
 		}
+		.snippetLine {
+			background : #33333328;
+			border-top : #333399 solid 1px;
+		}
```
5e-Cleric, line 17: lines 10 to 17 can be merged into :

```less
		.pageLine, .snippetLine {
			background : #33333328;
			border-top : #333399 solid 1px;
		}
```

5e-Cleric (approved):
overall works, got two coding thingies

client/homebrew/editor/editor.jsx
dbolack-ab: Debris I missed.

dbolack-ab: Should be clearer now - the snippet row should have numbers just as the pages did and that's the counter for it. I was unsure if it was wise to recycle the variable there so I created a new one.

client/homebrew/editor/editor.less
dbolack-ab: I went ahead and merged like this - I had left them separate in case ya'll though it should be a distinct style.

5e-Cleric:
Thank you @dbolack-ab 🎊

PR #3880, opened 2024-11-05
Fix Memoization

Gazook89:
This is just to correct some missed lines in the latest merge.  Adjusts the memoization at the end of brewRenderer.jsx.


calculuschild:
Thanks Gazook89!

PR #3890, opened 2024-11-11
Update Core.less and reset.less

Gazook89:
(i've no idea how this got attached to "give dismisskeys a default", i apologize...)

## Description
This PR focuses on changes on `core.less` and `reset.less` to make styling easier across the project, particularly around `button` elements.  The PR is actually a cherry-pick from the last commit on #3868, a commit that I wish I could benefit from right now rather than waiting on that one.

Currently, if you use a `button` in HB it gets styled like this, without any additional input from you:
image

All well and good when our site only has a few actual `button` elements, but a pain in the arse to unset all those properties if you want to use buttons in, say, a toolbar.  

This PR changes the default styling of buttons so that they basically lose their user-agent properties, but otherwise leaves them stripped bare.  The current green default button is then assigned the mixin `.colorButton()` and given the same parameter.  So when you actually *want* that style of button, you can ask for it.

Publish button, without using the mixin:
image

(obviously the mixin will be used here, this is just for an example).

Check that any place you expect a `button` to be it is still styled as you remember it.


ericscheid:
> but a pain in the arse to unset all those properties if you want to use buttons in, say, a toolbar

`.toolbar button { all: unset; }` or `.toolbar button { all: default; }`

calculuschild (approved):
This is fine with me but I'm not a CSS guy.

If anyone else wants to review this then I am ok merging.

5e-Cleric (approved):
I am okay with this change, but perhaps we should think about adding somewhere a comment with all mixins in use.

@Gazook89, feel free to merge this whenever you want.

PR #3899, opened 2024-11-19
Permit the \page Token to accept {} values.

dbolack-ab:
## Description

This PR allows users to apply style, class, and attribute values via the inline `{}` format to the base page for common inheritance.

To Do:
 - [x] Fix Attributes, They are not being inserted
 - [x] Fix Styles, they are not being inserted.
 - [ ] Add Tests

Solves #3901 

Usage:
Similar to an inline single mustache.

`\page{wide,banna=fail,color:black}`

  - [ ] Adding a style includes a style on the <div class='page'> instance.
  - [ ] Adding a class includes a class on the <div class='page'> instance.
  - [ ] Adding an HTML Attribute adds the attribute on the <div class='page'> instance.
  - [ ] Mixed forms of above work as expected
  - [ ] `\page` works without values.
  - [ ] Ensure CSS Properties ( --FOO ) can be set.


calculuschild:
Is there a concern that this doesn't apply to the very first page which doesn't have a `/page` break?

dbolack-ab:
> Is there a concern that this doesn't apply to the very first page which doesn't have a `/page` break?

No, though the included code also tolerates putting a `\page`  as the very first line for the occasion that is desirable ( this functionally moves \page to being a START-OF-PAGE instead of END-OF-PAGE marker.

calculuschild:
Would you mind creating an issue for this PR? So we have a place to document and discuss the problem this is aiming to solve.

5e-Cleric:
> > Is there a concern that this doesn't apply to the very first page which doesn't have a `/page` break?
> 
> No, though the included code also tolerates putting a `\page` as the very first line for the occasion that is desirable ( this functionally moves \page to being a START-OF-PAGE instead of END-OF-PAGE marker.

I cannot see this working in the deployment, although the rest seems to work fine.

While this syntax works for us, it is also worth mentioning that codemirror stops processing that line as a `.pageLine`, so we have no way of targeting that element for code highlight.

IF we did it like with tables, that you need the injection to be in the next line:

```
\page
{background:red}
```

That should preserve the code hightlight, and be somewhat intuitive and consistent.

I wish we could keep it in the line though

dbolack-ab:
> > > Is there a concern that this doesn't apply to the very first page which doesn't have a `/page` break?
> > 
> > 
> > No, though the included code also tolerates putting a `\page` as the very first line for the occasion that is desirable ( this functionally moves \page to being a START-OF-PAGE instead of END-OF-PAGE marker.
> 
> I cannot see this working in the deployment, although the rest seems to work fine.

What can you not see?

Regarding the syntax highlighting - yes, that's not there yet.

5e-Cleric:
> What can you not see?

adding `{background:red}` at the begginging of the document does not work.

dbolack-ab:
> > What can you not see?
> 
> adding `{background:red}` at the beggining of the document does not work.

No. It won't. This syntax *only* works on a `\page` line as described in usage.

Gazook89:
For the first page, @5e-Cleric :

image

Gazook89:
Also, I'll just add this discussion that was on this same topic, before we started using named spans and `:has()` css selector: https://github.com/naturalcrit/homebrewery/issues/1489.  I don't know if it adds anything but it bothered me that I could half-remember talking about this but not the details.

calculuschild:
> the included code also tolerates putting a \\page as the very first line for the occasion that is desirable ( this functionally moves \\page to being a START-OF-PAGE instead of END-OF-PAGE marker.

Is there any risk of this breaking existing brews? I'm not sure if there would be any documents out there depending on a `\page` at the start to make a blank page. Probably not common, right?

5e-Cleric:
> Is there any risk of this breaking existing brews? I'm not sure if there would be any documents out there depending on a `\page` at the start to make a blank page. Probably not common, right?

This should break no document. I have never seen or heard, or can think of any good reason to have an empty page at the top.

5e-Cleric:
> While this syntax works for us, it is also worth mentioning that codemirror stops processing that line as a `.pageLine`, so we have no way of targeting that element for code highlight.
> 
> IF we did it like with tables, that you need the injection to be in the next line:
> 
> ```
> \page
> {background:red}
> ```
> 
> That should preserve the code hightlight, and be somewhat intuitive and consistent.
> 
> I wish we could keep it in the line though

The rest of my comment still stands, this breaks the highlight in an irrecuperable way, could we consider that other syntax?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -180,7 +212,23 @@ const BrewRenderer = (props)=>{
...
+					const styleInst = styleSplit.split(':');
+					if(styleInst[0] && styleInst[1]) declaredStyles[styleInst[0].trim()] = styleInst[1].trim();
+				}
+			}
+
+			const styles = {
+				...declaredStyles,
+				...(!displayOptions.pageShadows ? { boxShadow: 'none' } : {}),
+				// Add more conditions as needed
+			};
```
5e-Cleric, line 228: Suggested by the toaster, looks cleaner. 

```suggestion
			// I assume there is a prettier way to do this.
const styles = (() => {
    const rawStyles = brewTemplates[index]?.attr?.styles || "";
    const declaredStyles = Object.fromEntries(
        rawStyles.split(";").map(style => {
            const [key, value] = style.split(":").map(s => s?.trim());
            return key && value ? [key, value] : [];
        }).filter(entry => entry.length)
    );

    return {
        ...declaredStyles,
        ...(displayOptions.pageShadows ? {} : { boxShadow: "none" }),
        // Add more conditions as needed
    };
})();

```

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-lines */
 /*eslint max-lines: ["warn", {"max": 300, "skipBlankLines": true, "skipComments": true}]*/
```
5e-Cleric, line 2: no need for two separate rules for eslint, decide on one.

dbolack-ab: This is what I get for letting VSCode resolve the warning. :)

client/homebrew/brewRenderer/brewRenderer.jsx
calculuschild: But use tabs.

calculuschild (changes requested):
A couple things noted just testing quickly:

- The page number at the end of a `/page` becomes inaccurate if it is placed at the first line:

image

- The "Jump" buttons are similarly broken in this case. Trying to jump from editor to brewRenderer will crash the page because it tries to jump to a page number that doesn't exist.

dbolack-ab:
> A couple things noted just testing quickly:
> 
> * The page number at the end of a `/page` becomes inaccurate if it is placed at the first line:
> * The "Jump" buttons are similarly broken in this case. Trying to jump from editor to brewRenderer will crash the page because it tries to jump to a page number that doesn't exist.

I have this solved - though there may be a more clever way to handle it.

calculuschild:
> // There is bound to be a more elegant way to do this. But this works for now.

I think I see a way to do this and keep it all inside of `markdown.js` so we don't need to duplicate the tag handling in `brewRenderer.jsx`. Let me poke at it a bit before we merge.

5e-Cleric:
@calculuschild should we merge? did you advance in your attempt?

calculuschild:
Please don't merge yet. I was out of town over the weekend and have not implemented yet. If i am right my solution would reduce this PR to ~10 lines.

dbolack-ab:
> > // There is bound to be a more elegant way to do this. But this works for now.
> 
> I think I see a way to do this and keep it all inside of `markdown.js` so we don't need to duplicate the tag handling in `brewRenderer.jsx`. Let me poke at it a bit before we merge.

I will look at this again and see if I can remember *why* exactly I duplicated it.

PR #3902, opened 2024-11-20
Upgrade server-side code to ESM (import vs require)

calculuschild:
## Description
Upgrades the server code and all dependencies to use more modern ES6 syntax. Involves rearranging a few things slightly, but 95% of this is just exchanging `require` for `import` where needed.

Removes tests for `content-negotiation.js` middleware. This test wasn't being run as part of Circle-CI and has actually been broken for a while now. Easier to just remove it than try to troubleshoot it *and* fix conflicts with ESM conversion.

Also adds a dependency `babel-plugin-transform-import-meta`. Previously we used `__dirname` in our `app.js`, but this is no longer available in ESM modules. Unfortunately, Jest also does not like the alternative, so this plugin allows Babel to transform it specifically for Jest tests.

Supersedes #3813 

Should allow us to keep up with packages that no longer support CJS, such as #3860 

This touches 49 total files. Behavior should be unchanged, but it's possible there is some function I forgot to convert over and so will cause an `undefined` error that I just haven't hit yet. ESM is also more sensitive to bad syntax (forgetting to put `let` or `const` in front of a variable, etc.). I have fixed the cases I ran into but please double-check any server interaction you can think of.


dbolack-ab:
I'm a bit at a loss as to how to test this more than the existing tests and just putting it to daily driver use.

Good place for an integration instance...

dbolack-ab (approved):
On visual skim this all looks good.

G-Ambatte:
I think we might have broken the content negotiation test when we added support for hosting images locally in `/staticImages`; I'm poking at it now.

PR #3907, opened 2024-11-22
[DX] Update ESLint rules

### LINKED ISSUE #3906: [DX] ESLint throws errors on Import Attributes
### Your idea:

When using the new syntax for `import` (e.g. `import X from Y with { type: 'json' };`), ESLint throws an unexpected token error on the `with` keyword.

G-Ambatte:
This PR resolves #3906.

This PR adds the `@babel/plugin-syntax-import-assertions` package as a dev dependency, and adds it to the list of Babel plugins in the config file. This prevents the "unexpected token" error on the `with` keyword.

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -5,6 +5,7 @@
   ],
   "plugins": [
     "@babel/plugin-transform-runtime",
-    "babel-plugin-transform-import-meta"
+    "babel-plugin-transform-import-meta",
+    "@babel/plugin-syntax-import-assertions"
   ]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
         "vitreum": "git+https://git@github.com/calculuschild/vitreum.git"
       },
       "devDependencies": {
+        "@babel/plugin-syntax-import-assertions": "^7.26.0",
         "@stylistic/stylelint-plugin": "^3.1.1",
         "babel-plugin-transform-import-meta": "^2.2.1",
         "eslint": "^9.15.0",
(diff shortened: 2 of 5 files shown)


calculuschild:
Just for clarity, this doesn't actually update the ESLint rules, but rather tells Babel to convert it to something different so ESLint doesn't see it.

This Babel Rule is also deprecated, since `assert` has been discarded in favor of `with`. https://babeljs.io/docs/v8-migration#babelplugin-syntax-import-assertions

So while this works, I think it may be more appropriate to find the ESLint setting (perhaps converting to their @stylistic plugin) rather than using Babel to convert to a deprecated syntax.

What do you think? At a minimum I think we should add a comment explaining why we are using a deprecated babel plugin since it's not obvious its just to make ESLint happy.

G-Ambatte:
Actually, the spec just reached Stage 4 (at the start of last month - Oct 2024) and it should now be live in ESLint? Searching the repo, I found a lot of rules in ESLint's `camelcase.js` from v9.14 and up, which we're already using... so it might be possible without requiring a new package.

The package WAS the recommended way to make it work. I'll leave this PR open while I look into whether the ESLint config can be modified to achieve the same result.

calculuschild:
You are right, this is a very recent spec change. Based on this link you shared in Gitter, I'm wondering if there is some deprecated ESLint Formatting Rule we use that is the culprit.

https://eslint.org/blog/2024/11/eslint-v9.14.0-released/

G-Ambatte:
It appears to be working now, without this PR merged... No additional package and no config changes required.

calculuschild:
Ok, I just checked this. I have ESLint 9.15.0 installed and I am not seeing this issue on my end. I think you may have just needed some kind of refresh or install to catch up to the latest version. I'm going to close this and the issue for now.

PR #3908, opened 2024-11-23
V4 proposed aligned paragraph tokens

dbolack-ab:
## Description

This implements the suggested syntax for V4 paragraph text alignment tokens from https://github.com/naturalcrit/homebrewery/discussions/2879

https://github.com/naturalcrit/homebrewery/discussions/2879

- insert `:- ` in front of a paragraph. It should align left. You should see a class of mdParagraphJustifyLeft on the paragraph tag
- insert `-: ` in front of a paragraph. It should align right. You should see a class of mdParagraphJustifyRight on the paragraph tag
- insert `:-: ` in front of a paragraph. It should align center. You should see a class of mdParagraphJustifyCenter on the paragraph tag


5e-Cleric:
i want to test this

calculuschild:
Three thoughts:

First, the table syntax this borrows from allows any number of hyphens in the construction. Is it valuable to allow that here as well?  For example:

```
"I am inevitable."
--------------: Michael Scott
```

Second, it valuable to allow this syntax on the first line of a paragraph 
by itself, to act as an optional familiar appearance more like the Table syntax?

```
This is a normal paragraph.

---------------------------:
This paragraph is going
to be right-aligned.

:--------------------------:
This paragraph is going
to be centered

This paragraph is unmarked,
and so defaults back to
left-aligned
```

Third, we should aim to have Markdown convert to pure HTML, so they do not rely upon CSS to function. In this case, we have the `align` HTML property we can use, the same way it is handled in Tables.


Please don't start writing code for the first two unless we get a consensus, so you don't have to undo it later.

ericscheid:
> ```
> This paragraph is unmarked,
> and so defaults back to
> left-aligned
> ```

Technically, since that ¶ is unmarked it should default back to the *default* .. which in many themes might be left-aligned, but could also conceivably be full-aligned in some themes.

I suggest

> ```
> :----------------
> This para is marked to be 
> explictly left aligned
>
> This paragraph is unmarked,
> and so is the default for the
> current theme
> ```

5e-Cleric:
> I chose to use a class instead of the property so that users could overload the feature with additional styling. If we do it with the property, users lose that option and must {{}} it as-well.
> 

To be fair, centered text is something common to most if not all themes, i'm not sure of the value of the overriding possibility. Not against it.

dbolack-ab:
> > I chose to use a class instead of the property so that users could overload the feature with additional styling. If we do it with the property, users lose that option and must {{}} it as-well.
> 
> To be fair, centered text is something common to most if not all themes, i'm not sure of the value of the overriding possibility. Not against it.

I'm guessing I explained that poorly? I dunno. The notion that a brew or theme could choose to override and extend what  `:-:` did struck me as useful.

Gazook89:
You could use the attribute selector `[align=center]` to apply styles rather than using a class

dbolack-ab:
> You could use the attribute selector `[align=center]` to apply styles rather than using a class

See above.

calculuschild:
Markdown should translate to plain HTML wherever possible. This is a case where we don't need CSS at all, so we should not need to make this one a special exception to the rule.

> The notion that a brew or theme could choose to override and extend what `:-:` did struck me as useful.

People can still override and extend the functionality of `:-:` if they want, using the selector suggested by Gazook89. A CSS class is not needed to accomplish that.

```
// Change all `:-:` paragraphs to actually be left-aligned, with a black border and red text.
p[align=center] {
  text-align: left;    // CSS will override the align="center" attribute
  border: solid black;
  color: red;
}
```

5e-Cleric:
I agree with calc here

5e-Cleric:
This has changed the behaviour of paragraphs:

image

The first two paragraphs are separated as two `p` elements with different alignment.

The second two paragraphs without the syntax are joined in the same `p` element.

The second behaviour is desired, we have two ways of separating them, a blank line or two spaces at the end, as the current behaviour stands, both are ignored if alignment is set.

dbolack-ab:
> This has changed the behaviour of paragraphs:

I believe I have everything running correctly now.

5e-Cleric:
The behaviour is solid, looks good on that front.

PR #3909, opened 2024-11-23
V4 proposed non-breaking space token

dbolack-ab:
## Description

Implements `:>`  for a series of non-breaking spaces analogous to `:` based hard breaks per V4 proposed discussion.

This serves to evaluate the utility of this feature.

Relates to discussion #2879 

Implements as an inline token. Use `:` followed by one `>` for each non-breaking space.
### Reviewer Checklist

  - [ ] Test any number of `>` after the `:` 
  - [ ] Verify it doesn't get parsed in code blocks.
  - [ ] ??


dbolack-ab:
> As long as these syntaxes don't conflict with existing documents, I think this one should be pretty safe. The implementation looks good at first glance as well.

That's largely why I let this loose without tests. I don't seem to be particularly good at guessing how the userbase will break the code. :)

> For example, `:>` probably shouldn't open the emoji auto-suggest (I don't think it does, so good).

I didn't spot that happening when I tested, but I could have overlooked it.

> Also potential conflicts with multiline and single line term : definition blocks. What should the following do?

I think that can be avoided with a slightly smarter regex to ensure that `::>` isn't matched.

calculuschild:
> I think that can be avoided with a slightly smarter regex to ensure that ::> isn't matched.

Probably, yes. And we should still add a test so we can lock down what we want the output to be in that case.

dbolack-ab:
> > I think that can be avoided with a slightly smarter regex to ensure that ::> isn't matched.
> 
> Probably, yes. And we should still add a test so we can lock down what we want the output to be in that case.

In the mean time I will sit on this to see if there are any other concerns.

5e-Cleric:
The syntax is desired, (i think i proposed it), i see no other conflicts with syntax apart from the one already noted, and it should not be a worry to break old brews. Not sure if we need to wait for v4.

5e-Cleric (approved):
Overall looks great! Add a test for `:>>:>>` too, which is a working syntax, and this should be good to go

calculuschild:
Can we add a test for `Term ::> Definition`?

What do we want this to do? Should the `&nbsp;` have priority or should the `<dl>` have priority?

And another test for

```
Term
::> Definition1
::>> Definition2
```

What do we want to happen here? I feel like @ericscheid would have an opinion.

On one hand, we could just say `<dl>` are a higher-order "block" structure so they should have priority over inline characters. `Term ::> Definition` becomes `<dl><dt>Term</dt><dd>> Definition</dd></dl>`

On the other hand, we could say the `&nbsp;` has priority, noting that emojis currently also have priority over DLs. (`Term ::thumbs-up:`) becomes `Term :👍`  and `Term ::> Definition` becomes `Term :%nbsp; Definition`. This is also the way this PR is currently implemented. I lean toward this just to follow the emojis behavior.

Either way, I think we have discussed previously that for V4, we will eventually update DL format to require a space before and after the `::`. The ambiguity would be gone then for both emojis and `&nbsp;`

calculuschild:
One more thing. Do we want to add a syntax highlighting to this?

Otherwise this PR looks good.

dbolack-ab:
> Can we add a test for `Term ::> Definition`?
> ```
> Term
> ::> Definition1
> ::>> Definition2
> ```

Done

> On one hand, we could just say `<dl>` are a higher-order "block" structure so they should have priority over inline characters. `Term ::> Definition` becomes `<dl><dt>Term</dt><dd>> Definition</dd></dl>`. This is more inline with standard Markdown philosophy.

This is the current behavior.

dbolack-ab:
> One more thing. Do we want to add a syntax highlighting to this?

We don't for `::::` so I would be inclined not to.

calculuschild:
Ok, last call for opinions on the ambiguous cases @5e-Cleric, @ericscheid, @Gazook89 , @G-Ambatte 

Otherwise I am merging as-is.

shared/naturalcrit/markdown.js
```diff
@@ -391,12 +391,33 @@ const forcedParagraphBreaks = {
...
+};
+
 const definitionListsSingleLine = {
 	name  : 'definitionListsSingleLine',
 	level : 'block',
-	start(src) { return src.match(/\n[^\n]*?::[^\n]*/m)?.index; },  // Hint to Marked.js to stop and check for a match
+	start(src) { return src.match(/\n[^\n]*?::[^:\n]*/m)?.index; },  // Hint to Marked.js to stop and check for a match
 	tokenizer(src, tokens) {
-		const regex = /^([^\n]*?)::([^\n]*)(?:\n|$)/ym;
+		const regex = /^([^\n]*?)::([^\:\>][^\n]*)(?:\n|$)/ym;
```
calculuschild, line 420: I believe these changes are causing the single-line definition list `Term ::> Definition` case to fail, because it gives `&nbp;` priority here.

calculuschild:
@dbolack-ab We were missing a test case for `Term ::> Definition` (single-line definition list). I added it just now and it is failing.

PR #3911, opened 2024-11-25
Fix regression in emoji auto-complete due to ESM update

dbolack-ab:
## Description

As above, converts overlooked requires to imports.

- Closes # N/A - per git

Attempt an autocomplete.


calculuschild:
Thank @dbolack-ab ! I don't think there are any other places the font are `require`'d so this should be good. Look functional from testing the deployment on my phone.

PR #3923, opened 2024-11-28
WIP: User Brew Theme Write-in

dbolack-ab:
## Description

A first pass at adding a write-in option for User Brews. Functional, but fugly. Provided for feedback.

Needs:

 - [ ] opinions on UI placement
 - [ ] opinions on best choice for displaying a write-in based User Brew ( flip to write-in box? Add to drop-down list? )
 - [ ] Sanity checking? on brew share-id input.
 - [ ] Should it only try to load after enter/cr?

- Closes #

### Reviewer Checklist

  - [ ] Hitting arrow toggle at end of the themes list toggles between input modes
  - [ ] Inputting a valid share-id changes the brew theming
  - [ ] Returning to the Pulldown list and selecting a theme works
  - [ ] Invalid shareIds are handled in a useful way for the user.


Gazook89:
Just for clarity, here is what it looks like at this time:

image

Note the little arrow icon on the right side of the theme dropdown...clicking that opens the below text input:

image

As stated in the PR the styling/structure isn't complete, but wanted to point this out because it took me a minute to even realize that arrow existed at all (particularly when i opened in Firefox, since the font awesome icons don't render).

Gazook89:
Regarding opinions on the actual questions about what this should look like:   

What about a dropdown menu option at the bottom that says "User Theme...".  Choosing that displays the text input for writing in your own, below the theme select.  

If something is already entered in the *user theme* input, and then they select an *included theme*, it should hide the user theme input again.  But in state, it shouldn't clear that user input quite yet, in case the user decides to go back to using the write-in option again-- they shouldn't have to enter it again.  But preserving this value is just a client-side thing...reloading the page wouldn't preserve it.

Does that make sense?

dbolack-ab:
> As stated in the PR the styling/structure isn't complete, but wanted to point this out because it took me a minute to even realize that arrow existed at all (particularly when i opened in Firefox, since the font awesome icons don't render).

Yeah, that arrow was duplicated from the thun nail hider and isn't the most obvious with the length of the fake pulldown.

Holding in state makes sense.

dbolack-ab:
This is functional, if ugly

5e-Cleric:
Working does work.

Styling wise, perhaps we could use text on the button instead: "use a brew as theme", or something of the sort.

dbolack-ab:
> Working does work.
> 
> Styling wise, perhaps we could use text on the button instead: "use a brew as theme", or something of the sort.

I updated to try that out - looks much more obvious. There's still a weird regression I somehow? made with the pulldowns I can't identify.

5e-Cleric:
Looks better, and it is somewhat easy to use, we may want to include something like `<small>The id is the last part of the URL in the share page of the brew: `.../share/{this part}`</small>`

calculuschild:
Is there a reason we can't just enter text directly into the top entry, like we do in the Language dropdown?

image

The Language Entry
image

Seems they should just use the same type of interaction?

Or if we add a button, we should add a matching button to the language dropdown, no?

5e-Cleric:
I don't think they should be together, The functionalities are very different, and themes by link are not saved in the dropdown. I'd keep the select away from the input.

The input + select works because we have a predefine list of language keys to show, this is different, you may use an id for the theme, or select one predetermined.

calculuschild:
> The functionalities are very different

The functionality is the same though?

In both, you can type in a custom value, or select from the dropdown.

dbolack-ab:
> Is there a reason we can't just enter text directly into the top entry, like we do in the Language dropdown?

Pithy reason?  I don't know how to do that. I have a lot of basic react to learn.

> Seems they should just use the same type of interaction? I.e. no little arrow or separate button. Just a normal editable textbox that shows a dropdown below when it is active.

I didn't even process I could type into that input.  If I did, I would *assume* it is a filter and can't take new input.

5e-Cleric:
I honestly thought it was an anti-pattern, but i just found out it is completely fine, i guess, since there is an aria role for it:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role

So i guess, lets go. We have a combobox component.

calculuschild:
I am going to work on merging this today, so don't touch the branch.

5e-Cleric:
So will you be changing this into a combobox?

If so, please add some description of this change, something like: `<small>You may paste the share id of a brew in this box, or select one of the ready options</small>`

I plan on releasing a notification just for this feature when it goes live, as it should be ground breaking.

calculuschild:
Yeah, I will make it a combobox, and have a placeholder text there that says what you can do.

calculuschild:
##### Changes so far:
- Use our existing comboBox component
- When selecting a user theme, also display the theme author (i.e. will show `calculuschild : theme name` instead of `!!! : theme name`.)
- Renamed some variables and deleted unused values after the transition to comboBox

##### ToDo
- [ ] Update the tests to handle author being in the themeBundle
- [ ] Allow pasting in the whole share link
- [ ] Add input validation so it doesn't bother fetching on bad input (not a valid share URL or ID)
- [ ] Don't return theme bundle if not tagged `meta:theme`

calculuschild:
@5e-Cleric , @dbolack-ab I have moved this to a combobox, and added a handful of other things (input validation and test cases, etc.)

Does one of you want to confirm this still works?

dbolack-ab:
> @5e-Cleric , @dbolack-ab I have moved this to a combobox, and added a handful of other things (input validation and test cases, etc.)
> 
> Does one of you want to confirm this still works?

Share id works. Full URL to share id works, but still has the "error" pop-up

dbolack-ab:
> > Share id works. Full URL to share id works, but still has the "error" pop-up
> 
> @dbolack-ab What error do you see, and what are you entering that causes it? I don't get any error but I am probably not doing the same steps.
> 
> Note that the URL validation is checking against the deployed URL. So if you are testing on homebrewery-pr-3923 it assumes are also using links with the same base URL. My guess is you are pasting in a homebrewery.naturalcrit.com/share/xyz link which would cause a validation mismatch?

That would indeed be it - I should have thought of that.

calculuschild:
Merging this. Thanks @dbolack-ab !

PR #3931, opened 2024-12-02
Convert any `tags` strings to arrays

### LINKED ISSUE #3927: Editor crashes when accessing Metadata Panel
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

When accessing the Metadata panel, the Homebrewery edit page crashes to a white screen.

Console shows an error occurred: `values.map is not a function`.

Affected brew link: https://homebrewery.naturalcrit.com/share/bD6vc20VG9JU

### Code

_No response_

G-Ambatte:
This PR resolves #3927.

This PR adds a check to `splitTextStyleAndMetadata` which checks if the `tags` property is a String, and if so, converts it to an Array. This prevents the TagInput crash discovered in #3927 and should allow any future functions to proceed with confidence that `tags` will properly be an Array.

### DIFF
--- a/shared/helpers.js
+++ b/shared/helpers.js
@@ -22,6 +22,7 @@ const splitTextStyleAndMetadata = (brew)=>{
 		brew.text = brew.text.slice(index + 5);
 	}
 
+	// Handle old brews that still have empty strings in the tags metadata
 	if(typeof brew.tags === 'string') brew.tags = brew.tags ? [brew.tags] : [];
 };
 


dbolack-ab (approved):
Looks good.

calculuschild:
Should we also change splitTextStyleAndMetadata to not be creating strings in the first place?

5e-Cleric:
> Should we also change splitTextStyleAndMetadata to not be creating strings in the first place?

Definitely

dbolack-ab:
> Definitely

Where is it creating strings that it shouldn't be?

calculuschild:
Ok, so it's not `splitTextStyleandMetadata`, where the empty string would be created, but rather here in `mergeBrewText`:

https://github.com/naturalcrit/homebrewery/blob/bea74c3b46d1404618661fc3d0e4ca59d49bd8bd/server/homebrew.api.js#L177-L191

However, as it is currently written, the YAML output does create `[]` arrays a we want. I think these empty strings are remnants of older brews before we fully implemented tags (the brew in question was last edited 2 years ago). If I remember correctly, we just had a placeholder value there which happened to be an empty string before changing the default to an array.

So... nothing to change then as far as preventing strings. We just have to handle old data in the wrong format, in which case this PR is complete. I will merge this.

G-Ambatte:
> However, as it is currently written, the YAML output does create `[]` arrays a we want. I think these empty strings are remnants of older brews before we fully implemented tags (the brew in question was last edited 2 years ago). If I remember correctly, we just had a placeholder value there which happened to be an empty string before changing the default to an array.

This is exactly the conclusion I came to.

PR #3932, opened 2024-12-03
Load tags from Homebrewery stub, not text metadata

### LINKED ISSUE #2692: gdrive metadata should override mongo-stub metadata on userpage
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

On the userpage, the metadata displayed should be from either the mongo record (stub or full), or the properties retrieved from the GDrive metadata API, with precedence given to the gdrive properties.

The precedence should be something like `metadata codefence in the text` > `Gdrive file props` > `mongo stub props`, with some brew props only coming from the `mongo` record (e.g. brew `views` count).

----

Currently, it appears that on the userpage the title and last-modified date (at least) is being sourced from the mongo stub record and not from the gdrive metadata. That is, if the gdrive file is modified, and renamed, then the userpage does not reflect that. 
Renaming the gdrive file does not change the title displayed on the userpage. 

image

image

Loading the brew in question into the homebrewery editor will reveal the updated title and description, sourced from the codefenced metadata section of the brew contents.

image

Saving the brew will rename the gdrive file to match the title in the metadata panel (which being loaded from the codefenced metadata content could be different from the new filename).

Similarly, the share link of the brew uses the title from the brew contents metadata.

image

## Proposals

1. the userpage should display title and last-modified from the gdrive metadata, and whatever other valid metadata from gdrive file metadata (i.e. without needing to fetch the full file content)
2. [optional] indicate if the gdrive last-modified is different from the mongo last-modified
3. [optional-plus] one-click "refresh metadata" button if difference is detected (additionally this could even indicate gdrive brews that don't even have a mongo stub, with the one-click button thus creating the stub)

### Code

_No response_

G-Ambatte:
This PR resolves #2692.

This PR removes `tags` from the metadata loading process of the `splitTextStyleAndMetadata` function, ensuring that only tags stored in Homebrewery's internal database stub are used.
However, the option to load these tags is retained, and used in the NEW FROM FILE function - so when recovering a brew from a downloaded backup, the `tags` should be retained.

### DIFF
--- a/client/homebrew/navbar/newbrew.navitem.jsx
+++ b/client/homebrew/navbar/newbrew.navitem.jsx
@@ -19,7 +19,7 @@ const NewBrew = ()=>{
 					style : ''
 				};
 				if(fileContent.startsWith('```metadata')) {
-					splitTextStyleAndMetadata(newBrew); // Modify newBrew directly
+					splitTextStyleAndMetadata(newBrew, true); // Modify newBrew directly
 					localStorage.setItem(BREWKEY, newBrew.text);
 					localStorage.setItem(STYLEKEY, newBrew.style);
 					localStorage.setItem(METAKEY, JSON.stringify(_.pick(newBrew, ['title', 'description', 'tags', 'systems', 'renderer', 'theme', 'lang'])));
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -70,6 +70,7 @@ const NewPage = createClass({
 			brew.renderer = metaStorage?.renderer ?? brew.renderer;
 			brew.theme    = metaStorage?.theme    ?? brew.theme;
 			brew.lang     = metaStorage?.lang     ?? brew.lang;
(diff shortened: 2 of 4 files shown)


calculuschild:
I have two main thoughts here:

1) This only affects tags specifically. I am concerned about leaving one-off special cases like this that could easily lead to confusion down the line, where certain metadata fields can get updated externally but other ones can't.

2) I'm not sure this actually resolves the concern raised in the linked issue, that more recent fenced metadata should override older stub metadata in case of a conflict in the `/user/` page. In fact this PR *removes* the working conflict resolution we have for all the other pages.

I feel there was some disconnect in the discussion on the issue page. But I would also like @ericscheid to weigh in since he raised the issue in the first place.

dbolack-ab:
I am in agreement on the notion we have split-metadata.

Anything that needs to be looked up belongs with the stub.  Everything else should be in the _\`\`\`metadata_ bit of the brew.  Even if it isn't something we search now, we should think on that for the future.

I believe Authors, Tags, and Title are the only candidates.

calculuschild:
Closing following discussion in #2692

Issue #3918, opened 2024-11-27
Vault button lost its formatting


5e-Cleric:
During one of the last PRs, @Gazook89 moved the styling of buttons to a less mixin, but the vault search button did not have that included in it. It needs to be added in. The site is functional and the button works, so it is only a styling matter.

dbolack-ab:
I think this kind of styling issue should be higher

PR #3934, opened 2024-12-03
Fix Vault styles

5e-Cleric:
- Fixes Vault button with missing styles
- Fixes vault filter bar font, that somehow got messed up
- turns checkboxes back to the default styling


5e-Cleric:
Should fix #3918

PR #3941, opened 2024-12-08
Create backend batch re-author framework

5e-Cleric:
Adds a backend api route for renaming brews on batch. The call will be coming from naturalcrit's renaming account.actions.js file.


5e-Cleric:
closing to reopen as upstream branch for automatic deployment

PR #3942, opened 2024-12-08
Create backend batch re-author framework 

5e-Cleric:
Adds a backend api route for renaming brews on batch. The call will be coming from naturalcrit's renaming account.actions.js file.

While testing in local, a cors policy was necessary for this to work, we should already have one to prevent unwanted access to our apis, but this also means we gotta whitelist everyone that users our api.


5e-Cleric:
I could use some reviews on this, to merge and test the renaming feature.

dbolack-ab:
Could the restrictions happen at the reverse proxy level instead of in CORs?

5e-Cleric:
> Could the restrictions happen at the reverse proxy level instead of in CORs?

I had to ask the toaster(chatgpt) what that meant, i have no idea how to set that up, but i must mention i only set up a CORS policy because my own calls from naturalcrit.com were being stopped, i need this policy in place anyway for those calls, so might as well.

server/middleware/check-client-version.js
```diff
@@ -1,6 +1,8 @@
 import packageJSON from '../../package.json' with { type: "json" };
 const version = packageJSON.version;
 
+//This should be only for internal calls, but right now prevents spam api calls, this should be done with a proper cors policy
```
calculuschild, line 4: This is not intended as a spam blocker. It's just checking that the version of Homebrewery on the client Browser matches the latest version, to avoid issues interacting with brews if things have changed between versions. If a cors policy is added to prevent spam API calls, this version check is still needed, but it could be moved to only apply to brew update calls.

server/app.js
```diff
@@ -55,6 +55,34 @@ app.use(bodyParser.json({ limit: '25mb' }));
 app.use(cookieParser());
 app.use(forceSSL);
 
+import cors from 'cors';
+
+// CORS Configuration
+const corsOptions = {
```
calculuschild, line 61: Please use tabs for indenting

server/app.js
```diff
@@ -312,6 +340,35 @@ app.get('/user/:username', async (req, res, next)=>{
 	return next();
 });
 
+//Rename Brews
```
calculuschild, line 343: Can this be something like "Change author name on brews" or something? We are not renaming the brews.

server/app.js
```diff
@@ -55,6 +55,34 @@ app.use(bodyParser.json({ limit: '25mb' }));
...
+const corsOptions = {
+    origin: (origin, callback) => {
+        const allowedOrigins = [
+            'https://homebrewery.naturalcrit.com',
+            'http://localhost:8000',
+            'http://localhost:8010',
+            'https://naturalcrit.com'
+        ];
+
+        const herokuRegex = /^https:\/\/.*\.herokuapp\.com$/; // Matches any Heroku app
```
calculuschild, line 70: We might want to limit this to `/^https:\/\/homebrewery.*\.herokuapp\.com$/` so other random apps can't get in.

server/app.js
```diff
@@ -55,6 +55,34 @@ app.use(bodyParser.json({ limit: '25mb' }));
...
 app.use(forceSSL);
 
+import cors from 'cors';
+
+// CORS Configuration
+const corsOptions = {
+    origin: (origin, callback) => {
+        const allowedOrigins = [
+            'https://homebrewery.naturalcrit.com',
+            'http://localhost:8000',
```
calculuschild, line 65: Is it possible for someone to send rogue commands from their local machine and maliciously rename authors without their permission?

server/app.js
```diff
@@ -312,6 +340,35 @@ app.get('/user/:username', async (req, res, next)=>{
 	return next();
 });
 
+//Rename Brews
+app.put('/api/user/rename', async (req, res) => {
+    const { username, newUsername } = req.body;
```
calculuschild, line 345: Might be good to check if `req.account` exists here to make sure only logged-in users are changing names. Not sure if we need to do anything special to get the logged in user if the request comes from naturalcrit but they should share the same cookie I think, so the JWT handler above should just work.

calculuschild (changes requested):
Left some review comments

server/middleware/check-client-version.js
5e-Cleric: Ideally, either that or only apply to internal calls.

server/app.js
```diff
@@ -476,12 +533,15 @@ app.get('/vault', asyncHandler(async(req, res, next)=>{
...
-app.use(asyncHandler(async (req, res, next)=>{
-	if (!req.route) return res.redirect('/'); // Catch-all for invalid routes
-		
-	const page = await renderPage(req, res);
-	if(!page) return;
-	res.send(page);
+app.use(asyncHandler(async (req, res, next) => {
+    if (!req.route && !req.path.startsWith('/api/')) {
+        return res.redirect('/');
+    }
```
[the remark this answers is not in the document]
5e-Cleric: Well, i added the change to stop my call rerouting in a loop, but i can test this again now that it is finished and see if i really need it. I'd keep it because any error in the api would end here, and this patch makes an error like that easier to identify. Not sure.

server/app.js
5e-Cleric: forgot to lint

server/app.js
5e-Cleric: true that

server/app.js
5e-Cleric: I should be able to change the allowed list based on the homebrewery domain as well.

5e-Cleric: patched, localhost will be added only if on local

server/app.js
5e-Cleric: limited it to ` /^https:\/\/(?:homebrewery-pr-\d+\.herokuapp\.com|naturalcrit-pr-\d+\.herokuapp\.com)$/;` to be completely sure.

server/app.js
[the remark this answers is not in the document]
calculuschild: Any error in the API should be caught in the API and send the  error response to the client. If the API call is still reaching this point that is a bug.

server/app.js
calculuschild: I think it's worth it to keep it inline with the other APIs we have for user account features. Should just be a 1-line addition here I believe.

server/app.js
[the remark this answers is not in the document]
5e-Cleric: change reverted

server/app.js
5e-Cleric: will fix this in the next PR, cannot test if the req.account is getting through in a deployment

5e-Cleric:
This should be good to merge, i could use a review.

server/app.js
```diff
@@ -243,6 +280,7 @@ app.get('/css/:id', asyncHandler(getBrew('share')), (req, res)=>{getCSS(req, res
 //User Page
 app.get('/user/:username', async (req, res, next)=>{
 	const ownAccount = req.account && (req.account.username == req.params.username);
+	console.log(req.account);
```
calculuschild, line 283: Just a note we should eventually remove all these logs of user account details after debugging this feature is complete.

server/middleware/check-client-version.js
```diff
@@ -1,14 +1,22 @@
...
-		return res.status(412).send({
-			message : `Client version ${userVersion} is out of date. Please save your changes elsewhere and refresh to pick up client version ${version}.`
-		});
-	}
+    if (origin && origin !== `http://${sameSite}` && origin !== `https://${sameSite}`) {
+        return next();  // Skip version check if the request is from another site, like naturalcrit.com
+    }
 
-	next();
+    const userVersion = req.get('Homebrewery-Version');
```
calculuschild, line 11: Please use tabs

server/middleware/check-client-version.js
```diff
@@ -1,14 +1,22 @@
...
+    const sameSite = req.get('Host');
 
-	if(userVersion != version) {
-		return res.status(412).send({
-			message : `Client version ${userVersion} is out of date. Please save your changes elsewhere and refresh to pick up client version ${version}.`
-		});
-	}
+    if (origin && origin !== `http://${sameSite}` && origin !== `https://${sameSite}`) {
+        return next();  // Skip version check if the request is from another site, like naturalcrit.com
+    }
```
calculuschild, line 9: I think it would be more appropriate to either move any cross-domain calls before we call `app..use(checkClientVersion)`, or move `checkClientVersion` to be path-specific so it only applies to so database writes where client/server version matters, like "update" or "new" or "delete":

```
router.use('/api', checkClientVersion);
router.post('/api', asyncHandler(api.newBrew));
router.put('/api/:id', asyncHandler(api.getBrew('edit', true)), asyncHandler(api.updateBrew));
router.put('/api/update/:id', asyncHandler(api.getBrew('edit', true)), asyncHandler(api.updateBrew));
router.delete('/api/:id', asyncHandler(api.deleteBrew));
router.get('/api/remove/:id', asyncHandler(api.deleteBrew));
router.get('/api/theme/:renderer/:id', asyncHandler(api.getThemeBundle));
```

becomes

```
router.post('/api', checkClientVersion, asyncHandler(api.newBrew));
router.put('/api/:id', checkClientVersion, asyncHandler(api.getBrew('edit', true)), asyncHandler(api.updateBrew));
router.put('/api/update/:id', checkClientVersion, asyncHandler(api.getBrew('edit', true)), asyncHandler(api.updateBrew));
router.delete('/api/:id', checkClientVersion, asyncHandler(api.deleteBrew));
router.get('/api/remove/:id', checkClientVersion, asyncHandler(api.deleteBrew));
router.get('/api/theme/:renderer/:id', asyncHandler(api.getThemeBundle));
```

server/app.js
5e-Cleric: This one can be removed now, the rest yes of course

server/middleware/check-client-version.js
5e-Cleric: will do

server/middleware/check-client-version.js
5e-Cleric: This makes sense too, got it

calculuschild (approved):

PR #3953, opened 2024-12-19
Fix admin added as author #3952

### LINKED ISSUE #3952: Using Script Removal Tool adds admin user to brew as an Author
### Your idea:

When using the new Script Removal Tool, the admin user is added to the brew as an Author.
While easily rectified, this should not happen.

G-Ambatte:
This PR resolves #3952.

This PR removes the account property from the brew update initiated by the script cleaning tool. This prevents the Admin user performing the cleaning from being added to the brew as an author.

### DIFF
--- a/server/admin.api.js
+++ b/server/admin.api.js
@@ -108,6 +108,9 @@ router.put('/admin/clean/script/:id', asyncHandler(HomebrewAPI.getBrew('admin', 
 
 	req.body = brew;
 
+	// Remove Account from request to prevent Admin user from being added to brew as an Author
+	req.account = undefined;
+
 	return await HomebrewAPI.updateBrew(req, res);
 });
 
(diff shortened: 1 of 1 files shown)


calculuschild:
Did you already test this works?

G-Ambatte:
> Did you already test this works?

Yes, when running my test branch, the script cleaning tool removed the scripts and did NOT add the current logged in user as an author.
It might be useful to test it in a deployment before we push it fully live.

calculuschild:
Deploying now

G-Ambatte:
Before cleaning:
image

After cleaning:
image

calculuschild:
That's good enough for me.

ericscheid:
>> Yes, when running my test branch, the script cleaning tool removed the scripts and did NOT add the current logged in user as an author.

What if the current user was already an author .. do they get removed, or is it truly “does not add as an author”?

G-Ambatte:
During the updateBrew process, the current user (from req.account.username) is added to the Authors array, which is uniqified (?) to ensure that there are no duplicate elements. 

This PR removes the account property from the outgoing request, ensuring that the Authors list is not changed by this update - the current user is NOT being removed from the list; rather the change ensures that they are never able to be added while performing a script tag cleaning.

PR #3960, opened 2024-12-22
Add google refresh info to error page #3955

### LINKED ISSUE #3955: Add steps to refresh credentials to 403 error page
## Description

When we have a credentials issue during an API call (i.e., saving a brew), the `error-navitem.jsx` displays an actual instruction to refresh credentials.

If the error occurs from visiting a page, we send them to the `/error` page with a 403 error, which does not have a special instruction to refresh credentials, just a generic message for HBErrorCode 01.

https://github.com/naturalcrit/homebrewery/blob/980a7bd57ee3baa899c35bffc3c2d492e4cf0bd9/client/homebrew/pages/errorPage/errors/errorIndex.js#L17-L21

They are the same error from Google, so we should update the error message in the `/error` page case to please refresh Google Credentials.

___

Note that HBErrorCode 01 also covers status 400 errors as well. "invalid grant - bad request". We could split that into a separate HBErrorCode.

Although... looking at what it means, it seems its also usually an issue with credentials. User revoked access to the app, has a corrupted refresh token, recently changed their Google password, etc. Logging in again would also fix those cases.

Also note, `error-navitem.jsx` is just treating any error with `google` in the name as a credentials issue. We could probably split that into special handling of HBErrorCode 01 (credentials) and 02 (missing file).

G-Ambatte:
This PR resolves #3955.

This PR adds guidance on how a user can refresh their Google credentials directly to the Homebrewery error message.

The instructions are added as a constant so that they are consistent between HB Error Codes 01 and 02, and can then be easily added to any other messages that would also benefit from the instructions.

### DIFF
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -70,7 +70,7 @@ const errorIndex = (props)=>{
 			- **The Google Account may be closed.** Google may have removed the account
 			due to inactivity or violating a Google policy. Make sure the owner can
 			still access Google Drive normally and upload/download files to it.
-			:
+			
 			If the file isn't found, Google Drive usually puts your file in your Trash folder for
 			30 days. Assuming the trash hasn't been emptied yet, it might be worth checking.
 			You can also find the Activity tab on the right side of the Google Drive page, which
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -7,7 +7,7 @@ const loginUrl = 'https://www.naturalcrit.com/login';
 
 const errorIndex = (props)=>{
 	const googleRefreshInstructions = dedent`
-	:
+	&nbsp;
 	
 	### Refreshing your Google Credentials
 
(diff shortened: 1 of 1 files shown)


G-Ambatte:
Google credential refresh instructions, as they appear at the bottom of a `HBErrorCode: 02` error page for a HB brew `File Not Found` error.

image

calculuschild:
For now, I moved it just into HBErrorCode 01, since I think credentials aren't going to help in the case the brew is actually missing (02). 

I think we should also separate the `errornavItem.jsx` handling of those two cases so as to not confuse users when refreshing credentials does not restore the file they deleted.


If we keep getting reports specifically on the 02 case and we find that refreshing creds works, we can put it back in there.

G-Ambatte:
> For now, I moved it just into HBErrorCode 01, since I think credentials aren't going to help in the case the brew is actually missing (02).
> 
> I think we should also separate the `errornavItem.jsx` handling of those two cases so as to not confuse users when refreshing credentials does not restore the file they deleted.
> 
> If we keep getting reports specifically on the 02 case and we find that refreshing creds works, we can put it back in there.

If I recall correctly, attempting to get the file with expired permissions results in a 404, exactly as if the file was missing. But we can deal with that in a later PR.

PR #3965, opened 2024-12-24
Implement content-visibility on pages

calculuschild:
## Description

Adds CSS properties `content-visibility` and `contain-intrinsic-size`, and changes `contain` to "strict` for Legacy and V3 `.page` components.

Dramatically speeds up reflow calculations on large brews, as pages outside of the viewport can be skipped.

Noticed extreme lag when testing #3845. This seems to solve *most* of it.

@5e-Cleric brought up a possible solution to speed up the brewRenderer earlier in the Gitter Chat

> Dec 12, 2024
> 5e-Cleric (Victor)
> https://www.trevorlasn.com/blog/css-content-visibility
> This one could be useful

Load this test brew (500+ pages) and compare lag with/without the PR upon changing zoom level or page layout. I don't think this will improve scrolling performance. Confirm normal page display, scrolling/jumping, toolbar behavior is otherwise unchanged.

[Test Brew](https://github.com/user-attachments/files/17656392/HB-NecromanticArmorofSaladDressing.1.txt)

### DIFF
--- a/themes/Legacy/5ePHB/style.less
+++ b/themes/Legacy/5ePHB/style.less
@@ -42,23 +42,25 @@ body {
 }
 .phb, .page{
 	.useColumns();
-	counter-increment : phb-page-numbers;
-	position          : relative;
-	z-index           : 15;
-	box-sizing        : border-box;
-	overflow          : hidden;
-	height            : 279.4mm;
-	width             : 215.9mm;
-	padding           : 1.0cm 1.7cm;
-	padding-bottom    : 1.5cm;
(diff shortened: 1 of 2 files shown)


5e-Cleric:
I see no effect on scroll, in fact it shows only the background on scroll, as opposed to mostly renderered pages when scrolling the same doc on live site.

When zooming out there is a significant performance improvement. Also when changing view type when zoomed out. Overall good changes.

5e-Cleric (approved):

PR #3968, opened 2025-01-02
Create delete route for account deletion

5e-Cleric:
## Description

Just one change, adds a deletion route in app.js, which makes a request to the delete method for all documents based on username.

The delete method should handle deletion or author removal automatically.

## Related Issues or Discussion

I might transform this into a clone of the delete brew function but working in batches

After this PR is merged i plan on moving some of the routes in app.js into other files, as that one is starting to get gargantuan.


server/app.js
```diff
@@ -376,6 +376,31 @@ app.put('/api/user/rename', async (req, res)=>{
...
 
+//Delete brews based on author
+app.delete('/api/user/delete', async (req, res)=>{
+	const { username } = req.body;
+	const ownAccount = req.account && (req.account.username == username);
+	
+	if(!username)
+		return res.status(400).json({ error: 'Username and newUsername are required.' });
+	if(!ownAccount)
+		return res.status(403).json({ error: 'Must be logged in to change your username' });
```
calculuschild, line 387: Change error text since we are not changing username.

server/app.js
```diff
@@ -376,6 +376,31 @@ app.put('/api/user/rename', async (req, res)=>{
...
+
+	try {
+		const brews = await HomebrewModel.getByUser(username, true, ['authors']);
+		await brews.forEach((brew)=>{
+			request.delete(`/api/${brew.googleId ?? ''}${brew.editId}`).send().end((err, res)=>{
+				if (err) reportError(err);
```
calculuschild, line 393: `reportError` is not defined. This line can be removed.

If you call `deleteBrew()` directly, any error will already be caught by your try/catch block and handled there.

server/app.js
```diff
@@ -376,6 +376,31 @@ app.put('/api/user/rename', async (req, res)=>{
...
+        	});
+		});
+
+		return res.json({ success: true, message: `Brews for ${username} renamed to ${newUsername}.` });
```
calculuschild, line 397: `newUsername` is not defined. Change success message.

server/app.js
```diff
@@ -376,6 +376,31 @@ app.put('/api/user/rename', async (req, res)=>{
...
+	} catch (error) {
+		console.error('Error renaming brews:', error);
+		return res.status(500).json({ error: 'Failed to rename brews.' });
```
calculuschild, line 400: We are not renaming brews. Change error message.

server/app.js
```diff
@@ -376,6 +376,31 @@ app.put('/api/user/rename', async (req, res)=>{
...
+		return res.status(400).json({ error: 'Username and newUsername are required.' });
+	if(!ownAccount)
+		return res.status(403).json({ error: 'Must be logged in to change your username' });
+
+	try {
+		const brews = await HomebrewModel.getByUser(username, true, ['authors']);
+		await brews.forEach((brew)=>{
+			request.delete(`/api/${brew.googleId ?? ''}${brew.editId}`).send().end((err, res)=>{
+				if (err) reportError(err);
+        	});
```
calculuschild, line 394: tabs

calculuschild (changes requested):
Left some fixes.

server/app.js
```diff
@@ -376,6 +376,45 @@ app.put('/api/user/rename', async (req, res)=>{
 	}
 });
 
+//Delete brews based on author
+app.delete('/api/user/delete', async (req, res) => {
+	const { username } = req.body;
+	console.log(username);
+	const ownAccount = req.account && (req.account.username == username);
+	
+	// if(!ownAccount) return res.status(403).json({ error: 'Must be logged in to change your username' });
```
calculuschild, line 385: Does this need to be uncommented? (and change the error message to "Must be logged in to delete your account!")

server/app.js
```diff
@@ -376,6 +376,45 @@ app.put('/api/user/rename', async (req, res)=>{
 	}
 });
 
+//Delete brews based on author
+app.delete('/api/user/delete', async (req, res) => {
+	const { username } = req.body;
+	console.log(username);
```
calculuschild, line 382: Can probably remove this console.log

server/app.js
```diff
@@ -376,6 +376,45 @@ app.put('/api/user/rename', async (req, res)=>{
...
+	try {
+		const brews = await HomebrewModel.getByUser(username, true, ['authors']);
+		//get the relevant fields, not just author you moron!
+		console.log(brews);
+		
+		for (let brew of brews) {
+			//attaching the brew to the request for the deleteBrew method to work
+			req.brew = brew;
+
+			await new Promise((resolve, reject) => {
```
calculuschild, line 396: Would it make sense to use `Promise.all` as you did in the username change function above?

server/app.js
```diff
@@ -376,6 +376,45 @@ app.put('/api/user/rename', async (req, res)=>{
...
+app.delete('/api/user/delete', async (req, res) => {
+	const { username } = req.body;
+	console.log(username);
+	const ownAccount = req.account && (req.account.username == username);
+	
+	// if(!ownAccount) return res.status(403).json({ error: 'Must be logged in to change your username' });
+
+	try {
+		const brews = await HomebrewModel.getByUser(username, true, ['authors']);
+		//get the relevant fields, not just author you moron!
```
calculuschild, line 389: I can't recall why we needed the other fields. Does this still need to be fixed?

calculuschild (changes requested):

calculuschild:
What else is needed on this in terms of testing? I recall there was some issue with CORS or something?

5e-Cleric:
@calculuschild this would be way easier if the account systems in local were connected like they are on live, otherwise its a mess, i can do it, but it is a mess.

server/app.js
5e-Cleric: i can't recall either, working on it

5e-Cleric: ok yes, because the deletebrew function needs at least the google id and edit id

server/app.js
5e-Cleric: done

server/app.js
5e-Cleric: yes, and yes, i had it commented because that would not work in local

5e-Cleric:
This should be done, but it cannot be tested thoroughly:

 to test in local you must make up a req.account object with at least a username
 same thing in a deployment since the cookies do not work through heroku deployments

5e-Cleric:
this should be released before the actual deletion api is set in naturalcrit.

calculuschild:
> @calculuschild this would be way easier if the account systems in local were connected like they are on live, otherwise its a mess, i can do it, but it is a mess.

They are connected in local in the same way. Log in via a naturalcrit server on localhost, and you will be logged in on the homebrewery at localhost.

If you are not seeing that there is a bug somewhere.

5e-Cleric:
> > @calculuschild this would be way easier if the account systems in local were connected like they are on live, otherwise its a mess, i can do it, but it is a mess.
> 
> They are connected in local in the same way. Or they should be; I used it extensively when I was implementing Google Drive integration. Log in via a naturalcrit server on localhost, and you will be logged in on the homebrewery at localhost.
> 
> If you are not seeing that there is a bug somewhere, or I am misunderstanding what you mean.

The logins are not connected for me at all, logins are kept or at least require hard reloads, and logging in homebrewery just requires you to type a username and it logs you in, while in naturalcrit you need to provide a valid username and password, how is that connected?

calculuschild:
> logging in homebrewery just requires you to type a username and it logs you in, while in naturalcrit you need to provide a valid username and password, how is that connected?

We added the username-only login as a convenience for people using an offline deployment of Homebrewery. It creates a simplified cookie and does not go through the user authentication process at all.But if you have both servers running, you should still be able to log in from the naturalcrit server, which generates a cookies that will work on the Homebrewery Server.

5e-Cleric:
Huh, well, nevertheless, the PR works

calculuschild:
> Huh, well, nevertheless, the PR works

Unfortunately, it still doesn't work. If there is more than one brew on an account, it only deletes the first one, and then crashes the Homebrewery side. The Naturalcrit side will continue to delete the account but the rest of the brews remain.

Homebrewery provides this log:

```
node:_http_outgoing:848
    throw new ERR_HTTP_HEADERS_SENT('remove');
          ^

Error [ERR_HTTP_HEADERS_SENT]: Cannot remove headers after they are sent to the client
    at ServerResponse.removeHeader (node:_http_outgoing:848:11)
    at ServerResponse.send (C:\Users\Trevor\Documents\GitHub\homebrewery\node_modules\express\lib\response.js:201:10)
    at Object.deleteBrew (file:///C:/Users/Trevor/Documents/GitHub/homebrewery/server/homebrew.api.js:529:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  code: 'ERR_HTTP_HEADERS_SENT'
}
```

Looks like this line does not wan to be called more than once, i.e., can't reuse the same `res` object to return multiple responses.
https://github.com/naturalcrit/homebrewery/blob/bfd14757c28103bb73ba7fabc4c870bc6d3901be/server/homebrew.api.js#L529

Before, you were directly deleting from homebrewModel, which *would* work, except that does not delete Google Drive documents, and also skips over some of the other possibilities (deleting yourself as author instead of the whole file if you are a co-author, etc.), and this is where I suggested we just reuse the api.deleteBrew(). However, I did not realize multiple responses would be an issue here.

What we may need to do is pull out the core part of `api.deleteBrew()` (maybe most of it) into its own function that does not rely on the req/res objects, and call that instead. Similar to how we have `api.excludePropsFromUpdate()` pulled out of `api.updateBrew()`. Then `api.deleteBrew()` becomes just a wrapper that takes in req, res, and pulls out the necessary values for the actual delete step.

calculuschild:
I need to go through and re-test this. When you get a chance can you resolve the conflicts since path-routes was made?

5e-Cleric:
conflicts resolved

PR #3922, opened 2024-11-27
Refactor sharepage as functional comp

5e-Cleric:
one more in the long list of refactors.

PR #3975, opened 2025-01-09
Apply fix to clear error.

### LINKED ISSUE #3974: SharePage: Error in console
### Your idea:

image

image

The issue is in SharePage.jsx, line 26.

dbolack-ab:
## Description

Replace an updateState with this.setState

- Closes #3974

### DIFF
--- a/client/homebrew/pages/sharePage/sharePage.jsx
+++ b/client/homebrew/pages/sharePage/sharePage.jsx
@@ -23,7 +23,9 @@ const SharePage = (props)=>{
 	});
 
 	const handleBrewRendererPageChange = useCallback((pageNumber)=>{
-		this.setState({ currentBrewRendererPageNum: pageNumber });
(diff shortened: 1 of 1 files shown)


calculuschild:
This will not work. 

Needs two changes:

1) use `setState`, without `this` (functional component has no "this")

2) must also set the other fields in state or they will be lost.

calculuschild (changes requested):
The code as written is still causing errors.

Needs two changes:

1) use `setState`, without `this` (functional component has no "this")

2) must also set the other fields in state or they will be lost, i.e. using prevState: `setState((prevState)=>{...`

5e-Cleric:
I forgot #3922 was merged before this

calculuschild:
Seems to be working now. Merging. Thanks @dbolack-ab !

⭐ 🎉 💯

PR #3978, opened 2025-01-11
Wrap titles in error messages with pre blocks to prevent rendering.

### LINKED ISSUE #3231: Brew title parsed as markdown on Error pages. e.g. images will be rendered.

dbolack-ab:
## Description

Attempt to prevent URLS  in titles ( which shouldn't be there! ) from rendering by wrapping it in a <pre> block.

> [!CAUTION]

- Closes #3231

Trigger error page on a document with an image URL in the Title.

### DIFF
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -89,7 +89,7 @@ const errorIndex = (props)=>{
 		
 		:
 
-		**Brew Title:** ${props.brew.brewTitle || 'Unable to show title'}
+		**Brew Title:** \`${props.brew.brewTitle || 'Unable to show title'}\`
 
 		**Current Authors:** ${props.brew.authors?.map((author)=>{return `[${author}](/user/${author})`;}).join(', ') || 'Unable to list authors'}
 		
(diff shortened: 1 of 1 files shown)


calculuschild:
I think here we want to actually use `<pre>` instead of backticks, so we don't also get `<code>`.

5e-Cleric:
nothing remains fixed:

https://homebrewery-pr-3978.herokuapp.com/edit/mG30Ekhd1AjL

image

calculuschild:
Mmmmm. I was mistaken about `<pre>`. I think we still don't want `<code>` since that has its own formatting, but lodash `_.escape()` Looks like it might do what we need.

calculuschild:
No, I am mistaken again. The document title is `cat warrior`, so escaping HTML does nothing; it is still just Markdown at that point and *then* gets converted to an HTML image.

calculuschild:
Added a simple custom escape function. Just inserts a `\\` before every character in the string to escape it. Then it cannot be parsed as markdown.

5e-Cleric:
Can confirm, this approach works, considering merging straight up.

PR #3979, opened 2025-01-11
Update Dedent

5e-Cleric:
Dedent tabs is deprecated and dedent is maintained again, this PR updates the repo accordingly. As far as i can see, all features work as intended without issue.


calculuschild:
`dedent` does not contain the fix added to `dedent-tabs` for interpolated newlines. This breaks Table of Contents generation.

5e-Cleric:
From Gitter:


> Yeah, the fix is not included. This breaks our Table of Contents generation.
> Basically including an indented string inside another string loses the indents of the inserted string.

```
x = dedent`
String 
${insertedString} // this loses any indents inside
String 
`
```
>
```
 insertedString = dedent`
  One
    Two
      Three`;
```

PR #3988, opened 2025-01-14
Fix issue with ':' in mustache block quoted text

### LINKED ISSUE #3886: CSS property setting syntax does not permit the use of a colon (:) in the assigned value.
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

In writing a how-to I discovered I was unable to assign a string value to a CSS property if it contained a `:` via the `{{}}` style setting syntax.

### Code

```gfm
{{footer,--chapterName:"'1:How-to'"}}
```

Generates 
```
<span class="inline-block footer" style="--chapterName:;">“‘1:How-to’”</span>
```

Instead of 

```
<span class="inline-block footer" style="--chapterName:‘1:How-to’;”></span>
```

dbolack-ab:
This fixes the issue with CSS Attribute assignment strings containing `:` in mustache blocks

## Description

This updates the regex to permit `:` in the right side of a CSS attribute/property or HTML attribute assignment.

- Closes #3886

```gfm
{{footer,--chapterName:"1:How-to"}}
```

Should now generate
```
<span class="inline-block footer" style="--chapterName:‘1:How-to’;”></span>
```

Instead of 

```
<span class="inline-block footer" style="--chapterName:;">“‘1:How-to’”</span>
```

etc. Should work for all four mustache situations.

### Reviewer Checklist

  - [ ] Verify with each mustache type.
- [ ] Does this require new tests?


ericscheid:
A very similar regex issue occurs with `{{footer,--chapterName:"'1=How-to'"}}`. Can that also be fixed in this PR?

dbolack-ab:
> A very similar regex issue occurs with `{{footer,--chapterName:"'1=How-to'"}}`.
> 
> Hmm .. testing the regex in regex101 suggests it should work already, except it doesn't.
> 
> https://regex101.com/r/9gEWyC/1

Yeah, I didn't test those because I wasn't trying to solve for them. :)

Let me see what I can find.

5e-Cleric:
this fix is in good place, just tested and it works, should go for a full testing to make sure we don't break anything else

5e-Cleric:
Consider this will have to merge with #4004

shared/naturalcrit/markdown.js
```diff
@@ -828,7 +828,7 @@ const processStyleTags = (string)=>{
 
 	const id         = _.remove(tags, (tag)=>tag.startsWith('#')).map((tag)=>tag.slice(1))[0]        || null;
 	const classes    = _.remove(tags, (tag)=>(!tag.includes(':')) && (!tag.includes('='))).join(' ') || null;
-	const attributes = _.remove(tags, (tag)=>(tag.includes('='))).map((tag)=>tag.replace(/="?([^"]*)"?/g, '="$1"'))
+	const attributes = _.remove(tags, (tag)=>(tag.includes('=')) && (tag.slice(0, tag.indexOf(':')).indexOf('=') > 0)).map((tag)=>tag.replace(/="?([^"]*)"?/g, '="$1"'))
```
calculuschild, line 831: Can you clarify what this added check is doing?

shared/naturalcrit/markdown.js
```diff
@@ -127,7 +127,7 @@ const mustacheSpans = {
 	start(src) { return src.match(/{{[^{]/)?.index; },  // Hint to Marked.js to stop and check for a match
 	tokenizer(src, tokens) {
 		const completeSpan = /^{{[^\n]*}}/;               // Regex for the complete token
-		const inlineRegex = /{{(?=((?:[:=](?:"['\w,\-()#%=?.: ]*"|[\w\-()#%.]*)|[^"=':{}\s]*)*))\1 *|}}/g;
+		const inlineRegex = /{{(?=((?:[:=](?:"['\w,\-()#%=?.: ]*"|'[\w,\-()#%=?.: ]*'|[\w\-()#%.]*)|[^"=':{}\s]*)*))\1 *|}}/g;
```
calculuschild, line 132: I'm not sure I understand why we needed to allow `'single quotes'` for this PR to work? If not needed can we remove this change?

dbolack-ab: I suppose you can.

shared/naturalcrit/markdown.js
dbolack-ab: I think it is debris I missed from a different fix attempt. I can't see any sane reason for it not to be true if the previous test is true.

shared/naturalcrit/markdown.js
dbolack-ab: Reverted in next commit.

dbolack-ab:
There is a hard regression in here I am attempting to locate.

calculuschild:
@dbolack-ab Ping me when you are ready for me to merge and I will try to do it today.

dbolack-ab:
@calculuschild Ping

dbolack-ab:
@5e-Cleric The issue was that the example was wrong. This pattern no longer requires triple quotes ( `"'foo'"` ), only double quotes.

5e-Cleric:
Ok, the suggested behaviour works, but breaks a prior feature: #3066 , that was merged to allow proper text content, very in line with the intention of this very PR:

`{{footer,--chapterName:"How'to" }}` works live but doesn't in this PR.

calculuschild:
> Ok, the suggested behaviour works, but breaks a prior feature: #3066 , that was merged to allow proper text content, very in line with the intention of this very PR:
> 
> `{{footer,--chapterName:"How'to" }}` works live but doesn't in this PR.

I think this commit mistakenly removed the single quote from the first half of the curly span regex, when it should have only removed from the second part.

https://github.com/naturalcrit/homebrewery/commit/33bb08391fac1845ae526fc24d8a914bed1c10b9

But it sounds like this should be added as a test case.

calculuschild:
> @5e-Cleric The issue was that the example was wrong. This pattern no longer requires triple quotes ( `"'foo'"` ), only double quotes.

There is an issue with this approach. The double quotes were needed to allow for values with spaces:

```
{{--myShadow;"5px 5px 15px rgba(0, 0, 0, 0.3)" my text}}
```

Now, the custom property is forced to be a string:

```
<span style="--myShadow;'5px 5px 15px rgba(0, 0, 0, 0.3)'";>my text</span>
```

Doing something like `box-shadow: var(--myShadow)` is no longer valid because `--myShadow` is a string instead of a shadow parameter.

@5e-Cleric Is this a concern that all quoted values for custom properties will be converted to strings? I feel this may be a breaking change for some styles by our users.

calculuschild:
Just a random thought. If triple-quoting is just not something we can bear to live with, perhaps we could just try to add some validation when we see a quoted string. Run a check to determine if it was intended as a CSS value or should be handled as a string. Found this somewhere:

```
function shouldQuoteCSSValue(value) {
  // Regular expressions for common CSS formats
  const cssNumberRegex = /^[0-9]+(\.[0-9]+)?(px|em|rem|vh|vw|%)?$/;
  const cssColorRegex = /^(#[0-9a-fA-F]{3,8}|rgba?\(.+\)|hsla?\(.+\))$/;
  const cssFunctionRegex = /^[a-zA-Z-]+\(.+\)$/;

  // If it matches known CSS patterns, no quotes needed
  if (cssNumberRegex.test(value) || cssColorRegex.test(value) || cssFunctionRegex.test(value)) {
    return false;
  }

  // Otherwise, assume it's a string
  return true;
}

// Example tests
console.log(shouldQuoteCSSValue("5px 5px 15px rgba(0, 0, 0, 0.3)")); // false
console.log(shouldQuoteCSSValue("5x shakalakala")); // true
console.log(shouldQuoteCSSValue("hello world")); // true
console.log(shouldQuoteCSSValue("rotate(45deg)")); // false
```

dbolack-ab:
So I am, at this point, confused. Should I revert this all the way back to just :?

5e-Cleric:
> > @5e-Cleric The issue was that the example was wrong. This pattern no longer requires triple quotes ( `"'foo'"` ), only double quotes.
> 
> There is an issue with this approach. The double quotes were needed to allow for values with spaces:
> 
> ```
> {{--myShadow;"5px 5px 15px rgba(0, 0, 0, 0.3)" my text}}
> ```
> 
> Now, the custom property is forced to be a string:
> 
> ```
> <span style="--myShadow;'5px 5px 15px rgba(0, 0, 0, 0.3)'";>my text</span>
> ```
> 
> Doing something like `box-shadow: var(--myShadow)` is no longer valid because `--myShadow` is a string instead of a shadow parameter.
> 
> @5e-Cleric Is this a concern that all quoted values for custom properties will be converted to strings? I feel this may be a breaking change for some styles by our users.
> 
> But also... changing quotes behavior is outside the scope of the original issue, which was just that `:` isn't allowed inside quotes. Maybe that can be pulled out into its own issue to be discussed and we can just apply the regex for `:` now?

I'm not sure if you are suggesting a change or expressing what the behaviour will be if this gets merged

calculuschild:
> I'm not sure if you are suggesting a change or expressing what the behaviour will be if this gets merged

I'm pointing out a breaking behavior change if this gets merged, and asking if it is acceptable.

We can always just fix the part with the `:` and leave any quote handling changes for later discussion.

calculuschild:
> So I am, at this point, confused. Should I revert this all the way back to just :?

@dbolack-ab I think we should revert back to just `:`, since that's what the original issue was about. Any changes to how quotes are handled is outside the scope of that issue. We can merge the `:` fix on its own and get it released.

But, we should start a separate to discuss the quote handling, and you can copy the code for that in there so it isn't lost. In general I like the idea of not needing triple quotes. But I don't want to push a breaking change on our users if it prevents certain uses of custom CSS variables.

5e-Cleric:
> > I'm not sure if you are suggesting a change or expressing what the behaviour will be if this gets merged
> 
> I'm pointing out a breaking behavior change if this gets merged, and asking if it is acceptable.
> 
> We can always just fix the part with the `:` and leave any quote handling changes for a separate discussion/PR.

As with any PR, breaking changes should be avoided, specially when the feature to be broken was implemented intentionally.

dbolack-ab:
Reworking

PR #3992, opened 2025-01-15
Implement suggested fix for 3448

### LINKED ISSUE #3448: Markdown variables consume too many characters if parentheses are nested with whitespace before final )
Not sure what is causing this issue, but it is a bug of some sort.

Some combination of links and parentheses inside tables merges cells together, sometimes splitting words.

the following code will work as a demonstration:
```
|  title 1  | title 2 | title 3 | title 4|
|-----------|---------|---------|--------|
|[foo](bar) |  Ipsum  |    )    |   )    |
```
results in this:
image

dbolack-ab:
## Description

Per issue, remove trim sequence and remove unneeded if wrapper.

```
|  title 1  | title 2 | title 3 | title 4|
|-----------|---------|---------|--------|
|[foo](bar) |  Ipsum  |    )    |   )    |
```

image

- Closes #3448

### Reviewer Checklist

  - [ ] Test as above


shared/naturalcrit/markdown.js
```diff
@@ -680,7 +681,7 @@ function MarkedVariables() {
 					}
 					if(match[8]) { // Inline Definition
 						const label = match[10] ? match[10].trim().replace(/\s+/g, ' ') : null; // Trim edge spaces and shorten blocks of whitespace to 1 space
-						let content = match[11] ? match[11].trim().replace(/\s+/g, ' ') : null; // Trim edge spaces and shorten blocks of whitespace to 1 space
+						let content = match[11] ? match[11] : null; // Trim edge spaces and shorten blocks of whitespace to 1 space
```
calculuschild, line 684: We can remove the comment since we aren't trimming here anymore.

Also I realize now we can just use:

```suggestion
						let content = match[11] || null;
```

calculuschild (changes requested):
Since this fixes a new edge case in our Markdown, we should also add this as a test in our variables tests, to make sure we don't break it in the future.

Otherwise looks pretty straightforward.

tests/markdown/variables.test.js
```diff
@@ -402,4 +402,10 @@ describe('Variable names that are subsets of other names', ()=>{
 		const rendered = Markdown.render(source).trimReturns();
 		expect(rendered).toBe('<p>14</p>');
 	});
+});
+
+describe('Don\'t Eat all the parentheticals!', ()=>{
+	const source='\n|  title 1  | title 2 | title 3 | title 4|\n|-----------|---------|---------|--------|\n|[foo](bar) |  Ipsum  |    )    |   )    |\n';
+	const rendered = Markdown.render(source).trimReturns();
+	expect(rendered).toBe('<table><thead><tr><th>title 1</th><th>title 2</th><th>title 3</th><th>title 4</th></tr></thead><tbody><tr><td><a href=\"bar\">foo</a></td><td>Ipsum</td><td>)</td><td>)</td></tr></tbody></table>');
```
calculuschild, line 410: Tests need to be inside an `it()` call. `describe()` is a grouping that contains the individual `it()` tests, but the `describe()` isn't a test itself.

calculuschild:
Alright, I can see the test passing now. Ready to merge!

Issue #3901, opened 2024-11-20
Allow setting of HTML Attribute and CSS features at the <div class="page"> level


dbolack-ab:
### Your idea:

Presently, when a user wants to change CSS parameters of a Page classed div, it has to be done via `page :has()` mechanisms. While this works pretty well, it's not the most intuitive thing for folks new to CSS. 
Its primary failing is it only permits setting CSS Properties to predefined values.

This issue proposes that we extend the single-branch mustache syntax to be appendable to a `\page` and inserted into the page classed div's style, class, and arbitrary add-on attributes as desired.

Additionally, to support setting values on the first page, we need to consider dropping a completely empty Page 1, treating the `\page` definition as the start of the page, not the end. By only dropping the empty page body, we maintain backward compatibility.

example syntax:

```
\page{--TOC:"'exclude'",wide,XCUSTOMATTR="'Foo'"}
```

* I may have the quoting syntax inverted in the example.

5e-Cleric:
I agree this would make things easier for users, if they find out about this feature.

5e-Cleric:
This was implemented

PR #4004, opened 2025-01-24
Alternate \page{curlies}

calculuschild:
## Description

Takes #3899 and moves the handling of curlies back into Markdown.js. Small adjustments to regex, etc.

Opening as a new PR as I was getting too many merge conflicts to easily track. Note that these changes *only* apply to V3. Legacy should be unaffected.

##### Usage:
Similar to an inline single mustache.

`\page{wide,banna=fail,color:black}`

Solves https://github.com/naturalcrit/homebrewery/issues/3901

*Reviewers, refer to this list when testing features, or suggest new items*
  - [ ] Adding a style includes a style on the `<div class='page'>` instance.
  - [ ] Adding a class includes a class on the `<div class='page'>` instance.
  - [ ] Adding an HTML Attribute adds the attribute on the `<div class='page'>` instance.
  - [ ] Adding an ID does ***not*** add it to the `<div class='page'>` instance.
  - [ ] Mixed forms of above work as expected
  - [ ] CSS styles with hyphens (e.g. `line-height`) work (Page is a React component, which requires camelCase input)
  - [ ] Curlies mix properly with other pages styles (e.g. the page "shadow" option from the toolbar)
  - [ ] `\page` on the first line of a document will *not* create a page break, and is counted as page 1
  - [ ] `\page` works without values.
  - [ ] Highlighting of `\page` still works
  - [ ] Page number display is accurate
  - [ ] Jumping to source/brew still works in every case (including Legacy)
  - [ ] Ensure CSS Properties ( --FOO ) can be set.

<details><summary>Copy the list here</summary>
<p>

```

  - [ ] Adding a style includes a style on the `<div class='page'>` instance.
  - [ ] Adding a class includes a class on the `<div class='page'>` instance.
  - [ ] Adding an HTML Attribute adds the attribute on the `<div class='page'>` instance.
  - [ ] Adding an ID does ***not*** add it to the `<div class='page'>` instance.
  - [ ] Mixed forms of above work as expected
  - [ ] CSS styles with hyphens (e.g. `line-height`) work (Page is a React component, which requires camelCase input)
  - [ ] Curlies mix properly with other pages styles (e.g. the page "shadow" option from the toolbar)
  - [ ] `\page` on the first line of a document will *not* create a page break, and is counted as page 1
  - [ ] `\page` works without values.
  - [ ] Highlighting of `\page` still works
  - [ ] Page number display is accurate
  - [ ] Jumping to source/brew still works in every case (including Legacy)
  - [ ] Ensure CSS Properties ( --FOO ) can be set.

```

</p>
</details>

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -40,7 +40,7 @@ const BrewPage = (props)=>{
 		...props
 	};
 	const pageRef = useRef(null);
-	const cleanText = safeHTML(props.contents);
+	let cleanText = safeHTML(props.contents);
 
 	useEffect(()=>{
 		if(!pageRef.current) return;
@@ -78,6 +78,20 @@ const BrewPage = (props)=>{
 		};
 	}, []);
 
+	// Extract any page styles from `\page{cssProp:value}`
+	if(cleanText.match(/^<pagebreak/)) {
+		const injectedTags = Markdown.extractHTMLStyleTags(cleanText.substring(0, cleanText.indexOf('\n')));
+		const styleObject  = injectedTags.styles?.split(';').reduce((acc, curr) => {
+			const [key, value] = curr.split(':').map(item => item.trim());
+			const camelCaseKey = key.replace(/-([a-z])/g, g => g[1].toUpperCase()); //Convert to camelCase for React
+			acc[camelCaseKey] = value;
+			return acc;
+		}, {});
+
+		props.style = {...props.style, ...styleObject};
+		cleanText = cleanText.substring(cleanText.indexOf('\n'));
+	}
+	
 	return <div className={props.className} id={`p${props.index + 1}`} data-index={props.index} ref={pageRef} style={props.style}>
 	         <div className='columnWrapper' dangerouslySetInnerHTML={{ __html: cleanText }} />
 	       </div>;
@@ -126,7 +140,7 @@ const BrewRenderer = (props)=>{
 	if(props.renderer == 'legacy') {
 		rawPages = props.text.split('\\page');
 	} else {
-		rawPages = props.text.split(/^\\page$/gm);
+		rawPages = props.text.split(/^(?=\\page(?:{[^\n{}]+})?$)/gm);
 	}
 
 	const handlePageVisibilityChange = (pageNum, isVisible, isCenter)=>{
@@ -183,6 +197,11 @@ const BrewRenderer = (props)=>{
 
 			return <BrewPage className='page phb' index={index} key={index} contents={html} style={styles} onVisibilityChange={handlePageVisibilityChange} />;
 		} else {
+			let pageText2 = pageText.substring(0, pageText.indexOf('\n'));
+			let butt  = Markdown.marked.lexer(pageText2);
+			console.log(butt)
+
+
 			pageText += `\n\n&nbsp;\n\\column\n&nbsp;`; //Artificial column break at page end to emulate column-fill:auto (until `wide` is used, when column-fill:balance will reappear)
 			const html = Markdown.render(pageText, index);
 
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -176,6 +176,7 @@ const BrewRenderer = (props)=>{
 			...(!displayOptions.pageShadows ? { boxShadow: 'none' } : {})
 			// Add more conditions as needed
 		};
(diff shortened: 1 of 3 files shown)


5e-Cleric:
Using an empty inject breaks the syntax, which doesn't make much sense, allow for `\page{}`

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -40,7 +40,7 @@ const BrewPage = (props)=>{
 		...props
 	};
 	const pageRef = useRef(null);
-	const cleanText = safeHTML(props.contents);
+	let cleanText = safeHTML(props.contents);
```
calculuschild, line 43: Should stay as `const`. Probably leftover from my rearranging things

calculuschild:
> Using an empty inject breaks the syntax, which doesn't make much sense, allow for `\page{}`

Fixed.

calculuschild:
Any final comments, @5e-Cleric , @dbolack-ab , @ericscheid ?

5e-Cleric:
Right now, when setting style for the *first* page, spaces are being allowed between page break and style inject, but not for other pages. Additionally, these spaces, while they don't break the regex for the inject, do break the syntax highlight.

5e-Cleric:
Also, class or id injection doesn't work for either ( `\page{#myPageId,myPageClass}`

5e-Cleric:
Will update these as i go
  - [ ] Adding a style includes a style on the `<div class='page'>` instance.
  - [ ] Adding a class includes a class on the `<div class='page'>` instance.
  - [ ] Adding an HTML Attribute adds the attribute on the `<div class='page'>` instance.
  - [ ] Adding an ID does ***not*** add it to the `<div class='page'>` instance.
  - [ ] Mixed forms of above work as expected
  - [ ] CSS styles with hyphens (e.g. `line-height`) work (Page is a React component, which requires camelCase input)
  - [ ] Curlies mix properly with other pages styles (e.g. the page "shadow" option from the toolbar)
  - [ ] `\page` on the first line of a document will *not* create a page break, and is counted as page 1
  - [ ] `\page` works without values.
  - [ ] Highlighting of `\page` still works
  - [ ] Page number display is accurate
  - [ ] Jumping to source/brew still works in every case (including Legacy)
  - [ ] Ensure CSS Properties ( --FOO ) can be set.

5e-Cleric:
`/page` in the first line in legacy creates a page, not in v3, i assume this is intended

5e-Cleric:
Custom properties `--foo` are not being set, furthermore, a custom property with the same name as a norrmal property is added as that normal property, so `--border` is added as `border`.

calculuschild:
> Right now, when setting style for the first page, spaces are being allowed between page break and style inject, but not for other pages. Additionally, these spaces, while they don't break the regex for the inject, do break the syntax highlight.

I see what is going on here...  First page is the only one that might go into the Marked parser without being filtered through the page split regex. Ok. So to align with the other injector behavior, I should allow spaces in every case.

> Also, class or id injection doesn't work for either ( \page{#myPageId,myPageClass}

Good catch; somehow I deleted the class part when doing merges. Will put that back in. ID's are intentionally left out, else it would mess with page selection on the toolbar, etc.

> /page in the first line in legacy creates a page, not in v3, i assume this is intended

Correct. This PR only applies to V3.

> Custom properties --foo are not being set, furthermore, a custom property with the same name as a norrmal property is added as that normal property, so --border is added as border.

I see what happened. When I added the camelCase conversion it also converts `--my-custom-variable` to `myCustomVariable`. I will tweak that line to exclude converting properties starting with `--`.

calculuschild:
@5e-Cleric I have fixed all of these cases you reported. Can you confirm?

5e-Cleric:
I don't think it is a problem, but the order of attributes is different, can we fix that?

image

calculuschild:
Can you clarify what steps result in the different order?

As for testing, I'm not fully sure how we would go about testing React output. I might put it off for later unless you know of a good strategy.

5e-Cleric:
> Can you clarify what steps result in the different order?

Making the same inject in page 1 vs page 2, with some text in the middle, not sure either.

calculuschild:
> Making the same inject in page 1 vs page 2, with some text in the middle, not sure either.

Interesting. I'm not sure how much control we have over the order React chooses here. Given that Page 1 is the only different one, my best guess is it has something to do with the very first "empty" render before any pages have been split and parsed, But this doesn't impact the behavior of the document at all so I don't know if it's worth hunting down.

I think I will merge this and we can decide if its worth the effort to clean up the ordering later.

PR #4005, opened 2025-01-24
Workaround for unclosed <pre> blocks before rendering.

### LINKED ISSUE #3206: Unclose Code Fence syntax reveals "columnSplit" div, other elements
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

With an empty brew, if you start a code block like below, you'll see a `&nbsp;` and `columnSplit`.  I'm not sure if this is really the worst thing, but it's sort of unexpected and a little jarring.  And likely even more so for someone not familiar with how the columnSplits work, because your editor is otherwise empty.

~~~

kjlkfjsd

```

~~~

image

### Code

_No response_

### PR DESCRIPTION
**Pull request by dbolack-ab**
## Description

This tests if there are unclosed \`\`\` blocks in a column and inserts a closure set if missing.

I'm not sure if this is a fix you *really* need but it does fix the issue posted.

- Closes #3206 

To Test, open a blank document and insert a backtick based preformatted block without  closing it. In production you should see the portions of the auto-inserted column break enclosed in the block due to it being auto closed when the parser hits the `<column>`. In test, it should be an empty block.

Using the 5ePHB theme makes this much more obvious.

### Reviewer Checklist

  - [ ] Correctly inserts backticks when the block is not closed and no `\column` is used after.
  - [ ] Correctly inserts backticks when the block is not closed and `\column` is used after.
  - [ ] Properly formatted blocks are not munged
  - [ ] Text with properly formatted blocks does not have extra backtick sets

### CODE REVIEW

**Comment by 5e-Cleric** [comment]
So, with 1 more backtick the issue persists...

**Comment by dbolack-ab** [comment]
> So, with 1 more backtick the issue persists...

So it fails with?

```
\`\`\`\`
```

?

calculuschild:
I think this would be much more straightforward to just take this line and, instead of injecting `\colulmn` into the markdown source, just  append the appropriate `\n<div class='columnSplit'></div>\n` to the output HTML.

https://github.com/naturalcrit/homebrewery/blob/69c633dabebd517fc0d44398142510b23b86d8fb/client/homebrew/brewRenderer/brewRenderer.jsx#L199

dbolack-ab:
> I think this would be much more straightforward to just take this line and, instead of injecting `\colulmn` into the markdown source, just append the appropriate `\n<div class='columnSplit'></div>\n` to the output HTML. Then there are no rogue elements in the markdown to begin with. I.e. clean up the input instead of trying to catch an error in the output. Then this becomes a 2-line PR.
> 
> https://github.com/naturalcrit/homebrewery/blob/69c633dabebd517fc0d44398142510b23b86d8fb/client/homebrew/brewRenderer/brewRenderer.jsx#L199

This seems to work. I tested for the presence of the string before adding it as I was unsure if adding a second to the end could cause an issue with the layout.

PR #4006, opened 2025-01-24
Workaround for react-frame-component v5

G-Ambatte:
This PR adds a Suspense boundary to the Frame component from `react-frame-component`. Local experimentation indicates that this may resolve the race condition that has been causing issues with rendering.


G-Ambatte:
Current iteration contains minimal code changes; some functions may now be obsolete.

G-Ambatte:
Experimentation now shows that this approach doesn't work; closing.

PR #4023, opened 2025-02-01
Small adjustments to the PHB theme snippets

5e-Cleric:
This introduces a few changes to make the PHB 2014 theme better at dealing with user's customization.

Specifically:

- Allows the backcover to actually fit in a4 page sizes (see: https://www.reddit.com/r/homebrewery/comments/1i56us9/adjusting_the_size_of_the_template_backcover/, https://www.reddit.com/r/homebrewery/comments/16r66m4/back_cover_page_with_a4_page_format/, https://www.reddit.com/r/homebrewery/comments/1dpsdrx/print_a4_settings_backcover/)
- Adapts the code for custom logos to work better (see: https://www.reddit.com/r/homebrewery/comments/1ege859/resizing_the_logo_dimensions/ , https://www.reddit.com/r/homebrewery/comments/157iw23/modify_logo_from_back_cover/)


calculuschild:
Do any of these changes risk messing up existing documents that already use these snippets? Such as the backcover on Letter size pages, etc.?

5e-Cleric:
> Do any of these changes risk messing up existing documents that already use these snippets? Such as the backcover on Letter size pages, etc.?

Can confirm this does not break existing documents, on letter size, just deletes two lines of code that were unnecessary for the letter size, and actually broke the document for other sizes.

dbolack-ab (approved):

calculuschild:
Ok, I have now tested this and I agree there is no effect on Letter size.

The one place where there is a difference is on the `.backCover` `width` property for A4 (or taller) page sizes. The back-cover image shifts to the right slightly as it expands to fit the vertical space on the page. This may not be a critical thing, but wanted to point it out in case this is a concern. @5e-Cleric thoughts?


Before:
image


After:
image

5e-Cleric:
The effect is small, but if you want i can probably prevent this.

5e-Cleric:
Actuallly, after messing around with this, no, users might change the width of the page to lots of things, the PHB is a letter size template, we may move the image wider to make sure it is not clipped, but i think we can leave the text width to the user. It is a pain to solve for all use cases.

calculuschild:
Ok. If you are ok with it as is we can merge this.

calculuschild (approved):

PR #4026, opened 2025-02-03
fix undefined value in errorIndex.js if brew doesn't exist

5e-Cleric:
With the recent brews issue existing, opening one of those links takes the user to an app error heroku page, instead of an error page, even though we will fix the recent brews error, this one can be easily avoided by preventing the split function to break the site by giving it a default string, this is a small change and it may help us in the future.

may test with `/edit/6WbHRBwXdC-v`


5e-Cleric:
This is a problem because the new function that escapes the title of the brew to make sure we don't render it as markdown is run for deleted brews too, which don't have a title.

5e-Cleric:
Can confirm the fix works, an error page is shown correctly.

calculuschild:
Looks good to me. I'll send this live right now.

PR #4034, opened 2025-02-07
Experimental event stream emitter

G-Ambatte:
This PR adds an event-stream route of `/stream`.
- One second after the server starts, a message of type `initStream` is emitted on this stream.

This PR also modifies the routes defined in `homebrew.jsx` to add a listener, subscribed to `/stream`.
- ~On detecting a message with the type `initStream`, a `window.location.reload()` is automatically performed.~ REMOVED
- When an edited brew is saved, an update message is sent to the stream. The Share Page listens for this and writes a message to the console when it occurs.


G-Ambatte:
To test this PR:
- build and start the server (e.g. `localhost:8000`)
- open at least one Homebrewery page (e.g. `localhost:8000/new`)
- leaving all pages open, stop then restart the server

Expected results:
- approximately one second after server restarts. all open pages will automatically reload

G-Ambatte:
> and then one second after that the author will be wanting to know what happened to the changes they made that were not saved.

This PR does not remove any listeners (like `onbeforeunload`), so the current iteration will still fire those functions before completing the page reload.
image

G-Ambatte:
> What is the intended use case for this? Is this more of a development tool (like liveReload), or something else?

Correct, that's the use case for this implementation. However, the underlying `/stream` endpoint allows for clients to respond to server notifications, of which "the server restarted so reload the page" is just the first example I thought of when we were discussing `liveReload` a few days ago.

Gazook89:
With Vite now, is this still needed?

5e-Cleric:
This is a whole different thing to livereload, this, if i understand correctly, would allow two things:

- Update sharepage when brew is saved
- Synchronize two brews being edited

am i wrong?

@G-Ambatte i assume this needs to be redone from scratch, but if we can get this up to date, i'd like this to be merged so we can work on the two points. How's it look?

G-Ambatte:
> i assume this needs to be redone from scratch

@5e-Cleric : Not rebuilt from scratch, just updated. I've removed the "live reload" type functionality now, but left a `console.log()` in the Share Page as an example of how to listen for a streamed message.

5e-Cleric:
Can confirm this works, and is awesome.

server/app.js
```diff
@@ -1,4 +1,4 @@
-/*eslint max-lines: ["warn", {"max": 500, "skipBlankLines": true, "skipComments": true}]*/
```
5e-Cleric, line 1: can be undone, file is now smaller

client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -39,6 +39,18 @@ const SharePage = (props)=>{
...
+		eventSource.addEventListener('message', (evt)=>{
+			const messageData = JSON.parse(evt.data);
+
+			if(messageData.eventType == 'brewUpdated'){
+				if(messageData.shareId == brew.shareId && messageData.version != brew.version) {
+					console.log(`brew has been updated, viewing ${brew.version}, new version is ${messageData.version}`);
+				}
+			}
+		});
+
```
5e-Cleric, line 53: Adding a comment
```suggestion

        // listen for changes in the brew version
		const eventSource = new EventSource('/stream');
		eventSource.addEventListener('message', (evt)=>{
			const messageData = JSON.parse(evt.data);

			if(messageData.eventType == 'brewUpdated'){
				if(messageData.shareId == brew.shareId && messageData.version != brew.version) {
					console.log(`brew has been updated, viewing ${brew.version}, new version is ${messageData.version}`);
				}
			}
		});

```

5e-Cleric (approved):
Looks great, i say this can be merged, once calc has taken a look.

server/app.js
```diff
@@ -1,4 +1,4 @@
...
+/*eslint max-lines: ["warn", {"max": 750, "skipBlankLines": true, "skipComments": true}]*/
```
5e-Cleric:
```suggestion
/*eslint max-lines: ["warn", {"max": 400, "skipBlankLines": true, "skipComments": true}]*/
```

Issue #4035, opened 2025-02-08
Break markdown extensions in markdown.js out into their own npm modules


dbolack-ab:
### Your idea:

It is plausible not all of these can be moved out, but here's the list:

* [X] justifiedParagraphs - marked-justified-paragraph
* [x] Definitionlists - marked-definition-lists
    * [x] definitionListsMultiLine 
    * [x] definitionListsSingleLine
* [x] forcedParagraphBreaks  - marked-hard-breaks
* [x] nonbreakingSpaces - marked-nonbreaking-spaces
* [x] superSubScripts - marked-subsuper-text
* [x] Mustaches
    * [x] mustacheSpans 
    * [x] mustacheDivs 
    * [x] mustacheInjectInline
    * [x] mustacheInjectBlock

calculuschild:
Mustaches might be tricky but I agree these are probably the easiest candidates.

dbolack-ab:
> Mustaches might be tricky but I agree these are probably the easiest candidates.

In each of the sub-listed Items I was assuming a single module for each so those common componants can run true.

That said, it may be better to make a distinct "markdown+" module that is essentially markdown.js and anything that doesn't move cleanly can stay put.

calculuschild:
Mustaches I think we might want to bundle into one package. I don't see a use case where you would want the DIV version but not the SPAN version. Same with injection. I.e. in my mind they are the same extension.

My comment on it being tricky is that I think it's integrated into other parts of the site so it might be harder to pull out.

Eventually, yeah someone could make a Markdown+ package that would just be a wrapper around all the other packages.

calculuschild:
Mustaches I think we might want to bundle into one package. I don't see a use case where you would want the DIV version but not the SPAN version. Same with injection. I.e. in my mind they are the same extension.

My comment on it being tricky is that I think it's integrated into other parts of the site so it might be harder to pull out.

Eventually, someone could make a Markdown+ package that would just be a wrapper around all the other packages. Could maybe include an actual spec document.

dbolack-ab:
> Mustaches I think we might want to bundle into one package. I don't see a use case where you would want the DIV version but not the SPAN version. Same with injection. I.e. in my mind they are the same extension.
> 
> My comment on it being tricky is that I think it's integrated into other parts of the site so it might be harder to pull out.
> 
> Eventually, someone could make a Markdown+ package that would just be a wrapper around all the other packages. Could maybe include an actual spec document.

[...]

PR #4043, opened 2025-02-11
Move Superscript/Subscript functions into their own module

dbolack-ab:
## Description

Moves the `^superscript^` and `^^subscript^^` extension into its own NPM module and swaps it in.

https://www.npmjs.com/package/marked-subsuper-text

- Related to #4035 

### Reviewer Checklist

  - [ ] Superscripts still work
  - [ ] subscripts still work


calculuschild:
We should put all of these npm packages under a group organization so we can all access them.

I am a member of the "naturalcrit" organization but not the owner (belongs to Scott Tolksdorf). I can see if he will turn over ownership.

In the mean time I also created a "Homebrewery" organization. I can invite you as members so you can publish packages under it.

package.json
```diff
@@ -114,6 +114,7 @@
     "marked-extended-tables": "^1.1.0",
     "marked-gfm-heading-id": "^4.0.1",
     "marked-smartypants-lite": "^1.0.3",
+    "marked-subsuper-text": "^1.0.1",
```
calculuschild, line 117: Is this package live? I can't see it in NPM.

shared/naturalcrit/markdown.js
```diff
@@ -333,35 +334,6 @@ const mustacheInjectBlock = {
 	}
 };
 
-const superSubScripts = {
-	name  : 'superSubScript',
```
calculuschild, line 337: I see the package is named "subSuper" but here it is superSub. Let's make it consistent either way.

calculuschild:
We should also move our test cases over to the package. Minus any test cases that are checking the interaction between multiple extensions.

shared/naturalcrit/markdown.js
dbolack-ab: Whoops. That's an error. Should be subSuper.

package.json
dbolack-ab: Should be? I can see it and pull it.

https://www.npmjs.com/package/marked-subsuper-text ?

shared/naturalcrit/markdown.js
dbolack-ab: Where I perform the import is it named `MarkedSubSuperText`

dbolack-ab:
> We should also move our test cases over to the package. Minus any test cases that are checking the interaction between multiple extensions.

I think this particular extension squeaked through without tests. The module does have some, however.

shared/naturalcrit/markdown.js
dbolack-ab: I have renamed the block type in the module for consistency.

package.json
calculuschild: Weird. It doesn't show up when I search for it. But I can open it. Will review over there as well as part of this PR.

calculuschild:
I added some issues to your package.

dbolack-ab:
> I added some issues to your package. Once those are handled I think this can merge.

I've updated for those. Let me know when you think it's good and I'll increment versions.

calculuschild:
@dbolack-ab Added a comment to the Readme.md issue. Needs one more tweak.

calculuschild:
Ok, looks good now. Do you need to re-publish the package and update the version number here?

calculuschild:
Ah I see you already did. Merging now!

PR #4047, opened 2025-02-12
Marked hard breaks

dbolack-ab:
## Description

As above.  https://github.com/dbolack-ab/marked-hard-breaks

#4035 

_Replace this line with instructions on how to test or view your changes, as well as any before/after
screenshots or recordings for UI changes._

### Reviewer Checklist

  - [ ] Verify Line Breaks work as before
  - [ ] Feature B does Y


.circleci/config.yml
```diff
@@ -67,15 +67,9 @@ jobs:
       - run:
           name: Test - Definition Lists
           command: npm run test:definition-lists
-      - run:
-          name: Test - Hard Breaks
-          command: npm run test:hard-breaks
       - run:
           name: Test - Non-Breaking Spaces
           command: npm run test:non-breaking-spaces
-      - run:
```
calculuschild, line 76: This is removing the tests for paragraph justification. Let's keep those here just to make sure we are still passing all tests before merging.

Gazook89:
Wouldn't we run into an issue where we are impacting existing brews, if we apply a style to the `<br>` element?  I would assume that is why we made our own class-- that it is less likely it would impact existing brews.

calculuschild:
> Wouldn't we run into an issue where we are impacting existing brews, if we apply a style to the `<br>` element? I would assume that is why we made our own class-- that it is less likely it would impact existing brews.

This approach wouldn't apply any style to `<br>`. The `.columnWrapper` already has the spacing set to get the same line height. The only change would be the HTML structure coming from the Markdown.

.circleci/config.yml
dbolack-ab: I had a stray commit here, I'll put it back - the justifications were done first and expected to get merged first so I didn't bother when it happened. :)

dbolack-ab:
> Before merging this, I do want to experiment with having this extension just output `<br>` instead of the `<div class="blank">` thing, as it has been bothering me for a while that we rely on a custom CSS class to get a simple line break to work. There was some reason we avoided it before (I think controlling the height of the blank lines?), but there may be ways around that we didn't consider. I want to try to resolve this before releasing it to the public.

This is easy enough done. Try now. I did NOT add new css code so ya'll could figure out what HB should do.

dbolack-ab:
I'm going to close this until the hardbreak change is decided just to not deal with the headache of keeping this current. It will be silly easy to redo

dbolack-ab:
Reopening pending V4

dbolack-ab:
closing in favor of https://github.com/marked-modules-transitional/hbmarkedwrapper/pull/4

PR #4052, opened 2025-02-14
Fix bug in save logic

### LINKED ISSUE #4051: EditPage save bug
On opening an Edit Page, when a change is made, the Save button changes to SAVE NOW. Undoing that change results in the Save button reverting to "SAVED", as there are no unsaved changes.
However, after the brew has been saved once, making a change still causes the Save button to change to SAVE NOW but undoing that change no longer causes the button to revert.

On investigation, the `this.savedBrew` is being updated with the response from the `updateBrew` API call, which no longer matches the object structure of `this.state.brew` - for example, it has no `text` property at all - so the `hasChanges()` function will always return true.

G-Ambatte:
This PR resolves #4051.

This PR changes how EditPage updates the `this.savedBrew` object with the results of the `updateBrew` API call, which allows the `hasChanges()` function call to work correctly again.

This PR also adds a default results for the Save button, as fixing the `hasChanges()` function has exposed a situation where there is no result for the Save button text.

### DIFF
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -256,13 +256,8 @@ const EditPage = createClass({
 		};
 		history.replaceState(null, null, `/edit/${this.savedBrew.editId}`);
 
-		this.setState((prevState)=>({
-			brew : { ...prevState.brew,
-				googleId : this.savedBrew.googleId ? this.savedBrew.googleId : null,
-				editId 	 : this.savedBrew.editId,
-				shareId  : this.savedBrew.shareId,
-				version  : this.savedBrew.version
-			},
+		this.setState(()=>({
+			brew        : this.savedBrew,
 			isPending   : false,
 			isSaving    : false,
 			unsavedTime : new Date()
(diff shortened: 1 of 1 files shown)


client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -247,16 +247,17 @@ const EditPage = createClass({
 			});
 		if(!res) return;
 
-		this.savedBrew = res.body;
+		this.savedBrew = {
```
calculuschild, line 250: The old version uses the whole brew (or stub) returned from the server. This only uses the googleId, editId, shareId, and version. Are these the only values that can change as a result of the `/update` API call? Just want to double-check we don't risk losing some brew metadata when saving here.

calculuschild: Clarified over Gitter: old code was resulting in `this.savedBrew` and `this.state.brew` not being updated with the same contents after a save. This PR guarantees both are identical after a save.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -336,6 +337,7 @@ const EditPage = createClass({
 		if(!this.state.isPending && !this.state.isSaving){
 			return <Nav.item className='save saved'>saved.</Nav.item>;
 		}
+		return <Nav.item className='save saved'>no changes.</Nav.item>;
```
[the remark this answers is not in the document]
G-Ambatte: The latest commit should have clarified the logic for the Save button.

calculuschild:
Looks good. Thanks @G-Ambatte ! 🎉 ⭐ 💯

PR #4069, opened 2025-03-02
Experimental brew load function

G-Ambatte:
This experimental PR demonstrates a process for loading brew data from `homebrew.jsx` rather than from `app.js`. This prevents user data (like `</script>`) from appearing in the emitted HTML source (and potentially crashing the web app).

As an example, the following is a brew that contains the title `TITLE`, text content of `TEXT`, and style content of `STYLE` - none of which appear in the `start_app` function call:

```
<script>start_app({"version":"3.17.0","url":"/share2/YcHkHfN17mRI","account":{"username":"sor_robertson","issued":"2025-03-01T05:27:16.385Z"},"enable_v3":true,"enable_themes":true,"config":{"local":true,"publicUrl":"https://homebrewery.naturalcrit.com","baseUrl":"http://localhost:8000","environment":"local","deployment":""},"data":{"id":"YcHkHfN17mRI","type":"share"}})</script>
```

### DIFF
--- a/client/homebrew/homebrew.jsx
+++ b/client/homebrew/homebrew.jsx
@@ -17,6 +17,10 @@ const ErrorPage = require('./pages/errorPage/errorPage.jsx');
 const VaultPage = require('./pages/vaultPage/vaultPage.jsx');
 const AccountPage = require('./pages/accountPage/accountPage.jsx');
 
+const request = require('superagent');
+const { splitTextStyleAndMetadata } = require('../../shared/helpers.js');
+
+
 const WithRoute = (props)=>{
 	const params = useParams();
 	const [searchParams] = useSearchParams();
@@ -63,7 +67,22 @@ const Homebrew = createClass({
 		global.enable_themes = this.props.enable_themes;
 		global.config = this.props.config;
 
-		return {};
+		return {
+			brew : this.props.brew
+		};
+	},
+
+	componentDidMount : async function() {
+		if(this.props.url.startsWith('/share2/')){
+			request.get(`/get/${this.props.data.id}`)
+				.then((data)=>{
+					const brew = data.body;
+					splitTextStyleAndMetadata(brew);
+					this.setState({
+						brew: brew
+					})
+				});
+		}
 	},
 
 	render : function (){
@@ -73,6 +92,7 @@ const Homebrew = createClass({
 					<Routes>
 						<Route path='/edit/:id' element={<WithRoute el={EditPage} brew={this.props.brew} userThemes={this.props.userThemes}/>} />
 						<Route path='/share/:id' element={<WithRoute el={SharePage} brew={this.props.brew} />} />
+						<Route path='/share2/:id' element={<WithRoute el={SharePage} brew={this.state.brew} />} />
 						<Route path='/new/:id' element={<WithRoute el={NewPage} brew={this.props.brew} userThemes={this.props.userThemes}/>} />
 						<Route path='/new' element={<WithRoute el={NewPage} userThemes={this.props.userThemes}/> } />
 						<Route path='/user/:username' element={<WithRoute el={UserPage} brews={this.props.brews} />} />
--- a/client/homebrew/pages/sharePage/sharePage.jsx
+++ b/client/homebrew/pages/sharePage/sharePage.jsx
@@ -49,7 +49,7 @@ const SharePage = (props)=>{
 		return ()=>{
 			document.removeEventListener('keydown', handleControlKeys);
 		};
-	}, []);
+	}, [brew]);
 
 	const processShareId = ()=>{
 		return brew.googleId && !brew.stubbed ? brew.googleId + brew.shareId : brew.shareId;
--- a/server/app.js
+++ b/server/app.js
@@ -432,6 +432,19 @@ app.get('/new', asyncHandler(async(req, res, next)=>{
 	return next();
 }));
 
+// Get brew data
+app.get('/get/:id', asyncHandler(getBrew('share')), (req, res)=>{
+	res.status(200).json(req.brew);
+});
+
+app.get('/share2/:id', (req, res, next)=>{
+	req.data = {
+		id   : req.params.id,
+		type : 'share'
+	};
+	return next();
+});
+
 //Share Page
 app.get('/share/:id', asyncHandler(getBrew('share')), asyncHandler(async (req, res, next)=>{
 	const { brew } = req;
@@ -567,7 +580,8 @@ const renderPage = async (req, res)=>{
 		enable_themes : config.get('enable_themes'),
 		config        : configuration,
 		ogMeta        : req.ogMeta,
-		userThemes    : req.userThemes
+		userThemes    : req.userThemes,
+		data          : req.data || undefined
 	};
 	const title = req.brew ? req.brew.title : '';
 	const page = await templateFn('homebrew', title, props)


ericscheid:
Also, it reduces the initial page load to just the app code, and consequently very large brews would at least present the HB UI framing earlier.

Con: it would increase the number of requests hitting the server. Not quite double, but certainly more (app+brew+css, instead of appbrew+css). Once this is working well though we could get the app.js to be cached (since it only really changes when we push code updates to prod).

calculuschild:
I agree with the idea in general. It makes sense to separate the brew contents from the appstart contents.

Given that we already have API paths for retrieving brews and things in homebrewapi.js, we should try to centralize any new brew fetching paths in there instead of in app.js.

G-Ambatte:
I've shifted the API get route to HomebrewAPI; to make it consistent with the other API routes I have added the `checkClientVersion` middleware, which then in turn required the API call to use the `request-middleware.js` function.

calculuschild:
Thinking over this some more, I have two thoughts:

1) Should the fetching logic be put in `editPage.jsx` and `sharePage.jsx` directly? `homebrewery.jsx` routes us to the correct page... seems we can let those pages handle their own initial brew fetch logic upon component mount? At that point they are all client-side so it would help keep things more self-contained.

2) We need to consider how this change will impact error handling. Current behavior will display an error popup if we are already in a page, or reroute to an error page if the error happens before the page loads (within `app.js`). Moving initial brew fetching to client side means we don't know if (for example) the user has authorship access until the page is already loaded, so we would need to add rerouting logic in the appropriate places.

G-Ambatte:
>     1. Should the fetching logic be put in `editPage.jsx` and `sharePage.jsx` directly? `homebrewery.jsx` routes us to the correct page... seems we can let those pages handle their own initial brew fetch logic upon component mount? At that point they are all client-side so it would help keep things more self-contained.

I was initially hoping that it would be as simple as putting a "brew fetch" function in the common-to-everything React Router component, and then manipulating the data in each page as needed for that page's purpose.
However I no longer think that this approach will be viable, so my intention is to switch the fetch logic to the individual page.
 
>     2. We need to consider how this change will impact error handling. Current behavior will display an error popup if we are already in a page, or reroute to an error page if the error happens before the page loads (within `app.js`). Moving initial brew fetching to client side means we don't know if (for example) the user has authorship access until the page is already loaded, so we would need to add rerouting logic in the appropriate places.

We'll need to catch any error and throw to the error page; even SharePage should throw if the brew is locked.

ericscheid:
A side thought on routing: we don't necessarily need new/additional routes beyond `/share/:id` or `/edit/:id` if we use content negotiation. That is, a request to `/edit/:id` for a `text/html` response should return the app.js UI frame, which then issues a request for the same URI but this time for `text/json`.

G-Ambatte:
The latest commits add the ErrorPage component to the SharePage. If an error occurs while fetching the brew data, the error information is presented in the ErrorPage component instead of the normal SharePage content. As best I have been able to determine, this is identical to the current ErrorPage presented for the same error.

G-Ambatte:
Weird issue I haven't been able to hammer out yet - the first page is always rendered using the Legacy renderer, even if the document is set to V3.

G-Ambatte:
I think that this is in a pretty good place right now for the Share Page, so it's worth getting more eyes on it before moving on to look at any further pages.

G-Ambatte:
## Current state of the PR

The Share Page functions as normal in a way that is not immediately noticeably different to the previous operation. However, in the event that the brew content contains something that would crash the Homebrewery app if passed as a parameter to the `start_app` call, this does NOT cause the Share Page to fail.

### Loading a brew with active content

EditPage (using previous data loading system):
image

SharePage (using new data loading system):
image

SharePage HTML:
image
The earlier work to create `safeHTML.js` prevents any of the active content from reaching the BrewRenderer and the HTML output.


### Functionality

- Share Page loads the brew and renders it correctly, including theme:
image

- If there is an error accessing the document, normal errors are thrown and logged by the server, and an Error Page is presented:
image
image
image

G-Ambatte:
Found a couple of missing features

To do:
- [ ] increment view count when Share Page is accessed by non-author
- [ ] remove Edit link when Share Page is accessed by non-author
- [ ] reinstate ogMeta tag data

server/homebrew.api.js
```diff
@@ -478,14 +493,42 @@ const api = {
...
+	returnBrew : async (req, res)=>{
+		return res.status(200).json(req.brew);
+	},
+	shareAuthorChecks : async (req, res, next)=>{
+		// If logged in user is not an author:
+		if(!req.brew.authors.includes(req?.account?.username)){
+			// Remove Edit ID
+			req.brew.editId = undefined;
+			// Increase view count and lastViewed property
+			await HomebrewModel.increaseView({ shareId: req.params.id });
```
calculuschild, line 506: Just thinking out loud:

This no longer handles the case to increase view count for unstubbed Google-only brews. I suppose this makes sense; view count/lastViewed only really matters in the context of the HB environment itself, and is not required to "rebuild" the document if HB goes down.

If `GoogleActions.increaseView()` is no longer being used anywhere, let's remove it or put a note somewhere to remove it later.

server/homebrew.api.js
```diff
@@ -478,14 +493,42 @@ const api = {
...
 		res.status(204).send();
+	},
+	returnBrew : async (req, res)=>{
+		return res.status(200).json(req.brew);
+	},
+	shareAuthorChecks : async (req, res, next)=>{
+		// If logged in user is not an author:
+		if(!req.brew.authors.includes(req?.account?.username)){
+			// Remove Edit ID
+			req.brew.editId = undefined;
```
calculuschild, line 504: This appears to be performing part of the `sanatizeBrew()` function (removing the `editId`), but not the other parts (removing `_id` and `__v`). Do we need to handle that here by bringing over `sanatizeBrew()`, or is it already covered somewhere in the fetching process?

server/homebrew.api.js
```diff
@@ -168,6 +169,20 @@ const api = {
 		return res.status(200).send(brew.style);
 	},
 
+	getMeta : async (req, res, next)=>{
```
calculuschild, line 172: I see what `getMeta` is doing, but I am concerned this basically means we have to request the brew twice when opening a page (once for the metadata and again inside the `sharePage.jsx`).

Not sure if there is a way to apply the metadata at the same time of the initial brew fetch from the `sharePage`, which would at least save a Mongo call.

But at a minimum, I think we can use `await api.getBrew('share', true)` so we only fetch the stub.

G-Ambatte: I've kept this separate at this stage because I'm not sure whether the metadata information needs to be returned in the initial response in order for the meta tags to be picked up correctly for social media previews.

I think that you are likely correct about `getBrew('share', true)`.

server/homebrew.api.js
G-Ambatte: I've updated the function in the latest commit to also remove the `_id` and `__v` properties.

server/homebrew.api.js
G-Ambatte: Good point - I thought that this had already been changed, but on checking now I see that `GoogleActions.increaseView` is called in the current version of the SharePage. However, is the view count something that should be stored outside of the brew stub, now that they exist?
I'm tempted to just replicate the functionality and leave the philosophical questions for another day.

5e-Cleric:
Let's block this until the unification of edit and new page is completed:  #1800

G-Ambatte:
While testing this today using this 584 page brew: https://homebrewery-pr-4069.herokuapp.com/share/R9uVUjwBDxeV

- Chrome v143 appears to render correctly
- Firefox v147 appears to crash and never renders the brew

Further investigation will be required to determine the cause.

Issue #4079, opened 2025-03-08
Vault search does not render pagination correctly when page size is changed


5e-Cleric:
This is a mistake i probably introduced way back, some searches trigger a recount of pages, some don't.

Will try to find the issue and fix it shortly.

PR #4080, opened 2025-03-08
Vault fixes

5e-Cleric:
For some time, I've wanted to fix some things i wasn't perfectly loving about the page, so this is the time to fix them:

- Fix bug with pagination not updating (Fixes issue #4079)
- Pagination now will not appear when less than 10 total brews are found
- Fixed issue with small text having an extra space character
- Fix styling for narrow screens and narrow panel
- display pagination on top part of the panel if the page size is over 10 brews
- Now, the _searching_ text will only appear if no brews are already displaying, otherwise, fast searches showed it for a few milliseconds, looking like a weird glitch
- lint all related files


calculuschild:
Ok, adding this to v3.18.0 along with the other fixes.

PR #4088, opened 2025-03-11
Add pageNumber brew variable

G-Ambatte:
This PR adds a `pageNumber` object to the brew, which is automatically set to the current page number. It can be overridden within the page, but the variable is reset at the start of the next page.


calculuschild (approved):

calculuschild (commented):
This approach seems good to me.

Just a couple thoughts:

1) Do we want to distinguish built-in variables with some kind of preface? 'HBPageNumber', etc? To avoid possible conflict with user variables later.

2) Can we add test cases for the new math functions?

G-Ambatte:
> * Do we want to distinguish built-in variables with some kind of preface? 'HBPageNumber', etc? To avoid possible conflict with user variables later.

I think so, if only to reduce the likelihood of collision with existing user variables. Whatever format we come up with will likely continue with any other system variables we introduce later, so I think that it merits some careful consideration.

I believe we should have a prefix that identifies it as an automatic variable; here's a few suggestions off the top of my head: `HB`/`homebrewery`, `sys`/`system`, `doc`/`document`, or `brew`.
There should then be a separator, such as `.`, `:`, `>`, or `-`.
Finally with the actual variable name.

So `pageNumber` (as in the current version of the PR) would become something like one of the following:

- `HB-pageNumber`
- `document>pageNumber`
- `sys.pageNumber`
- `brew:pageNumber`

Just writing those out examples once, I already find myself leaning towards the abbreviated prefixes, and away from separators that might be confusing, such as Math operators (`>` and `-`) and JavaScript object property identifiers (`.`).

I would welcome some additional opinions on the topic.

> * Can we add test cases for the new math functions?

I'll add tests shortly.

G-Ambatte:
The latest commit adds a snippet to insert the new variable-based auto page number.

shared/naturalcrit/markdown.js
```diff
@@ -902,7 +946,13 @@ let globalPageNumber = 0;
 const Markdown = {
 	marked : Marked,
 	render : (rawBrewText, pageNumber=0)=>{
-		globalVarsList[pageNumber] = {};					//Reset global links for current page, to ensure values are parsed in order
+		const lastPageNumber = pageNumber > 0 ? globalVarsList[pageNumber - 1].HB_PageNumber.content : 0;
+		globalVarsList[pageNumber] = {							//Reset global links for current page, to ensure values are parsed in order
+			'HB_PageNumber' : {									//Add document variables for this page
+				content  : !isNaN(Number(lastPageNumber)) ? Number(lastPageNumber) + 1 : lastPageNumber,
```
dbolack-ab, line 952: Are we doing this this way to allow the user to reset the value  arbitrarily?

G-Ambatte: Correct, this is to allow the user to set the page number, and have that affect future pages as well. If they set it to something other than a number, then that is also carried forward, but without being changed.

calculuschild:
I am ok with `HB_pageNumber`. All in favor?

G-Ambatte:
> I am ok with `HB_pageNumber`. All in favor?

I've changed the name now, from `HB_PageNumber` to `HB_pageNumber`. I have also added additional tests for the page variable, to ensure that it increments properly across page boundaries, including when it has been set to a new value (including a value that is not a number).

PR #4099, opened 2025-03-17
Fix minor v3.18.0 issues

G-Ambatte:
This PR seeks to revert the changes to hard breaks from PR #4058, in order to eliminate the issues as reported in [this Reddit post](https://redd.it/1jcor5e).

This PR also adds a `<div class='columnSplit'></div>` to the end of every BrewPage, which will restore the simulated `column-fill: auto` functionality to older versions of Chrome.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -39,7 +39,7 @@ const BrewPage = (props)=>{
 		...props
 	};
 	const pageRef = useRef(null);
-	const cleanText = safeHTML(props.contents);
+	const cleanText = safeHTML(`${props.contents}\n<div class="columnSplit"></div>\n`);
 
 	useEffect(()=>{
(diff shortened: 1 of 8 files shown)


dbolack-ab (commented):
I thought this was moved elsewhere, not removed. @calculuschild ?

5e-Cleric (approved):

5e-Cleric:
> I thought this was moved elsewhere, not removed. @calculuschild ?

Should i wait for calc to answer this before merging?

calculuschild:
> I thought this was moved elsewhere, not removed. @calculuschild ?

The extra `\column` was fully removed because as I was moving it feom the input Markdown to the output HTML, I realized it wasn't actually needed. It was there originally to force Chrome to imitate `column-fill: auto` which wasn't working properly at the time. It has since been fixed in Chrome.

However, this is apparently one of those cases where people stuck on Windows 7 and unable to update Chrome still need that fix in place.

PR #4100, opened 2025-03-18
Up to v3.18.1

5e-Cleric:
v3.18.1 does not bring any new feature, but it does fix a few standing issues since last version that were breaking:

colons are back to being rendered as `div class="blank"` instead of `br` elements.

Back to inserting a colsplit at the end of the page for automatic column break in older versions.

Fix CORS issue in local network.


5e-Cleric:
Looks alright to me, creating changelog and pushing fix

PR #4104, opened 2025-03-19
CodeEditor overflow issue fix

### LINKED ISSUE #4098: Top of snippetBar is hidden when using Ctrl+F in Codemirror section
### Renderer

v3.18.0

### Browser

Chrome Version 134.0.6998.89 (Official Build) (arm64))

### Operating System

MacOS (14.2.1 (23C71))

### What happened?

If you ctrl+f in the code or style part of the editor the top nav of the snippetBar nav gets hidden and can't be clicked (which makes moving between code and style impossible).

NOTE: This doesn't happen when the snippetBar is one row, only when it's over two rows (due to screen res / position of divider).

Image
Image

This can be rectified by refreshing the page.

5e-Cleric:
Should fix #4098 

Still not quite sure what the issue is, for some reason codeMirror needs a hardcoded height when flex is around it, this deletes our old update height function and handles the resizing with the same container query that handles the snippetbar breakpoint.

Basically, when the snippetbar is wrapped, the editor is set a height of 100% - snippetbar height.

We might be able to fix this programatically by passing the snippetbar height as a css var.

does fix the issue and reduce the lines of the codebase.


calculuschild:
I can confirm this fixes the bug, and I don't see anything else breaking related to editor height adjustment.

Going to merge...

PR #4106, opened 2025-03-19
Store toolbar state in local storage

### LINKED ISSUE #3835: Store toolbar state in localStorage
Both visibility and zoom state should be saved in localstorage with the id of the brew, to guarrantee a better UX, if a user does not want to have the toolbar there all the time, it should stay hidden between reloads, and same goes for the zoom level.

5e-Cleric:
Fixes #3835 

Stores a variable array in localstorage called `hb_toolbarState` and `hb_toolbarVisibility` which store each part separately, since they are used in different components.

Those states are loaded on mount using useEffect.

As for right now, all works except the page shadows, for some reason.

Might alter how the shadows work to fix this.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -117,6 +117,13 @@ const BrewRenderer = (props)=>{
 		pageShadows  : true
 	});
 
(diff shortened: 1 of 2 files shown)


5e-Cleric:
Closing in favor of #4156 since all these changes have been also applied there.

PR #4107, opened 2025-03-19
"Workaround" for bad \page mustache blocks

### LINKED ISSUE #4105: Double colon breaking page curly inject
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

Applying `{::after}` to `\page` in a document with content results in the webpage breaking and displaying a white/blank screen.

Typing `\page{::` into a blank document will get the same result.

### Code

```gfm

```

dbolack-ab:
## Description

This adds a simple test on the lexar results when run for `\page` inclusion. 

This does not *perfectly* solve the underlying issue here but it does catch any  and all situations where the mustache block was effectively invalid and so there were no `.injectedTags` results on the resulting token.

- Closes #4105

### Reviewer Checklist

  - [ ] Test with a page using a pseudo target such as `::after` in the mustache.


client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -186,12 +186,14 @@ const BrewRenderer = (props)=>{
 		} else {
 			if(pageText.startsWith('\\page')) {
 				const firstLineTokens  = Markdown.marked.lexer(pageText.split('\n', 1)[0])[0].tokens;
-				const injectedTags = firstLineTokens.find((obj)=>obj.injectedTags !== undefined)?.injectedTags;
```
calculuschild, line 189: What happens if you just use conditional chaining here:

```
injectedtags = firstLineTokens?.find((...

or

(obj)=>obj?.injectedTags ...
```

I believe one of the above should be basically the same result as this PR, but not require nesting `if` statements.

dbolack-ab: It is, I just prefer blocks as more obvious  - even if less elegant. :)

Switched it out for your solution.

calculuschild:
Ready to merge. Thanks @dbolack-ab!  ⭐ 💯 🎉

PR #4113, opened 2025-03-21
Fix CORS regex test for local installs

G-Ambatte:
A recent change to the CORS configuration accidentally broke local API calls. This PR changes how the regex test is conducted to correct this misconfiguration.


5e-Cleric (approved):

5e-Cleric:
Just tested in local and this does resolve the issue we've been facing, merging.

PR #4120, opened 2025-03-25
Fix `:::` BR option 1

calculuschild:
## Description

Tweak to the way lists are handled, to address the rendering differences when switching the `::::` to `<br>`. The source of the issue was that in normal text, `:` is always placed between `<p>` elements. But in lists, there is no block separating the list text and the `:::` spacing. An extra `<br>` is required to create a blank line when sitting in normal text, as opposed to a single `<br>` when between two block elements.

```
<p>text</p>
<br> // blank line
<p>text</p>
```

vs 

```
Text
<br><br> // one more br needed
Text
```

___

This PR tries to solve this by simply forcing *all* lists to wrap their text in `<p>` tags.

This does slightly affect spacing for lists that contain both plain text and other block elements without spaces between, though I am not certain this is a common thing. e.g.:

```
- text
  | table immediately after |
  |:-----------------------:|
  | cell                    |
```

The other impact could be for custom styles, as this moves he line spacing of lists from the `ul` element into the `li p` element.

Maybe this is too risky, in which case we just continue waiting for V4 before making this change.

___

### Note
Alternative, perhaps less disruptive approach in #4121

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -381,7 +381,12 @@ const forcedParagraphBreaks = {
 	tokenizer(src, tokens) {
 		const regex  = /^(:+)(?:\n|$)/ym;
 		const match = regex.exec(src);
+
 		if(match?.length) {
+			const lastToken = tokens[tokens.length - 1];
+			if(lastToken?.type == 'text')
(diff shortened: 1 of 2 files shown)


calculuschild:
Closing in favor of waiting for V4

PR #4134, opened 2025-04-02
Handle page overflow via soft page breaks

### LINKED ISSUE #3737: Auto Pagination / Page Breaks
### Your idea:

It would be nice to not need manual page breaks-- the document should automatically determine where page breaks are needed and create a new page.   Manual page breaks would still be allowed.

An example of a web document editor that manages to do this is Google Docs.

Likely, some method of calculating the height of elements and determining when that total height is greater than the page/container height, and then splitting the last element that overflows into a new page/container.  Calculating heights of whole elements is likely "easy", but a [library would likely be needed](https://www.npmjs.com/package/text-metrics) for calculating exactly where to split the text node of the last element.  

[Or, a library to handle it all (paged.js).](https://github.com/pagedjs/pagedjs)

There are likely a million edge cases where this will have poor results.  But I think it's worth a try at some point, if for no other reason than to have a good answer when someone else asks for it.

dbolack-ab:
## Description

This is a working proof of concept for soft page breaks. This handles rendered text overflow  by inserting an alternate tag for pagebreaks.

Blocked by: 
 https://github.com/calculuschild/marked-definition-lists/pull/139
 

### Reviewer Checklist

  - [ ] Add a page when there is enough text after to wrap into a third column. It should insert a `\softpage` tag
  - [ ] As above, but with a page with a {}. 


dbolack-ab:
I've found a bunch of failure modes I didn't see in early testing. Gonna let this sit then come back in a bit.

Gazook89:
So just right away I notice that if I go insert more paragraphs halfway through a document with a few pages, it doesn't really recalculate the soft breaks.  If a softbreak exists, and then i add some paragraphs before it, it can create a new softbreak but not remove the existing following softbreaks.

I would assume that basically every softbreak after the location of any change (aka cursor position) has to be wiped and recalculated?

Gazook89:
Sort of a half-baked idea, let me know what you think:

Can we have this text softbreak and flow between elements that share a class (or alternate html attribute)?  So we have a default content 'river' that is the basic `.page` columns, which is what you basically have now.  But if I want a second content river where text flows between different elements automatically....I could create a few divs, all with class `.sidebar` (or html attribute `data-river="sidebar"`), and then any of the content placed in the first div that overflow would just continue to the next div in the DOM with the same attribute?  

I don't know, probably not any actual purpose to this.  Not often you have two different flows of text running across pages.

dbolack-ab:
> I would assume that basically every softbreak after the location of any change (aka cursor position) has to be wiped and recalculated?

I had code originally that would recalc all lines between hard pages ( SOF, EOF, `\page` ) but it dropped out at one point for reasons I do not recall.

dbolack-ab:
> Sort of a half-baked idea, let me know what you think:
> 
> Can we have this text softbreak and flow between elements that share a class (or alternate html attribute)? So we have a default content 'river' that is the basic `.page` columns, which is what you basically have now. But if I want a second content river where text flows between different elements automatically....I could create a few divs, all with class `.sidebar` (or html attribute `data-river="sidebar"`), and then any of the content placed in the first div that overflow would just continue to the next div in the DOM with the same attribute?

I think it is plausible, though it'd be kinda fugly.

> 
> I don't know, probably not any actual purpose to this. Not often you have two different flows of text running across pages.

I dunno. I have GIANT sample doc I'm using to test this I stripped the `\page` from and it has a number of column tall divs that being able to split MIGHT be a good idea.. Dunno.

dbolack-ab:
The test work I am doing with adding line numbers will simplify this immensely

PR #4147, opened 2025-04-15
Add print-specific styling for facing and flow page layouts

### LINKED ISSUE #4146: Printing from Facing or Flow Layouts are not on page correctly
### Your idea:

When printing from the BrewRenderer when the layout is set to Facing or Flow, the Homebrewery page is not correctly scaled or positioned on the page.

Single Pages (correct):
Image

Facing Pages:
Image

Flow Pages:
Image

G-Ambatte:
This PR resolves #4146.

This PR adds print-specific styling for the facing and flow layouts, to correct the incorrect scaling and positioning as identified in #4146.

Single Page Layout:
image
image

Facing Layout:
image
image

Flow Layout:
image
image


5e-Cleric:
Uh... no? If you print using the facing and flow page layouts, i'd assume you want to print in two page sets.

5e-Cleric:
> > Uh... no? If you print using the facing and flow page layouts, i'd assume you want to print in two page sets.
> 
> You mean print multiple pages in the same sheet? I don't think so. PDF output should be the same, independent of the current view mode in the document.

Then i think we should tell users that to print they should pick the single page view, they can switch back after printing. OR, just switch that to single, print the page, and switch it back...

Gazook89:
Or just do what this PR does, right? No need to switch things back and forth if we just dictate styling for the print medium

G-Ambatte:
Printing multiple pages on a single sheet could be a future feature - but I would expect one for users to specifically and deliberately select. Otherwise we'll get issue reports of "why is my brew printing like this????"

Such a feature would be work for a future PR.

calculuschild:
Right. The viewer tools are for viewing convenience. They don't dictate PDF output.

If users want to print multiple pages on a sheet, that can be done through the print dialog.

G-Ambatte:
> If users want to print multiple pages on a sheet, that can already be done through the print dialog.

I've had mixed success trying to use this in the past, but ultimately that's not the issue I'm trying to fix with this PR.

client/homebrew/brewRenderer/brewRenderer.less
```diff
@@ -68,12 +68,16 @@
 @media print {
 	.toolBar { display : none; }
 	.brewRenderer {
-		height      : 100%;
-		padding-top : unset;
-		overflow-y  : unset;
+		height     : 100%;
+		padding    : unset;
+		overflow-y : unset;
+		&:has(.facing, .flow) {
```
calculuschild, line 74: Just one question. Does this `:has()` rule add anything different than the `padding : unset;` above?

calculuschild: Ah, I see it might be a specificity thing. If I remove the `:has` rule it doesn't work anymore because facing pages still have some other styling adding padding.

G-Ambatte: You've got it - it's to revert the more specific styling to the original single page values during printing.

PR #4151, opened 2025-04-20
Alternate Add d100/percentage support to dicefont icons

### LINKED ISSUE #3347: Add d100 support in diceFont

dbolack-ab:
## Description

This is added as a new icon but inlined with the df icons for user convenience.

Replaces #3463

Add values of 00 through 100 on the D10 Dice Icon using `df_dper_#` ( example df_dper_75 )
I used dper as the die signifier instead of dpercent for user brevity. Not hugely married to it.

Provides replacement d10 values along with adding a `0` option so that the two sets using the same d10 die outline match better.

- Closes #3347 

If you want or need to regenerate the font, the source images are in `d100_Sources/`. There is a readme that should cover the needs to do so.

[Test Document](https://homebrewery-pr-4151.herokuapp.com/share/QyiI4rfEEMtN)

image


calculuschild:
Since this uses the d10 as the base shape, could we just extend the existing `df_d10` label with the higher numbers?

Also if possible, Can the new icons be appended into the existing diceFont.woff2 file? Rather than a whole separate font file just for the d100s.

dbolack-ab:
> Since this uses the d10 as the base shape, I wonder if it would make more sense to just extend the existing `df_d10` label with the new numbers? For some reason `dper` feels like it would be unintuitive to find in a search considering all the other dice follow a certain naming convention.

Probably. Given that's an arbitrary label, I assumed there'd be a disagreement with any choice I made and went with the one I thought least likely to stick, so ya'all'd discuss. :) I considered d100, dpercent, dper. I thought d10 was a bad choice, but I suppose there's not really a situation where confusion could kick in.

> Also if possible, can the new icons be appended into the existing diceFont.woff2 file? Rather than a whole separate font file just for the d100s.

Probably if I spun up a 10? year old VM and was able to build the font tool dicefot used. Otherwise, I think I'd need to build the whole thing as a replacement face.  I could build up a new font from all of the source files, plus the new stuff.

> Last, I think there is still some alignment and bold/outline difference on the font. Comparing the d10_10 vs the dper_10, I would expect them to be identical:

They should be, but I see I still have tweaking to do.

calculuschild:
I think I could help merge the font files if you can just make the tweaks to align the new and old icons.

dbolack-ab:
Before I go trying to pixel push - does this reconstructed look okay? 
[dpercent-Kremlin-v1.0 (1).zip](https://github.com/user-attachments/files/19887438/dpercent-Kremlin-v1.0.1.zip)

If so, I will clean up the related code to make sure the glyphs match the emoticon tags and update the PR.

If not, I'll have to slice and dice the ... dice.

5e-Cleric:
> I think I could help merge the font files if you can just make the tweaks to align the new and old icons.

@calculuschild are you still planning on this?

calculuschild:
Yeah. Still planning on it.

PR #4152, opened 2025-04-20
Glossaries

### LINKED ISSUE #3370: Dynamic Glossary collation
### Your idea:

Dynamic Glossary collation follows the pattern set up in Rich Indexes and definition lists.

Arbitrarily placed markdown is placed in the brew text and then collated and ordered by a snippet which outputs any number of glossary lists.

Markdown Format:
The markdown takes one of two shapes, each as a single line entry.
1. `# Glossary Term :: Glossary Definition` - This creates a term/definition pair and places it on the default glossary list `Glossary`
2. `#GlossaryList Glossary Term :: Glossary  Definition` - This creates a term/definition pair and places it on the user-created  glossary list `GlossaryList`

Markdown details:
In both cases, the markdown is consumed on parsing.

Snippet Details:
The snippet runs its own parsing of the markdown from the brew and places it in the designated Glossary List. 
Duplicate terms will have their Definitions collected for shared output.  
Each list will be ordered by term. 
Output for each list will have the list name as an H2 header, followed by the list rendered as a PanDoc style <DL> list followed by a `\page`

dbolack-ab:
## Description

Implement dynamic collated Glossaries.

Usage:

`Term // Definition`
`#Glossary Name: Term // Definition`

If no Glossary Name is provided, a default of `Glossary` is used

Glossary is generated by snippet, inserted at cursor, with one Glossary per `\page` if multiple are present.

- Closes #3370

Reference document:

https://homebrewery.naturalcrit.com/share/d3CFfIOGARxm


5e-Cleric:
I am trying to review this PR, but to be honest, i have very little idea of what to review, i see the glossary snippet, which outputs three page breaks and the glossary.

In a normal book, which syntax should be used for what? Prepare the terms vs display the list?

dbolack-ab:
> I am trying to review this PR, but to be honest, i have very little idea of what to review, i see the glossary snippet, which outputs three page breaks and the glossary.

That's absolutely fair. I thought we covered this well enough in the discussion topic, but it bears restating if it's not clear.

The purpose of this PR is to implement a tool for assembling and ordering any number of Glossaries while placing the glossary entries at arbitrary locations  ( presumed to be where term comes up in the text ) collected and poured by automation. 

You could certainly do this by hand with little effort and the notion of multiple glossaries is somewhat uneeded, but as most of this code is duplicated from the index code, it was mostly free.

And remember, this isn't purely a definitions-of-words construct. It is for anything that would be well represented by a Definition list, such as a list of locations, NPCs, or medicinal herbs.

> 
> In a normal book, which syntax should be used for what? Prepare the terms vs display the list?

If you have a single glossary, use the first syntax. If you have multiple, use the second where "Glossary Name" is the glossary to add the term to.

5e-Cleric:
> That's absolutely fair. I thought we covered this well enough in the discussion topic, but it bears restating if it's not clear.

You probably did, i forgot there was a discussion about it.


Okay, thank you, so, can we separate the snippet in two, with corresponding names? something like simple glossary and compound glossaries, i guess?

We really ought to work on tooltips for these things, or a brew explaining each snippet.

dbolack-ab:
> Okay, thank you, so, can we separate the snippet in two, with corresponding names? something like simple glossary and compound glossaries, i guess?

I mean, we could? But I don't understand the point. It's the same code.

> 
> We really ought to work on tooltips for these things, or a brew explaining each snippet.

5e-Cleric:
> If you have a single glossary, use the first syntax. If you have multiple, use the second where "Glossary Name" is the glossary to add the term to.

you have two syntaxes, so each should be a different snippet, no?

ericscheid:
see comments on syntax, ease of use, snippet plurality, and more

https://github.com/naturalcrit/homebrewery/issues/3370#issuecomment-3540424138

dbolack-ab:
> > If you have a single glossary, use the first syntax. If you have multiple, use the second where "Glossary Name" is the glossary to add the term to.
> 
> you have two syntaxes, so each should be a different snippet, no?

Pretty sure we're not talking about the same snippet concept at this point.

I have only added a menu entry to collect and pour the Glossary entries.

5e-Cleric:
Lets keep the conversation in the issue, because eric shared some insights there https://github.com/naturalcrit/homebrewery/issues/3370#issuecomment-3540424138

Issue #1807, opened 2021-11-06
Feature request: Index maker


5e-Cleric:
So when i say index people thinks of the ToC, but i am talking about the actual index on the PHB:
![image](https://user-images.githubusercontent.com/91662974/140615723-7f74055c-d4de-4f80-867a-6ff809a84811.png)

Could something like this be implemented? something simmilar to the ToC, just getting a few keywords, maybe the ones with the class=index? or the ones we target with **** or -word-?

I feel it is the last bit to add for hb to really look like the PHB, though i get no more than a few people have brews that large as to need one of those.

ericscheid:
Collecting a list of words and compiling a concordance isn't too hard, but also isn't very helpful.

A good index will have things like concept groups (`casting a spell` has sub-entries of `area of effect`, `attack roll`, etc), and cross references (like `carrying capacity: see under lifting and carrying`).

Indexing is difficult.

calculuschild:
Yeah, I think making an autogenerator for an index is going to be very tricky to pull off.

However we can definitely add a snippet for an index with the necessary styling and some example text inside.

5e-Cleric:
oh man... my brew is 70+ pages long, it is a pain to track words, lucky we still have chrome's search for words

ericscheid:
As a thought .. we might be able to leverage `{curly-spans}` here, at least as a means of marking up spans of text with a class.

> Every character has six ***{index*** ability scores}. These can be ***{index,topic="ability score",entry="rolling"*** rolled} or pulled from a ***{index,topic="ability score"*** standard array}, and these can be modified by ***{index,entry="ability score",entry="increase",see="racial traits"*** racial traits}.

Gazook89:
@ericscheid I think you want to be sure you have double brackets on those spans e.g. `{{index ability scores}}`  otherwise you have got the injector syntax.

5e-Cleric:
This issue is closed, as will never be done, instead a template snippet will be done, as tracked in #1096

ericscheid:
Re-opening as PR #3113 is making a brave attempt.

dbolack-ab:
Working code needs an eval.

PR #4153, opened 2025-04-20
Brew Snippet variables.

dbolack-ab:
## Description

Brew snippet variables are Brew variable assignments parsed by the lexer but not stored in the stack. They are intended to be assignments picked up by snippet parsing of the brew for the snippets' use only.

example:

`[var]#content`

## Use Cases:

Indexing Snippet
Glossary Snippet

#3370 
#1807 

### Reviewer Checklist

  - [ ] Create a Snippet variable, verify it is not expressable via brew var syntax
  - [ ] Verify Brew var assignments still work and can be expressed


dbolack-ab:
On further discussion, this is not the best route forward. Implementing helped determine that.

PR #4156, opened 2025-04-21
toolbar, small fixes

### LINKED ISSUE #3835: Store toolbar state in localStorage

5e-Cleric:
these are some small fixes i've wanted to do for some time, nothing breaking:

- [x] small changes to the styling to prevent overlapping with the left positioned buttons
- [x] have the fit to width function fit as many pages as it is showing now (facing pages) 

also Fixes #3835, storing the toolbar state into localStorage

Stores a variable array in localstorage called `hb_toolbarState` and `hb_toolbarVisibility` which store each part separately, since they are used in different components.

Those states are loaded on mount using useEffect.

As for right now, all works except the page shadows, for some reason.

Might alter how the shadows work to fix this.


5e-Cleric:
Both buttons should work as initially intended just fine:

- fit to width
  - single page : fit 1 page width
  - facing      : fit 2 pages width + column gap
  - flow        : fit 1 page width

- Fit entire page
  - single page : fit 1 page longest'd dimension
  - facing      : fit 2 pages by longest dimension (1 page size vs (2 page width + column gap)
  - flow        : fit as many pages as they fit, while fitting the longest dimension of 1 page

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -167,11 +190,11 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
 						<h1>Options</h1>
 						<label title='Modify the horizontal space between pages.'>
 							Column gap
-							<input type='range' min={0} max={200} defaultValue={10} className='range-input' onChange={(evt)=>handleOptionChange('columnGap', evt.target.value)} />
+							<input type='range' min={0} max={200} defaultValue={displayOptions.colGap || 10} className='range-input' onChange={(evt)=>handleOptionChange('columnGap', evt.target.value)} />
```
calculuschild, line 193: should this be `displayOptions.columnGap`?

5e-Cleric: i changed that in #4106 and have ported it over, i might have messed up that name there, i'll fix it.

5e-Cleric:
I closed #4106 and will keep developing both together, which are almost ready.

calculuschild:
Doing one more pass on this PR. Three notes:

1) When the toolbar is hidden, it becomes _very_ difficult to see how to reopen it:

image

Keeping in mind that the `hbToolbar_visibility` localstorage will default to `false` if it doesn't exist, this means the toolbar will disappear for most users with the button being nearly invisible. I might suggest the default be `true`, and make the visibility button more obvious when it is hidden.

2) The `hbToolbar_visibility` key doesn't seem to actually be working as expected. It behaves as if it's only checking for the key existence, rather than the boolean value. I.e.: If the value does not exist at all, the page loads with toolbar hidden. If the value does exist (true or false), the page loads with the toolbar visible. Do we need some type casting here?

https://github.com/user-attachments/assets/28316be6-66d0-4384-a042-6dee7b28b79c


3) The "Jump To" button no longer scrolls to the top of the page. There is some padding offset or something:

##### Without this PR:

image


##### With this PR:

image

5e-Cleric:
Hm, i will resolve this issues shortly (not today though), thanks for the notes.

5e-Cleric:
> The "Jump To" button no longer scrolls to the top of the page. There is some padding offset or something

I'm not being able to replicate this issue, at all, have compared live, stage, and my local and do not see this bug.

The rest are fixed though

5e-Cleric:
closed it by mistake

calculuschild:
> I'm not being able to replicate this issue, at all, have compared live, stage, and my local and do not see this bug.

Here's a video showing this being reproduced on the PR deployment:

https://github.com/user-attachments/assets/fc8726a9-8637-4ffd-8e71-db66e7d5a012

We used to account for the height of the toolbar when jumping, but now we do not.

5e-Cleric:
> > I'm not being able to replicate this issue, at all, have compared live, stage, and my local and do not see this bug.
> 
> Here's a video showing this being reproduced on the PR deployment:
> 
>  Recording.2025-05-07.142712.mp4 
> We used to account for the height of the toolbar when jumping, but now we do not.


That is the current behavior on the live site, I believe Gambatte got rid of that padding at some point, don't ask me why.

https://github.com/user-attachments/assets/6202dfc4-7e51-44d4-8231-907a78e15b02

calculuschild:
Weird. I don't know why it wasn't showing that for me before. Ok, I guess this is alright then. Will merge.

Thanks @5e-Cleric !

calculuschild (approved):

Gazook89:
Can you double check that the fit/fill buttons work?  I am working on the toolbar in my own branch right now, and finding that I'm getting some console errors around the zoom/fit buttons.  I confirmed the errors exist on the [staging site]() as well, and I went back to commit 9b4047f3f9a7265b8dff8b43c8bef75ce12f6992 which is part of this branch but before anything was done with these buttons and the errors don't exist.  A few commits later on the branch, and they do.  

I'm looking it over to see if I can find what the exact fix is, while still preserving the intent of the PR.   The errors in the console relate to getting `displayOptions.zoomLevel` as `NaN` or `null`.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -55,15 +60,30 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
...
+			else
+				desiredZoom = (iframeWidth / (widestPage + 20)) * 100;
 
 		} else if(mode == 'fit'){
-			let minDimRatio;
 			// find the page with the largest single dim (height or width) so that zoom can be adapted to fit it.
-			if(displayOptions.spread === 'facing')
-				minDimRatio = [...pages].reduce((minRatio, page)=>Math.min(minRatio, iframeWidth / page.offsetWidth / 2), Infinity);    // if 'facing' spread, fit two pages in view
+			let minDimRatio;
+			if(displayOptions.spread === 'active')
```
Gazook89, line 71: `active` isn't a valid value for displayOptions.spread-- should it be 'facing'?  Changing this to 'facing' doesn't fix the wider issue but it's a start

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -55,15 +60,30 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
...
+						iframeWidth / page.offsetWidth,
+						iframeHeight / page.offsetHeight
+					),
+					Infinity
+				);
 			else
-				minDimRatio = [...pages].reduce((minRatio, page)=>Math.min(minRatio, iframeWidth / page.offsetWidth, iframeHeight / page.offsetHeight), Infinity);
+				minDimRatio = [...pages].reduce(
+					(minRatio, page)=>Math.min(minRatio,
+						iframeWidth / ((page.offsetWidth * 2) + parseInt(displayOptions.columnGap)),
```
Gazook89, line 82: I think another issue is that `displayOptions.columnGap` isn't always set, so it's coming back null and screwing up the `minDimRatio` returned here.  If I adjust the column gap first, i can then get the 'fit entire' button to work as expected.

Gazook89:
Here is what I have, that works for me, but I'm not sure I've preserved whatever your goal was with the facing pages thing-- though to me it seems fine:

toolbar.jsx
```jsx
if(displayOptions.spread === 'facing'){
		minDimRatio = [...pages].reduce(
			(minRatio, page)=>Math.min(minRatio,
				iframeWidth / page.offsetWidth,
				iframeHeight / page.offsetHeight
			),
			Infinity
		);
	} else {
		minDimRatio = [...pages].reduce(
			(minRatio, page)=>Math.min(minRatio,
				iframeWidth / ((page.offsetWidth) + parseInt(displayOptions.columnGap)),
				iframeHeight / page.offsetHeight
			),
			Infinity
		);
	}
```

And in brewRenderer.jsx
```jsx
const [displayOptions, setDisplayOptions] = useState({
		zoomLevel    : 100,
		spread       : 'single',
		startOnRight : true,
		pageShadows  : true,
		columnGap    : 10,
		rowGap       : 10
	});
```

5e-Cleric:
Huh, maybe i made a mess while merging the two PRs? i seem to remember that i tested it after and it worked fine. Huh, working on it

5e-Cleric:
@Gazook89 i cannot confirm the issues you mention in a local branch
they do happen in stage, but not in a local updated branch

to clarify, i tested this before and after clearing the cache.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
5e-Cleric: Then why was there code to set it to active before i even made the change? i must be missing something.

image

Gazook89: It was my mistake. Should be single

PR #4158, opened 2025-04-21
Change page overflow from hidden to clip

### LINKED ISSUE #4157: Navigating via headings overview changes page layout
This is the bug:

https://github.com/user-attachments/assets/2fb90eaf-31dd-487a-96d4-9bd0fbb878f2

I only caused it by navigating to a header inside a page, it goes away in a reload, not sure how is this possible

G-Ambatte:
This PR resolves #4157.

This PR changes the `overflow` property of `page` elements from `hidden` to `clip`. This prevents unintended scrolling within the page when navigating by ID.

### DIFF
--- a/themes/V3/Blank/style.less
+++ b/themes/V3/Blank/style.less
@@ -52,7 +52,7 @@ body { counter-reset : page-numbers 0; }
 	width                  : 215.9mm;
 	height                 : 279.4mm;
 	padding                : 1.4cm 1.9cm 1.7cm;
-	overflow               : hidden;
+	overflow               : clip;
 	background-color       : var(--HB_Color_Background);
 	text-rendering         : optimizeLegibility;
 	contain                : strict;


G-Ambatte:
From the MDN reference docs (https://developer.mozilla.org/en-US/docs/Web/CSS/overflow):

### Hidden

> Overflow content is clipped at the element's padding box. There are no scroll bars, and the clipped content is not visible (i.e., clipped content is hidden), but the content still exists. User agents do not add scroll bars and also do not allow users to view the content outside the clipped region by actions such as dragging on a touch screen or using the scroll wheel on a mouse. **_The content can be scrolled programmatically_** (for example, by linking to anchor text, by tabbing to a hidden yet focusable element, or by setting the value of the [scrollLeft](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollLeft) property or the [scrollTo()](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo) method), in which case the element box is a scroll container.

### Clip

> Overflow content is clipped at the element's overflow clip edge that is defined using the [overflow-clip-margin](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-clip-margin) property. As a result, content overflows the element's padding box by the [<length>](https://developer.mozilla.org/en-US/docs/Web/CSS/length) value of overflow-clip-margin or by 0px if not set. Overflow content outside the clipped region is not visible, user agents do not add a scroll bar, and **_programmatic scrolling is also not supported._** No new [formatting context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_display/Block_formatting_context) is created. To establish a formatting context, use overflow: clip along with [display: flow-root](https://developer.mozilla.org/en-US/docs/Web/CSS/display#flow-root). The element box is not a scroll container.

Emphasis mine.

5e-Cleric (approved):

PR #4170, opened 2025-04-30
Tweak icon height

### LINKED ISSUE #3318: Fix spacing in userpage
image
As you can see in this image from my own userpage, brews with the HB icon in them grow taller than those stored in Drive.

The issue here is the image has a set height of 24px, overflowing the parent element with a height of 17.33px (yes, decimal pixels are a thing)

As a fix, changing that height in brewitem.less should fix it, another option is to change the image source link to use our `/assets/naturalCritLogoRed.svg`, and add a shadow, which should decrease the size.

This is a simple issue, ideal for new developers to try their hand at it.

dbolack-ab:
## Description

Adjusts the icon height to be in line with FA icons on the page. 

Note that it's a lot washed out and hard to read - though I dunno if I would call that an issue with the scaling or what is being scaled.

I expect feedback will tweak this tweak. :)

- Closes #3318


Gazook89:
Can you show a screenshot that this works on your machine?  On mine, I duplicated a `brewItem` element via console tools, and then copied an element containing the google drive icon from the user page of the live site and pasted in place of the homebrewery icon.  

image

There is a height difference of a little more than 3 pixels between the two.  I'm not sure why, yet.  The icons themselves (the HB logo and the Drive logo) **are the same size, 19.27px**.  So somewhere else there is a difference, even though these are the same brewItem, just cloned.  

I cannot find any difference except that the `.info` box is slightly different between the two, in heights, but nothing else.

dbolack-ab:
> I cannot find any difference except that the `.info` box is slightly different between the two, in heights, but nothing else.

I assume this has to do with the source file for the HB icon. I swapped it to an SVG.

5e-Cleric:
its still there, and it is the icons, but somehow the dev tools reports they are the exact same height.

I knot it is the icons, because hiding them all with display:none makes the size to change and be the same.

5e-Cleric:
A perhaps complex solution for this is to just remove them from the normal flow of the page:

image

this seems to do the trick (i'm using a browser extension to apply css without needing to rebuild or anything.

dbolack-ab:
@5e-Cleric Try this update.

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
```diff
@@ -115,15 +115,18 @@
 		}
 	}
 	.googleDriveIcon {
-		height  : 18px;
 		padding : 0px;
 		margin  : -5px;
+		height  : 18px;
+		width   : 20.1406px;
```
5e-Cleric, line 121: is there any good reason for this very very specific number?

5e-Cleric (approved):

client/homebrew/pages/basePages/listPage/brewItem/brewItem.less
dbolack-ab: Matches the dimensions I was seeing for the Google one.

5e-Cleric: but the google icon has the same exact height set..?

calculuschild: I removed the width in testing and don't see any real impact. I would suggest removing it.

PR #4195, opened 2025-05-11
Correct missed Font Awesome version number references

dbolack-ab:
## Description

Updates a few specific references from 5 to 6


5e-Cleric:
This is good as a quick fix, but i think all of these should work rendering the icon programatically instead, so this does not need more updating in the future, but i can do that in a later, separate PR. I'll merge this quickly.

5e-Cleric (approved):

PR #4196, opened 2025-05-11
Add function to remove other authors

5e-Cleric:
This PR introduces a check for ownership in app.js, which is prop drilled into editPage and there used to render the author list differently depending on if the user is author or owner.

## UI:

image

Ownership is checked again on the server side, just in case.


5e-Cleric:
I don't get why are the tests failing...

dbolack-ab:
The problem is 

```
router.put('/api/:id/:author', checkClientVersion, asyncHandler(api.deleteAuthor));
```

I don't understand enough about the mongoose router - it looks like it is colliding with the 

```
router.put('/api/prune/:id/:author', checkClientVersion, asyncHandler(api.deleteAuthor));
```
Changing like below works:

```
router.put('/api/prune/:id/:author', checkClientVersion, asyncHandler(api.deleteAuthor));
```

dbolack-ab:
> i can't find such a code in my local instance, where is that from?

/server/homebrew.api.js

510:router.put('/api/:id/:author', checkClientVersion, asyncHandler(api.deleteAuthor));

5e-Cleric:
huh, well, i fixed that, but the tests are still failing

dbolack-ab:
Weird. It passes locally.

```
[BABEL] Note: The code generator has deoptimised the styling of /mnt/Extra/home/dbolack/Code/homebrewery-fork/node_modules/lodash/lodash.js as it exceeds the max of 500KB.
 PASS  server/homebrew.api.spec.js (9.625 s)
[BABEL] Note: The code generator has deoptimised the styling of /mnt/Extra/home/dbolack/Code/homebrewery-fork/node_modules/react-dom/cjs/react-dom.development.js as it exceeds the max of 500KB.
 PASS  tests/routes/static-pages.test.js
 PASS  tests/html/safeHTML.test.js
 PASS  server/admin.api.spec.js
 PASS  tests/markdown/mustache-syntax.test.js
 PASS  tests/markdown/variables.test.js
 PASS  tests/markdown/definition-lists.test.js
 PASS  tests/markdown/emojis.test.js
 PASS  tests/markdown/basic.test.js
 PASS  tests/markdown/paragraph-justification.test.js
 PASS  tests/markdown/non-breaking-spaces.test.js
 PASS  tests/markdown/hard-breaks.test.js
 PASS  server/middleware/content-negotiation.spec.js
------------------------------|---------|----------|---------|---------|-------------------------------------------------------------------------------
File                          | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s                                                             
------------------------------|---------|----------|---------|---------|-------------------------------------------------------------------------------
All files                     |   60.92 |    51.41 |   52.65 |      62 |                                                                               
 client                       |    90.9 |     62.5 |     100 |     100 |                                                                               
  template.js                 |    90.9 |     62.5 |     100 |     100 | 1-5                                                                           
 client/homebrew/brewRenderer |   96.15 |    91.66 |     100 |     100 |                                                                               
  safeHTML.js                 |   96.15 |    91.66 |     100 |     100 | 8                                                                             
 client/homebrew/utils        |   28.57 |      100 |       0 |   33.33 |                                                                               
  request-middleware.js       |   28.57 |      100 |       0 |   33.33 | 6-9                                                                           
 server                       |   46.56 |       30 |   33.71 |   47.49 |                                                                               
  admin.api.js                |   61.14 |    55.26 |   41.86 |      62 | 22,33,54-58,64-72,78,83-92,98-116,121-128,133-148,153-163,304,349-350,375-381 
  app.js                      |      35 |    18.03 |    23.4 |   36.43 | ...17,422-429,434-456,461-509,516-520,530-534,573,581-588,592,606-624,628-630 
  brewDefaults.js             |     100 |      100 |     100 |     100 |                                                                               
  config.js                   |       0 |        0 |       0 |       0 |                                                                               
  forcessl.mw.js              |      40 |    33.33 |     100 |      25 | 3-6                                                                           
  googleActions.js            |    7.07 |     4.54 |       0 |    7.14 | 14-23,41-342                                                                  
  homebrew.api.js             |    68.3 |       52 |   67.56 |   70.28 | 53-86,104,133-137,338-437                                                     
  homebrew.model.js           |      20 |        0 |       0 |      20 | 8-9,36-43,47-53,57-63                                                         
  notifications.model.js      |   88.46 |       50 |     100 |   91.66 | 30,53                                                                         
  static-assets.mv.js         |    37.5 |        0 |      50 |    37.5 | 5-16                                                                          
  token.js                    |      10 |      100 |       0 |      10 | 6-22                                                                          
  vault.api.js                |      20 |        0 |       0 |   21.42 | 8-9,18-19,23-29,33,37-79,84-102                                               
 server/middleware            |   61.53 |       75 |      75 |   61.53 |                                                                               
  check-client-version.js     |       0 |        0 |       0 |       0 | 4-13                                                                          
  content-negotiation.js      |     100 |      100 |     100 |     100 |                                                                               
 shared                       |      35 |    29.54 |   33.33 |   37.14 |                                                                               
  helpers.js                  |      35 |    29.54 |   33.33 |   37.14 | 11-27,39-54,67-69,78-85,109-115,120-135                                       
 shared/naturalcrit           |   91.75 |    83.61 |   92.68 |   92.12 |                                                                               
  markdown.js                 |   91.75 |    83.61 |   92.68 |   92.12 | 124,132,150,167,336,380,439,602,696,715-716,720,842-897                       
 themes/fonts/iconFonts       |     100 |      100 |     100 |     100 |                                                                               
  diceFont.js                 |     100 |      100 |     100 |     100 |                                                                               
  elderberryInn.js            |     100 |      100 |     100 |     100 |                                                                               
  fontAwesome.js              |     100 |      100 |     100 |     100 |                                                                               
  gameIcons.js                |     100 |      100 |     100 |     100 |                                                                               
------------------------------|---------|----------|---------|---------|-------------------------------------------------------------------------------
Jest: "server/homebrew.api.js" coverage threshold for statements (70%) not met: 68.3%

Test Suites: 13 passed, 13 total
Tests:       250 passed, 250 total
Snapshots:   0 total
Time:        16.133 s
```

dbolack-ab:
> huh, well, i fixed that, but the tests are still failing

Particularly weird because looking at the test that failed, it passed.

calculuschild:
> Particularly weird because looking at the test that failed, it passed.

The log results above are showing a failing result. The tests themselves gave the correct output and passed, but the coverage test is not passed:

> Jest: "server/homebrew.api.js" coverage threshold for statements (70%) not met: 68.3%

5e-Cleric:
tests passing! please don't let me do tests anymore.

server/app.js
```diff
@@ -378,7 +378,7 @@ app.get('/edit/:id', asyncHandler(getBrew('edit')), asyncHandler(async(req, res,
 	req.brew = req.brew.toObject ? req.brew.toObject() : req.brew;
 
 	req.userThemes = await(getUsersBrewThemes(req.account?.username));
-
+	req.isOwner = req.account && (req.account.username == req.brew.authors[0]);
```
calculuschild: I think we can simplify a lot by letting `editPage.jsx` calculate `isOwner` rather than doing it on the server and prop drilling everything down.

server/homebrew.api.spec.js
```diff
@@ -382,7 +381,68 @@ describe('Tests for api', ()=>{
 			await expect(fn(req, null, next)).rejects.toEqual({ 'HBErrorCode': '51', 'brewId': '1', 'brewTitle': 'test brew', 'code': 404, 'message': 'brew locked' });
 		});
 	});
+	describe('Get CSS', ()=>{
```
calculuschild, line 382: Is there a particular reason some of the test cases have been moved around? It's ok if there is. It just isn't clear if it's relevant to deleting authors or if there is something special I should be checking here.

server/homebrew.api.js
```diff
@@ -534,6 +553,7 @@ router.post('/api', checkClientVersion, asyncHandler(api.newBrew));
 router.put('/api/:id', checkClientVersion, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
 router.put('/api/update/:id', checkClientVersion, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
 router.delete('/api/:id', checkClientVersion, asyncHandler(api.deleteBrew));
+router.put('/api/prune/:id/:author', checkClientVersion, asyncHandler(api.deleteAuthor));
```
calculuschild: should be able to put the `api.getBrew()` middleware inline here like we do on the other API calls. Then you don't need to do special handling inside `api.deleteAuthor`.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -156,6 +156,19 @@ const MetadataEditor = createClass({
...
 
+	handleDeleteAuthor : function(author){
+		if(!confirm('Are you sure you want to delete this author? They will lose all edit access to this brew.')) return;
+		request.put(`/api/prune/${this.props.metadata.editId}/${author}`)
+			.send()
+			.end((err, res)=>{
+				if(err) {
+					this.props.reportError(err);
+				} else {
+					window.location.reload();
```
calculuschild: This feels a little funny that deleting an author requires a window reload, but adding an invited author does not.

Is there a reason we can't just re-use the same approach all of our other metadata changes follow? I.e., make the change, call `this.props.onChange(this.props.metadata, 'fieldName');`, and then let the change be applied to the server when the user saves?

calculuschild (changes requested):

server/app.js
5e-Cleric: req.account is calculated in server, so we ought to prop drill that down for the check anyway, no? Is this any worse?

server/homebrew.api.spec.js
5e-Cleric: I seem to recall i moved them around so they were in the same order as the functions they were testing are in the file, i'd better check

client/homebrew/editor/metadataEditor/metadataEditor.jsx
5e-Cleric: Well, if i delete an author, and then close the site, the author is not deleted, but a user might presume it is.

server/app.js
5e-Cleric: nevermind, forgot about globals, yes, this will be made in the metadata editor itself.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
calculuschild: Ok. I mean, the same thing would happen for any metadata change. Update title, change language, invite an author. Currently none of those "stick" until the next save. This is fine for now though.

server/homebrew.api.spec.js
5e-Cleric: image
Here, this is the current order in my PR, you'll notice they now match the order
(the ones that have its pair)

server/homebrew.api.js
5e-Cleric: special handling? its the same handling as in deletebrew

5e-Cleric: If i am understanding this right, if i add the `asyncHandler(api.getBrew('edit', false))`, the brew will come with the request, right?

Might do that change for deltebrew as well

client/homebrew/editor/metadataEditor/metadataEditor.jsx
5e-Cleric: I can make the change, all these feel weird to me, they should trigger a save, would you agree?

calculuschild: They do trigger a save. Just not immediate. An autosave will fire after a few seconds.

server/homebrew.api.js
calculuschild: Yes.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
5e-Cleric: ah, i guess i could do that.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -156,6 +157,15 @@ const MetadataEditor = createClass({
 			});
 	},
 
+	handleDeleteAuthor : function(author){
+		if(!confirm('Are you sure you want to delete this author? They will lose all edit access to this brew.')) return;
```
Gazook89: Consider using "remove" rather than "delete" here.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -194,16 +204,38 @@ const MetadataEditor = createClass({
...
+			<div className='field authors'>
+				<label>authors</label>
+				<ul className='list'>
+					{authors.map((author, i)=>(
+						<li className='tag' key={i}>{author}
+							{i>0 &&
+							<button
+								onClick={()=>this.handleDeleteAuthor(author)}
+								className='delete'
+								title={`Remove ${author} as an`}
```
Gazook89: Missing the word "author" at the end of this `title` attribute.

client/homebrew/editor/metadataEditor/metadataEditor.less
```diff
@@ -164,7 +161,21 @@
...
+	.authors.field {
+		.list > .tag:first-child {
+			position:relative;
+			
+			&::after {
+				content:'';
+				position:absolute;
+				border-radius:inherit;
+				inset:-1px;
+				border: 2px solid goldenrod;
```
Gazook89: Personally, I think a border is an odd choice here.  I'm not a fan of it aesthetically, but it also crosses the line into a "keyboard focus indicator", like when you are tabbing through inputs/buttons and you get the typical outline indicator to show which element you are on.  The current styling for that focus indicator in the properties editor is set as just a thing black outline, but even if this goldenrod is not consistent with that it just gives that seem *feeling*, IMO. 

I'm not sure I've got any great ideas, but for now I'd just go with bolding the username text.

Gazook89: Also, if keeping the border, I'm not sure if there is a reason for `::after` at all.  It could be done with a single line:

```
.list > .tag:first-child {
  position: relative;
  outline: 2px solid goldenrod;
}
```

but maybe i missed something here.

client/homebrew/editor/metadataEditor/metadataEditor.less
```diff
@@ -164,7 +161,21 @@
...
+			}
+		}
+		button {color:#B00;}
```
Gazook89: Use one of the colors we have defined in `colors.less` for consistency, like `redLight`.  The colors are already imported at the top of this file. 

And, if we are going to color these buttons red, we should just apply that to the broader `.tag` styling so that its consistent with the tags above.

If it were entirely up to me, I'd leave these buttons black, and flip to red on hover (with a slight easing transition, both ways), `transition: color .2s linear;`.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -194,16 +204,38 @@ const MetadataEditor = createClass({
...
+				</div>
 			</div>
-		</div>;
+		);
+		return (
+			<div className='field authors'>
+				<label>authors</label>
+				<ul className='list'>
+					{authors.map((author, i)=>(
+						<li className='tag' key={i}>{author}
```
Gazook89: Instead of conditionally rendering the 'remove' button, you might consider conditional rendering the entire tag:  either an "owner tag" or a "author tag".  Then, only author tags would have the remove button.

This would allow you to give the "owner tag" a `title: "Owner"` attribute, which would make it more clear why that tag is special and why it is has different styling.  It would also be more straightforward for the styling itself, you could give each a `.owner` or `.author` tag and apply CSS based on that rather than using `:nth-child(1)`.  

Also, each of these author tags could be a link to that users profile page.

Gazook89 (changes requested):
Left a few design related comments.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
5e-Cleric: oops

client/homebrew/editor/metadataEditor/metadataEditor.less
5e-Cleric: They are far too small for the color change to be noticeable, in my screen the buttons are smaller than the cursor, so you barely see the change.

client/homebrew/editor/metadataEditor/metadataEditor.less
5e-Cleric: fun fact, all the usernames were already bold.

5e-Cleric:
## Last Iteration:

image

Gazook89 (approved):
What about this:

- crown squashed vertically just a bit and rotated with `transform: rotate(-25deg) scaleY(0.6);` and positioned to hang on the corner
- set only the tag that is the current user to bold; so not the *owner*, but the currently logged in user *if* they are one of the authors.  Logically I think if you are viewing the editor the current user *must* be in that list (or at least in the Invited Authors list).
- remove the `text-decoration` on the hyperlink (use `any-link` selector).  
- change the `title` attribute on the user profile link to say "Author - Click to open [username]'s profile in a new tab" and set the link to open in a new tab.
- and because there are now two actions in the tag (a link and a button), I really think that the 'remove' button needs some sort of hover effect...either changing the "x" color or fading in a background color of some sort for that button.  

image

----

All that said, none of that is key to the functionality and isn't really a blocker to the PR for me.

PR #4207, opened 2025-05-22
Fix Codemirror Overflow

### LINKED ISSUE #4098: Top of snippetBar is hidden when using Ctrl+F in Codemirror section

5e-Cleric:
should fix #4098

I am still confused, i think the error is i somehow reverted the fix and made a mess while testing.

The base height of the editor has to take into account the height of the snippetbar. A Js solution can be established to programatically take the height of the snippetbar and use it, or this change, that will need fixing if the height or width of the bar changes.


5e-Cleric:
Look, i know i should be using ref for this, but this approach with vanilla js is shorter and easier.

calculuschild:
I don't have access to test this at the moment. If someone else can confirm this works we can merge and get it deployed. We can clean up later.

calculuschild:
Ok, I was able to test this now. Will merge and send out live.

PR #4208, opened 2025-05-22
bring back icons to admin

5e-Cleric:
imports the font to admin, although perhaps we should import it through the shared folder and move the font there.


calculuschild:
This is fine for now. I'll send this out so it's working and you can refactor it later if you want.

PR #4210, opened 2025-05-23
Add column, columnbreak, and pagebreak compatibulity to Legacy

dbolack-ab:
## Description

This updates the Legacy Renderer workflow to include use of `\pagebreak`, `\column`, and `\columnbreak`.

Column compatibility is performed by substituting the column label for a set of ` ```\n```\n ` before parsing.  Pagebreak comes from extending the previous split to regex.


calculuschild:
Looks good. I just shuffled one line around.

PR #4213, opened 2025-05-25
Add basic CSS cache busting

G-Ambatte:
This PR changes the name of the compiled CSS file, created during the build process, and modifies the HTML template to point at this file.

The new filename includes the current Homebrewery version (i.e. `bundle-v3.19.0.css`), so any future versions will automatically import their associated compiled CSS; this should eliminate the issues experienced during the recent v3.19.0 update.

### DIFF
--- a/client/template.js
+++ b/client/template.js
@@ -1,3 +1,6 @@
+import packageJSON from '../package.json' with { type: 'json' };
+const version = packageJSON.version;
+
 const template = async function(name, title='', props = {}){
 	const ogTags = [];
 	const ogMeta = props.ogMeta ?? {};
@@ -15,7 +18,7 @@ const template = async function(name, title='', props = {}){
 		<head>
 			<meta name="viewport" content="width=device-width, initial-scale=1, height=device-height, interactive-widget=resizes-visual" />
 			<link href="//fonts.googleapis.com/css?family=Open+Sans:400,300,600,700" rel="stylesheet" type="text/css" />
-			<link href=${`/${name}/bundle.css`} type="text/css" rel='stylesheet' />
+			<link href=${`/${name}/bundle-${version}.css`} type="text/css" rel='stylesheet' />
 			<link rel="icon" href="/assets/favicon.ico" type="image/x-icon" />
 			${ogMetaTags}
 			<meta name="twitter:card" content="summary">
--- a/scripts/buildHomebrew.js
+++ b/scripts/buildHomebrew.js
@@ -10,6 +10,9 @@ import babel          from '@babel/core';
 import babelConfig    from '../babel.config.json' with { type : 'json' };
 import less           from 'less';
 
+import packageJSON from '../package.json' with { type: 'json' };
+const version = packageJSON.version;
+
 const isDev = !!process.argv.find((arg)=>arg === '--dev');
 
 const babelify = async (code)=>(await babel.transformAsync(code, babelConfig)).code;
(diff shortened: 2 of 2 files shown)


G-Ambatte:
> I don't think the issues were only CSS, is html also cached?

I believe that any query can be cached; it would be trivial to add the same `version` parameter to the names of the bundled JS  and SSR files.

Additionally, `version` doesn't have to be Homebrewery version - it could be replaced by any fingerprinting hash, in the event that we feel the need to decouple cache updating from the Homebrewery project's official version number.

5e-Cleric:
I am not very well versed in cache, will then the previous bundle be erased automatically if the new file has a different name?

G-Ambatte:
> I am not very well versed in cache, will then the previous bundle be erased automatically if the new file has a different name?

Correct, my understanding is that eventually the previous bundle will age out and be dropped during normal garbage collection, depending on the cache settings.

5e-Cleric:
> > I am not very well versed in cache, will then the previous bundle be erased automatically if the new file has a different name?
> 
> Correct, my understanding is that eventually the previous bundle will age out and be dropped during normal garbage collection, depending on the cache settings.

we're not talking about eventually here, we're talking about immediately erasing it so the user only gets the updated one, no?

G-Ambatte:
> we're not talking about eventually here, we're talking about immediately erasing it so the user only gets the updated one, no?

The filename changes with each release, so the browser only requests the updated one.

calculuschild:
@G-Ambatte What was the original issue this was trying to solve? Something with v3.19.0? Just so I can test against it?

G-Ambatte:
> @G-Ambatte What was the original issue this was trying to solve? Something with v3.19.0? Just so I can test against it?

When the site's styling changes in a version update, users often must perform a hard reload in order to clear the cached version and get the new CSS.

This PR puts the Homebrewery version into the CSS filename at build time, and changes the template to request the current version. Because the CSS filename changes when the version changes, the previous version will no longer be a valid match in the cache, and the updated & correct version of CSS will always be requested with no additional user actions required.

---

This has been an issue for years now, but I don't know that it's ever been deemed serious enough to actually open an issue about. I recall multiple Reddit posts following updates that were solved by "do a hard reload".

calculuschild:
@G-Ambatte  Right, I understand the concept of cache busting. But it seems like there was something specific that happened with 3.19.0 that you were concerned about. What was that?

> this should eliminate the issues experienced during the recent v3.19.0 update.

Also keep in mind, this PR only affects the CSS of the site UI. Brew Themes are separate CSS files that are not included in this CSS bundle.

G-Ambatte:
> it seems like there was something specific that happened with 3.19.0 that you were concerned about

Not specifically, I saw a post-update Reddit report that was solved by "do a hard refresh" and happened to be in a position where I was able to put together a PR while it was still on my mind.

G-Ambatte:
> Brew Themes are separate CSS files that are not included in this CSS bundle.

If Brew Theme CSS is handled in the build scripts, then we could expand the PR to also include those files in a similar fashion.

G-Ambatte:
@calculuschild I had a dig around and I believe these were the comments that triggered the PR:

https://www.reddit.com/r/homebrewery/comments/1kt3jxd/v319_is_here/mtqqt0r/

5e-Cleric (approved):
This has been sitting too long, if you remake it for hacktoberfest, it can be merged within the week for sure.

G-Ambatte:
> This has been sitting too long, if you remake it for hacktoberfest, it can be merged within the week for sure.

That's the plan, I am trying to make sure that I give the PRs a genuine review before I create the PR, with the benefit of being a few weeks/months older and thus presumably wiser.
I'll target this one next.

G-Ambatte:
Taking this one down for now; I'll review and re-create as appropriate.

PR #4220, opened 2025-05-27
fix dev background not working

5e-Cleric:
This PR moves the background to the homebrew.jsx component to allow for its use in any site, and removes redundant backgrounds.

### DIFF
--- a/client/homebrew/homebrew.jsx
+++ b/client/homebrew/homebrew.jsx
@@ -1,7 +1,7 @@
 /* eslint-disable camelcase */
 import 'core-js/es/string/to-well-formed.js'; //Polyfill for older browsers
 import './homebrew.less';
-import { useState, useEffect } from 'react';
+import React, { useState, useEffect } from 'react';
 import { StaticRouter as Router, Route, Routes, useParams, useSearchParams } from 'react-router';
 
 import HomePage    from './pages/homePage/homePage.jsx';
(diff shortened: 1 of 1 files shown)


5e-Cleric:
Can confirm it works in any pages except admin. error, vault, user, account, all with brewrenderer...

calculuschild:
Hi @5e-Cleric . I merged your changes to make `homebrew.jsx` a functional component, so this has some conflicts to resolve.

But also, can you help me understand what this is trying to fix so I can test? I'm not sure what part of the dev background is broken. (I'm sure it was mentioned in the Gitter but I can't recall).

5e-Cleric:
the dev background was set in brewRenderer, so it is not displayed in the userpage, vault, account page..

this PR drops duplicate background colors and the blue background is now set in the base page, homebrew.jsx, instead of having the blue background in multiple layers over each other.

also that way, the dev background changes at that level

Gazook89:
Is it worth just dropping the 'Local' text out of there?  To me it's just really distracting.  I don't love the color, either, but I get that it might need to be eye-catching.

FWIW, i just use a [browser extension that adds a dismissable border for any list of URLs](https://chromewebstore.google.com/detail/jjccpcminoppplpmcfghflolejbdkekm?utm_source=item-share-cb):

image

5e-Cleric:
We need some indication of whether the server is local or a deployment, so i'd keep the text, yes, i know it is ugly and distracting, but that means we shall not forget about it. I'd rather it stayed.

PR #4225, opened 2025-05-28
Fix calc operator regex

### LINKED ISSUE #3175: Curly syntax not working with Calc numeric operators
Right now, the curly syntax for the `calc()` function, breaks because the RegEx pattern does not expect mathematical symbols.

This brew should work as an example: https://homebrewery.naturalcrit.com/share/oDUY9Cl2cDd6

Preview for later/ if the brew was deleted:
image

5e-Cleric:
Should fix #3175

Widens the regex to account for `+` and `*` characters

### DIFF
--- a/shared/naturalcrit/markdown.js
+++ b/shared/naturalcrit/markdown.js
@@ -185,7 +185,7 @@ const mustacheSpans = {
 	start(src) { return src.match(/{{[^{]/)?.index; },  // Hint to Marked.js to stop and check for a match
 	tokenizer(src, tokens) {
 		const completeSpan = /^{{[^\n]*}}/;               // Regex for the complete token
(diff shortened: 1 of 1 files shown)


calculuschild:
> As I understand the regex, this change is only being applied to the match text inside `"` quotation marks. Should the change also be applied to the second match criteria, which is for unquoted text?

I agree. `{{background:lightgreen,height:calc(57px?*3) }}` should also be valid, without using quotes. Addition and subtraction need spaces, and so must be quoted, but I am going to add + to the non-quoted case anyway. I don't think it causes any harm, and then both parts of the regex can be equal. Might be a way we can extract common regex from all of these and simplify this for the future as well.

calculuschild:
Ok, I added +*/ to the non-quoted case as well. Otherwise all seems good. Merging!

PR #4226, opened 2025-05-28
refactor homebrew.jsx to be functional

5e-Cleric:
refactors the component to be functional, comes with some lines of difference.

Works just fine locally.


calculuschild:
All seems to work. Thanks!

PR #4227, opened 2025-05-28
Add Transfer mechanism for inserting Brew vars into css vars plus HB_pageNumber workaround

dbolack-ab:
## Description

This adds a new operator to mustache blocks that copies the value of a brew variable into a designated CSS property variable. The primary use case for this is use of variable values in `:before` and `:after` blocks, where Brew vars cannot be resolved.

This also includes a workaround for ensuring that the CSS version of HB_pageNumber exists and is valid in a `\page` mustache.

## Syntax:

{...,brewVariable>cssPropertyWithoutDashes,...}

Example:

`\page {MnMMono,HB_title>bookTitle}`

### Brew Contents:

```markdown
[exampleBrew]:I am the new value


\page {exampleBrew>exampleCSS}
```

### Style Contents:

```css
:root {
  --exampleCSS: "I should be overridden!";
}


.page {

  &:before {
    content: var(--exampleCSS);
    bottom: 5px;
    left: 40px;
    background: unset;
  }
}
```

### Results 

Page 1 will show `I should be overridden!` in the upper left.
Page 2 will show `I am the new value` in the upper left.

This is not the most astounding example of using the feature.

### Reviewer Checklist

  - [ ] Duplicate example above into a new brew using the 5ePHB theme.
  - [ ] Verify Page 1 and page 2 have text per the Results.


calculuschild:
Interesting idea. I am curious what the benefit would be over the existing way we can already do this:

```
[exampleBrew]:I am the old value

{{mySpan,--exampleCSS:$[exampleBrew]}}

::

[exampleBrew]:I am the new value


{{mySpan,--exampleCSS:$[exampleBrew]}}
```

dbolack-ab:
> Right. I would agree there's not a need for a new syntax. Would be worth checking if there is any issue with `\page{--css...}` I didn't test that one.

Do you mean like ?

{{mySpan,--exampleCSS:"'My value'"}}

Gazook89:
I think he meant the injection straight into the `.page` node.

This doesn't work:
```
[myVariable]:15px

\page{width:$[myVariable]]}
```

or this:

```
[myVariable]:red

\page{--bg-color:$[myVariable]]}

// css/style editor

.page {

background:var(--bg-color);

}
```

Gazook89:
This works, to change the pseudo-selector `::before`'s background color via a variable on the `note` in the brew editor:

```

[myVariable]:green

{{note,--color:$[myVariable]

}}

```
css / style editor
```
.page .note::before {
  content: 'hello';
  color: blue;
  position: absolute;
  height: 10px;
  width: 40px;
  background: var(--color);
  
  @property --color {
    syntax: "<color>";
    inherits: false;
    initial-value: #c0ffee;
  }
  
}
```

Note, this uses the new CSS Custom Property `@property` rule which gives a greater ability to restrict it to types.  I believe it is only supported in Chrome right now.  **But**, that was just to trying it out...it works without the `@property` declaration as well.

I *could not* get it to work for setting the `content` property of the pseudo-selector `::before`, from the brew editor, though.

dbolack-ab:
> I think he meant the injection straight into the `.page` node.
> 
> This doesn't work:
> 
> ```
> [myVariable]:15px
> 
> \page{width:$[myVariable]]}
> ```

Oh, no, that won't work unless the variable was previously parsed. That section of code only applies the lexer.

dbolack-ab:
reopened, specifically for `\page {}` scenario.

dbolack-ab:
Didn't need to reopen this - with the variable preload "fix" in place, the `\page` mechanism should always have access to a Brew Variable.

PR #4237, opened 2025-06-02
Add defaults to the toolbar for rowGap and columnGap

5e-Cleric:
Toolbar has some previously undiagnosed issues related to the defaults of some values. This fixes that.


5e-Cleric:
fxes the issues just reported in the closed pr #4156 , in a simple manner, it works fine

PR #4242, opened 2025-06-04
Change Table of Contents generator to use system variables

G-Ambatte:
This PR adds a system variable for each page, with the name `HB_pageNumber_X`, where `X` is the number of the page, as found in the page's ID, e.g. `<div class='page' id='p1'>`. The value of this variable is set to whatever the `HB_pageNumber` variable is for that page.

The Table of Contents generator is modified to use these new variables in the displayed content, while using the actual page ID for the link. This allows the Table of Contents to provide functional in-document navigation, while displaying whatever custom numbering scheme that the user has applied.


Gazook89:
What about pages that don't have a page number on them?  Just thinking about the many brews where someone hasn't put any pageNumber snippet on any pages.  Seems like we should still create a ToC.  Maybe this works, I haven't checked out the branch to test it, so correct me if I'm wrong.

G-Ambatte:
> What about pages that don't have a page number on them? Just thinking about the many brews where someone hasn't put any pageNumber snippet on any pages. Seems like we should still create a ToC. Maybe this works, I haven't checked out the branch to test it, so correct me if I'm wrong.

The `HB_pageNumber` variable is the one automatic variable that we have right now, used for the "Variable Auto Page Number" so it will always have a value. Currently, it is set to `1` when the first page is rendered, and then on each subsequent page, incremented by one if the previous page's `HB_pageNumber` variable resolves to a number; if not, the previous page's non-number value is simply repeated.
So, in the event that the user has not modified these numbers at all, then the variables will simply resolve to their respective normal page numbers; i.e. `HB_pageNumber_1` => `1`, `HB_pageNumber_2` => `2`, and so on, for as many pages as exist.

G-Ambatte:
Tests are currently failing; as the current implementation adds an extra block to the source text of every rendered page, this changes the result of almost every single variable test.

G-Ambatte:
I've changed the process to add the `HB_pageNumber_N` variables after the `Marked.preprocess` but before `Marked.walkTokens`, which means that Marked render output is no longer modified - so all tests now complete normally.

G-Ambatte:
Marking this one as `blocked` until #4243 is resolved.

dbolack-ab (approved):
I *think* we should test to see if the global Var value exists first - it won't now - but this future proofs for alternate "numbering" systems.

HOWEVER - I don't see that as a blocker for approval such as my opinion counts. :)

I would also remove the blocker as that PR was closed.

dbolack-ab:
Do you have a functioning example document that shows how this might look when a user changes numbering? I'm having trouble visualizing

PR #4243, opened 2025-06-05
Switch markdown.js to use external npm version of markdown variables

dbolack-ab:
## Description

Moves the variables code to a module, makes small adjustments accordingly.

Supports #4035


dbolack-ab:
Thoroughly unsure why the module can't be seen remotely

5e-Cleric:
@calculuschild can i get a deployment here?

@dbolack-ab as long as the conflicts are solved and i can do a quick test this should go fast.

PR #4258, opened 2025-06-17
Demonstration of using markdown-extended-variables for a JSON document mail merge

dbolack-ab:
## Description

This demonstrates how marked-extended-variables adds an iteration modifier against JSON filled Brew variables.

```
[TestJson]: {"a": [{"a":"A","b":"B"},{"a":"C","b":"D"}]}

|Column 1|Column2|
|--------|-------|
+[TestJson.a]:|$a|$b|
```

image

## Usage

`+[JSON Target]:Formatting string`

`+` is used to toggle the Brew variable behavior into iteration mode.

The JSON target is the Brew variable name. Subdocuments of the JSON variable may be targeted for iteration by using dot notation. Start with the Brew variable name then append each child document as needed. In the example above, the variable is `TestJson` and the subdocument is `a`;  

The Formatting string, as of now, is a simple text string that can replace $variableName with a value from the iterated array's members.

## Assumptions.

This code assumes that the JSON Target is either an object or an array of like objects. If an Object, it is treated as a single item array. 

## Known Broken things

Math doesn't work. This code chose to make a new block output type and pass things into it than output a series of lookups and add flattened JSON targeting into the `$[var]` calls.  I will likely change that in the future.


dbolack-ab:
With working -if suboptimal - math


```

[TestJson]: {"a": [{"a":"1","b":"B"},{"a":"7","b":"D"}]}

|Column 1|Column2|
|--------|-------|
+[TestJson.a]:|$[$a*2]|$b|

$[TableNum]: 0

$[TableNum]: $[TableNum + 1]

Table $[TableNum]: Horse Type and Quality

$[1 + 3 * 5 - (1 / 4)]

```

5e-Cleric:
This is me just rambling, but considering the output, you didn't think to put the word TABLES anywhere in the PR? not the title or the description?
I would have named it `marked-variables from JSON to table`.

Again, this is just me rambling.

calculuschild:
> This is me just rambling, but considering the output, you didn't think to put the word TABLES anywhere in the PR? not the title or the description?
> I would have named it `marked-variables from JSON to table`.
> 
> Again, this is just me rambling.

I don't think the point of this PR is to generate tables. It just happens to be the example use case in the top comment.

The point is to allow importing JSON to markdown variables

dbolack-ab:
> I don't think the point of this PR is to generate tables. It just happens to be the example use case in the top comment.
> 
> The point is to allow importing JSON to markdown variables

Pretty much this.

Issue #4259, opened 2025-06-18
Crashes with a white screen


AntRomAco:
### Renderer

v3

### Browser

other

### Operating System

Windows

### What happened?

While editing a document and autosaving, the page has a high chance of crashing with a white screen. I'm using Opera GX on Windows 10.

Console logs:
Failed to decode downloaded font: https://homebrewery.naturalcrit.com/themes/V3/fonts/5e/Nodesto%20Caps%20Condensed.woff2
5eYoOujfNlzb:1 OTS parsing error: invalid sfntVersion: 1008813135
bundle.js:84126 Error Updating Local Brew
/api/update/5eYoOujfNlzb:1 
            
            
           Failed to load resource: net::ERR_CONNECTION_RESET
bundle.js:35932 TypeError: Cannot read properties of undefined (reading 'status')
    at ReactClassComponent.render (bundle.js:82698:27)
    at jj (bundle.js:35947:189)
    at hj (bundle.js:35946:149)
    at Vk (bundle.js:36035:172)
    at Uk (bundle.js:36023:389)
    at Tk (bundle.js:36023:320)
    at Ik (bundle.js:36023:180)
    at Nk (bundle.js:36014:88)
    at Gk (bundle.js:36011:429)
    at J (bundle.js:39490:203)
Li @ bundle.js:35932
bundle.js:36026 Uncaught TypeError: Cannot read properties of undefined (reading 'status')
    at ReactClassComponent.render (bundle.js:82698:27)
    at jj (bundle.js:35947:189)
    at hj (bundle.js:35946:149)
    at Vk (bundle.js:36035:172)
    at Uk (bundle.js:36023:389)
    at Tk (bundle.js:36023:320)
    at Ik (bundle.js:36023:180)
    at Nk (bundle.js:36014:88)
    at Gk (bundle.js:36011:429)
    at J (bundle.js:39490:203)
bundle.js:84126 Error Updating Local Brew
/api/update/5eYoOujfNlzb:1 
            
            
           Failed to load resource: net::ERR_CONNECTION_RESET

AntRomAco:
Also, here's the document I'm editing if anyone wants to check it:
https://homebrewery.naturalcrit.com/share/XVO3vIdQHQrT

calculuschild:
Based on the error here, it looks like it is crashing here:

[...]

PR #4280, opened 2025-06-27
Diff Saving

calculuschild:
## Description

Sends only a diff patch to the server instead of the full document text. Drastically reduces package size from client to our server, with the tradeoff that we need to now fetch the full brew contents from Mongo/Google instead of just the stub so we can apply the text diff before saving.

#4259 

Try to make the saving break. With this PR, saving should be atomic by blocking new saves until the previous completes.


ericscheid:
Does this also anticipate save-conflicts?

calculuschild:
> Does this also anticipate save-conflicts?

We still check that the brew version number matches on the server before applying a save update, so that hasn't changed. I do need to implement something on the edit page to prevent sending a new save when the previous one is still not returned. We don't want to send Diff(C, B) when Diff(B, A) may not have succeeded yet.

PR #4286, opened 2025-06-30
Add brew snippets to local save history

### LINKED ISSUE #4276: Add Snippets to save history
### Your idea:

The current user side history only handles the brew and style tabs.

dbolack-ab:
## Description

As above. Straightforward member add

Closes #4276 

### Reviewer Checklist

  - [ ] Verify snippets are added to the save bundle in  browser.


calculuschild (changes requested):
This makes sure snippets are *saved*, but I don't think it lets you *restore* snippets.

Might need to update this function as well to explicitly pull out snippets when the backup is restored:

https://github.com/naturalcrit/homebrewery/blob/b5ff26f8575ad73ce6f10968663fc7b72da22a4c/client/homebrew/pages/editPage/editPage.jsx#L187-L195

On that same note, it looks like only `text` and `style` are explicitly restored. We probably want to restore `title` as well in that function. But definitely not `savedAt` or `version` as that would lead to things being overwritten and conflicting with our server version check. `shareId` should never change so we don't need that one either.

G-Ambatte:
> This makes sure snippets are _saved_, but I don't think it lets you _restore_ snippets.

@calculuschild is correct. We could delete the undesired properties from `newData` and then spread the object in the `setState` call so any future properties that are added to the object will automatically be included in the restore.

calculuschild:
> We could delete the undesired properties from `newData` and then spread the object in the `setState` call so any future properties that are added to the object will automatically be included in the restore.

I would agree, though I do see `brew.shareId` and `brew.savedAt` are needed for some internal logic in the `versionHistory.js`, so I don't know if we can actually remove those here.

dbolack-ab:
Honestly, I hadn't spotted that functionality - I thought you only had enough implemented to go in and manually grab the lost data. :)

calculuschild:
Looks good. Merging!

Issue #3205, opened 2023-12-23
Brew language setting should reflect into og:locale, and userpage card elements


ericscheid:
### Your idea:

We can now set the language of a brew in the brew Properties editor:
<img width="807" alt="image" src="https://github.com/naturalcrit/homebrewery/assets/2605467/b8d4cbe2-b265-48df-b220-9eabb1d999eb">

This does appear in the html of the editor preview pane and the preview page:
<img width="611" alt="image" src="https://github.com/naturalcrit/homebrewery/assets/2605467/a644a480-512b-4536-98ff-0a821030f88a">

It does not however appear in the Open Graph elements for the share page:
<img width="1149" alt="image" src="https://github.com/naturalcrit/homebrewery/assets/2605467/46b7925a-f971-473c-b711-9908ed4a2ea9">

According to the OG spec, the `og:locale` defaults to `en_US`. It does not specify to assume the surrounding context (which also wouldn't matter here since the brew.language setting is applied to the `<div class="pages">` element, not the `<html>` element.

**Proposal: the brew.language property should be reflected into the open graph elements.** 
Note that it appears that the language codes of open graph use underscores, not hyphens (i.e. `en_US`, not `en-US`).

Additionally, the user page has a UI where each brew is presented in a card, showing the brew.title and brew.description.

<img width="1158" alt="image" src="https://github.com/naturalcrit/homebrewery/assets/2605467/09e1c84a-d22c-4245-8644-3f5562050b1a">

Since the `brewItem` card also contains UI elements (e.g. "1 minutes ago", plus various tooltips), the brew.language should be applied to the `h2` title element and the `<p class="description">` element only.

dbolack-ab:
This currently isn't quite a solve as we do not include territories in most of our language options.

Additionally, I cannot make heads or tails of the second proposal.

Gazook89:
If the idea is to set a locale based on language, that seems like a mistake. And I’m not sure why locale is important, like language maybe is? I kind of assume I’m missing something.

dbolack-ab:
I assume opengraph spec fulfillment. I'm just trying to kill low hanging fruit

ericscheid:
> If the idea is to set a locale based on language, that seems like a mistake. And I’m not sure why locale is important, like language maybe is?

In OG speak, `locale` is broadly synonymous with `language` (it also includes locale based formating, like numbers). 

1. `OG:locale` vs `html:lang` has some nuance to it, just to muddy the waters

[...]

PR #4287, opened 2025-06-30
Add brew locale to opengraph localization

dbolack-ab:
## Description

This attempts to solve #3205 by adding the brew's language in as a locale.

This is not entirely complete as our language choices are largely not in the *language_TERRITORY* format meaning we need to do some mapping and possibly update the languages lists to include territories where we presently do not.

- #3205

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -88,6 +88,14 @@ pre {
 ## changelog
 For a full record of development, visit our [Github Page](https://github.com/naturalcrit/homebrewery).
 
+### Wednesday 7/09/2025 - v3.19.3
+
+{{taskList
+##### calculuschild
+* [x] Restoring original saving behavior; will continue investigating why save was failing for some users in background
+}}
+
+
 ### Wednesday 7/09/2025 - v3.19.2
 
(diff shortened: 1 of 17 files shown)


server/app.js
```diff
@@ -383,6 +383,7 @@ app.get('/edit/:id', asyncHandler(getBrew('edit')), asyncHandler(async(req, res,
 		title       : req.brew.title || 'Untitled Brew',
 		description : req.brew.description || 'No description.',
 		image       : req.brew.thumbnail || defaultMetaTags.image,
+		local       : req.brew.lang,
```
calculuschild: @dbolack-ab I think this is meant to be "locale". Can you confirm?

5e-Cleric: tagging @ericscheid who reported the issue and suggested the fix to confirm too.

dbolack-ab: > @dbolack-ab I think this is meant to be "locale". Can you confirm?

It should. Pushing update.

calculuschild:
Merging!

PR #4320, opened 2025-07-10
Brew ID validations

G-Ambatte:
Given the appearance of some strange requests in the logs, I thought it best to prevent them from actually doing anything by applying some basic validations to the IDs derived from the request parameters.

Currently, the validation parameters are as follows:

- Brew ID : 7 to 14 characters in the ranges `A-Za-z0-9_-` (derived from the NanoID valid character set)
- Google ID: 33 or 44 characters in the ranges `A-Za-z0-9+/`, starting with a `1` (best analysis of existing Google IDs in the database, combined with online reports of Google ID construction)

Should NanoID or Google decide to change the nature of their IDs at any point in the future, these validations would need to be updated to reflect.

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -50,10 +50,15 @@ const api = {
 		}
 
 		// ID Validation Checks
+		// Homebrewery ID
+		// Typically 12 characters, but the DB shows a range of 7 to 14 characters
 		if(!id.match(/^[A-Za-z0-9_-]{7,14}$/)){
 			throw { name: 'ID Error', message: 'Invalid ID', status: 404, HBErrorCode: '11', brewId: id };
 		}
-		if(googleId && !googleId.match(/^1(?:[A-Za-z0-9+\/]{32}|[A-Za-z0-9+\/]{43})$/)){
+		// Google ID
+		// Typically 33 characters, old format is 44 - always starts with a 1
+		// Managed by Google, may change outside of our control, so any length between 33 and 44 is acceptable
+		if(googleId && !googleId.match(/^1(?:[A-Za-z0-9+\/]{32,43})$/)){
 			throw { name: 'Google ID Error', message: 'Invalid ID', status: 404, HBErrorCode: '12', brewId: id };
 		}
 
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -50,7 +50,7 @@ const api = {
 		}
 
 		// ID Validation Checks
-		if(!id.match(/^[A-Za-z0-9_-]{7,12}$/)){
+		if(!id.match(/^[A-Za-z0-9_-]{7,14}$/)){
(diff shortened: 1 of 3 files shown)


G-Ambatte:
Analysis of data in MongoDB shows that there are entries in the database for values of `editId` and `shareId` that are between 7 and 12 characters long, so I have adjusted the validation regex to reflect.

G-Ambatte:
Further analysis has revealed the ID is between 7 and 14 characters.

server/homebrew.api.js
```diff
@@ -48,6 +48,15 @@ const api = {
...
 		}
+
+		// ID Validation Checks
+		if(!id.match(/^[A-Za-z0-9_-]{7,14}$/)){
+			throw { name: 'ID Error', message: 'Invalid ID', status: 404, HBErrorCode: '11', brewId: id };
+		}
+		if(googleId && !googleId.match(/^1(?:[A-Za-z0-9+\/]{32}|[A-Za-z0-9+\/]{43})$/)){
+			throw { name: 'Google ID Error', message: 'Invalid ID', status: 404, HBErrorCode: '12', brewId: id };
+		}
+
```
calculuschild: Could you just add some comments to clarify the regex? 

```
// Homebrewery ID (nanoid): 7–14 characters, alphanumeric (and _-)
// Google Drive File ID: starts with '1', followed by 32 or 43 alphanumeric (and +/)
```

Just in case something changes in the future and we need to decode what happened.

server/homebrew.api.js
```diff
@@ -48,6 +48,15 @@ const api = {
 			}
 			id = id.slice(googleId.length);
 		}
+
+		// ID Validation Checks
+		if(!id.match(/^[A-Za-z0-9_-]{7,14}$/)){
+			throw { name: 'ID Error', message: 'Invalid ID', status: 404, HBErrorCode: '11', brewId: id };
+		}
+		if(googleId && !googleId.match(/^1(?:[A-Za-z0-9+\/]{32}|[A-Za-z0-9+\/]{43})$/)){
```
calculuschild: I do have very slight wariness on specifying exact character lengths here for Google. I guess if it changes we would know pretty quickly with a bunch of user reports. But I'm thinking maybe we don't want to subject ourselves to a panic moment like that if we just loosen this up a little to just > 33 chars.

I will let you make the call on that.

G-Ambatte: I suspect that it would be too loose a definition if it doesn't include a maximum length - perhaps something like `/^1(?:[A-Za-z0-9+\/]{32,43})$/` which will allow any number of characters between 33 and 44?

The new error types will very quickly identify the issue, should a genuine Google ID fail.

G-Ambatte:
While working on some additional test cases this morning, I realized that a brew ID can never be more than 12 characters long, because the `getId` function will always split it into a Google ID and a Brew ID. So the brews that have IDs that are 13 and 14 characters long can never be reached.

calculuschild:
Looks good. Thanks!

PR #4333, opened 2025-07-13
Add forceSSL middleware tests

G-Ambatte:
This PR adds tests for the Force SSL middleware, bring coverage testing for this file to 100%.

image

### DIFF
--- a/server/forcessl.mw.spec.js
+++ b/server/forcessl.mw.spec.js
@@ -1,0 +1,74 @@
+import forceSSL from './forcessl.mw';
+
+describe('Tests for ForceSSL middleware', ()=>{
+
+	it('should call next() when NODE_ENV is set to local', ()=>{
+		const nodeEnv = process.env.NODE_ENV;
+		process.env.NODE_ENV = 'local';
+
+		const nextFn = jest.fn();
+
+		forceSSL(null, null, nextFn);
+
+		process.env.NODE_ENV = nodeEnv;
+
+		expect(nextFn).toHaveBeenCalled();
+	});
+
+	it('should call next() when NODE_ENV is set to docker', ()=>{
+		const nodeEnv = process.env.NODE_ENV;
+		process.env.NODE_ENV = 'docker';
+
+		const nextFn = jest.fn();
+
+		forceSSL(null, null, nextFn);
+
+		process.env.NODE_ENV = nodeEnv;
+
+		expect(nextFn).toHaveBeenCalled();
+	});
+
+	it('should return 302 when header not HTTPS', ()=>{
+		const nodeEnv = process.env.NODE_ENV;
+		process.env.NODE_ENV = 'test';
+
+		const req = {
+			header : ()=>{ return true; },
+			get    : ()=>{ return 'test'; },
+			url    : 'URL'
+		};
+
+		const res = {
+			redirect : jest.fn((code, url)=>{})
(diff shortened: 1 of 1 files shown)


calculuschild (changes requested):
I see some common setup/teardown steps here that we can probably move out. I propose this:

```

import forceSSL from './forcessl.mw';

describe('Tests for ForceSSL middleware', () => {
    let originalEnv;
    let nextFn;

    beforeEach(() => {
        originalEnv = process.env.NODE_ENV;
        nextFn = jest.fn();
    });

    afterEach(() => {
        process.env.NODE_ENV = originalEnv;
        jest.clearAllMocks();
    });

    it('should call next() when NODE_ENV is set to local', () => {
        process.env.NODE_ENV = 'local';
        forceSSL(null, null, nextFn);
        expect(nextFn).toHaveBeenCalled();
    });

    it('should call next() when NODE_ENV is set to docker', () => {
        process.env.NODE_ENV = 'docker';
        forceSSL(null, null, nextFn);
        expect(nextFn).toHaveBeenCalled();
    });

    it('should return 302 when header not HTTPS', () => {
        process.env.NODE_ENV = 'test';

        const req = {
            header: () => null,
            get: () => 'test',
            url: 'URL'
        };

        const res = {
            redirect: jest.fn()
        };

        forceSSL(req, res, nextFn);
        expect(res.redirect).toHaveBeenCalledWith(302, 'https://testURL');
    });

    it('should call next() when header is HTTPS and NODE_ENV is not local or docker', () => {
        process.env.NODE_ENV = 'test';

        const req = {
            header: () => 'https'
        };

        forceSSL(req, {}, nextFn);
        expect(nextFn).toHaveBeenCalled();
    });
});
```

Secondly, I we are basically testing that we redirect to https:// only if we are not at https:// already, but the description of "calls next" is kind of vague. I would suggest naming these more explicitly according to what that logic path means. i.e.:

- should not redirect to https:// if environment is 'local'
- should not redirect to https:// if header is already HTTPS.
- should redirect to https:// if header is NOT already HTTPS

etc.

And then in the tests, checking that `next()` was called only confirms that the middleware ended. It is probably more meaningful to check that `res.redirect` was *not* called in those cases. `expect(res.redirect).not.toHaveBeenCalled();`

G-Ambatte:
I've been doing some light reading since I wrote this a day or two ago, and the NODE_ENV = 'test' may be unnecessary, as it appears that Jest already sets this - see https://jestjs.io/docs/environment-variables

I also originally wrote this under the impression that `next()` would NOT be called if the middleware had thrown an error, so was using it as a psuedo-success indicator. Unfortunately, further reading has shown that this is not the case - see https://expressjs.com/en/guide/writing-middleware.html, specifically the last line of the Overview paragraph:

> Starting with Express 5, middleware functions that return a Promise will call next(value) when they reject or throw an error. next will be called with either the rejected value or the thrown Error.

With these new understandings, I'll take another run at the tests.

---

> I've been doing some light reading since I wrote this a day or two ago, and the NODE_ENV = 'test' may be unnecessary, as it appears that Jest already sets this - see https://jestjs.io/docs/environment-variables

It appears that this is only the case if NODE_ENV is not already set; as this is set permanently on my development computer, I will continue to set it explicitly.

calculuschild:
Looks good! Thanks for the changes.

PR #4335, opened 2025-07-13
Fix test option 1 - Shift splitTextStyleAndMetadata to after version check

G-Ambatte:
This PR fixes the current test failures by moving the `splitTextStyleAndMetadata` function to after version check.

The test is currently failing because, as currently written, the test object does not have a `text` property - as the version check occurred before the `splitTextStyleAndMetadata` function call at the time it was written, the `text` property was never accessed and thus was not needed on the test object.
This PR returns the `splitTextStyleAndMetadata` to the same location in the code as when the test was written.

However, if this function call needs to remain before the version check, this PR can be closed in favour of the second PR (#4336).

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -339,7 +339,6 @@ const api = {
 		// Initialize brew from request and body, destructure query params, and set the initial value for the after-save method
 		const brewFromClient = api.excludePropsFromUpdate(req.body);
 		const brewFromServer = req.brew;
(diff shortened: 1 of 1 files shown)


G-Ambatte:
Leaving this at P2, as it doesn't affect actual server functionality; I was considering P1 because the current test issue does affect all other PRs.

calculuschild:
Closing in favor of #4336

PR #4336, opened 2025-07-13
Fix test option 2 - Add text property to version check test object

G-Ambatte:
This PR fixes the current test failures by adding the `text` property to the test object used in the version check.

The test is currently failing because, as currently written, the test object does not have a text property - as the version check occurred before the splitTextStyleAndMetadata function call at the time it was written, the text property was never accessed and thus was not needed on the test object.
This PR adds the text to the object passed for the test, which stops `splitTextStyleAndMetadata` from throwing an error during testing.

However, if this function call can be moved to after the version check, this PR can be closed in favour of the first PR (#4335).

### DIFF
--- a/server/homebrew.api.spec.js
+++ b/server/homebrew.api.spec.js
@@ -1056,7 +1056,7 @@ brew`);
 	describe('updateBrew', ()=>{
 		it('should return error on version mismatch', async ()=>{
 			const brewFromClient = { version: 1 };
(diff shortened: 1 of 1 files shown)


calculuschild:
Let's go with this one. I moved it up to enable some logging inside the version check. I have since removed those logs, but they may come back if we continue investigating diffing solutions.

PR #4338, opened 2025-07-14
Increase safeHTML testing to 100% coverage

G-Ambatte:
This PR adds an additional test to the safeHTML tests, which brings the coverage for this file to 100%.

image


calculuschild:
Seems good. Thanks!

PR #4348, opened 2025-07-19
Add indexes to HomebrewSchema

G-Ambatte:
This PR adds the indexes currently in use in the Production database to the HomebrewSchema file. It also modifies the DB connection string so that these indexes are automatically built when Homebrewery is started with the NODE_ENV is set to an option from the `local_environments` configuration setting - at time of writing, these default to `local` or `docker`.

### DIFF
--- a/server/db.js
+++ b/server/db.js
@@ -29,7 +29,7 @@ const disconnect = async ()=>{
 const connect = async (config)=>{
 	return await Mongoose.connect(getMongoDBURL(config), {
 		retryWrites : false,
-		autoIndex   : (config.get('node_env') == 'local' || config.get('node_env') == 'docker')
+		autoIndex   : (config.get('local_environments').includes(config.get('node_env')))
 	})
 		.catch((error)=>handleConnectionError(error));
 };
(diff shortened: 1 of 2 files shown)


G-Ambatte:
At this stage, I have kept this PR marked as a Draft to ensure that no staging environment is automatically created for it. While it _should_ never affect the indexes in the Production database, I think it best to get a second set of eyes on it before upgrading to full PR status.

G-Ambatte:
Index state before starting the server, with `NODE_ENV = 'production'`:
image

Index state after starting the server, `NODE_ENV = 'production'`:
image

Index state after starting the server, `NODE_ENV = 'local'`:
image

5e-Cleric:
> At this stage, I have kept this PR marked as a Draft to ensure that no staging environment is automatically created for it. While it _should_ never affect the indexes in the Production database, I think it best to get a second set of eyes on it before upgrading to full PR status.

A deployment will not be automatically created if you promote it to branch, that only happens when the branch is in the main repo, the rest are manually deployed as far as i know.

G-Ambatte:
> A deployment will not be automatically created if you promote it to branch, that only happens when the branch is in the main repo, the rest are manually deployed as far as i know.

With this assurance, I'll push to full PR.

G-Ambatte:
I'll open this PR again after a fresh pass over the code; possibly as soon as tomorrow.

PR #4352, opened 2025-07-22
Add Google service account to file permissions

G-Ambatte:
Currently, when Homebrewery creates Google files, it sets them to "Anyone with the link can edit".
This PR **also** grants the Homebrewery service account access to the file as a `writer`, with the expectation that we will be able to tighten the restrictions further once we have confirmed the permissions are added correctly.

The expectation is that this will not change the functionality at all, but will allow us to improve security on the Homebrewery Google files.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@babel/preset-react": "^7.28.5",
         "@babel/runtime": "^7.28.4",
         "@dmsnell/diff-match-patch": "^1.1.0",
-        "@googleapis/drive": "^18.0.0",
+        "@googleapis/drive": "^19.2.0",
         "@sanity/diff-match-patch": "^3.2.0",
         "body-parser": "^2.2.0",
         "classnames": "^2.5.1",
@@ -2295,9 +2295,9 @@
       }
     },
     "node_modules/@googleapis/drive": {
-      "version": "18.0.0",
-      "resolved": "https://registry.npmjs.org/@googleapis/drive/-/drive-18.0.0.tgz",
-      "integrity": "sha512-nk4TirsHttwZSOjBEBjltCPDKUqwFso59G3WitNE+EGNVSVseSEq981f8Dmjq2ah0/fk3i206wuCU4PUCwcoTQ==",
+      "version": "19.2.0",
+      "resolved": "https://registry.npmjs.org/@googleapis/drive/-/drive-19.2.0.tgz",
+      "integrity": "sha512-XzhQ+CNwW54umLecZLzQ5Q2RborJD1+u2YzJIT/BOpne+VJfsCyxWVG0nxYQmcGdiFXJPeW5+V3sdP/A/nt3eA==",
       "license": "Apache-2.0",
       "dependencies": {
         "googleapis-common": "^8.0.0"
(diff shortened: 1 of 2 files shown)


5e-Cleric:
Should this change break anything? why are we not merging if not?

G-Ambatte:
> Should this change break anything? why are we not merging if not?

This change does not remove any of the current permissions, but adds permission for the Homebrewery service account to the created file.

G-Ambatte:
With this change, a new file on Google should have three permissions: the author (owner), Homebrewery service (editor), and everyone (editor).

With this set up, if the Everyone permission is modified, the Homebrewery service account should still be able to see the document.
Both permissions would need to be revoked in order to lose access.

G-Ambatte:
I have created a new brew using the deployment and saved it to my Google Drive. The file has the following permissions set on creation:

image

5e-Cleric:
So, my question still stands unanswered, why are we not merging if this doesn't break anything?

server/app.js
```diff
@@ -330,7 +333,7 @@ app.get('/user/:username', async (req, res, next)=>{
 			}
 
 			//Remaining unstubbed google brews display current user as author
-			googleBrews = googleBrews.map((brew)=>({ ...brew, authors: [req.account.username] }));
+			googleBrews = googleBrews.map(async (brew)=>({ ...brew, authors: [req.account.username] }));
```
calculuschild: Any particular reason this map function has become async?

server/app.js
```diff
@@ -318,6 +318,9 @@ app.get('/user/:username', async (req, res, next)=>{
 		// If stub matches file from Google, use Google metadata over stub metadata
 		if(googleBrews && googleBrews.length > 0) {
 			for (const brew of brews.filter((brew)=>brew.googleId)) {
+				const permissionCheck = await GoogleActions.checkPermissions(auth, brew);
+				brew.permissionCheck = permissionCheck;
```
calculuschild: I don't see that this `permissionCheck` property gets used anywhere. Is it needed?

server/googleActions.js
```diff
@@ -241,6 +264,26 @@ const GoogleActions = {
 		return obj.data.id;
 	},
 
+	checkPermissions : async (auth, brew)=>{
```
calculuschild: Following the comment above, I don't see that the output from this `checkPermissions` is used for anything.

server/app.js
G-Ambatte: None that I currently recall, at least. I'll remove this from the PR.

server/app.js
G-Ambatte: I believe that is a hanging artifact from a different Google action that I was working on - as the User's auth is used to get the list of brews for the UserPage, I believe this was from a branch where that auth was also used to get the permissions on the file. The list of permissions could then be parsed for to check that `everyone` had the `writer` permission.

It's unnecessary and should not be in this PR, so I will remove it. The Permissions Check will have it's own PR and should not have bled into this one.

server/googleActions.js
G-Ambatte: As per previous. I'll remove this as well.

G-Ambatte: This is resolved in the latest commit.

server/app.js
G-Ambatte: This should be removed in the latest commit.

5e-Cleric (approved):
Looks good, will leave this one for calc to merge, because i am not sure about it.

G-Ambatte:
I need to double check this one once I'm back in front of my Dev machine, I was having issues checking the latest changes on the deployment. Should have an update on the next 24 hours or so.

G-Ambatte:
~~Setting this one to `blocked` until #4526 is merged.~~

Resolved, `blocked` tag removed.

server/googleActions.js
```diff
@@ -227,9 +232,27 @@ const GoogleActions = {
...
+				resource : {
+					type         : 'user',
+					emailAddress : clientEmail,
+					role         : 'writer'
+				},
+				fileId : obj.data.id,
+				fields : 'id',
+			})
+			.catch((err)=>{
+				console.log('Error updating permissions');
```
Gazook89: Is there more that could be done with this error, more description?  Even amending to "Error updating permissions on Google Drive account".  Just nice to get as much information as possible in the console, for when this error happens and it's a few years from now and we aren't 100% sure what the cause is anymore.

G-Ambatte: Updated error message to `Error updating permissions on Google Drive file`

5e-Cleric (approved):

5e-Cleric:
So many approved reviews! merging! thanks @G-Ambatte 🎉

PR #4363, opened 2025-07-26
Add database connection check middleware

G-Ambatte:
This PR adds middleware that checks that Mongoose database connection is in the `connected` state before processing requests. If the connection is not `1`, a basic error message is returned.

From https://mongoosejs.com/docs/api/connection.html#Connection.prototype.readyState, the potential states are as follows:

    0 = disconnected
    1 = connected
    2 = connecting
    3 = disconnecting

This PR also adds console messages for database connectivity events.

---

While MongoDB is running during normal operation, everything continues as normal.
image

Current result when DB is unavailable:
image
image

New results when DB is unavailable:
image
image

---

This PR also adds tests to ensure that a 503 page is returned if the `readyState` is set to `99`, and also that `next()` is called if `readyState` is set to `1`.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -32,6 +32,7 @@ import { splitTextStyleAndMetadata } from '../shared/helpers.js';
 
 //==== Middleware Imports ====//
 import contentNegotiation from './middleware/content-negotiation.js';
+import dbCheck from './middleware/dbCheck.js';
 import bodyParser         from 'body-parser';
 import cookieParser       from 'cookie-parser';
 import forceSSL           from './forcessl.mw.js';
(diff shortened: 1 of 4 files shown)


G-Ambatte:
I changed the result on error, so that instead of returning an error message, the user is redirected to the normal error page.

If the user attempts to refresh the page, they now sees the following on DB Connection error:
image

If the user attempts to make an API call (e.g. save a brew from Edit Page), they see this notification:
image

Server logs show the following:
image

calculuschild:
Is this a known problem? Can you make an issue for it please?

G-Ambatte:
> Is this a known problem? Can you make an issue for it please?

Issue #4412 raised and linked to this PR.

G-Ambatte:
Closing this PR now, I'll open a new PR after reviewing the state of the branch.

PR #4368, opened 2025-07-30
Fix Account page FA icon font weights

G-Ambatte:
This PR changes the `font-weight` property to `900`, which should correct the broken icons.


dbolack-ab (approved):
Confirmed fix.

PR #4371, opened 2025-07-30
Keep the cursor in the active view page after a divider width change.

### LINKED ISSUE #2963: Resizing the editor causes the current cursor position to leave the view
### Your idea:

Noted by someone in the DOMT discord, when you resize the editor pane, the current cursor position moves out of the editor visual viewport.  In smaller brews, it's not much of a problem.  But as the brew length increases, and the 'thumb' of the scrollbar gets smaller, even a little bit of resizing can cause a user to lose their place.

In my opinion, the ideal is keep the current position in the visual viewport constant.  This would mean that when you resize you can then continue editing as if you had not done the resize at all.

dbolack-ab:
## Description

This captures the cursor position when the scroll bars begin moving and restores it when done.

The code makes a best effort to keep the cursor in about the same place on the screen.

- Closes #2963 

Find a largish brew, place the cursor on a line near the top, bottom, or middle. Resize the scroll bar. When done resizing, the cursor should be placed at the same line number and character. The cursor's Y position in the frame should be roughly the same place on the screen as before.

### Reviewer Checklist

  - [ ] Test with the cursor near the top of frame. Resize the frame so the cursor slides above the viewable pane. Verify the  cursor placement after resizing is on the correct line of text, correct character, and same approximate space on the view pane.
  - [ ] Test with the cursor near the bottom of frame. Resize the frame so the cursor slides below the viewable pane. Verify the  cursor placement after resizing is on the correct line of text, correct character, and same approximate space on the view pane.
  - [ ] Test with the cursor near the middle of the frame. Resize the frame so the cusor moves significantly. Verify the  cursor placement after resizing is on the correct line of text, correct character, and same approximate space on the view pane.


calculuschild:
Ok, just playing around.... It looks like the reason for the sudden jump when you let go of the divider bar in the first place, is a call we make to `this.codeEditor.current?.updateSize();` which in turn calls `codeMirror.refresh();`. If I just comment that out, we don't have the weird jumping to begin with, and Codemirror seems to be able to hold onto the current view position just fine.

Is that a better solution?

5e-Cleric:
> Ok, just playing around.... It looks like the reason for the sudden jump when you let go of the divider bar in the first place, is a call we make to `this.codeEditor.current?.updateSize();` which in turn calls `codeMirror.refresh();`. If I just comment that out, we don't have the weird jumping to begin with, and Codemirror seems to be able to hold onto the current view position just fine.
> 
> Is that a better solution?

would love to know why we call that in the first place.

PR #4379, opened 2025-08-05
Inline SVG proof of concept

dbolack-ab:
## Description

This inserts all non-local SVGs as inline HTMLElements


5e-Cleric:
@calculuschild can we get a deployment of this, please?

Issue #4119, opened 2025-03-25
Standardize cookie and storage key names


5e-Cleric:
As per right now, our localStorage and cookie key names are as follows:

- `nc_session`

Localstorage keys:
- `AUTOSAVE_ON`
- `HOMEBREWERY-DEFAULT-SAVE-LOCATION-username`
- `HOMEBREWERY-EDITOR-THEME`
- `HOMEBREWERY-LISTPAGE-SORTDIR`
- `HOMEBREWERY-LISTPAGE-SORTTYPE`
- `HOMEBREWERY-LISTPAGE-VISIBILITY`
- `HOMEBREWERY-LISTPAGE-VISIBILITY`
- `dismiss_notif_drive`
- `dismiss_notif_faq`
- `dismiss_notif_update_v3.18`
- `dismiss_v134 update`
- `hbAdminTab`
- `homebrewery-new`
- `homebrewery-new-meta`
- `homebrewery-new-style`
- `homebrewery-recently-edited`
- `homebrewery-recently-viewed`
- `liveScroll`
- `naturalcrit-pane-split`

I think we should pick one syntax and keep it strict, i suggest:

`HB_name`

In that format, some would look like:

- `HB_default_save_location_username`
- `HB_liveScroll`
- `HB_autosave`

We should avoid "unsigned" names like `livescroll`, but we also don't need names so long like the save location.

Would love to hear everyone's opinion on this as well as ideas on how to fix this for users without bloating the codebase with checks for every time a user enters the site.

calculuschild:
> Would love to hear... ideas on how to fix this for users without bloating the codebase with checks for every time a user enters the site.

Unfortunately the best I can think of is a centralized "check the user's localstorage and fix any discrepancies" call that runs when they open a page. We leave it for a few months, and then remove it when we feel somewhat confident nobody is going to lose important localstorage data.

G-Ambatte:
There was, at one point, discussion of shifting user preferences from local storage to live in MongoDB, so as to ensure a consistent experience across devices. Rather than standardizing the local storage key names to different local storage keys, we could instead pursue the centralization of user setting configuration data into MongoDB.

dbolack-ab:
> There was, at one point, discussion of shifting user preferences from local storage to live in MongoDB, so as to ensure a consistent experience across devices. Rather than standardizing the local storage key names to different local storage keys, we could instead pursue the centralization of user setting configuration data into MongoDB.

That would probably be a better choice.

[...]

PR #4380, opened 2025-08-06
Create localstorage access wrappers to transition to standardized naming scheme

dbolack-ab:
## Description

This PR creates two PRs and a map in the helpers.js for transitioning the rather haphzardly named keys to a standardized scheme.

As proof of concept/implementation, the liveScroll key is updated.

#4119 

### Reviewer Checklist

  - [ ] Examine the localStorage for the test instance, manually set liveScroll to true
  - [ ] Reload the page, see that liveScroll has been removed and HB_livesScroll is now true
  - [ ] Verify scrolllock toggle is still preserved between loads.

### DIFF
--- a/client/components/splitPane/splitPane.jsx
+++ b/client/components/splitPane/splitPane.jsx
@@ -22,7 +22,7 @@ const SplitPane = (props)=>{
 	useEffect(()=>{
 		const savedPos = localStorageGetWrapper(window, storageKey);
 		setDividerPos(savedPos ? limitPosition(savedPos, 0.1 * (window.innerWidth - 13), 0.9 * (window.innerWidth - 13)) : window.innerWidth / 2);
-		setLiveScroll(localStorageGetWrapper(window, 'liveScroll') === 'true');
+		setLiveScroll(localStorageGetWrapper(window, 'HB_liveScroll') === 'true');
 
 		window.addEventListener('resize', handleResize);
 		return ()=>window.removeEventListener('resize', handleResize);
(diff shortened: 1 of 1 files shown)


5e-Cleric:
Do not take this as an attack but i don't see how this solves anything. The name is still arbitrarily defined in a file.

dbolack-ab:
> Do not take this as an attack but i don't see how this solves anything. The name is still arbitrarily defined in a file.

In what scenario would they not be arbitrarily named in a file? I don't understand this question.

This PR establishes a straightforward arrangement for transitioning current named values to any future standardized scheme that may be required.

calculuschild:
Unfortunately I think this starts to add way too many hooks into places all over the code. And this is just for the splitPane.jsx. Imagine implementing all of those `logalStorageGetWrapper` into *every* function call where localstorage is read.

I know @5e-Cleric said this in the original issue:
> Would love to hear... ideas on how to fix this for users without bloating the codebase with checks for every time a user enters the site.

But I think a centralized "check the user's localstorage and fix any discrepancies" call would be better than multiple calls to a wrapper all over the code.

G-Ambatte:
> a centralized "check the user's localstorage and fix any discrepancies" call

`homebrew.jsx` is where we set up the React Router, it's literally the first page hit on the front end which is why it's where we set up the globals. I'm pretty sure we could call such a function from there and it would check/update the keys with every page load - assuming the function removes the old keys after transitioning the content, there would be no keys left to check after the first execution.

PR #4414, opened 2025-09-09
Centralize updating key names

### LINKED ISSUE #4119: Standardize cookie and storage key names

G-Ambatte:
This PR will resolve #4119 .

This PR adds a function to `homebrew.jsx` which checks a map object for any local storage entries with a matching name. When found, the key data is copied to the new name provided by the map object.

- [x] new keys are created
  - [x] new keys are populated with data from old keys
- [x] all old key references are changed to the relevant new key
- [ ] old keys are deleted from local storage

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,8 +75,9 @@ it using the two commands:
 1. `npm install`
 1. `npm start`
 
-You should now be able to go to [http://localhost:8000](http://localhost:8000)
-in your browser and use The Homebrewery offline.
+When the Homebrewery server is started for the first time, it will modify the database to create the indexes required for better Homebrewery performance. This may take a few moments to complete for each index, dependent on how much content is in your local database - a brand new, empty database should be done in seconds.
+
+On completion, you should be able to go to [http://localhost:8000](http://localhost:8000) in your browser and use The Homebrewery offline.
 
 If you had any issue at all, here are some links that may be useful:
 - [Course](https://learn.mongodb.com/courses/m103-basic-cluster-administration) on cluster administration, useful for beginners
@@ -144,4 +145,5 @@ your contribution to the project, please join our [gitter chat][gitter-url].
 [github-mark-duplicate-url]: https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/about-duplicate-issues-and-pull-requests
 [github-pr-docs-url]: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request
 [gitter-url]: https://gitter.im/naturalcrit/Lobby
+
 
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -40,11 +40,8 @@ const Editor = createClass({
 				style : ''
 			},
 
-			onTextChange  : ()=>{},
-			onStyleChange : ()=>{},
-			onMetaChange  : ()=>{},
-			onSnipChange  : ()=>{},
-			reportError   : ()=>{},
+			onBrewChange : ()=>{},
+			reportError  : ()=>{},
 
 			onCursorPageChange : ()=>{},
 			onViewPageChange   : ()=>{},
@@ -438,7 +435,7 @@ const Editor = createClass({
 					language='gfm'
 					view={this.state.view}
 					value={this.props.brew.text}
-					onChange={this.props.onTextChange}
+					onChange={this.props.onBrewChange('text')}
 					editorTheme={this.state.editorTheme}
 					rerenderParent={this.rerenderParent}
 					style={{  height: `calc(100% - ${this.state.snippetbarHeight}px)` }} />
@@ -451,7 +448,7 @@ const Editor = createClass({
 					language='css'
 					view={this.state.view}
 					value={this.props.brew.style ?? DEFAULT_STYLE_TEXT}
-					onChange={this.props.onStyleChange}
+					onChange={this.props.onBrewChange('style')}
 					enableFolding={true}
 					editorTheme={this.state.editorTheme}
 					rerenderParent={this.rerenderParent}
@@ -467,7 +464,7 @@ const Editor = createClass({
 				<MetadataEditor
 					metadata={this.props.brew}
 					themeBundle={this.props.themeBundle}
-					onChange={this.props.onMetaChange}
+					onChange={this.props.onBrewChange('metadata')}
 					reportError={this.props.reportError}
 					userThemes={this.props.userThemes}/>
 			</>;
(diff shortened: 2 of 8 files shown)


client/homebrew/utils/updateLocalStorage/updateLocalStorageKeys.js
```diff
@@ -0,0 +1,22 @@
+import localStorageKeyMap from './localStorageKeyMap.json' with { type: 'json' };
+
+const updateLocalStorage = function(){
+	// Return if no window and thus no local storage
+	if(typeof window === 'undefined') return;
+
+	// Return if the local storage key map has no content
+	if(Object.keys(localStorageKeyMap).length == 0) return;
```
calculuschild, line 9: Why would this ever be true? Aren't we importing this right at the top?

calculuschild:
I think this is a fine way to do this. If you want to go ahead and delete the old keys from localstorage (so next page load we don't scan over the old key again and overwrite our good value with a stale entry), and change the other pages to use the new names I approve this.

G-Ambatte:
> I think this is a fine way to do this.

There is one key which is constructed from dynamic data: `HOMEBREWERY-DEFAULT-SAVE-LOCATION-username`. In order to accommodate this particular key, the approach might need to change slightly - possibly functions to return the original and new names to construct the dynamic key name.

EDIT: JSON can't store functions... but we can probably keep the static keys in JSON, and add the dynamic key(s) in a function before processing.

> If you want to go ahead and delete the old keys from localstorage (so next page load we don't scan over the old key again and overwrite our good value with a stale entry), and change the other pages to use the new names I approve this.

Can do.

client/homebrew/utils/updateLocalStorage/updateLocalStorageKeys.js
G-Ambatte: It _should_ never be true. However, as we're importing a JSON file which may have different content at some point in the future, I thought I should probably include an escape hatch for empty files.

G-Ambatte:
I've added a check before the local storage key is updated so it will no longer be updated if the key already exists. I have not yet uncommented the line in order to activate the functionality to remove the existing keys.

calculuschild:
> EDIT: JSON can't store functions... but we can probably keep the static keys in JSON, and add the dynamic key(s) in a function before processing.

We could just make it a .js file and hold a normal javascript object. Kind of like we do with our `snippets.js` dropdown menu definitions.

G-Ambatte:
> We could just make it a .js file and hold a normal javascript object. Kind of like we do with our `snippets.js` dropdown menu definitions.

I've changed the JSON file to a JS file that exports a `getLocalStorageMap` function, so all of the code that defines the old and new names of the local storage keys now exists in a single file.

---

**BUG SQUASHING:** I've noticed that saving from the New Page does not currently clear the content. After some investigation, it appears that what is actually happening is that the keys ARE cleared, however when the page loads on the Edit Page, the absent keys are repopulated from the old key names (because the new keys no longer exist).
So this issue will be resolved once we activate the key deletion.

client/homebrew/utils/updateLocalStorage/localStorageKeyMap.js
```diff
@@ -0,0 +1,35 @@
...
+		'homebrewery-recently-viewed' : 'HB_nav_recentlyViewed',
+
+		'hb_toolbarState'      : 'HB_renderer_toolbarState',
+		'hb_toolbarVisibility' : 'HB_renderer_toolbarVisibility'
+	};
+
+	if(global?.account?.username){
+		const username = global.account.username;
+		 localStorageMap[`HOMEBREWERY-DEFAULT-SAVE-LOCATION-${username}`] = `HB_editor_defaultSave_${username}`;
+	}
```
calculuschild, line 30: Do we need to also map the case where there is no user logged in? On the newPage.jsx it would have been: "HOMEBREWERY-DEFAULT-SAVE-LOCATION-''

Maybe this is ok, because I can't think of a case where this "no-username" key would actually be created, since it is only set from a logged-in user's accountPage.

client/homebrew/utils/updateLocalStorage/updateLocalStorageKeys.js
```diff
@@ -0,0 +1,23 @@
...
+	const localStorageKeyMap = getLocalStorageMap();
+	if(Object.keys(localStorageKeyMap).length == 0) return;
+
+	const storage = window.localStorage;
+
+	Object.keys(localStorageKeyMap).forEach((key)=>{
+		if(storage[key] && !storage[localStorageKeyMap[key]]){
+			const data = storage.getItem(key);
+			storage.setItem(localStorageKeyMap[key], data);
+			// storage.removeItem(key);
```
calculuschild, line 17: Just make sure to uncomment before we merge.

client/homebrew/utils/updateLocalStorage/localStorageKeyMap.js
G-Ambatte: If there is no user is logged in, then they cannot possibly be linked to a Google account, so the only option is to save to Homebrewery.

calculuschild:
Do you want to close and reopen this for hacktoberfest?

client/homebrew/utils/updateLocalStorage/localStorageKeyMap.spec.js
calculuschild, line 1: I'm not sure we need testing for this one; feels a little much since it's just a find-and-replace of localstorage keys, (not validating any API or Markdown spec) and these tests are a lot of lines for what amounts to just checking the size of the mapping array.

Was CircleCI complaining about code coverage?

G-Ambatte: I was trying to keep ahead of the coverage testing complaints. I'll push a new PR soon that handles the updated pages.

G-Ambatte:
Closing this PR in favor of an updated PR that handles the new pages correctly.

PR #4430, opened 2025-09-27
Clean Up Common features of new/home/edit

calculuschild:
## Description

Just further cleanup and getting common names and structures set up between all three pages while keeping the unique bits functional.


5e-Cleric (approved):
So much for keeping PRs small..

Looks good to me.

PR #4434, opened 2025-10-01
Fix dev background

5e-Cleric:
For some reason when the developing background was added, it was added just for the edit page and new, not for userpage, vault, or account page.

This PR moves the background from the brewRenderer up into homebrew.jsx


client/homebrew/homebrew.jsx
```diff
@@ -42,15 +41,35 @@ const Homebrew = (props)=>{
...
 	global.enable_themes = enable_themes;
 	global.config        = config;
 
+	const backgroundObject = ()=>{
+		if(!isClient) return null;
+		if(config.deployment) {
+			return {
+				backgroundImage : `url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='100px' width='200px'><text x='0' y='15' fill='%23fff7' font-size='20'>${config.deployment}</text></svg>")`
+			};
+		} else if(config.local) {
```
G-Ambatte: More of a style comment than actual code functionality: `else` is not required here. As the preceding `if` will `return` out of the function when true, anything after the first `if` completes is effectively an `else` condition.

client/homebrew/homebrew.jsx
```diff
@@ -42,15 +41,35 @@ const Homebrew = (props)=>{
...
 	global.version       = version;
 	global.enable_v3     = enable_v3;
 	global.enable_themes = enable_themes;
 	global.config        = config;
 
+	const backgroundObject = ()=>{
+		if(!isClient) return null;
+		if(config.deployment) {
+			return {
+				backgroundImage : `url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='100px' width='200px'><text x='0' y='15' fill='%23fff7' font-size='20'>${config.deployment}</text></svg>")`
```
G-Ambatte: Potentially a DRY issue... What do you think of something like this?
```
if(!isClient) return null;
if(config.deployment || config.local){
  const bgText = config.deployment || 'Local';
  return {
    backgroundImage : `url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='100px' width='200px'><text x='0' y='15' fill='%23fff7' font-size='20'>${bgText}</text></svg>")`
  };
}
return null;
}
```

G-Ambatte (approved):
As this PR is fully functional in the deployment; comments provided are suggestions for alternative implementations of the same functionality, so ultimately stylistic rather than functionality.

It's a good change to cover use cases that I didn't consider when implementing the original functionality.

client/homebrew/homebrew.jsx
5e-Cleric: Hey, the original code was yours, if i recall, i just took it to the correct place, but, yeah, i will give this a go

client/homebrew/homebrew.jsx
```diff
@@ -42,15 +41,32 @@ const Homebrew = (props)=>{
 		brews
 	} = props;
 
+	const [isClient, setIsClient] = useState(false);
+
+	useEffect(()=>{
+		setIsClient(true);
+	}, []);
+
```
calculuschild: Two comments related to `isClient`:

1) Can we just use `typeof window !== undefined` instead of this whole state/useEffect thing?
2) Why do we care if we are in a client? Can we just remove that altogether?

5e-Cleric: if i remember correctly, the following function runs before the client is created, and will throw errors. UseEffect updates this meaningless state when the client is loaded, which then allows for the rest to be checked reliably. But don't quote me on that, that's not my function, its @G-Ambatte's, i moved it, didn't make it.

Have had that happen in other cases though.

calculuschild: Ok. Just looking at the diff, `isClient` did not exist at all before this PR, and there was no check for being in the client; we were just relying on the config to tell us if we were in a deployment environment. I would assume we are still safe to just use the config environment without introducing a new state or a check for being in the client.

5e-Cleric:
Huh, well, i know i added it for a reason, but since removing it doesn't seem to break anything... out it goes.

calculuschild:
Looks good to me. I'm going to merge!

Issue #4424, opened 2025-09-22
Add separation between scroll bar and text in Editor panel


Falconloft:
### Your idea:

Issue: When trying to click the end of a text line, it is exceedingly easy to click the scroll bar instead. 

Fix: Adjust the margin from -50px to -40px. -45px may even be enough, but after testing, -40 seems better.

```
.CodeMirror-scroll {
    overflow: scroll !important;
    margin-bottom: -50px;
    margin-right: -40px;
    padding-bottom: 50px;
    height: 100%;
    outline: none;
    position: relative;
    z-index: 0;
}
```

Gazook89:
I vote just removing the custom styling of the scroll bars entirely, so you can see the border again. 

This is what I did in this [expansive PR](https://github.com/naturalcrit/homebrewery/pull/4122), demo here (if the demo still works): https://homebrewery-pr-4122.herokuapp.com/

Falconloft:
> I vote just removing the custom styling of the scroll bars entirely, so you can see the border again.

I'm good with that too.

5e-Cleric:
Custom styles for the scrollbar completely removed, will be coming with the next update, we are behind schedule on that.

PR #3388, opened 2024-04-02
Custom scrollbar fixes

5e-Cleric:
Should be easy to review, again.

Issue #3355, opened 2024-03-18
Custom scrollbar


ericscheid:
### Your idea:

[5e-Cleric](https://github.com/5e-Cleric) writes..

> PR #3346 adds in a scrollbar that i personally like more than the basic one:
> 
> <img width="284" alt="image" src="https://github.com/naturalcrit/homebrewery/assets/91662974/59342880-7dd4-4503-854a-b5c14b645c39">
> 
> The color is the inverse color of the background. 
> 
> Everyone will have an opinion on the color, so feel free to comment.
> 
> A border radius can be added, but sacrificing the space between the bar and the edge of the screen

ericscheid:
Adding as an issue for ease of finding in the future. 

(All PRs really should start from an issue, fwiw)

calculuschild:
Making note of this exchange about having the scrollbar auto-hide, so it doesn't get lost:

> On MacOS, I have the option of having scroll bars auto-hide, but this css has the thumb always visible. Blech.

> It would likely need a onScroll event on the brewRenderer to change the opacity of the scroll bar (with a fade transition to make it smooth).

> Right. I don't think there is a CSS-only way to detect the scrolling state. It is possible with Javascript. Found this example that uses event listeners:
>
> ```
>    const scrollableDiv = document.getElementById("scrollable");
>    let scrollingTimeout;
>
>    scrollableDiv.addEventListener("scroll", () => {
>      clearTimeout(scrollingTimeout);
>      scrollableDiv.style.setProperty("--scrollbar-thumb-color", "black");
>      scrollingTimeout = setTimeout(() => {
>        scrollableDiv.style.setProperty("--scrollbar-thumb-color", "rgba(90, 90, 90, 0.5)");
>      }, 300); // Adjust the delay as needed
>    });
> ```

G-Ambatte:
A minor layout issue with the new scrollbar - the page numbers appear under the scroll bar.

![image](https://github.com/naturalcrit/homebrewery/assets/19826659/8ebfbe72-9894-4e59-aaa3-2e533c3224ff)

Gazook89:
I know this is merged already, and there was an opportunity to comment before it was, so this is sort of pissing in the wind--- but I'm not a huge fan of this change.  

[...]

PR #3346, opened 2024-03-06
Custom scrollbar 

5e-Cleric:
This PR adds in a scrollbar that i personally like more than the basic one:

image

The color is the inverse color of the background. 

Everyone will have an opinion on the color, so feel free to comment.

A border radius can be added, but sacrificing the space between the bar and the edge of the screen

PR #4435, opened 2025-10-01
remove custom scrollbar styles

5e-Cleric:
# Contents
Ever since the custom scrollbar was added to the editor panel in (#3346) and fixed (#3388), a number of people has complained about it, and reported in some systems it still hides some text, therefore, this PR removes it.

Related issues:

#4424
#3355

# As for testing

There is not much to test, its a CSS only PR:

## Before

image

## After the change

image


G-Ambatte (approved):
Works perfectly in deployment.

Hopefully this will resolve the known issue for Mac users that the end of the text disappears under the scrollbar, but unfortunately I can't test that as I don't have a Mac device.

5e-Cleric:
Gazook tested and sees the issue fixed, merging!

PR #4436, opened 2025-10-01
Add indexes to Mongo schema

G-Ambatte:
This PR adds the indexes used on the Production MongoDB to the schema, and modifies the MongoDB connection properties so that these indexes will be built during Local/Docker installations without requiring user action.

The autobuild property is  only `true` for local installations, so this change is intended to have no effect on the Production install.


5e-Cleric (commented):
Looks good, but i can't approve until i test it on local, provide explicit testing steps, please.

G-Ambatte:
> Looks good, but i can't approve until i test it on local, provide explicit testing steps, please.

I can provide better instructions tomorrow, but in short:

1. Connect to your local Homebrewery DB using Compass or similar
2. Delete all indexes
3. Build the PR branch and start a local Homebrewery instance as normal (NODE _ENV set to local or docker)
4. Once the server is started, return to Compass and refresh the indexes view. You should observe that many indexes have now been created.

EXTRA CREDIT: Repeat the above steps for NODE_ENV set to production and observe that NO indexes are created.

PR #4438, opened 2025-10-02
Fix default save location functionality

### LINKED ISSUE #4437: New Page no longer respect default Save Location
While working on an unrelated PR, I discovered that saving from the New Page is no longer respecting the selected Default Save Location - in my case, my account is linked to Google but the default save location is Homebrewery, and New Page was incorrectly attempting to save to Google.

After a considerable amount of experimentation, I have determined that the SAVEKEY is being incorrectly set, as the `global.account.username` appears to be unset at the time that the SAVEKEY is defined, resulting in it being set to `HOMEBREWERY-DEFAULT-SAVE-LOCATION-` and having no username set at all. This results in the `localStorage` call returning no information.

G-Ambatte:
This PR resolves #4437.

The root cause of the issue was SAVEKEY being defined while the `global.account.username` is not defined. By shifting the definition of SAVEKEY to just before it is used to retrieve the data from local storage, the username is populated and all of the expected functionality returns.

### DIFF
--- a/client/homebrew/pages/newPage/newPage.jsx
+++ b/client/homebrew/pages/newPage/newPage.jsx
@@ -23,7 +23,7 @@ import { printCurrentBrew, fetchThemeBundle, splitTextStyleAndMetadata } from '.
 const BREWKEY  = 'homebrewery-new';
 const STYLEKEY = 'homebrewery-new-style';
 const METAKEY  = 'homebrewery-new-meta';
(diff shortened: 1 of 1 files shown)


calculuschild (approved):
Ok. So we can't access globals until the component mounts. Interesting.

Looks good to me!

G-Ambatte:
> Ok. So we can't access globals until the component mounts. Interesting.
> 
> Looks good to me!

I'm interested to see how/if this learning will affect the local storage key clean up. I'll find it once I get to that PR, hopefully this weekend.

PR #4445, opened 2025-10-03
Sed for parsing mustache bocks does not allow some expected characters - take 2

### LINKED ISSUE #4201: Sed for parsing mustache bocks does
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

A number of regular shift-state symbols are not accepted by the regex when trying to use them in a CSS property.  

Example:

```
\page {--chapterName:"'Dungeons & Dragons'"}
```

### Code

```gfm

```

dbolack-ab:
## Description

Resubmit the PR for additional allowed characters, minus braces.

There is still an issue with divs I cannot spot.

---- From the original ----

The regex for mustaches prevents using common and not-so-common characters that users may wish to assign to CSS property vars when those vars are used for strings.

\&\:\!\@\$\^\*\<\>\;\:\[\]\{\}\-\_\+\=

- Closes #4201 

### Reviewer Checklist

  - [ ] Test using \&\:\!\@\$\^\*\<\>\;\:\[\]\-\_\+\= characters in --property assignment string
  example: `\page {MnMCinnabar,pattern1,--chapterName:"'Blue : ! @ $ * < > ; : [ ] - _ + ='"}`
  - [ ] Test using \&\:\!\@\$\^\*\<\>\;\:\[\]\-\_\+\= characters in --property assignment string in a div block
  example: `\page {MnMCinnabar,pattern1,--chapterName:"'Blue : ! @ $ * < > ; : [ ] { } - _ + ='"
  
  }`
  - [ ] Verify syntax highlighting still hilights
  - [ ] Verify other mustache assignments still are set up.


5e-Cleric:
It seems like `=` is not valid

dbolack-ab:
> It seems like `=` is not valid

Yeah, not sure what broke there. Will give it a poke later.

The example tests were also bad.

shared/naturalcrit/markdown.js
```diff
@@ -185,7 +185,7 @@ const mustacheSpans = {
 	start(src) { return src.match(/{{[^{]/)?.index; },  // Hint to Marked.js to stop and check for a match
 	tokenizer(src, tokens) {
 		const completeSpan = /^{{[^\n]*}}/;               // Regex for the complete token
-		const inlineRegex = /{{(?=((?:[:=](?:"['\w,\-+*/()#%=?. ]*"|[\w\-+*/()#%.]*)|[^"=':{}\s]*)*))\1 *|}}/g;
+		const inlineRegex = /{{(?=((?:[:=](?:"['\w,\-()#%=?. \&\:\!\@\$\^\*\<\>\;\:\[\]\-\_\+\=]*"|[\w\-()#%.]*)|[^"=':{}\s]*)*))\1 *|}}/g;
```
calculuschild: For legibility, most of these added symbols do not need the backslash escape.

In trying to see why this was failing some cases, note that between `" "` double quotes, we already had `-`, `+`, `=`, `*`, but have lost `/`

Also note that without quotes, we have lost `+` and `*` and `/`.

Maybe this gets us closer:
`/{{(?=((?:[:=](?:"['\w,\-+*/()#%=?.&:!@$^<>;:\[\]_= ]*"|[\w\-+*/()#%.]*)|[^"=':{}\s]*)*))\1 *|}}/`

Same comment applies to the other two cases.

shared/naturalcrit/markdown.js
```diff
@@ -242,7 +242,7 @@ const mustacheDivs = {
 	start(src) { return src.match(/\n *{{[^{]/m)?.index; },  // Hint to Marked.js to stop and check for a match
 	tokenizer(src, tokens) {
 		const completeBlock = /^ *{{[^\n}]* *\n.*\n *}}/s;                // Regex for the complete token
-		const blockRegex = /^ *{{(?=((?:[:=](?:"['\w,\-+*/()#%=?. ]*"|[\w\-+*/()#%.]*)|[^"=':{}\s]*)*))\1 *$|^ *}}$/gm;
+		const blockRegex = /^ *{{(?=((?:[:=](?:"['\w,\-()#%=?.\&\:\!\@\$\^\*\<\>\;\:\[\]\-\_\+\= ]*"|[\w\-()#%. ]*)|[^"=':{}\s]*)*))\1 *$|^ *}}$/gm;
```
calculuschild: This Regex mistakenly allows ` ` to the non-quote case. Which means users can input `color:re d`, which is invalid.

calculuschild:
As for this test case:

```
{{--chapterName:"'Blue : ! @ $ * < > ; : [ ] - _ + ='"
  Hi!
}}
```

The one character that seems to fail for me is `>`. I'm not sure why.

Edit: It's prematurely ending the opening DIV tag:

image

Maybe worth removing <> for now; We can get the rest of the characters working, and then add some HTML encoding or something for the others. I feel like it's going to need more thought because that may have weird interactions with variable values if they are encoded.

shared/naturalcrit/markdown.js
dbolack-ab: Fixed.

shared/naturalcrit/markdown.js
dbolack-ab: Testing with this one.

dbolack-ab:
> The one character that seems to fail for me is `>`. I'm not sure why.
> 
> Edit: It's prematurely ending the opening DIV tag:

> Maybe worth removing <> for now; We can get the rest of the characters working, and then add some HTML encoding or something for the others. I feel like it's going to need more thought because that may have weird interactions with variable values if they are encoded.

Agreed. removing for now. This PR was *mostly* concerned with getting characters that would likely be used in chapter heads in ( so mostly hyphen and ampersand ) and I kinda dogpiled the others that were missing...

calculuschild:
Ok, I think this one is alright now. I don't see any failures.

PR #4446, opened 2025-10-03
Inline SVG proof of concept

dbolack-ab:
## Description

This inserts all non-local SVGs as inline HTMLElements


calculuschild:
I am still wary of having network operations inside our markdown parsing step (potentially very slow keyboard lag). Just putting that out there.

But even so, I can't seem to get this to work. SVGs I use in an image tag just display as a normal ``. Can you give an example text to demonstrate this working?

dbolack-ab:
> I am still wary of having network operations inside our markdown parsing step (potentially very slow keyboard lag). Just putting that out there.

Not gonna argue that, though my reading says that fetch should hit the browser cache so it should be minimal after the first load which I expect for most users (while editing) will happen before they insert the URL.

> But even so, I can't seem to get this to work. SVGs I use in an image tag just display as a normal ``. Can you give an example text to demonstrate this working?

This is currently being bitten by CORS when it wasn't earlier. Either CORS has tightened up again or I managed to test with a compliant site. 

This kind of feature may have to get pushed behind a proxy.

Meh.

PR #4448, opened 2025-10-03
Add DB check middleware

### LINKED ISSUE #4412: Add database connection event monitoring
### Your idea:

In it's current implementation, Homebrewery must successfully connect to the database before it starts running the Express server to return web pages.
However, there is currently no logic in place to detect if the database disconnects or otherwise becomes unavailable. In the event that this happens, any number of errors may occur, depending on the user activity at the time. The wide spread of these errors will make it difficult to diagnose the actual root cause.

My suggestion is that we implement database connection event monitoring, and in the event that the database state is not `connected`, throw a specific error.

G-Ambatte:
This PR resolves #4412.

This PR adds middleware that checks that Mongoose database connection is in the `connected` state before processing requests. If the connection is not `1` (connected), an error is thrown.

From https://mongoosejs.com/docs/api/connection.html#Connection.prototype.readyState, the potential states are as follows:

    0 = disconnected
    1 = connected
    2 = connecting
    3 = disconnecting

This PR also adds console messages for database connectivity events.

---

While the database must be connected when the Express web server starts, there are no checks or logic to ensure that the database is still running during normal operation. This PR adds that logic and adds logging that will make investigating database connection issues simpler for future troubleshooters.

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -531,11 +531,13 @@ const api = {
 	}
 };
 
-router.post('/api', checkClientVersion, dbCheck, asyncHandler(api.newBrew));
-router.put('/api/:id', checkClientVersion, dbCheck, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
-router.put('/api/update/:id', checkClientVersion, dbCheck, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
-router.delete('/api/:id', checkClientVersion, dbCheck, asyncHandler(api.deleteBrew));
-router.get('/api/remove/:id', checkClientVersion, dbCheck, asyncHandler(api.deleteBrew));
-router.get('/api/theme/:renderer/:id', dbCheck, asyncHandler(api.getThemeBundle));
+router.use(dbCheck);
+
+router.post('/api', checkClientVersion, asyncHandler(api.newBrew));
+router.put('/api/:id', checkClientVersion, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
+router.put('/api/update/:id', checkClientVersion, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
+router.delete('/api/:id', checkClientVersion, asyncHandler(api.deleteBrew));
+router.get('/api/remove/:id', checkClientVersion, asyncHandler(api.deleteBrew));
+router.get('/api/theme/:renderer/:id', asyncHandler(api.getThemeBundle));
 
 export default api;


5e-Cleric:
We should catch the error and show a popup for it, for instance, with this in place, it works, but the error the user sees is: `Looks like there was a problem saving.Report the issue here.`

Do you wanna add that in this PR or later in another?

5e-Cleric (approved):
As it stands, it works and doesn't break anything

G-Ambatte:
> We should catch the error and show a popup for it, for instance, with this in place, it works, but the error the user sees is: `Looks like there was a problem saving.Report the issue here.`
> 
> Do you wanna add that in this PR or later in another?

I'll take a quick look at how that notification is set up before making a call.

server/middleware/dbCheck.spec.js
calculuschild, line 1: I think you can trim down the testing a lot here:

```
import mongoose from 'mongoose';
import dbCheck from './dbCheck.js';
import config from '../config.js';

describe('dbCheck middleware', () => {
  const next = jest.fn();

  afterEach(() => jest.clearAllMocks());

  it('should skip check in test mode', () => {
    config.get = jest.fn(() => 'test');
    expect(() => dbCheck({}, {}, next)).not.toThrow();
    expect(next).toHaveBeenCalled();
  });

  it('should call next if readyState == 1', () => {
    config.get = jest.fn(() => 'production');
    mongoose.connection.readyState = 1;
    dbCheck({}, {}, next);
    expect(next).toHaveBeenCalled();
  });

  it('should throw if readyState != 1', () => {
    config.get = jest.fn(() => 'production');
    mongoose.connection.readyState = 99;
    expect(() => dbCheck({}, {}, next)).toThrow(/Unable to connect/);
  });
});
```

G-Ambatte: I've implemented this in the latest commit.

G-Ambatte:
> We should catch the error and show a popup for it, for instance, with this in place, it works, but the error the user sees is: `Looks like there was a problem saving.Report the issue here.`
> 
> Do you wanna add that in this PR or later in another?

Added now; looking in to this revealed that I had the format of the error object slightly wrong.

G-Ambatte:
In the Edit Page, you should now get this error:
image

> **DISREGARD**
> ~~...huh. Testing in Firefox, on the exact same brew in the exact same conditions, throws the exact same error at the exact same place in the code, but instead the error message is this:~~
> image
>
> ~~I'll dig into this a little bit, but my suspicion is that it's not specific to this PR.~~
**DISREGARD**

**UPDATE:** I performed a hard reload with cache disabled and it began working correctly - the Firefox error message now matches the Chrome error message.

server/homebrew.api.js
```diff
@@ -530,11 +531,11 @@ const api = {
...
-router.put('/api/update/:id', checkClientVersion, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
-router.delete('/api/:id', checkClientVersion, asyncHandler(api.deleteBrew));
-router.get('/api/remove/:id', checkClientVersion, asyncHandler(api.deleteBrew));
-router.get('/api/theme/:renderer/:id', asyncHandler(api.getThemeBundle));
+router.post('/api', checkClientVersion, dbCheck, asyncHandler(api.newBrew));
+router.put('/api/:id', checkClientVersion, dbCheck, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
+router.put('/api/update/:id', checkClientVersion, dbCheck, asyncHandler(api.getBrew('edit', false)), asyncHandler(api.updateBrew));
+router.delete('/api/:id', checkClientVersion, dbCheck, asyncHandler(api.deleteBrew));
+router.get('/api/remove/:id', checkClientVersion, dbCheck, asyncHandler(api.deleteBrew));
+router.get('/api/theme/:renderer/:id', dbCheck, asyncHandler(api.getThemeBundle));
```
calculuschild: I think we can do something like `router.use(dbCheck);` in one place instead of calling it out in each individual route. Or in app.js do `app.use(dbCheck)` if we think it should be a universal thing.

What do you think?

G-Ambatte: I had `app.use(dbCheck)` initially, then thought that it might be better to apply it individually, so routes that don't touch the DB don't cause an unnecessary check.
That said, those routes are very limited and not something that would be visited regularly: Change Log, FAQ, Migrate, etc. New and Home technically don't use the DB - unless they save, which uses the API route which also has it's own dbCheck.

At the very least, I think adding `router.use(dbCheck)` will apply the check to all of these API routes in a single line.

G-Ambatte: I've applied the change to `router.use(dbCheck)` in the latest commit. It appears to work perfectly.

PR #4449, opened 2025-10-03
Alternate Add d100/percentage support to dicefont icons

### LINKED ISSUE #3347: Add d100 support in diceFont

dbolack-ab:
## Description

This is added as a new icon but inlined with the df icons for user convenience.

Replaces #3463

Add values of 00 through 100 on the D10 Dice Icon using `df_dper_#` ( example df_dper_75 )
I used dper as the die signifier instead of dpercent for user brevity. Not hugely married to it.

Provides replacement d10 values along with adding a `0` option so that the two sets using the same d10 die outline match better.

- Closes #3347 

If you want or need to regenerate the font, the source images are in `d100_Sources/`. There is a readme that should cover the needs to do so.

[Test Document](https://homebrewery-pr-4151.herokuapp.com/share/QyiI4rfEEMtN)

image


themes/fonts/iconFonts/diceFont.js
```diff
@@ -90,7 +91,108 @@ const diceFont = {
 	'df_solid_small_dot_d6_3' : 'df solid-small-dot-d6-3',
 	'df_solid_small_dot_d6_4' : 'df solid-small-dot-d6-4',
 	'df_solid_small_dot_d6_5' : 'df solid-small-dot-d6-5',
-	'df_solid_small_dot_d6_6' : 'df solid-small-dot-d6-6'
+	'df_solid_small_dot_d6_6' : 'df solid-small-dot-d6-6',
+	'df_dper_00'              : 'df dpercent-00',
```
calculuschild: I would prefer we make these a continuation of the `df_d10_...` naming pattern. It's the same d10 body just with higher numbers inside. So for example `df_d10_00 : df d10-00`.

d100_Sources/buildFont.bash
calculuschild, line 1: This is handy as a tool on its own, but for the same reasons we don't add photoshop files or font-forge projects, I don't think we need to store the raw source artifacts or generation scripts to the repository. The compiled font file all we really want; everything else feels more like tooling that should live in its own repo, or offline to be pulled in as needed.

themes/fonts/iconFonts/diceFontD100.woff2
calculuschild, line 1: Try to convert to a .woff2 if you can. Better compression for no loss, and follows what our other fonts use.

d100_Sources/buildFont.bash
dbolack-ab: That's how stuff gets lost, but okay

themes/fonts/iconFonts/diceFontD100.woff2
dbolack-ab: Was already, just wrong extension

calculuschild:
Just made a couple other cleanups. Merging now!

PR #4451, opened 2025-10-04
Update markdown.js to use external npm version of markdown variables

dbolack-ab:
## Description

Moves the variables code to a module, makes small adjustments accordingly.

Supports #4035

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,19 +10,19 @@
       "hasInstallScript": true,
       "license": "MIT",
       "dependencies": {
-        "@babel/core": "^7.27.1",
-        "@babel/plugin-transform-runtime": "^7.28.0",
-        "@babel/preset-env": "^7.28.0",
+        "@babel/core": "^7.28.4",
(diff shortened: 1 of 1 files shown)


calculuschild:
Let's make sure any external Markdown packages like this are fully transferred over to naturalcrit before committing. I want to make sure we all have access to the repo if something needs to be fixed.

dbolack-ab:
I'm 99% sure you're the owner or in a position to reown everything. I spot checked.

Let me know if I missed any.

calculuschild:
This one is getting some errors deploying. I have converted this package over to naturalcrit and am just going over that package to see where it is crashing.

dbolack-ab:
I have had weird intermittent errors with NPM not finding the module.

calculuschild:
Got the tests passing, and all seems to be working now. Some of the cross-page behavior was not quite working.

In the marked-variables package I also found some opportunities for cleanup and got the package to clear the tests in about half the time by simplifying a lot of regex and removing some duplicate calculations. Yee-haw.

PR #4452, opened 2025-10-04
Standardize local storage key names #4119

### LINKED ISSUE #4119: Standardize cookie and storage key names

G-Ambatte:
This PR resolves #4119.

This PR adds a function to update the local storage key names from their current names to a new standardized name format.
In the current iteration, the function also checks for a specific key (`HB_deleteKeys`); if found, the old keys are removed after the new key is populated.


5e-Cleric:
Is it just me or are the names in your code also not standard, `ADMIN_TAB`, and `HB_editor_splitWidth`, what is the standard here?

calculuschild:
> Is it just me or are the names in your code also not standard, `ADMIN_TAB`, and `HB_editor_splitWidth`, what is the standard here?

`ADMIN_TAB` is a variable name. `HB_editor_splitWidth` is a Localstorage key string.

These are two different things.

5e-Cleric:
> > Is it just me or are the names in your code also not standard, `ADMIN_TAB`, and `HB_editor_splitWidth`, what is the standard here?
> 
> `ADMIN_TAB` is a variable name. `HB_editor_splitWidth` is a Localstorage key string.
> 
> These are two different things.

you're right, i didn't pay the appropiate attention to the code.

5e-Cleric (approved):
After rereviewing the code, everything seems upstanding.

I don't think  we'll ever be able to remove the localstorage key map or update function.

G-Ambatte:
> `ADMIN_TAB` is a variable name. `HB_editor_splitWidth` is a Localstorage key string.

I didn't mention it, but the PR also standardizes the use of a variable for the key name on each page. Many were already set up in this way but some were specifying the string in every call to local storage.

G-Ambatte:
> I don't think  we'll ever be able to remove the localstorage key map or update function.

I think after a year or so, anyone who actually cares about the content of the local storage keys will also be someone who has visited the site in that time.

PR #4455, opened 2025-10-04
Permanently enable old local storage key deletion

### LINKED ISSUE #4454: Local Storage key deletion behind switch
### Your idea:

The recent PR to update the local storage key names was merged with the opt-in key deletion function still in place. In order to ensure that the old keys are correctly removed, this function needs to be bypassed so that the deletion function always runs.

Without this key deletion in place, data will (effectively) not be cleared from New Page when the data is saved (and the user is moved to the Edit Page), which will at best be confusing for users.

G-Ambatte:
This PR resolves #4454.

This PR permanently enables the key deletion functionality of the recently merged local storage key update function.


calculuschild:
Ok, looks good. Thanks for catching this!

Issue #4475, opened 2025-10-14
Error


dafide78-max:
### Error Code

```shell
Error: {"originalUrl":"/api/1STlecb40dX5EqeNeMywCBq-aIGSPz3VtSIdLFFQw55WP"}


Error: {"originalUrl":"/api/1STlecb40dX5EqeNeMywCBq-aIGSPz3VtSIdLFFQw55WP"}
    at Request.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:82249:21)
    at Emitter.emit (https://homebrewery.naturalcrit.com/homebrew/bundle.js:1968:20)
    at XMLHttpRequest.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:82543:10)
{
  "status": 500,
  "method": "DELETE",
  "url": "/api/1STlecb40dX5EqeNeMywCBq-aIGSPz3VtSIdLFFQw55WP"
}
```

### Your description of what happened:

I don't know what happened

dafide78-max:
I really don't understand your reply.... Please let me know what i should do to solve the problem because i lost many of my works and i can't save any new because after few days i lose it. Also i can't delete anything. What's going on!!!!????

5e-Cleric:
Right, this error has been reported many times before, at least 10 times. For logistics, we are not going to have the 10 error reports, so all discussion about it will be held in #4313

Yesterday we merged some code that should help us get to the bottom of what is happening to create this error, but we suspect it has to do with older brews with older Ids. 

As far as other reporters go, no one has mentioned this to be more than a one time thing, are you seeing this issue every day?

dafide78-max:
Good morning,
yes i'm seeing it every day from months....

------ Messaggio originale ------
Da "Víctor Losada Hernández" ***@***.***>
A "naturalcrit/homebrewery" ***@***.***>
Cc "dafide78-max" ***@***.***>; "Author" 
***@***.***>
Data 17/10/2025 14:32:20
Oggetto Re: [naturalcrit/homebrewery] Error (Issue #4475)

PR #4476, opened 2025-10-15
handle getid uncatched error

5e-Cleric:
Catches a getId error further down, should fix a myriad of saving errors, like https://github.com/naturalcrit/homebrewery/issues/4475


calculuschild:
I don't think this will *fix* those errors, but it should get us more info to help diagnose them.

Merging!

Issue #3103, opened 2023-10-16
PHB footers should default to all caps?


calculuschild:
Copied from #2289:

> This footer should transform text to CAPS automatically, in the PHB theme, reopening the issue to track that change.
> If a javascript change is not optimal, this CSS will do it:
> 
> ```
> .footnote {
>   text-transform:uppercase;
> }
> ```

G-Ambatte:
This is an interesting question - SHOULD the text transform to caps? This behavior isn't clearly indicated by the `footnote` tag, so might be surprising/confusing to users.

I would suggest an alternative styling of `.footnote.caps { text-transform: uppercase; }` and then adjust the snippet to apply `{{footnote,caps [...]}}` instead.

5e-Cleric:
I disagree on that, a footnote has a series of properties, and most off the time, all caps is one of them, i thought it did until i sent that message.

But i don't really care, it won't make my brews any longer, both options are okay.

This is a perfect issue for new users to contribute.

Gazook89:
It’s a stylistic change, to match PHB— I don’t think it is onerous/unexpected that the footnote would be all caps. Same deal as the small caps on the first line after h1 headers. 

THAT SAID, we could have a `no-caps` class that removes the uppercase property (rather than a class that adds it). This could also be applied to a paragraph that has the first line capitalized by using injection curlies, rather than/in addition to the method offered in the Style Editor snippets.

G-Ambatte:
I think that the only thing that we can rely on is that users expect that the current behaviour of the footnote will continue in the future, so the styling of `.footnote` should ideally remain unchanged, thus preventing any unexpected changes to existing documents - even if that change is only "this tiny piece of text is all caps now".
It does seem like this would be perfect for the Theme Versioning change that I mentioned in a comment on that issue earlier today ([here](https://github.com/naturalcrit/homebrewery/issues/1899#issuecomment-1806940577)) to allow users to stay on their current version if they so desired. But that's a pretty enormous change, and ideally Themes will be live and available to users for some time before we even consider beginning work on that feature.

With that said, I'm not strongly for or against either position, and the actual change itself is only a handful of lines of styling; I'm certain that we can quickly knock out a PR to apply the desired change, whatever it might be.

I also really enjoy the discussion; it's great that we have enough people active enough to discuss the issue.

5e-Cleric:
It is nice the possibility of having a discussion, but we better use our time in interesting features, rather than missing nuances.

[...]

PR #4478, opened 2025-10-15
Add uppercase style to footnotes

5e-Cleric:
Fixing #3103

adds `text-transform: uppercase;` to the .footnote class, to perfect the theme some more.


5e-Cleric:
will wait for a negative response for 24 hours, if there isn't there it goes.

calculuschild:
Ok, this seems good to me. Users can do `{{footnote,text-transform:unset` if they really want to deviate from the theme, but this shouldn't *break* any existing brews by throwing off page layouts.

PR #4480, opened 2025-10-15
Fix cursor on head navigation

### LINKED ISSUE #2689: brewTitle in top navbar should not have a "click-me" cursor
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

On all pages that have a main navbar which has a title (i.e. pretty much all pages except the literal home page) .. the page/brew title has `cursor: pointer` as a css rule. Clicking the title does nothing. 

image

There is also a CSS rule for `.brewTitle.navItem`
image
so we can simply add `cursor: unset;` to that.

## Solution

Add `cursor: unset` to the CSS rule for `.brewItem.navItem`.

### Code

_No response_

5e-Cleric:
Fixes #2689

### DIFF
--- a/client/homebrew/navbar/navbar.less
+++ b/client/homebrew/navbar/navbar.less
@@ -37,7 +37,10 @@
 
 		&:has(.brewTitle) {
 			flex-grow : 1;
-			min-width : 300px; 
+			min-width : 300px;
+		}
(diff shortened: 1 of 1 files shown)


calculuschild:
Looks good. Merging!

PR #4481, opened 2025-10-15
fix save button glitch

5e-Cleric:
We had implemented a few animations for the save button, that were causing a visual flicker on screen, simplified the code to use a simpler animation just when needed.

Check that save button on homepage or new.


calculuschild:
This seems to work ok for me. I don't see the flickering.

PR #4505, opened 2025-10-28
Disable some codeEditor hotkeys if not in a gfm language editor.

### LINKED ISSUE #3319: Some shorcuts should only work in brew text editor
you can insert a new page break in the style editor, this doesn't make any sense, similarly, there are other shortcuts that should be limited to work in the text editor only.

dbolack-ab:
## Description

Disables most of the Homebrewery added hotkeys in the editor when not in GFM language mode.

- Closes #3319 

### Reviewer Checklist

  - [ ] Hotkeys ( insert page, bold, etc ) that should not be present in Styles editor do not insert
  - [ ] Hotkeys still work in Editor and Snippets menu


Gazook89:
Upfront, this all looks good to me, the idea and code. Has my approval.


<details>
<summary>Two additional thoughts:</summary>

1. I support just removing a bunch of the hotkeys.  We don't advertise them anywhere, so they are near-impossible to know they even exist.  We could add some listing of hotkeys, either as a summonable modal or hints in relevant snippet bar menus etc.  Or, just assume they won't be missed.  I made probably most of the hotkeys myself as the first javascript code contributions I did, back when they were easy repeatable pattern that helped me learn JS.   I'm not sure they actually have any utility.

Second thought I have barely put any thinking into, so feel free to skip, but.... 

2. Instead of writing the `if` statement within each hotkey function, we could have a wrapper function that allows us to set which editor it is for.  Something like this:

```
extraKeys         : {
    ...
    'Shift-Cmd-Example'     : mdHotkey(this.makeSuper),
    'Shift-Cmd-Example2'      : cssHotkey(this.makeSuper),
    'Shift-Cmd-Example3'      : mdHotkey(this.makeSub) || cssHotkey(this.makeSub),
    ...
}
```

And those function basically just check the editor...and so in example 3 if the mdHotkey() function returns false then it runs the cssHotkey function.  

I don't know, probably better ways to run with that.  

</details>

dbolack-ab:
> Upfront, this all looks good to me, the idea and code. Has my approval.
> 
> Two additional thoughts:
> 1. I support just removing a bunch of the hotkeys.  We don't advertise them anywhere, so they are near-impossible to know they even exist.  We could add some listing of hotkeys, either as a summonable modal or hints in relevant snippet bar menus etc.  Or, just assume they won't be missed.  I made probably most of the hotkeys myself as the first javascript code contributions I did, back when they were easy repeatable pattern that helped me learn JS.   I'm not sure they actually have any utility.

This might be best to move to a Discussion topic. If we do so, I think this PR can hold off till that conversation resolves

> 2. Instead of writing the `if` statement within each hotkey function, we could have a wrapper function that allows us to set which editor it is for.  Something like this:

If we want to go that route - where we have alternate definitions based on the pane ( in places where the same concept happens but the execution is different - see Comment hotkey for example ), we should create distinct objects for each pane and pass that in as extrakeys ( that would probably be cleaner than this implementation ) field in the buildEditor call.

Gazook89:
Agreed on both points. I don’t personally want to open the new discussion at this time, and for point two it probably would necessitate some thought on what hotkeys are useful in style and snippet editors (though snippet and the normal editor are both GFM, snippet may have things like a hot key for “new snippet”). 

Full steam ahead with what you have.

dbolack-ab:
> Agreed on both points. I don’t personally want to open the new discussion at this time, and for point two it probably would necessitate some thought on what hotkeys are useful in style and snippet editors (though snippet and the normal editor are both GFM, snippet may have things like a hot key for “new snippet”).

I've added a "tab" property and switch to helper functions and referencing the property instead of using the language.

I think this will make it cleaner *if* we decide to gio more agressive with the hotkeys and possibly overloading based on tab.

calculuschild:
@dbolack-ab this has conflicts again. Not sure what happened. Could you resolve it again and then ping one of us to merge it?

dbolack-ab:
> @dbolack-ab this has conflicts again. Not sure what happened. Could you resolve it again and then ping one of us to merge it?

A bunch of `this.codemirror.somemethod` were changed to `this.codemirror?.somemethod` :)

calculuschild:
Alrighty. I think this one's OK to merge now. Thanks @dbolack-ab! 💯 ⭐ 🎉

PR #4512, opened 2025-11-04
Move renderWarnings from shared to client

Gazook89:
## Description

Moving this component to the component folder so it's closer to where it is actually used.  Not moving to the `homebrew` folder because ultimately i think pretty much everything in that folder should move to the components as well (`homebrew` isn't a helpful folder distinction).

Deletes a directory, moves 2 files.

none

Check that it still renders by using a non-chrome browser.

### Reviewer Checklist

- [ ] still renders with a non-chrome browser (orion).


dbolack-ab:
Looks good. Can we get a test instance?

calculuschild:
> `homebrew` isn't a helpful folder distinction).

`homebrew` is probably one of the most important folder distinctions, though maybe the name is poorly chosen. `homebrew` is the container of the core React app, if that makes sense. We have he Admin app, we have some reusable components, we have common icon resources, and we have the Homebrewery app. It may make more sense if the folder were renamed "Homebrewery".

PR #4513, opened 2025-11-04
Move SVG components to components folder

Gazook89:
## Description

Move the SVGs out of the `shared` directory, closer to where they are actually used.  They are svg jsx files, setup as components, so i've moved them to the components directory.  The hope is that we aren't hunting for files between `client` and `shared` as much.

Deleted `combat.svg.jsx` because we don't use it anywhere.  `homebrew.svg.jsx` isn't used anywhere in the the homebrewery currently, either, but i've kept it because I think it could be useful down the line (though in my UI Overhaul PR I believe i further modify that svg component).  

renamed the remaining two to be more descriptive of the image for additional clarity.

none

The only thing that really changes here is the naturalcrit icon, which is used in the account page and the navbar, and both still work.

### Reviewer Checklist

- [x] all pages still work without crashing due to missing component.


calculuschild:
Looks good. Merging!

PR #4526, opened 2025-11-11
Fix transferring brews to/from Google storage

G-Ambatte:
A recent change to the Edit Page appears to have broken the ability to transfer brews to or from Google Drive storage.

The root cause of this issue was that the state of `saveGoogle` was being updated, but the `trySave` function was being called before the state update had completed, resulting in the incorrect values being used inside the `trySave` function and thus no transfer parameters being included in the `update` API call.


calculuschild:
Ok, this is fine for now. As part of future merging effort between home/new/edit I will probably change the saving functions to actually take in the relevant values as parameters so we aren't dealing with functions trying to synchronize state. Functional components are nice but they do introduce some timing issues like this due to slightly different component lifecycle.

Issue #4388, opened 2025-08-19
Add Save Reminder to NewPage


G-Ambatte:
### Your idea:

While working in the New Page, there are limited notifications that the work the user has completed is not being saved anywhere outside of local storage, or that local storage may be wiped through unintentional actions (cloning a brew, etc).

I suggest that we implement some basic measures to remind the users that the contents of NewPage are NOT SAFE:

1. We pop a SAVE button over the BrewRenderer, similar to the button that appears on the HomePage when the initial editor content has been changed.
<img width="329" height="137" alt="Image" src="https://github.com/user-attachments/assets/d6da09e0-cd90-4566-8474-3f5a5a265d4e" />

2. We implement a time-based reminder to save, essentially identical to the EditPage's autosave-disabled reminder.
<img width="232" height="101" alt="Image" src="https://github.com/user-attachments/assets/9864aae6-cc00-4711-bddb-0f86031af64e" />

3. These notifications should be frequent, regular, and unmissable to the point of borderline nuisance. The point is to ensure that no user working on the NewPage has any reason to believe that their document is saved securely.

5e-Cleric:
I'd use the alert method, or whatever substitute they offer

G-Ambatte:
> I'd use the alert method, or whatever substitute they offer

I believe `alert()` is a blocking function. We probably want a non-blocking method, so as not to interfere with anything that is already running.

5e-Cleric:
Yes but, we probably want users to be forced to make the decision, no? Save now or face the risk of losing everything. Or maybe just for closing.

PR #4527, opened 2025-11-12
Add unsaved-change warning to Home page editor

MiniX16:
## Description

This PR adds a security warning before any action that could cause the user to lose unsaved work on the home page.

Added warning:
image

This warning appears before any action that would cause the user to lose their current work.

### Reviewer Checklist

- [ ] The warning appears whenever it should.
- [ ] Identify opportunities for simplification and refactoring.

### DIFF
--- a/client/homebrew/pages/homePage/homePage.jsx
+++ b/client/homebrew/pages/homePage/homePage.jsx
@@ -70,23 +70,14 @@ const HomePage =(props)=>{
 			}
 		};
 
-		const handleBeforeUnload = (e)=>{
-			if(unsavedChangesRef.current) {
-				e.preventDefault();
-				e.returnValue = '';
-				return '';
-			}
-		};
-		
-		const previousBeforeUnload = window.onbeforeunload;
-
-		window.onbeforeunload = handleBeforeUnload;
-
 		document.addEventListener('keydown', handleControlKeys);
-
+		window.onbeforeunload = ()=>{
+			if(unsavedChangesRef.current)
+				return 'You have unsaved changes!';
+		};
 		return ()=>{
 			document.removeEventListener('keydown', handleControlKeys);
-			window.onbeforeunload = previousBeforeUnload;
+			window.onbeforeunload = null;
 		};
 	}, []);
 


Gazook89 (approved):
Why does this need to create a ref for unsaved changes when we already have unsavedChanges as a state variable?  Can't we just check the current state directly?  I see that this pretty much follows the same function on editPage.jsx so likely it's just me not understanding, but the question remains.  However, since it follows that same formula, and works on my local machine, it looks good to me.

G-Ambatte:
I think there is potentially value in adding a blocking onClose notification; however, Issue #4388 is about adding reminders in the same fashion as the EditPage's "autosave off" timed reminders, i.e. "You've been working on this for more then 15 minutes without saving. Save now?"

I have not reviewed the actual code yet, but I believe that the issue was not actively being worked on because changes were being made to the underlying page code to create a unified Editor.

G-Ambatte:
> Why does this need to create a ref for unsaved changes when we already have unsavedChanges as a state variable? Can't we just check the current state directly?

Calc made the following comment in `editPage.jsx`, which I think explains it pretty well:
image
I did spend some time experimenting with just using state, and was able to confirm that it just doesn't work with State only - the Ref is absolutely mandatory.

> I see that this pretty much follows the same function on editPage.jsx so likely it's just me not understanding, but the question remains. However, since it follows that same formula, and works on my local machine, it looks good to me.

For reasons that I have not yet been able to unravel, the method used here (`window.addEventListener('beforeunload', ()=>{});`) **_does not_** work for me in Chrome v142 for Windows 11 x64. However, the method used in the EditPage.jsx (`window.onbeforeunload = ()=>{};`) **_does_** work.

**Functional Code example:**
image

So I would suggest that this PR use the same method. This should also simplify the upcoming merge of the underlying structures of Home and New.

---

However, this does not address the issue raised in #4388 - that issue is for the New Page, not the Home Page; and it is about popping non-blocking reminders during editing, not showing a blocking warning on page close.
So I would unlink this PR from that particular issue.

That said, I see value in this PR and it should be merged once those changes are made.

client/homebrew/pages/homePage/homePage.jsx
```diff
@@ -69,13 +70,28 @@ const HomePage =(props)=>{
...
 
+		const handleBeforeUnload = (e)=>{
+			if(unsavedChangesRef.current) {
+				e.preventDefault();
+				e.returnValue = '';
+				return '';
+			}
+		};
+		
+		window.addEventListener('beforeunload', handleBeforeUnload);
```
G-Ambatte: For unknown reasons, this doesn't work for me (Chrome v142, Windows 11 x64). However `window.onbeforeunload()` works, as used in EditPage.jsx.

G-Ambatte (approved):
Looks good! I've also updated the PR to the current master, so it should be just about ready to merge.

Reloading with changes present:
image

Closing with changes present:
image

calculuschild:
Just made a small tweak to closer match how we do this in `editPage.jsx`. Thanks @MiniX16 ! ⭐ 🎉 💯

PR #4533, opened 2025-11-15
Fix small bugs towards 3.20

5e-Cleric:
Fixes snippetbar Wrap
fixes snippetbar being hidden definitively
remove new brew dropdown from new route

### DIFF
--- a/client/homebrew/editor/editor.jsx
+++ b/client/homebrew/editor/editor.jsx
@@ -85,16 +85,15 @@ const Editor = createClass({
 				editorTheme : editorTheme
 			});
 		}
-  		const snippetBar = document.querySelector('.editor > .snippetBar');
-  		if (!snippetBar) return;
+		const snippetBar = document.querySelector('.editor > .snippetBar');
+		if (!snippetBar) return;
 
-  		this.resizeObserver = new ResizeObserver(entries => {
+		this.resizeObserver = new ResizeObserver(entries => {
 			const height = document.querySelector('.editor > .snippetBar').offsetHeight;
 			this.setState({ snippetBarHeight: height });
-			console.log('setting state to ', height );
-  		});
+		});
 
-  		this.resizeObserver.observe(snippetBar);
+		this.resizeObserver.observe(snippetBar);
 	},
 
 	componentDidUpdate : function(prevProps, prevState, snapshot) {
@@ -118,7 +117,7 @@ const Editor = createClass({
 	},
 
 	componentWillUnmount() {
-  		if (this.resizeObserver) this.resizeObserver.disconnect();
+		if (this.resizeObserver) this.resizeObserver.disconnect();
 	},
 
 	handleControlKeys : function(e){
@@ -436,7 +435,6 @@ const Editor = createClass({
 	},
 
 	renderEditor : function(){
-		console.log('state at rendering is: ',this.state.snippetBarHeight);
 		if(this.isText()){
 			return <>
 				<CodeEditor key='codeEditor'


client/homebrew/pages/newPage/newPage.jsx
```diff
@@ -228,7 +228,6 @@ const NewPage = (props)=>{
 				{error
 					? <ErrorNavItem error={error} clearError={clearError} />
 					: renderSaveButton()}
-				<NewBrewItem />
```
calculuschild, line 231: This should not be removed. It is intentionally added as part of making the nav bar more common across new/edit/home.

calculuschild (changes requested):

calculuschild:
@5e-Cleric Can we get a screenshot of before/after, or steps to test? It's hard to visualize what the changes here are doing.

5e-Cleric:
> @5e-Cleric Can we get a screenshot of before/after, or steps to test? It's hard to visualize what the changes here are doing.

Sure.
## issue 1: snippetbar is not wrapping at the correct breakpoint

1. open https://homebrewery-stage.herokuapp.com/new, notice snippetbar does not wrap correctly at the desired breakpoint (750px)
2. open the deployment, notice it now does

## issue 2: snippetbar being hidden sometimes

1. open live site or stage, move editor panel until snippetbar is wrapped
2. reload the site
3. without moving the splitpane, enter text into the editor until the cursor goes of screen to the bottom
4. notice, half the snippetbar is hidden
5. this is fixed in this deployment

## issue 3: save now button animation is triggered on page load without changes to save, causing a flicker
1. open staging site, https://homebrewery-stage.herokuapp.com/new, see the button flicker and dissapear
2. open deployment, notice that doesn't happen

sorry if that sounds condescending, its just very simple.

client/homebrew/pages/newPage/newPage.jsx
5e-Cleric: reverted

client/homebrew/editor/editor.jsx
```diff
@@ -437,7 +447,7 @@ const Editor = createClass({
 					onChange={this.props.onBrewChange('text')}
 					editorTheme={this.state.editorTheme}
 					rerenderParent={this.rerenderParent}
-					style={{  height: `calc(100% - ${this.state.snippetbarHeight}px)` }} />
+					style={{  height: `calc(100% - 26px)` }} />
```
calculuschild: There is some conflict going on here where we have two different sources trying to edit the height of the .codeMirror component. One is the resizeObserver, and then again here in renderEditor we immediately overwrite that value.

client/homebrew/editor/editor.jsx
```diff
@@ -85,7 +84,18 @@ const Editor = createClass({
...
-		this.setState({ snippetbarHeight: document.querySelector('.editor > .snippetBar').offsetHeight });
+  		const snippetBar = document.querySelector('.editor > .snippetBar');
+  		if (!snippetBar) return;
+
+  		this.resizeObserver = new ResizeObserver(entries => {
+    		// throttle/debounce inside if you like
+    		const h = snippetBar.offsetHeight;
+				const editor = document.querySelector('.editor .codeEditor');
+				if (!editor) return;
+				editor.style.height = `calc(100% - ${h}px)`;
```
calculuschild: It is not advised to edit the DOM directly without going through React. This desynchronizes the actual DOM from what React has in its copy and can lead to some weird render bugs or components getting "lost". Is there a way to do this without directly modifying the DOM like this? I think the normal approach would be to set a state here, and then use that state down in the render step to affect the output.

client/homebrew/editor/editor.jsx
```diff
@@ -408,11 +422,7 @@ const Editor = createClass({
 	},
 
 	//Called when there are changes to the editor's dimensions
-	update : function(){
-		const snipHeight = document.querySelector('.editor > .snippetBar').offsetHeight;
-		if(snipHeight !== this.state.snippetbarHeight)
-			this.setState({ snippetbarHeight: snipHeight });
-	},
+	update : function(){},
```
calculuschild, line 423: Can you clarify why we need a resizeObserver instead of using the existing `update` function we have here? Is `update()` not triggering correctly?

client/homebrew/editor/editor.jsx
5e-Cleric: the resize observer comes later, but yes i could remove this i guess.

client/homebrew/editor/editor.jsx
5e-Cleric: update does not run when the snippetbar changes size after mount, when css container kicks in

client/homebrew/editor/editor.jsx
5e-Cleric: Tried resizer observer setting state, this is the result, it just doesnt do it in time i guess:
image

If you want me to push these changes so you can check on your own, just ask.

calculuschild:
Ok, I can merge this now. So it looks like the `update()` function *does* get called, but only when the divider bar moves. Doesn't account for vertical scrolling that hides the snippet bar, which makes sense why we need the observer now.

PR #4534, opened 2025-11-16
Update New Item on NavBar

G-Ambatte:
This PR:

- fixes an issue with incorrect local storage key names when checking for content before loading from a text file backup
- changes the name of "NEW FROM BLANK" to "RESUME EDITING"
- adds a new button "NEW FROM BLANK" which will clear the local storage content before redirecting to the New Page editor
- consolidates the functions for checking local storage between NEW FROM BLANK and NEW FROM FILE
- provides a checkLocalStorage function which may be used in the future to check if RESUME EDITING is valid (if no local storage content, then there will be nothing to resume).

image


G-Ambatte (commented):
I have marked this PR as a P2; while there is the potential for data loss to potentially mark it as P1, it is only because a warning prompt will not be shown before overwriting the data in local storage, so I have reduced it to P2.

calculuschild:
Looks good! Merging.

PR #4538, opened 2025-11-19
fix .resetCounter

5e-Cleric:
Surprised no user had reported it yet, resetCounter was not working as supposed to. 

As it is live, it only sets the counter for that page instead of actually resetting the count as it should.

Its a small fix.


calculuschild:
I can confirm this works. Interesting find.

Issue #4551, opened 2025-11-29
"Back Cover Page" PHB snippet is offset by 1 pixel


Wynnies:
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

Using the back cover page snippet and changing the image, reveals the offset of 1 pixel from the left side from the back cover page. It is barely visible with the example image given when you select the snippet, so changing it makes it more visible.

<img width="669" height="857" alt="Image" src="https://github.com/user-attachments/assets/fecd7915-0dd0-45c8-8e6e-6576dbac5f4d" />

### Code

```gfm
{{backCover}}

# The Necromancer

Experience a rich and immersive story that adapts to your actions and decisions. Every choice you make has consequences, for good or ill. Will you be a hero or a villain? A leader or a follower? A friend or a foe? The choice is yours.
:
Explore a vast and dynamic world that changes according to your actions. You can shape the environment, influence the politics, and alter the history of the game world. But be careful, as every change has a ripple effect that may have unforeseen consequences.
:
Enter a world of wonder and danger, where you can find allies and enemies among the various races and factions that inhabit it. You can choose to join or oppose any of them, or forge your own path. The game world is alive and responsive to your actions.
___

For use with any fantasy roleplaying ruleset. Play the best game of your life!

![background image](https://i.imgur.com/dX3n2tK.jpeg){position:absolute,bottom:0,left:0,height:100%}

{{logo
![](/assets/naturalCritLogoWhite.svg)

Homebrewery.Naturalcrit.com
}}
```

G-Ambatte:
Drilling into this, it appears that the `backCover.png` asset has a 1px transparent line down the left edge.

<img width="1478" height="927" alt="Image" src="https://github.com/user-attachments/assets/458538d2-d024-4948-aedf-516247023771" />

We could tweak the `background-position-x` property to adjust the position: making it `-1px` seems to resolve the issue.

Alternatively, we could just fill the hole in the asset.

5e-Cleric:
This has been fixed

PR #4552, opened 2025-12-01
Fix backCover image asset

G-Ambatte:
This PR reslves #4551.

This PR moves the image data 1px to the left in the image, eliminating a 1px transparent gap that was (presumably) unintentionally left in the image.


dbolack-ab:
Do we need a test instance on this one?

calculuschild:
Yeap looks good. Thanks!

PR #4570, opened 2025-12-29
De-Imgur example images

dbolack-ab:
## Description

This relocates the handful of imgur based images to /assets. 

There *may* be value in one or more of these snippets using a full URL to make sure it is clear to users who are not yet familiar with Markdown how to construct the image markup.

### DIFF
--- a/themes/V3/Blank/snippets/licenseDTRPGCC.gen.js
+++ b/themes/V3/Blank/snippets/licenseDTRPGCC.gen.js
@@ -179,10 +179,10 @@ export default {
 		`;
 	},
 	// Verify Logo redistribution
-	monteCookLogoDarkLarge              : `![Cypher System Compatible](https://homebrewery.naturalcrit.com/assets/license_logos/CSCDarkLarge.png)`,
-	monteCookLogoDarkSmall              : `![Cypher System Compatible](https://homebrewery.naturalcrit.com/assets/license_logos/CSCDarkSmall.png)`,
-	monteCookLogoLightLarge             : `![Cypher System Compatible](https://homebrewery.naturalcrit.com/assets/license_logos/CSCLightLarge.png)`,
-	monteCookLogoLightSmall             : `![Cypher System Compatible](https://homebrewery.naturalcrit.com/assets/license_logos/CSCLightSmall.png)`,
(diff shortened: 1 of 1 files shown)


calculuschild:
I agree with the idea here. Some thoughts:

- Yes, a full URL is probably a good idea for these examples. It might actually be best to just make them all a full URL for consistency.
- We should rename the files while we are at it, instead of `4a8179462cfdf39054a418efd4cb743e.jpg`
- This is probably also the time to transition away from the AI-generated images. I made those on my own laptop with Stable Diffusion + Photoshop back before AI-gen became a more critical issue, but I think the common consensus between the devs now is we don't want that to be part of our brand. Do you guys think we can find some good "actually free" artwork to sub in here that still looks ok for the theme, and won't come with any licensing/hosting restrictions? @G-Ambatte @Gazook89 @5e-Cleric @dbolack-ab ?

5e-Cleric:
Without a doubt, now if everyone will agree on the quality, that's a different question i can't answer.

I'd ask you all again to review https://homebrewery.naturalcrit.com/share/WOd9MWxphMwo
attaching a preview as well just in case:

image


now, this is for the covers, if everyone agrees i'll start searching for assets for everything else.

This should not block the current PR, let's de-imgur the images and later swap them as we go.

calculuschild:
These are all really nice. Good finds.

Swapping the images should definitely block this though. We don't want to make a snippet that points to temporary images we intend to delete later.

Otherwise we get users with broken links in their brews, and we get complaints that our example images have stopped working.

5e-Cleric:
I'd use the assets used in this brew:

https://homebrewery.naturalcrit.com/share/VIkdYqx9_6LE

you may download the assets from my imagekit URLs, or i can provide them.

Word of advice: we probably want to downscale some of these, at least to the max quality output of the browser render process, 300DPI, which for a letter sized page is 2551 px × 3295 px, some of them are larger than that.

dbolack-ab:
Should I incorporate these things with this PR or is a separate one preferable?

5e-Cleric:
Put them in this one, no need for fuss with this one, should have done it months ago

calculuschild:
Alright, I'll go ahead and merge this. We'll just need to swap in the public domain art before deploying the next live version.

PR #4571, opened 2025-12-29
Fix Admin API tests

G-Ambatte:
The Admin API unit tests have been failing when run locally.

Investigation has shown that the new `dbCheck` middleware has been capturing the API calls because the database connection state is not ready (`1`). This is because the test file does not actually connect to the database at all, instead mocking all database calls.

This is NOT an issue when testing in CircleCI, as the `dbCheck` middleware contains an exception for test environments. However, when testing locally with the environment set to "local", the tests will fail.

This PR manually sets the `readyState` to `1` before each Admin API test, then restores the original value afterwards.


5e-Cleric (approved):
Seems alright to me

Issue #4581, opened 2026-01-09
Constant Message - "The server is out of sync with the saved brew. Please save your changes elsewhere, refresh, and try again."


Loneluxe:
### Error Code
/api/update/D1HNthxZDnPV:1  Failed to load resource: the server responded with a status of 409 ()
bundle.js:88786 Error Updating Local Brew
(anonymous) @ bundle.js:88786
bundle.js:87145 Error: {"message":"The server copy is out of sync with the saved brew. Please save your changes elsewhere, refresh, and try again."}

```
Error: {"message":"The server copy is out of sync with the saved brew. Please save your changes elsewhere, refresh, and try again."}
    at Request.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:82252:21)
    at Emitter.emit (https://homebrewery.naturalcrit.com/homebrew/bundle.js:1968:20)
    at XMLHttpRequest.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:82546:10)
{
  "status": 409,
  "method": "PUT",
  "url": "/api/update/D1HNthxZDnPV"
}
```
D1HNthxZDnPV:1 Failed to decode downloaded font: https://homebrewery.naturalcrit.com/themes/V3/fonts/5e/Nodesto%20Caps%20Condensed.woff2
D1HNthxZDnPV:1 OTS parsing error: invalid sfntVersion: 1008813135
bundle.js:87145 Error: {"message":"The server copy is out of sync with the saved brew. Please save your changes elsewhere, refresh, and try again."}
```

### Your description of what happened:

Been having a hiccup with one of my compendiums where no matter what I did, the message "The server is out of sync with the saved brew. Please save your changes elsewhere, refresh, and try again" kept displaying.  I went to the r/homebrewery subreddit and someone pointed at line 1095 stating the asterisk was causing the issue.  I removed the text on the line and couldn't save.  I removed the framework around the area and still couldn't save.  I copy and pasted the whole work into a new document and still couldn't save.  I am at a loss for what could be going on and the person aiding me pointed me here.  

Compendium Link: https://homebrewery.naturalcrit.com/share/9bCoZ6qEiid6

Line 1095 is under the Death Dragon (Young) statblock.

G-Ambatte:
[...]

PR #4583, opened 2026-01-11
Normalize brew text before hashing for saving

G-Ambatte:
## Description

This PR should resolve the issue with hash mismatch error causing Out of Sync error messages.

MD5 hashes were being generated in the EditPage from the editor's text before the entire brew object was sent to the `/update/` API route.
However, the API route compared the hash in the received data to a hash generated from a *normalized* version of the saved text (literally `text.normalize('NFC')`). In cases where the normalized version of the text differed from the original text, this would result in a different hash being created, and thus cause a hash mismatch, even though the text was correct.

Discussed via Gitter.
Issue #4581 also provided a brew which was able to reproduce this error.

### ORIGINAL REPRODUCTION

**EDIT:** *Now outdated, see minimal reproduction below*

~~1. Clone this example brew: https://homebrewery.naturalcrit.com/share/IiLsN0Pyqv5I~~

~~2. Save.~~

~~You will immediately see that the Out of Sync error appears.~~

~~I will attempt to create a more minimal  reproduction.~~

### MINIMAL REPRODUCTION

1. Navigate to https://homebrewery.naturalcrit.com/new

2. Paste the following character into the Content Editor: `ē`

3. Save.

Once the Edit Page loads, you will immediately see the Out of Sync error appear.

### DIFF
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -216,7 +216,7 @@ const EditPage = (props)=>{
 			text      : brew.text.normalize('NFC'),
 			pageCount : ((brew.renderer === 'legacy' ? brew.text.match(/\\page/g) : brew.text.match(/^\\page$/gm)) || []).length + 1,
 			patches   : stringifyPatches(makePatches(encodeURI(lastSavedBrew.current.text.normalize('NFC')), encodeURI(brew.text.normalize('NFC')))),
-			hash      : await md5(lastSavedBrew.current.text),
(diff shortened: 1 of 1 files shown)


G-Ambatte (commented):
Marked as a P1 because it prevents brews from saving for reason that is apparent to the user.

dbolack-ab (approved):
Chased this down in parallel and came to the same fix.

calculuschild:
I can confirm this fixes the issue on the reported brew. Merging!

PR #4604, opened 2026-01-23
Vitreum-to-vite

5e-Cleric:
This project is here for me to track my own progress, and not because i need or can use a review, there's way too much stuff to do yet, its a mess as of right now.

I'll try and split whatever i can but i don't think its gonna be easy to review anyway.

Current version is stable, ALL site features should be tested.
So far i have barely encountered bugs, listed in the comment below.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9954,6 +9954,22 @@
         "url": "https://opencollective.com/mongoose"
       }
     },
+    "node_modules/mongoose/node_modules/gcp-metadata": {
+      "version": "7.0.1",
+      "resolved": "https://registry.npmjs.org/gcp-metadata/-/gcp-metadata-7.0.1.tgz",
+      "integrity": "sha512-UcO3kefx6dCcZkgcTGgVOTFb7b1LlQ02hY1omMjjrrBzkajRMCFgYOjs7J71WqnuG1k2b+9ppGL7FsOfhZMQKQ==",
+      "license": "Apache-2.0",
+      "optional": true,
+      "peer": true,
+      "dependencies": {
+        "gaxios": "^7.0.0",
+        "google-logging-utils": "^1.0.0",
+        "json-bigint": "^1.0.0"
+      },
+      "engines": {
+        "node": ">=18"
+      }
+    },
     "node_modules/mongoose/node_modules/mongodb": {
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/mongodb/-/mongodb-7.0.0.tgz",


5e-Cleric:
Issues i have noticed:
- Codemirror takes a while to load
- custom syntax highlighting doesn't load after the codemirror load, resizing is needed for it to load
- markdown rendering seems to be different than the live version, or some css is missing

calculuschild:
Just took a look at the deployment. At first glance looks pretty good. I'm not seeing the markdown highlighting delay yet.

Haven't looked at code yet.

5e-Cleric:
image

On first load, codemirror is not ready when the highlight function is added, and no update is fired until the componentDidUpdate fires, so there is no highlighting

5e-Cleric:
Adding here a comment to specify why each change is made:

- Imports updated in css files, since the shared css files need to be imported otherwise vite complains, this might be just in dev version, must check.
- admin.jsx and homebrew.jsx's exports changed to export default
- added main.jsx files as entries for props, standard for vite.
- absolute URLs made relative
- BrewRenderer and HeaderNav styles must be input in the html for the brew renderer iframe in prod to work
- trySaveRef initialized as null instead of the function which is not defined at the time of the assignment (no idea how this was working earlier)
- Brought two files from vitreum, headtags.js and a file this one imports, now are joined in one
- global object is now non-existent, as a temporary measure it is set as an alias for `window.__INITIAL_PROPS__`
- making all tests that import app.js async
- modernize snippet gen files and make them be more consistent
- swaps template.js for an index.html file

client/homebrew/pages/homePage/homePage.jsx
```diff
@@ -27,7 +27,8 @@ const { both: RecentNavItem } = RecentNavItems;
 
 
 // Page specific imports
-import { Meta }                               from 'vitreum/headtags';
+import Headtags   from '../../../../vitreum/headtags.js';
+const Meta = Headtags.Meta;
```
calculuschild, line 31: I noticed `newPage.jsx` does not make this change and still uses `import { Meta }`. Does newPage.jsx need to be updated as well?

package.json
```diff
@@ -154,10 +150,7 @@
     "jest-expect-message": "^1.1.3",
     "jsdom": "^28.1.0",
     "jsdom-global": "^3.0.2",
-    "postcss-less": "^6.0.0",
     "stylelint": "^16.25.0",
-    "stylelint-config-recess-order": "^7.3.0",
-    "stylelint-config-recommended": "^17.0.0",
```
calculuschild, line 160: Not sure the stylelint dependencies should be removed.

5e-Cleric: huh? i don't recall having done this, perhaps it came from master?  I'll look into it

client/homebrew/pages/homePage/homePage.jsx
5e-Cleric: The meta component is actually not being used in newpage, not sure why it is imported, probably a relic, i'll remove the import as its not being used.

5e-Cleric:
This is done, can't advance anything more, nor i should

server/admin.api.spec.js
```diff
@@ -551,43 +503,6 @@ describe('Tests for admin api', ()=>{
...
-
-				expect(response.status).toBe(500);
-				expect(response.body).toEqual({
-					HBErrorCode : '70',
-					code        : 500,
-					message     : `Cannot find a locked brew with ID ${testBrew.shareId}`,
-					name        : 'Brew Not Found',
-					originalUrl : `/api/lock/review/request/${testBrew.shareId}`
-				});
-			});
```
calculuschild, line 589: This handful of tests appears to have been fully removed. Can they be added back, or is there a particular reason for their removal?

5e-Cleric: Uh, did i lose a test? damn, working on it.

package.json
calculuschild: This still needs to be fixed before merge.

package.json
```diff
@@ -12,10 +12,9 @@
     "url": "git://github.com/naturalcrit/homebrewery.git"
   },
   "scripts": {
-    "dev": "node --experimental-require-module scripts/dev.js",
-    "quick": "node --experimental-require-module scripts/quick.js",
-    "build": "node --experimental-require-module scripts/buildHomebrew.js && node --experimental-require-module scripts/buildAdmin.js",
-    "builddev": "node --experimental-require-module scripts/buildHomebrew.js --dev",
+    "start": "node server.js",
+    "build": "vite build",
+    "preview": "vite preview",
```
calculuschild: Which one should be used for dev work then? "Preview"?

server/admin.api.spec.js
5e-Cleric: i'm so blind

package.json
5e-Cleric: start, i should remove preview since its useless for us.

calculuschild:
Alright, so I've looked at all the files. Besides the stylelint dependencies that still need to be restored, I don't see anything else that looks out of place.

I do have some observations and questions while trying to run this locally:

1) Previously, we had a "production" mode and a "dev" mode. They did a lot of the same thing, but dev mode would auto-detect and restart the server when files changed. Do we still have the same functionality here? Are both actions now just sharing the common "build" and then "start" scripts? I don't fully follow the Vite way of working.

2) I notice the "build" script now produces a lot more additional files in the `/build` folder, with some weird auto-generated filenames. I'm not certain if this is normal for Vite build or if all of these are needed?

That said, I don't think either of these are *breaking* any functionality that I could find, so I think after fixing Stylelint we should be ok to merge, and we can fiddle more later.

image

calculuschild (approved):

5e-Cleric:
Alright, adressing the last comments:

## First question

> 1. Previously, we had a "production" mode and a "dev" mode. They did a lot of the same thing, but dev mode would auto-detect and restart the server when files changed. Do we still have the same functionality here? Are both actions now just sharing the common "build" and then "start" scripts? I don't fully follow the Vite way of working.

TLDR: Yes we have the same functionality in 2 commands instead of 5.

Previously we had the following:
- `build`, which bundled the code into loadable chunks
- `start`, which ran server.js, serving the site without any dev reload functions
- `dev`, which.. i've never used, i seem to recall it was broken
- `quick`, ditto
- `builddev`, which bundled the code AND launched liveReload

Now we have:
- `build`, which does the same, bundles the code, even if differently
- `start`, which still runs server.js, and serves the site.

Now start includes vite's ***Hot Module Reload*** when run locally, which makes it effectively a dev script when run locally. It will import the actual jsx files too.

The main source of confusion is i think that the `start` command works differently locally than it does in prod. Locally it imports all the files separately and does what live reload did, but better. In prod, it uses the bundle so everything is neat and tidy.

It removes complexity, specially so many custom scripts that all did very similar things(dev.js, quick.js), for no very good reason. 

### What is this Hot Module Reload?
Its a vite feature, an api that react leverages to do what we did with livereload, specifically the official [vite-plugin-react](https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react) uses [react-refresh](https://www.npmjs.com/package/react-refresh) to update the modules. Quoting from their npm page:

> This package implements the wiring necessary to integrate Fast Refresh into bundlers. Fast Refresh is a feature that lets you edit React components in a running application without losing their state. It is similar to an old feature known as "hot reloading", but Fast Refresh is more reliable and officially supported by React.


---

## Second question

> 2. I notice the "build" script now produces a lot more additional files in the `/build` folder, with some weird auto-generated filenames. I'm not certain if this is normal for Vite build or if all of these are needed?

The build process is handled automatically by Vite, vite uses Rollup. As far as i can tell we can customize the building process and configure how chunks are split, using the vite.config.js file.
There's more information about this in the [Docs](https://vite.dev/guide/build). Personally, i'd leave it as is and revisit it if we want to go into specifics, but i know so very little about the topic i don't feel confident explaining much. If you want to explore that world, be my guest.

Based on the contents of said files, i believe they are either part of our code editor or a dependency that is now split because the new bundler disagrees with the previous bundler on how to split things.

> That said, I don't think either of these are _breaking_ any functionality that I could find, so I think after fixing Stylelint we should be ok to merge, and we can fiddle more later.


I agree, should be all done by end of day.

calculuschild:
Ok, merging this! Huge effort to make this work, which is very much appreciated. 💯 ⭐ 🎉 

Feels great to be using a modern bundler that is actually maintained.

Issue #3712, opened 2024-09-05
\column break is not working, even on a new blank brew


toddgallant:
### Renderer

v3

### Browser

Firefox

### Operating System

MacOS

### What happened?

when using a \column break it is not work
<img width="2015" alt="image" src="https://github.com/user-attachments/assets/c5d3ead6-0782-48ab-9b6e-15ed9a8e01d3">

### Code

```gfm
\page
test

\column

test
```

G-Ambatte:
Correct.

The very _minute_ that Firefox supports column breaks, it will work on Homebrewery.

toddgallant:
👍

Issue #2473, opened 2022-10-28
Column break not working


Gozer-the-Gozerian:
I am using the /column command directly from the text editor window, but it is not causing a column break in the sheet.

Gazook89:
Which browser are you using? It won’t work in Firefox. 

Also, should be `\column`

Gozer-the-Gozerian:
Ahh...that was it.  Thanks!

Issue #543, opened 2017-11-02
Column Break Doesn't Work


IGJoshua:
So it seems like the column break isn't doing what it's supposed to do. And by that I mean it's not doing anything.

Perfect example is the first page when you first open it up. The line saying `Big things coming in v3.0.0` should appear on the right column, but it doesn't.

### Additional Details

**Brew code to reproduce** : <details><summary>Click to expand</summary><code><pre>

Some test text.

\`\`\`
\`\`\`

Some more test text.

</pre></code></details>

IGJoshua:
I've since discovered that this issue exists primarily in FireFox. Everything works as expected in Google Chrome.

calculuschild:
That makes sense, since only Chrome is supported.

PadgeM:
I'm running Chrome and it's not working for me.  It does the column break but it ends up destroying the note behind it.

calculuschild:
@PadgeM Do you still notice this issue? Do you have a specific code example or screenshot? I am unable to reproduce this problem and will be closing the issue unless we hear more.

PR #4615, opened 2026-01-31
Fix column breaks on Firefox

A1Asriel:
## Description

This PR adds a simple (and kinda stupid) solution to the longstanding column break issue on Firefox. Since it doesn't support the native `column-break-after` CSS property, this adds a 100vh bottom margin to `.columnSplit` elements and fill all available space to push everything below to the next column. Empty `@-moz-document url-prefix()` at-rule prevents Chromium and other browsers from enabling this hack.

It doesn't seem to break anything - both web version and print preview look fine and similar to the Chromium rendering (except for font rendering differences).

Solves (couldn't find an open issue for this bug):
- #543
- #2473
- #3712

*Current VS fixed version on Firefox Nightly 149.0a1 (2026-01-31)*
firefox

*Current VS fixed version on Chromium 146.0.7663.0 (Developer Build)*
chrome

Note: the only change I made to the document in the fixed version was replacing Imgur links with another host as it was throwing 403 in the locally hosted app for some reason.

### Reviewer Checklist

  - [x] Column breaks are working on Firefox
  - [ ] Page layout is not affected on Chromium
  - [ ] No side effects arise on Firefox


Gazook89:
Sorry, I can't pull it down right now, but does it continue to work when a PDF is generated?

client/homebrew/brewRenderer/brewRenderer.less
```diff
@@ -59,6 +59,12 @@
 		}
 		&-corner { visibility : hidden; }
 	}
+
+	@-moz-document url-prefix() {
+		.columnSplit {
+			margin-bottom: 100vh;
+		}
+	}
```
5e-Cleric, line 67: We shall not use the non-standard and deprecated syntax `moz-document`, instead the supports at-rule is made for this use.
```suggestion

	@supports (break-after:always) {
		.columnSplit {
			margin-bottom: 100vh;
		}
	}
```

dbolack-ab: Should this be a `@supports not (break-after: always) {` ?

5e-Cleric: That would make more sense, but apparently chrome only supports `-webkit-column-break-after`, for some reason, and firefox supports `break-after` only, (still ignores the hell out of it tho)

dbolack-ab: No, it should not. :)

dbolack-ab: Both the original and suggested change work delightfully well for me locally.

A1Asriel: > We shall not use the non-standard and deprecated syntax `moz-document`, instead the supports at-rule is made for this use.

Huh, I didn't know about this rule, but I suppose it fits better! Testing it showed the exact same results as well.

A1Asriel:
> Sorry, I can't pull it down right now, but does it continue to work when a PDF is generated?

Yes it does. The left one is generated via Firefox and the right one - through Chromium, for reference.

5e-Cleric:
Simple to review and test, merging now, thanks for the contribution @A1Asriel

PR #4635, opened 2026-02-15
Refactor-tag-system

5e-Cleric:
This PR remakes the tagInput component, including the following new features:

- Input pattern and regex validation
- tag suggestion system from a manually filled list of over 200 options, all of which highly used in our database at this moment
- Normalized keywords such as dnd or P2E into correct forms
- tags are now properly trimmed

the file `tagSuggestionList.js` is being kept as reference and should be deleted before merge, the file `curatedTagSuggestionList.js` is the correct file.


5e-Cleric:
> Works pretty well!
> 
> I have just noticed these issues:
> 
> 1) The default sorting for some reason places Meta:V3 and Meta:Legacy at the end, separate from the other Meta tags. Not sure if this is intentional? 

> Image
> 
> 2) You can't enter any of the System tags with `&` or `/` in the name, like `system: D&D Basic`. It just does nothing if you click it. I assume this is due to the valuePatterns filter.
> 
> 3) The curated tag names are not consistent in their capitalization. Can they be normalized?
> 
> Image
> 
> 4) The tags from `tagSuggestionList.js` don't appear in the list. That's ok for now but just in case it wasn't intended.

The list has been manually curated into another file, reduced from 2000 down to 200, its been manually sorted, that is why 1, 3 and 4 happen, not much people care to tag according to renderer, so i see no reason for them to be at the top

2 is an actual issue i'll take care of 

as for 3, we need some capitalization rules, do we force everything into lowercase? or just each word? just all words but the first?
I can normalize the results and i can edit the list but myself i don't know which should be the norm.

5e-Cleric:
all 1 2 and 3 are taken care of, i refer you to my now updated PR description last line about issue number 4

calculuschild:
With tests passing now I think it's good to merge! 💯  🎉  ⭐

PR #4646, opened 2026-02-23
Custom Snippets: Allow Inline

Gazook89:
## Description

Removes the additional line after a snippet, so it doesn't effectively create a `\n` at the end of the snippet.  This allows snippets to be inline with text.

Discussed in Gitter Feb 21 2026

To see current behavior in the live site, create a new document and a custom snippet:
```
\snippet test 1
{{red car}}
\snippet test 2

{{blue
van
}}

```
Write in your brew a sentence, and in the sentence add the *test 1* snippet.  It will break the sentence to a new line after the snippet.
```
Here is {{red car}}
my sentence.
```
Now, with this PR, do the same, and see the result is 
```
Here is {{red car}} my sentence.
```

### Reviewer Checklist

- [x] Verify new features are functional
  - [x] inline/single line snippets now no longer break lines
- [x] Verify old features have not broken
  - [x] multi-line snippets still work as expected
- [x] Test for edge cases / try to break things

### DIFF
--- a/shared/helpers.js
+++ b/shared/helpers.js
@@ -19,7 +19,7 @@ const brewSnippetsToJSON = (menuTitle, userBrewSnippets, themeBundleSnippets=nul
 						userSnippets.push({
 							name : snippetName,
 							icon : '',
-							gen  : snipSplit[snips + 1],
+							gen  : snipSplit[snips + 1].replace(/\n$/, ''),
 						});
 					}
 				}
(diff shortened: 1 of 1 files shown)


dbolack-ab:
I think this may need to test for multiline and not remove it for those.

Gazook89:
> I think this may need to test for multiline and not remove it for those.

I had an oversight (testing against live site, and something else).  Fixed my error and squashed the commits.  This should work now, tested on local.

Doesn't impact multi-lines, as it now just removes the last `\n` in the snippet gen property.

5e-Cleric:
@calculuschild can we get a deployment?

Gazook89:
It is possible to just `checkout` a branch by pulling it down to your local machine.  Or, this is a single line of change (repeated for local and theme snippets), so just amending your local `master` for a minute would do the trick, without a heroku deployment.

Issue #4655, opened 2026-02-28
Outdated CSS syntax highlighting in codemirror


Gazook89:
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

Codemirror's syntax highlighting for the CSS mode is outdated, and per Marijn, not likely going to be fixed since the focus has been on CM6 for awhile now.  We can move to CM6 someday, too, but maybe in the meantime we can use [`patch-package`](https://www.npmjs.com/package/patch-package) and make little tweaks ourselves.  

Some bad highlights (lowlights?):

- native nesting (both `&` prefixed and non-prefixed)
- `corner-shape` property and `scoop` value (chromium)
- `text-box` property
- `initial-letter` property
- `:is()` and `:where()` selectors   (their internal selectors show as errors)
- `:nth-child()` marks `even` and `odd` as errors
- probably more...

`patch-package` lets you work in the node_module file where you are making fixes.  Then run patch-package and point it to that file, and it generates a patch file in a `patches/` directory at the root.  Every time you run `npm install` it then makes the changes to the affected file automatically post install.  

This is coming up because a user in Discord was unsure why things were showing as errors (all red) when in fact they worked.  Bothered me for awhile now too.

### Code

```gfm

```

5e-Cleric:
@calculuschild can we get a quote on when do you expect to have CM6 ready?

calculuschild:
LMAO. I have not started anything relating to CM6.

I vaguely remember maybe Jeddai or someone had made some initial attempt years ago? But as far as I know there is no active branch with CM6 work at the moment.

Gazook89:
I think the confusion is perhaps the feeling that the Share and Edit merge needed to happen before CM6 happened.

5e-Cleric:
I wish there was a "this" emoji, yeah the merge should happen before CM6 otherwise we ought to implement the editor twice

5e-Cleric:
Alright nevermind, #4689 can close this.

PR #4657, opened 2026-03-01
Patch for CM5 native CSS nesting highlighting

Gazook89:
## Description

As described in issue #4655 CM5 doesn't highlight nested selectors.  

The CSS mode included with CM5 has two sub-modes, `x-LESS` and `x-SCSS`, and they have a `allowNested` property set to true.  So I took that and popped it into the base CSS mode as well, and it worked. 

This adds `patch-package` to the dev dependencies, and after any `npm install` it patches `node_modules/codemirror/mode/css` to modify the CSS mode to include `allowNested` option, set to `true`.  

Could this be done in a cleaner way, perhaps by modifying the language mode within our editor instance?  Maybe.  I tried.  I couldn't find where I could access or add a `allowNested` option.  

There are a few other things we could modify in the css mode in the same way like newer css properties, selectors, etc.  This is just a trail balloon, though.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,6 +75,7 @@
         "jest-expect-message": "^1.1.3",
         "jsdom": "^28.1.0",
         "jsdom-global": "^3.0.2",
+        "patch-package": "^8.0.1",
         "postcss-less": "^6.0.0",
         "stylelint": "^16.25.0",
         "stylelint-config-recess-order": "^7.3.0",
@@ -175,7 +176,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       },
@@ -199,7 +199,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=20.19.0"
       }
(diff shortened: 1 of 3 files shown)


calculuschild:
Patching the npm package feels a bit sketchy. I can see the CSS mode doesn't expose the allowNested option, but can we do something like this in our codeMirror setup to kind of wrap the old CSS mode in a custom mode and change it there?

```
// Get original mode
const origCssMode = CodeMirror.modes.css;

// Override it with a version that exposes the allowNested mode
CodeMirror.defineMode("css", function(config, parserConfig) {
  const mode = origCssMode(config, parserConfig);
  mode.allowNested = true;        // <- see if something like this works?
  return mode;
});
```

Gazook89:
Yes that seems to work just fine.  I tried to figure something out like that, but couldn't get it work, so thanks.  Feel like a dummy.

calculuschild:
> Yes that seems to work just fine. I tried to figure something out like that, but couldn't get it work, so thanks. Feel like a dummy.

Ok, in that case could you remove the patch-package dependency and then we can merge this?

PR #4665, opened 2026-03-07
Removing Systems

5e-Cleric:
Prior work has gone into this project, this completes the task for now, removing all references to systems in our UI and stop the creation of brews with such tag.

It also allows the manual change in the database  of tags and systems by stopping the override from the text field we do.

Next step in this PR would be to move systems to tags as they come in, but i haven't made it work yet.

### DIFF
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -43,7 +43,7 @@ const migrateSystemsToTags = (brew) => {
 		'5e': 'system:D&D 5e',
 		'4e': 'system:D&D 4e',
 		'3.5e': 'system:D&D 3.5e',
-		'Pathfinder': 'system:Pathfinder'
+		'Pathfinder': 'system:Pathfinder 2e'
 	};
 	const systemTags = brew.systems.map(s => systemMap[s]);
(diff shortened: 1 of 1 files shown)


5e-Cleric:
This all works, any opened brew has it systems moved into tags and the tags field deleted. Could use a review before merging.

server/homebrew.api.js
```diff
@@ -31,6 +31,19 @@ const isStaticTheme = (renderer, themeName)=>{
 // 	});
 // };
 
+const migrateSystemsToTags = (brew) => {
+	if (!('systems' in brew)) return brew;
+	if (!Array.isArray(brew.systems) || brew.systems.length === 0) {
+		brew.systems = undefined;
+		return brew;
+	}
+	const systemTags = brew.systems.map(s => `system:${s}`);
```
ericscheid: These systems values should be tagged as follows

- 5e → system:D&D 5e
- 4e → system:D&D 4e
- 3.5e → system:D&D 3.5e
- Pathfinder → system:Pathfinder 2e

That way they get converted to the canonical forms that we auto-suggest to authors, and we don't split the taxonomy unnecessarily.

PR #4669, opened 2026-03-10
Common render save button

### LINKED ISSUE #4675: cloning newpage issues
### What should happen
When a brew is cloned, the url should be `/new:id`
The save orange button should be displayed immediately

### What is happening
The save button is not appearing because there are no changes registered
The url only displays `/new/`

calculuschild:
More work toward making the save action more unified between `new`, `home`, `edit` pages.

This focuses specifically on reworking the **SAVE** button in the navbar to be identical, using parameters to switch between the differences rather than three unique implementations of the component.


Gazook89:
If they are going to be the same on each page, and just take arguments, can you split it out entirely as a component in a single file, and import it where needed?

Or, is this towards the ultimate goal of unifying the three pages themselves into one file (and thus not need to split this into it's own file)?

I reviewed what you have so far, and it seems fine, but not sure if you are going further since it's in draft status still.

calculuschild:
Superseded by other PRs toward New/Edit/Home merging. #4458

PR #4674, opened 2026-03-11
bring back error pages

5e-Cleric:
Brings back the error pages which were broken with vite spa


5e-Cleric:
Thanks @G-Ambatte for the fix on this via gitter chat.

PR #4676, opened 2026-03-11
swap ai assets for public domain 

5e-Cleric:
This PR swaps our ai art on the site for public domain art, i am unsure about a few assets (inside cover & watercolor mask assets

### DIFF
--- a/themes/V3/Blank/snippets/imageMask.gen.js
+++ b/themes/V3/Blank/snippets/imageMask.gen.js
@@ -6,8 +6,7 @@ export default {
 	center : ()=>{
 		return dedent`
 			{{imageMaskCenter${_.random(1, 16)},--offsetX:0%,--offsetY:0%,--rotation:0
-			  ![image description](${domain}/assets/roman_theatre.webp){height:100%}
-			  <!-- Art: The Roman Theater at Taormina, 1828, by Louise-Joséphine sarazin de Belmont -->
+			  ![The Roman Theater at Taormina, 1828, by Louise-Joséphine sarazin de Belmont](${domain}/assets/roman_theatre.webp){height:100%}
 			}}
 			<!-- Use --offsetX to shift the mask left or right (can use cm instead of %)
 			     Use --offsetY to shift the mask up or down


5e-Cleric:
image
Not sure what the final decision was with these, i could use some advice.

calculuschild:
> i am unsure about a few assets (inside cover & watercolor mask assets/

I think these all look great. If this is complete then I am ok with merging this. Is it ready?

5e-Cleric:
Before merging we need to agree on attribution method:

1. Attribution as alt text 
`Name, year and author`
2. Attribution as a comment after the image
```
![alt text](link)
<!-- Name, year and author -->
```

themes/V3/5ePHB/snippets/coverpage.gen.js
```diff
@@ -143,10 +147,11 @@ export default {
 
 			For use with any fantasy roleplaying ruleset. Play the best game of your life!
 
-			![background image](https://homebrewery.naturalcrit.com/assets/shopvials.jpg){position:absolute,bottom:0,left:0,height:100%}
+			![background image](${domain}/assets/ruined_tower.webp){position:absolute,bottom:0,right:-250px,height:100%}
+			<!-- Art: Italian Coast Scene with Ruined Tower, 1838, Thomas Cole -->
```
calculuschild: For consistency do we want "***by*** Thomas Cole" like the other attributions?

5e-Cleric: yeah, got me there

calculuschild:
> Before merging we need to agree on attribution method:

Hm good question. Let's put it to a vote in the Gitter chat.

As the image Alt Text it's just kind of convenient, and shows the user how they can utilize the alt text as a label of sorts. Adding the HTML comment might be confusing and "cluttery" as it introduces another extraneous element.

On the other hand, the attributions are a bit long which might look overwhelming in the Alt Text when paired with a long URL on the same line.

I *think* I would lean more toward putting it in the Alt Text but I don't have a strong opinion.

Poll here: https://matrix.to/#/!IuzKyDJxDuanSbtTQt:gitter.im/$bXnuCxzkYdEWrAaKfzZoPOb-7xgiOOHIS9zSt2LfsiA?via=gitter.im&via=matrix.org&via=integrations.ems.host

ericscheid:
Maybe also slip in "Public Domain", as a gentle and subtle reminder to not steal art.

calculuschild:
From the poll, it looks like majority favors option 1 of putting it in the Alt Text.

5e-Cleric:
Merging then! AI images, good riddance!

PR #4708, opened 2026-03-31
Make splitTextStyleAndMetadata async

dbolack-ab:
## Description

When loading particularly large brews, splitTextStyleAndMetadata does not have time to complete before the brew gets passed along, introducing errors. ( https://homebrewery.naturalcrit.com/share/MVz83vplntkj ) 

This should solve for that occasion.


5e-Cleric:
This looks fine, i'll merge this in a few days, just waiting in case someone has an objection.

i got a question, if it is async, won't the rest of the process keep going before this function arrives? i know this might be a dumb question.

calculuschild:
> i got a question, if it is async, won't the rest of the process keep going before this function arrives? i know this might be a dumb question.

Adding async does two things: It makes the function return a promise instead of the immediate result, so yes the rest of the thread can keep running and the promise may or may not be ready in time. The second thing is it allows us to use "await" inside. Once we do that, it still returns a promise but the thread has to wait for us.

So in our case it's really mainly useful not to make the external function async, but to force a wait on an internal function.

What I'm not clear on with this PR is how awaiting `splitTextStyleAndMetadata()` would help, since that function is already fully synchronous. @dbolack-ab can you confirm which loading step is running out of time? Do you know which outer function the error occurs in?

dbolack-ab:
> What I'm not clear on with this PR is how awaiting `splitTextStyleAndMetadata()` would help, since that function is already fully synchronous. @dbolack-ab can you confirm which loading step is running out of time? Do you know which outer function the error occurs in?

It happens in the baseline brew load from "disk."

G-Ambatte:
> > @dbolack-ab can you confirm which loading step is running out of time? Do you know which outer function the error occurs in?
> 
> It happens in the baseline brew load from "disk."

I'm not sure what making this async is intended to achieve, as `splitTextStyleAndMetadata` happens *after* the brew is loaded from any source - DB, local storage, Google Drive, text file, or anything else.

If the issue is a brew not loading properly, it'd be good to know where the file is being loaded from, and what route is being used to access it - can we reproduce the actual issue?

dbolack-ab:
> If the issue is a brew not loading properly, it'd be good to know where the file is being loaded from, and what route is being used to access it - can we reproduce the actual issue?

The problem is that when it occurs, parts of the HB front matter ( ` ```Css`, ` ```meta`, etc. ) aren't parsed out of `brew.text` before it is transferred to the editor.
 
It doesn't happen often - but splitTextStyleandMetadata needs to finish before the brew text is copied to the body.

5e-Cleric (approved):
I trust you to have found all references to the function, this is approved since it should fix how we work larger documents and not impact the rest.

G-Ambatte (changes requested):
I was able to reproduce the reported issue by injecting an artificial 10 second delay into `splitTextStyleAndMetadata`.

image
image

I believe that what has been identified here is a race condition between `splitTextStyleAndMetadata` and the React render of the page; by introducing a similar artificial delay into the asynchronous version in this PR, I received identical results.

Image

_although if I've messed up my delay code, please feel free to let me know_

In short, while this PR **HAS** identified a real issue, as best that I can determine, the provided fix does not actually resolve it.

PR #4713, opened 2026-04-04
Create baseline for HTML Export and Embedding

dbolack-ab:
## Description

This is the beginning of a series? of PRs for handling the myriad "we would like the HTML from the /share in a different shape" requests we've had over the years.

The end goal is to provide the following:

1. An API endpoint that provides the pages of the brew in such a way that HTML to PDF tools can correctly ingest the output to create a PDF.
2. An API endpoint suitable for embedding in webpages
3. An HTML file download of the `iframe` contents—with relative-path munging.
4. A zip file containing the HTML file and all assets. Paths will be adjusted to assets as placed in the archive. 

This PR:
* Changes the Get PDF top bar menu item to an Export menu. The export menu contains the relocated PDF export, a new entry for downloading the HTML file, and a new entry for downloading the zip file.
* Creates a /embed endpoint based on the old /share endpoint, mimicking the behavior of the old /print endpoint
* Creates functionality for the HTML download
* Stubs functionality for the HTML Zip download

Related Issues: #661, #802, #956

Notes: 
Offsite fonts will always be an issue with the HTML download due to CORS restrictions on fonts.

## Testing and Evaluating

I'm struggling to determine how best to test the two new features in this PR. Suggestions appreciated.

[Embed Test URL:](https://homebrewery-pr-4713.herokuapp.com/embed/_XHaNL9L0Vve)

## Problems still to be solved ( in later PRs )

* Ensuring all assets load on the endpoint in a way detectable by external scrapers ( playWright, etc. )
* Zip container download of rendered brew


client/homebrew/navbar/print.navitem.jsx
Gazook89, line 1: I would suggest picking one color for "Export", and it's submenu items, rather than a menagerie of colors for what are all related items.

client/homebrew/navbar/pdf.navitem.jsx
Gazook89, line 1: Is this better as just a part of the `print.navitem.jsx` file, rather than breaking out the pdf button out as it's own thing?

client/homebrew/homebrew.jsx
Gazook89, line 1: I think `share={true}` could be more descriptive.  The prop would be more clear as `navbar={true}`, since that is effectively what it is doing.   Saying the `/share` route is `share={true}` seems redundant and could be determined by a `window.url` check on the individual pages without passing that info through a prop.  Passing through a prop is fine, I just think the prop could have a more descriptive name.

Gazook89 (changes requested):
I'm just now actually looking at this PR, I was hoping someone smarter than me would jump in, but here we are.

Very surface level:  There are a couple changes that I'm not sure are necessary in this PR, and being the only changes on their files, it might be helpful to just revert these changes to reduce the "files changed" and make it more palatable to review.  If the changes *are* totally necessary, please ignore.
- headerNav.jsx
- codeEditor.jsx

Then there are a couple notes on individual files.  This is a first pass for me, I didn't dive deep into the guts quite yet.

Regarding testing:  Does this need to be on the live site in order to effectively test it?  Can the navbar changes be thrown behind a flag, made available via the browser console, and then see if we can get embeds working on our own blogs?

client/homebrew/navbar/print.navitem.jsx
dbolack-ab: Then they wouldn't be color-coded?

5e-Cleric:
@calculuschild can we get a deployment? And an opinion?

PR #4733, opened 2026-04-11
V4 metadata - Initial

dbolack-ab:
## Description

Adds new frontmatter brew metadata fields per discussion.

Uses defaults:

```
bleedSize:
  top: .125in
  bottom: .125in
  outer: .125in
  inner: .125in
safetySpace:
  top: .25in
  bottom: .25in
  outer: .25in
  inner: .5in
trimSize:
  width: 8.5in
  height: 11in
columns: 2
columnGutter: .125in
license: None
legalAuthors: 
```
**NOTE** The safe space and bleed have been adjusted to inner and outer from left and right to better represent print realities. Template writers will be responsible for handling the left vs right binding side.

DTRPG also recommends additional safe space for text, but I think we should leave that to the template writers.

#4709 

Ensure all necessary tests are in place


Gazook89:
Can this metadata be added now, in v3, but just not exposed via the interface at all?  Is there any risk to that?  I'm just thinking that if we can get this worked out and merged then it is one less thing that has to happen "all at once" with v4.  

----

One thing I was unsure about was "safetySpace", because I was thinking it was the same as "margin".  It's not, and for reference in case anyone is curious, [here is a link](https://www.greenerprinter.com/blog/understanding-bleeds-margins-and-safe-zones-in-printing/).  Basically, the space between the paper edge and the main content is the margin, and the space between paper edge and any critical graphic element (which might not be the main "content") is the "safe zone"-- meaning, don't put important stuff in that zone.

----

Should `columns: 2` be the default, or just a single column, and otherwise leave it to the templates/theme?

-----

Do you prefer this PR be narrowly focused on Page Layout related metadata, rather than ballooning it to include any other additions or deletions that may come about?

dbolack-ab:
> Can this metadata be added now, in v3, but just not exposed via the interface at all? Is there any risk to that? I'm just thinking that if we can get this worked out and merged then it is one less thing that has to happen "all at once" with v4.

I don't see a reason not to do so, other than demarcation. But I see that as a valid reason.

> One thing I was unsure about was "safetySpace", because I was thinking it was the same as "margin". It's not, and for reference in case anyone is curious, [here is a link](https://www.greenerprinter.com/blog/understanding-bleeds-margins-and-safe-zones-in-printing/). Basically, the space between the paper edge and the main content is the margin, and the space between paper edge and any critical graphic element (which might not be the main "content") is the "safe zone"-- meaning, don't put important stuff in that zone.

Safety space is the margin of error for the trim blades when the raw paper is cut. 

> Should `columns: 2` be the default, or just a single column, and otherwise leave it to the templates/theme?

I think all of these values should default to being inherited from the theme, and we'll need a mechanism for that, overriding, and resetting to the template.

> Do you prefer this PR be narrowly focused on Page Layout related metadata, rather than ballooning it to include any other additions or deletions that may come about?

Trying to bite-size as much as possible. :")

server/homebrew.api.js
```diff
@@ -220,6 +220,38 @@ const api = {
...
 		metadata.snippets = snippetsArray.length > 0 ? snippetsArray : undefined;
+
+		metadata.bleed = {
+			top    : brew?.bleed?.top || '.125in',
+			bottom : brew?.bleed?.bottom || '.125in',
+			inner  : brew?.bleed?.inner || '.125in',
+			outer  : brew?.bleed?.outer || '.125in',
+		};
+
+		// Copy Required Safe Space or Use defaults
```
5e-Cleric: these comments are unnecessary, the line speaks for itself, reduce to a comment about all of them, or remove completely

shared/helpers.js
```diff
@@ -93,6 +93,40 @@ const splitTextStyleAndMetadata = (brew)=>{
 		const metadata = yaml.load(metadataSection);
 		Object.assign(brew, _.pick(metadata, ['title', 'description', 'renderer', 'theme', 'lang']));
 		brew.snippets = yamlSnippetsToText(_.pick(metadata, ['snippets']).snippets || '');
+
+		// Copy Required Bleed Space or Use defaults
```
5e-Cleric: same as earlier, unnecessary comment

5e-Cleric (approved):

shared/helpers.js
```diff
@@ -93,6 +93,40 @@ const splitTextStyleAndMetadata = (brew)=>{
...
+		brew.bleed = {
+			top    : _.pick(metadata, ['bleed']).bleed?.top || '.125in',
+			bottom : _.pick(metadata, ['bleed']).bleed?.bottom || '.125in',
+			inner  : _.pick(metadata, ['bleed']).bleed?.inner || '.125in',
+			outer  : _.pick(metadata, ['bleed']).bleed?.outer || '.125in',
+		};
+
+		// Copy Required Safe Space or Use defaults
+		brew.safetySpace = {
+			top    : _.pick(metadata, ['safetySpace']).safetySpace?.top || '.25in',
```
calculuschild: Using `_.pick` here is redundant. `pick` just creates a copy of the metadata object with only the `bleed` field. We don't need to copy the object before grabbing the property.

```suggestion
			top    : metadata.bleed?.top || '.25in',
```

That said, I would propose moving all of these default settings over to `/server/brewDefaults.js` where we already made an effort to combine all the various places where we were setting defaults. The defaults themselves are applied in `server/homebrew.api.js` at the point of retrieving a brew, using `_.defaults`.

server/homebrew.api.js
```diff
@@ -220,6 +220,38 @@ const api = {
 		const metadata = _.pick(brew, ['title', 'description', 'tags', 'renderer', 'theme']);
 		const snippetsArray = brewSnippetsToJSON('brew_snippets', brew.snippets, null, false).snippets;
 		metadata.snippets = snippetsArray.length > 0 ? snippetsArray : undefined;
+
+		metadata.bleed = {
+			top    : brew?.bleed?.top || '.125in',
```
calculuschild: I would propose moving all of these default settings over to `/server/brewDefaults.js` where we already made an effort to combine all the various places where we were setting defaults. Then we don't have multiple possible sources of truth for the default values, and they only need to be applied in one common place instead of at multiple points along the journey.

server/homebrew.api.js
```diff
@@ -220,6 +220,38 @@ const api = {
...
+		metadata.safetySpace = {
+			top    : brew?.safetySpace?.top || '.25in',
+			bottom : brew?.safetySpace?.bottom || '.25in',
+			outer  : brew?.safetySpace?.outer || '.25in',
+			inner  : brew?.safetySpace?.inner || '.5in',
+		};
+
+		// Set brew paper trim size or default to US Letter
+		metadata.trimSize  = {
+			width  : brew?.trimSize?.width || '8.5in',
```
calculuschild: Can I propose just `trim`, to match with `bleed`?

calculuschild (changes requested):

server/homebrew.api.js
dbolack-ab: I think the longer form ( or paperSize ) makes it clearer on a fast read that this is a dimension, not a method call that lost its parentheses...

calculuschild: Thats fair, sure. Can we make bleedSize too or something then?

shared/helpers.js
```diff
@@ -93,6 +93,15 @@ const splitTextStyleAndMetadata = (brew)=>{
 		const metadata = yaml.load(metadataSection);
 		Object.assign(brew, _.pick(metadata, ['title', 'description', 'renderer', 'theme', 'lang']));
 		brew.snippets = yamlSnippetsToText(_.pick(metadata, ['snippets']).snippets || '');
+
+		brew.bleed = { top: metadata.bleed?.top, bottom: metadata.bleed?.bottom, inner: metadata.bleed?.inner, outer: metadata.bleed?.outer };
```
5e-Cleric: Do we need to redefine the object here? i may be missing something but it looks like just recreating the object instead of setting the brew.bleed object.

`brew.bleed = metadata?.bleed;` would work fine, no?

server/homebrew.api.js
dbolack-ab: Done.

shared/helpers.js
dbolack-ab: I'm always a bit fuzzy on when JS decides to copy vs reference, particularly with objects. 

That said, it'd probably be better to do

```
brew.bleedSize = {...metadata.bleedSize};
```

Yes?

5e-Cleric (approved):
Looks good, merging!

PR #4734, opened 2026-04-12
Add Navigation API check

G-Ambatte:
This PR adds a check to see if the Navigation API exists before attempting to hook an Event Listener to it.
In the event that the API does NOT exist, it instead creates a small function that runs once per second which checks if the current URL has changed; in the event that it has, it calls `scrollToHash` using the current URL's `hash` property.

This should resolve the issue that users have been reporting on Reddit and Discord of Homebrewery crashing when attempting to load on older browsers.


5e-Cleric (approved):
Looks fine.

calculuschild (approved):

Gazook89:
This seems to cause a lot of heartache-- can at least this fix be pushed live so the site stops crashing?  @calculuschild

PR #4740, opened 2026-04-19
Preload Marked Variables

dbolack-ab:
## Description

This aims to fix the hoisting issue with Markdown variables that are declared outside of the view range of page 1.

This might not be an optimal fix. I can see opportunities for tuning.


calculuschild:
Can you give an example use case to test this?

Gazook89:
> example use case

I think [this brew](https://homebrewery.naturalcrit.com/share/ZcLv6yOfbrEa) illustrates the issue, that if the variable is assigned a value far down a brew it's value isn't hoisted up.  I noticed that there *wasn't* an issue when the use of the variable was on page 1 (and the assignment was very far down), but there *was* an issue when I used the variable on page 11 (and the assignment was also very far down).

So this example has the variables at the end of page 11, and at the end of page 251 (second to last page).  variable is "car", value is "bmw".  

Apologies to @dbolack-ab if this is *not* a good example.

----

I did test this in the deployment and it works for me.  I haven't reviewed the code, though.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -233,7 +234,8 @@ const BrewRenderer = (props)=>{
 			renderedPages[props.currentEditorCursorPageNum - 1] = renderPage(rawPages[props.currentEditorCursorPageNum - 1], props.currentEditorCursorPageNum - 1);
 
 		_.forEach(rawPages, (page, index)=>{
-			if((isInView(index) || !renderedPages[index]) && typeof window !== 'undefined'){
+			const forceRender = checkHoists && (page.match(/([!$]?)\[((?!\s*\])(?:\\.|[^\[\]\\])+)\]/g));
```
Gazook89: It might be helpful in the future to break out this regex into a `const` with a comment about what it is.  I see now that it's for checking for *variable* syntax, but that comes easier with the context that this is a PR for fixing Variable related-- in the future that will be less forthcoming.  Doesn't need to be broken into it's own `const`, just an idea.

client/components/codeEditor/codeEditor.jsx
```diff
@@ -407,4 +407,4 @@ const CodeEditor = forwardRef(
 	},
 );
 
-export default CodeEditor;
+export default CodeEditor;
```
Gazook89: There doesn't seem to be any actual change in this file, just the line ending.

Gazook89 (approved):
Solves the stated issue for me (in Chrome and Safari).  Didn't see any side effects but didn't expect any just based on the small changes.

5e-Cleric (approved):
Code seems fine, with gazook's review, this might be the first PR from 2026 we can merge following the rules, lol.

5e-Cleric:
@dbolack-ab is this ready for merge?

dbolack-ab:
> @dbolack-ab is this ready for merge?

As best I am aware.

PR #4743, opened 2026-04-20
Add CSS formatter

5e-Cleric:
This PR includes Prettier as a CSS formatter, activated using 2 keybinds:
- <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>F</kbd>  (<kbd>Option</kbd> + <kbd>Shift</kbd> + <kbd>F</kbd> on Mac).
- <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>F</kbd>

### Icon button
image


5e-Cleric:
I could make it so if there is a selection made it only formats the selection, not sure if we want that.

5e-Cleric:
Its failing the jsdom in safehtml tests again, i don't get it:

```js
PS C:\Repositories\homebrewery> npm run test:safehtml

> homebrewery@3.21.0 test:safehtml
> jest tests/html/safeHTML.test.js --verbose

 FAIL  tests/html/safeHTML.test.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation, specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    C:\Repositories\homebrewery\node_modules\entities\dist\decode.js:1
    import { replaceCodePoint } from "./decode-codepoint.js";
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      1 | import globalJsdom from 'jsdom-global';
    > 2 | globalJsdom();
        |            ^
      3 | import safeHTML from '../../client/homebrew/brewRenderer/safeHTML';
      4 |
      5 | test('Exit if no document', function() {

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1314:40)
      at Object.require (node_modules/parse5/dist/tokenizer/index.js:4:1)
      at Object.require (node_modules/parse5/dist/parser/index.js:1:1)
      at Object.require (node_modules/parse5/dist/index.js:1:1)
      at Object.<anonymous> (node_modules/jsdom/lib/jsdom/browser/parser/html.js:3:16)
      at Object.<anonymous> (node_modules/jsdom/lib/jsdom/browser/parser/index.js:4:20)
      at Object.<anonymous> (node_modules/jsdom/lib/jsdom/living/nodes/Element-impl.js:10:27)
      at Object.<anonymous> (node_modules/jsdom/lib/generated/idl/Element.js:3720:14)
      at Object.<anonymous> (node_modules/jsdom/lib/jsdom/living/interfaces.js:21:12)
      at Object.<anonymous> (node_modules/jsdom/lib/jsdom/browser/Window.js:6:31)
      at Object.<anonymous> (node_modules/jsdom/lib/api.js:15:26)
      at globalJsdom (node_modules/jsdom-global/index.js:28:15)
      at Object.<anonymous> (tests/html/safeHTML.test.js:2:12)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        1.836 s
Ran all test suites matching tests/html/safeHTML.test.js.
(node:26340) Warning: `--localstorage-file` was provided without a valid path
(Use `node --trace-warnings ...` to show where the warning was created)
PS C:\Repositories\homebrewery> 

```

Gazook89:
I suggest making this into a button in the Style Editor snippet bar, like the "editor tools" of the Brew Editor.  Shortcut-only fuctionality is effectively hidden from probably the vast majority of users.  And using a 3-key shortcut is always more difficult than it needs to be.  I would even skip adding hotkeys for it at all.

It can be a button up near the top, and possibly presented in a way that can display a toggle for "auto formatting" either on Save or when switching to another editor tab or similar.  The toggle could be side-by-side with the format button, or a "..."/"cog" icon button up there to open a little "anchored" menu (akin to the settings button in the Preview Toolbar) with additional settings.

5e-Cleric:
> I suggest making this into a button in the Style Editor snippet bar, like the "editor tools" of the Brew Editor. Shortcut-only fuctionality is effectively hidden from probably the vast majority of users. And using a 3-key shortcut is always more difficult than it needs to be. I would even skip adding hotkeys for it at all.
> 
> It can be a button up near the top, and possibly presented in a way that can display a toggle for "auto formatting" either on Save or when switching to another editor tab or similar. The toggle could be side-by-side with the format button, or a "..."/"cog" icon button up there to open a little "anchored" menu (akin to the settings button in the Preview Toolbar) with additional settings.

I was thinking a button on the divider, like the arrows, perhaps that ought to be a full bar with buttons too. tagging @ericscheid, our UI expert

Gazook89:
The arrow buttons are buttons that interact across the editor and the preview. This prettier button functions only on editor, and at this point, only on one particular editor. 

We have tools that operate on the editor like undo/redo, fold, etc already. 

I understand that there may be concern about "one more button" in the snippet bar. That is reasonable. I do think the answer to that is just a rework of the snippet bar, though, rather than just trying to tack buttons elsewhere.

5e-Cleric:
> The arrow buttons are buttons that interact across the editor and the preview. This prettier button functions only on editor, and at this point, only on one particular editor.
> 
> We have tools that operate on the editor like undo/redo, fold, etc already.
> 
> I understand that there may be concern about "one more button" in the snippet bar. That is reasonable. I do think the answer to that is just a rework of the snippet bar, though, rather than just trying to tack buttons elsewhere.

well you open that issue so we can discuss it, meanwhile i'd like to have this function be there, button or not. What i sure don't want is for this already made PR to stop and wait for the snippetbar rework.

calculuschild:
We can add the function without a button for now. That's a solid chunk on its own that we can lock in.

Then in a followup PR I do agree with Gazook though that eventually a button should be added rather than hotkey only (or remove the hotkey). By the undo/redo buttons seems like the most logical place. The divider is for interaction between the panels.

That said I don't consider myself a UI expert at all.

Gazook89:
A tricky bit with hotkey-only is finding the hotkey that works:  `cmd` `shift` `f` is the hotkey for "Always Show Toolbar in Fullscreen" in the View menu of Chrome on MacOS.  

Tried testing this PR and couldn't get it to work, after also trying to figure out how to "shortcut capture" or prevent the browser from taking that command.

Gazook89:
I wonder, does Chrome have the same hotkey collision on Windows?  I guess not, if you aren't having this problem.  But does Chrome have a *Always Show Toolbar in Fullscreen*  option on Windows, and if so, is it a different hotkey or just have no hotkey?

I'm wondering if you can test the basic problem (hotkey overlap) and the possible solution (hotkey capture), just with a different combo.

5e-Cleric:
> I wonder, does Chrome have the same hotkey collision on Windows? I guess not, if you aren't having this problem. But does Chrome have a _Always Show Toolbar in Fullscreen_ option on Windows, and if so, is it a different hotkey or just have no hotkey?
> 
> I'm wondering if you can test the basic problem (hotkey overlap) and the possible solution (hotkey capture), just with a different combo.

have never used fullscreen

Gazook89:
> have never used fullscreen

me neither, but the option exists in the menus regardless, on mac.  I'm wondering if Windows Chrome has the same options, and if they have the same hotkey (obvioulsy `ctrl` and not `⌘`)

image

5e-Cleric:
This is working and now has a rudimentary animation and a snippetbar button.

@Gazook89 any idea for a better shortcut than <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>F</kbd>? would you prefer <kbd>Shift</kbd> + <kbd>Alt</kbd> + <kbd>F</kbd>?

moving it to that.

5e-Cleric:
Auto formatting can come later on another PR

client/components/codeEditor/customKeyMaps.js
```diff
@@ -45,7 +81,8 @@ const wrapSelection = (prefix, suffix)=>(view)=>{
 	}
 
 	view.dispatch({
-		changes
+		changes : { from, to, insert: text },
```
G-Ambatte: Is this functional? `text` is defined inside the `for` loop so it seems like it should be out of scope here.
As far as I can tell, with this change the `changes` array is never actually used?

client/components/codeEditor/customKeyMaps.js
```diff
@@ -45,7 +81,8 @@ const wrapSelection = (prefix, suffix)=>(view)=>{
...
+		selection
```
G-Ambatte: Where is `selection` defined at this point?

client/homebrew/editor/editor.jsx
```diff
@@ -333,6 +337,7 @@ const Editor = createReactClass({
 					redo={this.redo}
 					foldCode={this.foldCode}
 					unfoldCode={this.unfoldCode}
+					formatCode={this.isStyle() ? this.handleFormatCode : null}
```
G-Ambatte: It may be beneficial to have the tab check inside the `handleFormatCode` function, in case we ever decide to have a similar function for any other tab (i.e. snippet tab?)

package-lock.json
```diff
@@ -190,6 +191,7 @@
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.0.tgz",
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "license": "MIT",
+      "peer": true,
```
G-Ambatte: Not about this PR specifically, but these `peer: true` lines are due to a bug in NPM 11.6.1 and 11.6.2, and was corrected in 11.6.3. Unfortunately, 11.6.2 was bundled with Node 24 by default.
If you update to at least 11.6.3 and run `npm i` afterwards, you should see these lines removed.

client/homebrew/editor/editor.jsx
```diff
@@ -154,6 +154,10 @@ const Editor = createReactClass({
 		this.codeEditor.current?.injectText(injectText);
 	},
 
+	handleFormatCode: function () {
```
G-Ambatte:
```suggestion
	handleFormatCode : function () {
```

G-Ambatte: To keep the Linter happy

client/components/codeEditor/customKeyMaps.js
G-Ambatte: Doesn't appear to be functional

G-Ambatte (commented):
Currently breaks the `wrapSelection` function, probably unintentionally?

---

As best I can determine, the Style Editor cleaning function works as intended.

On the UI front, this PR raises the question of how to handle editor tools that do not function on certain editor tabs.

G-Ambatte (changes requested):

client/homebrew/editor/editor.jsx
5e-Cleric: No, this is how the snippetbar determines if the function is disabled, i didn't make it so. (or maybe i did and can't remember).

The button checks if the function exists, and if it does it renders the button enabled.

client/components/codeEditor/customKeyMaps.js
5e-Cleric: I had solved wrapselection twice in two PRs, this is them conflicting and bad conflict resolution on my part, i am removing all of these changes.

client/components/codeEditor/customKeyMaps.js
5e-Cleric: I had solved wrapselection twice in two PRs, this is them conflicting and bad conflict resolution on my part, i am removing all of these changes.

5e-Cleric:
@Gazook89 a secondary keybind should solve all our problems, right? or should i remove the Ctrl one? does codemirror or chrome's browser on mac win on precedence?

client/homebrew/editor/editor.jsx
G-Ambatte: This has previously been achieved by checking the value of the `view` prop passed to the SnippetBar component - in this case, `this.props.formatCode ? 'active' : ''` would instead be `this.props.view == 'style' ? 'active' : ''`.

Overall, this is a minor stylistic thing to make the new code consistent with the other existing buttons/tabs. The new functionality works and the style code gets formatted when the button is clicked.

G-Ambatte (changes requested):

client/homebrew/editor/editor.jsx
5e-Cleric: Teh code tools alongside it literally follow the same format:

```jsx
<div className='codeTools'>
    <div className={`editorTool foldAll ${this.props.foldCode ? 'active' : ''}`}
         onClick={this.props.foldCode} >
         <i className='fas fa-compress-alt' />
    </div>
    <div className={`editorTool unfoldAll ${this.props.unfoldCode ? 'active' : ''}`}
         onClick={this.props.unfoldCode} >
         <i className='fas fa-expand-alt' />
    </div>
    <div className={`editorTool formatCode ${this.props.formatCode ? 'active' : ''}`}
         onClick={this.props.formatCode} >
         <i className='fas fa-wand-magic-sparkles' />
    </div>
    <div className={`editorTheme ${this.state.themeSelector ? 'active' : ''}`}
         onClick={this.toggleThemeSelector} >
	 <i className='fas fa-palette' />
         {this.state.themeSelector && this.renderThemeSelector()}
    </div>
</div>
```

besides, this will be available in all editors at some point, when i finish a formatter for markdown.

5e-Cleric:
As mentioned before, that change would need to be soon undone, so i won't be implementing it. Since i want formatting to be available in all tabs in the near future.

client/components/codeEditor/codeEditor.less
Gazook89, line 1: Can you wrap the animation in a [`prefers-reduced-motion` query](https://css-tricks.com/almanac/rules/m/media/prefers-reduced-motion/), such that it only does the animation when it is set to `no-preference`?

client/components/codeEditor/extensions/customKeyMaps.js
```diff
@@ -176,6 +225,11 @@ export const generalKeymap = Prec.high(keymap.of([
 	{ key: 'Mod-d', run: deleteLine },
 ]));
 
+export const cssKeymap = Prec.highest(keymap.of([
+	{ key: 'Mod-Shift-f', run: formatCSS },
+  	{ key: 'Alt-Shift-f', run: formatCSS },
```
Gazook89, line 246: So I had to go back and look at previous edits and commits to double check this:

We went back and forth a bit about hotkeys for mac.  `Mod-Shift-F` *does* work on Mac.  The problem is that your first commit to which I earlier tested was actually set to `alt-shift-f`, and the PR comment said it was `cmd-shift-f`. 

That said, on MacOS, `Mod-shift-f` works fine.  If that means you can remove `alt-shift-f` as an option, I would recommend doing that.

client/components/codeEditor/extensions/customKeyMaps.js
```diff
@@ -2,7 +2,56 @@
...
+			tabWidth: 2,
+			useTabs: false,
+			printWidth: 100,
+			singleQuote: false,
+			trailingComma: 'all',
+			bracketSpacing: true,
+			endOfLine: 'lf'
+		});
+		formatted = formatted.replace(
+			/([^{]+)\{\s*\n\s*([^;\n]+:[^;\n]+;)\s*\n\s*\}/g,
```
Gazook89: Can you clarify what this is doing, and add a comment to the code to the same effect?

client/components/codeEditor/extensions/customHighlight.js
Gazook89, line 1: I don't believe this file has anything to do with the focus of the PR, right?  Just caught in the crossfire of the linter?

Gazook89 (changes requested):
It may be worth revisiting the placeholder/default text of the Style Editor so that it conforms to the Prettier standard.  Right now, if running the new format function, it cleans it up.  But on the other hand, maybe it's *good* that right away we have "poorly formatted" code that demonstrates the utility of the button.

----

I think the animation has whimsy, but I dunno-- too much whimsy?  In the future, i think this is where notification toasts would be helpful.  Once a function like this is completed, a toast popover saying "Style Editor formatted by Prettier.  [undo]" with a little undo button would make it very clear.

I don't think that's necessary immediately, but noting it as a possible improvement for later.

-----

It doesn't quite format the way *I* want them to be formatted, but I just need to adjust my preferences I think and overall this will be a boon to many users, I think.  

A couple of easy comments on the files but pretty much looks good.

client/components/codeEditor/codeEditor.less
5e-Cleric: that is a great idea, and i should change the animation too, any ideas?

client/components/codeEditor/extensions/customKeyMaps.js
5e-Cleric: Right, this is to catch single declaration rules to keep them to one line.

Apparently prettier can't do that by itself, it was requested and i agreed. 

I am still looking for a better way of doing this.

client/components/codeEditor/extensions/customKeyMaps.js
5e-Cleric: I'd rather keep it, as it is easier on my hands and the keybind in my IDE :p, if you don't mind much.

client/components/codeEditor/extensions/customHighlight.js
5e-Cleric: It totally did, and it is easier to leave it be if you don't mind.

client/components/codeEditor/extensions/customKeyMaps.js
Gazook89: I see, [that long thread on the Prettier repo](https://github.com/prettier/prettier/issues/5948) where they refuse to enable that....

In that thread, someone has linked a[ Prettier plug-in for this capability](https://github.com/zetafield/prettier-plugin-css-singleline), though.  Not sure if you saw it and rejected it as an unneeded dependency, or just didn't see it, but here you go.  The actual script for doing this work isn't even terribly long, it could just be pulled in to our codebase as a helper module.  Note, there is *one* issue on the repo, which we could probably fix.  

But if your code you already have is working well enough, I'm not stressed about it.

client/components/codeEditor/extensions/customKeyMaps.js
Gazook89: Then maybe we drop `mod-shift-f` and stick to only `alt-shift-f`?

client/components/codeEditor/codeEditor.less
Gazook89: I think personally I'm not entirely *for* any animation, at a knee-jerk level.  However, I *do* support whimsy, and this is whimsical.   But I don't have any particular ideas on how to improve it without really spending way too much time on it.

client/homebrew/editor/snippetbar/snippetbar.less
```diff
@@ -79,6 +79,12 @@
 					color     : grey;
 					&.active { color : inherit; }
 				}
+				&.formatCode {
+					.tooltipLeft('Clean your Code');
+					font-size : 0.75em;
```
Gazook89: This should be removed.  I believe it's an artifact from before a recent PR that removed font-size declarations from each button and just set one higher up the cascade.

In any case, it's making the icon smaller than the others.

Gazook89 (changes requested):

client/homebrew/editor/snippetbar/snippetbar.less
5e-Cleric: If anything its larger than the others.... 

image

I'll keep this size

5e-Cleric: image

Gazook89: > If anything its larger than the others....

The Format Code button, that looks like a magic wand?  from your screenshots it is clearly smaller than the other buttons.  So both visually, and provably via the code, it is smaller.

5e-Cleric: I'm an idiot, was looking at the pallete....

5e-Cleric:
This has gone through several reviews, all has been adressed, some stuff left for the future such as a toast message and an explicit undo button for this function (which would just call the undo function).

I am merging this now, and hurray to me.

PR #4746, opened 2026-04-20
fix newpage keybind

5e-Cleric:
It was a matter of precedence of the keybinds, this fixes any shortcut that coincided with the default keymap, mainly italics and newpage input.

This PR also changes the foldAll function back to folding just the ranges of pages.

Headings still have precedence over pages for folding.


calculuschild:
Ok, this seems to work. I'll merge this and send out.

PR #4750, opened 2026-04-21
add back cm5-themes

### LINKED ISSUE #4749: Rectangular Selection not working in v3.22.0
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

Rectangular selection (alt + left-drag) does not seem to be working in v3.22.0. It looks like CM6 might need the `rectangularSelection` and `crosshairCursor` extensions installed for this (from <https://codemirror.net/docs/ref/#view.rectangularSelection>).

### Code

```gfm

```

5e-Cleric:
Adds my own library (https://www.npmjs.com/package/codemirror-5-themes).

This library ships codemirror 5 themes, manually edited and transformed by postCSS to work with CM6

I have yet to add them all but it shouldn't take that long.


package.json
calculuschild: We can remove this then, right?

calculuschild (approved):

package.json
5e-Cleric: yeah

calculuschild:
Thanks for the quick fixes!

Issue #4059, opened 2025-02-18
Freezes when I interact


Zachariah1978:
### Error Code

```shell
Error: Service Unavailable


Error: Service Unavailable
    at Request.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:78608:21)
    at Emitter.emit (https://homebrewery.naturalcrit.com/homebrew/bundle.js:1967:20)
    at XMLHttpRequest.<anonymous> (https://homebrewery.naturalcrit.com/homebrew/bundle.js:78902:10)
{
  "status": 503,
  "method": "POST",
  "url": "/api"
}
```

### Your description of what happened:

I copy my files from Google Docs and then paste it here to work on it. All day today, whenever I do that, it freezes up. Then after five minutes it loads. However, if I so much as push the space bar it refreezes. This has only been happen today, but I'm on a time limit to finish before this Sunday for our game.
If I just open up and start typing fresh there is no problem. BUT I cannot retype 200 pages from scratch everytime I want to work on things.
Please, let me know if there is something I can do about this problem.
OR, if there is another website as good (or somewhat as good as) your for creating D&D homebrew content.
Thank you for your time.

G-Ambatte:
What browser are you using? Can we see the text that you are pasting into the Homebrewery editor, or do you have a link to a saved document?

G-Ambatte:
I did some digging and found what I suspect is a file that you managed to create earlier today. If so, on line 101, in the **Crossbows** table, the divider for the final column is malformed - it is `|:---- :` instead of `|:----:|`. This _shouldn't_ cause the issue that I'm/we're seeing, but I can confirm that removing the space (even without adding the final `|`) causes the editor to operate normally again.

Specifically:

```
| Weapon             | Wield    | Damage | Basic Range | Reload | Properties | Master Range | Reload | Wt | Cost |
|:-------------------|:---------|:------:|:-----------:|:------:|:----------:|:------------:|:------:|:--:|:---- :
```

needs to become:

```
| Weapon             | Wield    | Damage | Basic Range | Reload | Properties | Master Range | Reload | Wt | Cost |
|:-------------------|:---------|:------:|:-----------:|:------:|:----------:|:------------:|:------:|:--:|:----:|
```

If this was not your brew and you are reporting something else entirely, please disregard.

Zachariah1978:
I usually copy from my Google Doc, open Homebrewery, and Paste.

Here is the document in total.

[...]

PR #4756, opened 2026-04-22
Performance Mode for large brews

Frederlk:
## Summary

Opt-in Performance Mode that keeps typing and preview responsive on very large brews (50k+ lines, 800+ pages), plus a batch of always-on hot-path fixes that smooth out the editor for everyone. Marked draft / [WIP] per `contributing.md` to invite early feedback on approach before polish.

## Motivation

On documents that push a few megabytes of markdown, the editor and preview get choppy:
- `_.isEqual(currentBrew, lastSavedBrew)` deep-walks the whole brew on every keystroke.
- The preview re-renders on every toolbar tweak because `useMemo` depended on the entire `displayOptions` object.
- `safeHTML`'s DOM-scan runs for every page on every keystroke.
- The save path re-normalizes NFC three times, re-encodes the full text for `makePatches`, and gzips synchronously on the main thread.
- CodeMirror's highlight plugin tokenizes the full document on every update.

### Performance Mode (opt-in toggle in Properties)
- `useDeferredValue` on the preview text so typing stays snappy; the active page still renders promptly.
- Viewport-only highlighter in CodeMirror — tokenize only visible lines once docs exceed 2k lines.
- Debounced + `requestIdleCallback` + `startTransition` onChange pipeline.
- Debounced page-map recompute (250ms, 1s maxWait).
- Skip per-keystroke `Markdown.validate` (still runs on save).
- Cross-tab pref sync + one-time suggestion dialog when opening a large brew without the toggle.

### Always-on hot-path fixes (benefit everyone)
- `brewRenderer`: memoize page split; narrow `renderedPages` deps from `displayOptions` → `displayOptions.pageShadows` so toolbar changes (zoom, spread, gap) no longer invalidate the 800-page render cache.
- `safeHTML`: LRU cache (~100 entries) keyed on input HTML — during typing only the cursor's page rebuilds; neighbors cache-hit and skip the DOM scan.
- `codeEditor`: rope `doc.iterLines()` for the page map (vs `toString().split('\n')`), binary-search `findPageFromPos`, value round-trip identity short-circuit in the `[value]` effect, skip page-map scan on non-brewText tabs.
- `brewsEqual` whitelist replaces `_.isEqual` for change detection; field parity covered by a unit test against `DEFAULT_BREW` so adding a brew field without updating `brewsEqual` fails CI.
- Save path: normalize NFC once (was 3×), skip `makePatches` on unchanged text (metadata-only saves), schedule post-save `Markdown.validate` via `requestIdleCallback`, off-thread gzip via `CompressionStream` with `fflate` fallback for Safari <16.4 / SSR.
- `Snippetbar.shouldComponentUpdate` whitelist to avoid re-running `loadHistory(brew)` (IndexedDB `getMany`) on every text flush.
- `flushPending` ref on `Editor`/`CodeEditor` so Ctrl+S, blur, and beforeunload drain any debounced text into React state before the save reads it.

### Side fix: scroll preservation across tab switches
Pre-existing bug surfaced by the faster tab switches: scroll reset to top every time. Now each tab's `EditorState` is persisted with a CM6 `view.scrollSnapshot()` (the `isSnapshot:true` fast path in `DocView.scrollIntoView`), survives virtualized-viewport clamping, and restores to the exact previous scrollTop. The phantom `CodeEditor` on the Properties tab is kept on a stable `tab` + matching `value` so entering/leaving Properties no-ops the tab-change effect and doesn't corrupt `docsRef`.

## Tests

Added unit tests for the extracted helpers:
- `tests/brew/brewsEqual.test.js` — field parity against `DEFAULT_BREW`, equality semantics, array handling.
- `tests/brew/compress.test.js` — round-trip through `CompressionStream` and the fflate fallback, gzip magic bytes.
- `tests/html/safeHTML.test.js` — cache hits, LRU eviction, pure-function behavior against a reset `div`.

All 19 suites / 282 tests green via `npm test`.

## Lint override justification

`codeEditor.jsx` exceeds the 600-line max (now 640 lines) because Performance Mode's debounced/idle/transition pipeline, viewport-only highlighter, binary-search page lookup, and scroll snapshot persistence are all tightly coupled through the component's refs (`viewRef`, `docsRef`, `pageMap`, `performanceModeRef`, `lastEmittedRef`). Splitting would have to pass all of them across module boundaries and break apart the `useEffect` ordering that makes the tab-switch handoff correct. The file now carries `/* eslint max-lines: ["error", { "max": 650 }] */` with a header comment explaining why.

### Issues this likely addresses (user reports of the symptoms this PR targets)

- [#3830](https://github.com/naturalcrit/homebrewery/issues/3830) — *Lag when scrolling editor panel as of 3.16.0* (closed). Linked from two Reddit threads. Partial fixes landed via #3831 and #3965; Performance Mode + the always-on hot-path wins here chase the same root cause from a different angle.
- [#4059](https://github.com/naturalcrit/homebrewery/issues/4059) — *Freezes when I interact* (closed). ~200-page brew pasted from Google Docs, typing re-froze for 5 minutes. Performance Mode's deferred preview + viewport-only highlighter + debounced page-map specifically target the freeze path described there.
- [#4259](https://github.com/naturalcrit/homebrewery/issues/4259) — *Crashes with a white screen* (closed). Large brew, autosave path, Opera GX. Our save-path changes (off-thread `CompressionStream` gzip, skip `makePatches` on unchanged text, idle-scheduled validate) reduce the main-thread stall during autosave that contributed to these reports.
- [#3474](https://github.com/naturalcrit/homebrewery/issues/3474) — *Tab freezes on H1 header off the page (FF/MacOS)* (open). Different root cause (CSS `column-span`), orthogonal; Performance Mode doesn't fix this.

### Issues this PR deliberately does not solve

- [#3317](https://github.com/naturalcrit/homebrewery/issues/3317) — *Getting the current Editor Page for page rendering happens out of sync* (open). My `pageMap` + binary-search `findPageFromPos` change touches the same machinery, but the underlying state-propagation race described in #3317 is still present. I kept the fix out of scope so this PR stays reviewable; worth a follow-up.

### Prior-art PRs — how this PR layers on (and where it improves the tradeoff)

- [#3831](https://github.com/naturalcrit/homebrewery/pull/3831) *memoize `renderPages()` and `renderStyle()`* (merged).
  - **What it did**: `useMemo(renderPages, [text, displayOptions])`.
  - **What this PR does better**: narrows the `displayOptions` dep to `displayOptions.pageShadows` only. Zoom, spread, and gap live on the `.pages` wrapper as pure CSS — they shouldn't invalidate the 800-page render cache, but under #3831's dep list they did. Every toolbar tweak was re-running the per-page markdown → HTML pipeline. With the narrower dep it's a no-op. Same mechanism, correct invalidation boundary.

- [#3965](https://github.com/naturalcrit/homebrewery/pull/3965) *Implement content-visibility on pages* (merged).
  - **What it did**: CSS `content-visibility: auto` so offscreen pages skip layout/paint in the preview.
  - **What this PR does better / complements**: attacks the editor side, which `content-visibility` can't reach. `useDeferredValue` defers React work for the whole preview; viewport-only tokenization skips CodeMirror work for offscreen lines. Together: CSS skips offscreen *renderer* work, React + CM6 skip offscreen *editor* work — two non-overlapping wins on the same frame budget.

- [#1874](https://github.com/naturalcrit/homebrewery/pull/1874) *Tell CodeMirror to batch custom highlights before updating the browser* (merged).
  - **What it did**: batches highlight DOM writes after tokenization.
  - **What this PR does better**: attacks the cost *upstream* of the batcher. On a 50k-line brew, #1874 still tokenizes all 50k lines on every update — it just flushes the resulting decorations in one pass. Performance Mode's viewport-only tokenizer tokenizes ~150 visible lines instead of 50,000. #1874 reduced DOM churn for large highlight sets; this PR stops producing most of that set in the first place.

- [#3845](https://github.com/naturalcrit/homebrewery/pull/3845) *Intersection Observers for "Current Page(s)"* (merged).
  - **What it did**: renderer-side page tracking via `IntersectionObserver`.
  - **What this PR does better / complements**: independent editor-side lookup. `pageMap` uses CM6's rope `doc.iterLines()` (no full-doc `toString()` + `split('\n')` allocation), and `findPageFromPos` is binary-search O(log n) instead of linear O(n). Cursor → page on an 800-page doc goes from ~800 comparisons per cursor move to ~10. Different axis from #3845, stacks cleanly.

- [#3833](https://github.com/naturalcrit/homebrewery/pull/3833) *Speed up typing lag* (closed, not merged).
  - **What it tried**: move the Theme/Style `<style>` block into the iframe `<head>` to cut "Recalculate Styles" cost during every brewRenderer output.
  - **How this PR relates**: different layer (React / CM6 / save), not CSS cascade. #3833's approach is still viable and could layer on top of this PR for additional wins — the two don't conflict. Worth a reviewer noting if it's still worth reviving.

### Conceptual difference from all prior perf work

Prior PRs (#3831, #3965, #1874, #3845) are **always-on** optimizations that trade constant factors — memoize what you can, batch what you must, virtualize what you can afford. They don't change the algorithmic work the editor performs per keystroke.

This PR introduces an **opt-in mode** that changes *what* work the editor performs: defer the preview, skip per-keystroke HTML validation, debounce the React state propagation, tokenize only the viewport. These are user-visible tradeoffs (preview can lag a frame, HTML errors surface on save rather than on every keystroke) that aren't acceptable as defaults but are very much acceptable at 50k lines when the alternative is a frozen tab. Every user who's ever filed a "freezes on large brew" issue has implicitly said yes to this tradeoff; this PR just makes that tradeoff explicit and recoverable via a toggle.

## Test plan

- [x] `npm run verify` passes (lint errors limited to pre-existing camelcase in theme files; stylelint baseline unchanged; all 282 jest tests green).
- [x] Manual: scroll brewText to line 10,000 → Style → Snippets → Properties → Snippets → Style → Text. Lands back at line 10,000.
- [x] Manual: toggle Performance Mode on a 50k-line brew; typing stays responsive and preview catches up without blocking input.
- [x] Manual: Ctrl+S during perf-mode typing flushes pending text before save.
- [ ] Reviewer: sanity-check the `editor.jsx` phantom-CodeEditor change (Properties tab) against any edge cases I've missed.
- [ ] Reviewer: validate the `CompressionStream` fallback path (tested in jest with a stubbed global, worth a manual verify in Safari).

## Out of scope for this PR

- Local-dev env tweaks (`.env` support in `package.json` dev script, loopback CORS/SSL bypass on `server/app.js` / `server/forcessl.mw.js`) were kept on a separate local branch.
- No dependency updates beyond an already-present `fflate` (pulled in as a gzip fallback path).

### DIFF
--- a/client/components/codeEditor/codeEditor.jsx
+++ b/client/components/codeEditor/codeEditor.jsx
@@ -1,6 +1,11 @@
-/* eslint max-lines: ["error", { "max": 400 }] */
+/* eslint max-lines: ["error", { "max": 600 }] */
+// Bumped from 500: Performance Mode adds the viewport-only highlighter, the debounce/idle/
+// startTransition pipeline for onChange and recomputePages, and binary-search page lookup.
+// Splitting the file would scatter tightly-coupled state (debounced fns, refs, useEffects)
+// across modules without making either half easier to read.
 import './codeEditor.less';
-import React, { useEffect, useRef, forwardRef, useImperativeHandle } from 'react';
+import React, { useEffect, useRef, useMemo, forwardRef, useImperativeHandle, startTransition } from 'react';
+import _ from 'lodash';
 
 import {
 	EditorView,
@@ -40,10 +45,18 @@ import { generalKeymap, markdownKeymap } from './customKeyMaps.js';
 import foldOnPages from './customFolding.js';
 import { customHighlightStyle, tokenizeCustomMarkdown, tokenizeCustomCSS } from './customHighlight.js';
 import { legacyCustomHighlightStyle, legacyTokenizeCustomMarkdown } from './legacyCustomHighlight.js';
+import { PAGEBREAK_REGEX_V3 } from '@shared/pageBreaks.js';
 
-const PAGEBREAK_REGEX_V3 = /^(?=\\page(?:break)?(?: *{[^\n{}]*})?$)/m;
+// Run a callback during browser idle time, with a setTimeout fallback for older browsers
+// (Safari < 14). Used in perf mode to keep doc.toString() off the input thread.
+const scheduleIdle = (fn)=>{
+	if(typeof requestIdleCallback === 'function')
+		requestIdleCallback(fn, { timeout: 200 });
+	else
+		setTimeout(fn, 0);
+};
 
-const createHighlightPlugin = (renderer, tab)=>{
+const createHighlightPlugin = (renderer, tab, performanceMode = false, pageMapRef = null)=>{
 	//this function takes the custom tokens created in the tokenize function in customhighlight files
 	//takes the tokens defined by that function and assigns classes to them
 	//it also creates page number and snippet number widgets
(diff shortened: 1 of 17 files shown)


5e-Cleric:
This _reeks_ of AI, anything to say about that?

Gazook89:
Echoing @5e-Cleric, it seems pretty evident that this is the dreaded "new contributor opens large PR across multiple files in single commit" activity that is increasingly common, but so far HB has managed to dodge.  

I hesitate to immediately dismiss it as it's typically good to have more contributors and engagement, but it's uncomfortable to start this way.  

My immediate thought when starting out is: who has 800+ page brews, and *should* they?  That's not any official position on that, just my personal disbelief and opinion.  I would rather find some way to nudge people towards splitting their brews....for their sake, and for any audience they might have.  Generating content just for content is kind of blegh.

I am betting that there *are* improvements in this submission, some value that we can pull from this.  As noted, it contains stuff that helps all users (probably).  But it *feels* icky to get that feedback in this way.

Just my two cents.

*edit:  i know that 800+ brews exist....i just kind of shudder at the thought*

calculuschild:
Yeah. It's AI.

I'm not opposed to the ideas proposed, but we need live person who can understand the code and implement feedback.

The way this PR goes about the improvements is too chaotic to effectively review, and I don't think we have the resources to chase an AI around and around fixing its own mistakes.

@Frederlk, if you want to submit a smaller PR that addresses one of the performance cases, I would be happy to review it, and we can work to incrementally apply more improvements in subsequent PRs.

You can consult an AI behind the scenes if you need help understanding something, or for brainstorming, but please don't just have the AI generate the whole PR if it's going to be this complex. I'm not going to review a PR if it's clear the author hasn't looked at the code.

Frederlk:
Got it. Thanks for the heads up.

I intended to submit a smaller PR initially, but after seeing the performance gains from these improvements, I felt encouraged to do more. I will split the PR into smaller chunks and close this one.

Over the last few years, I created a massive homebrew bestiary file for my campaigns. But it became so laggy that I stopped extending it, though with the performance mode, I will be able to continue expanding it.

At some point I tried splitting it into multiple files previously, but it felt inconvenient.

calculuschild:
No worries. Your other PR is on a better scale; I'll be reviewing it shortly.

The reality is this project does not always move quickly for large PRs as we don't often have consistent blocks of time to sit down and absorb 500 lines of code changes. I also don't know about the other contributors here, but I have extensive exposure to AI-generated PRs at my job, and they are a pain to deal with, particularly once they start touching more than a couple files or trying to handle multiple tasks at once. I don't want to battle that here in my free time, if that makes sense.

PR #4763, opened 2026-04-23
Make common "wrapSelection" function

calculuschild:
## Description

This PR does three things:

1) Creates a common `wrapSelection()` function for hotkeys that wrap a prefix/suffix around the current selection; this should make it easier to update common functionality for this type of hotkey.

2) If text was selected, the selection grows or shrinks to accomodate the new/removed prefix/suffix. This means you can repeatedly press the hotkey to redo/undo the wrap action without needing to reselect the text.

3) If no text was selected, it places the prefix/suffix around the cursor.

### DIFF
--- a/client/components/codeEditor/customKeyMaps.js
+++ b/client/components/codeEditor/customKeyMaps.js
@@ -17,125 +17,28 @@ const indentLess = (view)=>{
 	return true;
 };
 
-const makeBold = (view)=>{
+const wrapSelection = (prefix, suffix) => (view) => {
 	const { from, to } = view.state.selection.main;
 	const selected = view.state.doc.sliceString(from, to);
 
-	let text, cursor;
+	let text, selection;
 
 	if(from === to) {
-		text = '****';
-		cursor = from + 2;
-	} else if(selected.startsWith('**') && selected.endsWith('**')) {
-		text = selected.slice(2, -2);
-		cursor = from + text.length;
-	} else {
-		text = `**${selected}**`;
-		cursor = from + text.length;
+		text = prefix + suffix;
+		selection = { anchor: from + prefix.length, head: from + prefix.length };
+	}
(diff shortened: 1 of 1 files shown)


Gazook89:
Would this benefit from having a third parameter, that allows choosing a final cursor position?  

I'm looking at `makeSpan`, `makeDiv`, `makeComment` etc that wrap some selected text, but that should *not* select the entirety of the resulting string but instead should place the cursor somewhere within.  

I'm thinking something like `wrapSelection('<!-- ', ' -->', 6)`, where the cursor would be in that 6th position.  But I don't know how complex is too complex for this simple function and what should just be done the normal codemirror way.  Because maybe then we want a selection range?

I'm working on updating some of these keymaps since they lost their final cursor positioning after the editor update, and will probably go ahead with just doing it the 'normal way', but I can change tacks if needed.

Gazook89:
One more thing:  When experimenting with eliminating our custom keymap for `makeComment`, I realized I might just remove that whole hotkey because CodeMirror can do that itself. In the course of looking at this, I notice that when i select some text and do the hotkey, it adds the html comment syntax but then *keeps the original selection of the syntax*.  Then, when i do the hotkey again, *it removes the html comment syntax while retaining the original selection, again*.  

It removes that HTML comment syntax without needing that syntax included in the selection.  Notably, it *also* works when the HTML comment syntax is inside the selection, like how your `wrapSelection` function works.

What's nice about this native behavior is that I can make something a comment, and then just start typing my comment...I don't have to move my selection inside the new syntax, and yet I can also toggle that syntax repeatedly.

I think we should update this new `wrapSelection` function to work the same way, though I'm sure there some complexity-- figuring out what should be sliced out and what should be kept in when it's not explicity part of the selection.  Might need to find the source code in CM to hook into that, or just copy it.

Issue #4761, opened 2026-04-23
Typing `:: ` (colon colon space) removes all custom highlighting in the editor


calculuschild:
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

Pick any document with custom highlighting (homepage works fine).

Find a blank line, and type `:: `. All custom highlighting (curly braces, super/sub, etc,) dissappears until the page is reloaded.

5e-Cleric:
wow

PR #4764, opened 2026-04-23
Fix multiline Definition List highlight breaking

calculuschild:
## Description

This PR does three things:

1) Simplifies some regex that was equivalent to just `.*` to reduce catastrophic backtacking
2) Prevents crashing and losing custom highlights if Description entry length was empty
3) Only begins multiline highlighting if there is a valid Term with content


calculuschild:
fixes #4761

PR #4593, opened 2026-01-20
Template class persistence

dbolack-ab:
## Description

This adds the concept of Template class persistence at the `\page` level.

If a `\page` has a {} block, the parsed attributes are examined. If one is named `hbtemplate` the value is stored in a lookup table.

If a page lacks a Template , the lookup table is walked backwards until a template name is found *or* the beginning of the document is found. 

If a template name is found, it is added to the attributes for the page.

examples:

```
\page {hbtemplate=colophon}
```
This page would use the colophon template.

```
\page {hbtemplate=chapter}
\page
```
Both of these pages will use the chapter template.

*Reviewers, refer to this list when testing features, or suggest new items *
- [ ] Verify new features are functional
  - [ ] Adding `hbtemplate=foo` shows in the page div classes.
  - [ ] Adding `hbtemplate=foo` shows in pages following that do not have an hbtemplate set 
  - [ ] Adding `hbtemplate=foo` shows in pages following that do not have an hbtemplate set until a page adds its own `hbtemplate`
- [ ] Verify old features have not broken
  - [ ] If no hbtemplate attribute is used, behaves as before.
- [ ] Test for edge cases / try to break things
- [ ] Identify opportunities for simplification and refactoring
- [ ] Check for code legibility and appropriate comments

PR #4769, opened 2026-04-24
Better dropdowns

### LINKED ISSUE #3840: Snippet dropdowns cut off behind the divider bar
## Description

This is not a new issue, but testing #3635 made it more obvious since it uses a very narrow editor panel.

Snippet dropdowns are being cut off behind the divider bar:

image

### LINKED ISSUE #4870: use ESC to dismiss menus
### Your idea:

When mousing to edit brew content .. if the mouse drifts into the menu bar space above the menus appear on hover. Getting the mouse pointer back to the brew text editor space involves careful navigation around this sudden obstacle.

Make it so **if the user presses the `ESC` key the menus are dismissed**.

Gazook89:
## Description
We have a problem with the snippet menus getting cut-off by the divider bar, since those are normal absolutely-positioned elements within a pane that gets clipped.  

This is a basic implementation of new menus using the Anchor Positioning API.  It creates a new component Dropdown.jsx.  It could be used elsewhere, such as in the Navigation bar, but for now I've limited it to just the snippet menus to avoid going overboard.  The original nav components are still in, as a result, but could be replaced further down the line.  

## improvements

The menus are not clipped by the pane, they exist in the `top layer`.

The menus are "click to open" rather than on hover.  This is more accessible for anyone with motor disability, but helpful for everyone.  For developers, you can now click a menu open, and then actually inspect it without having to use the dev tools to manually force the "hover" state.  

Menus are dismissed by clicking anywhere off the menu (or on the original trigger).  ~~**Question: do people prefer that the menu closes after clicking a snippet, or leave the menu open?**~~  I've made it so menus close after selection by default, with an attribute `no-dismiss="true"` that can be added for a menu if you want it to remain open.  It's not used in this PR, but could be somewhere else.  The attribute takes `"true"` as a string because react custom attributes only take strings.  

## next steps

~~The menus have no complex algorithms for determining placement right now, but they could (fairly easily).~~ The menus right now default to opening down, aligned to the left edge of the trigger.  Submenus open to the right, aligned to the top edge of the trigger.  Both menus and submenus can flip inline if no space. 

Anchor Positioning is now supported in the big browsers since about the beginning of this year.  ~~If we want to support further back, that's possible using resize observers, position calculations, absolute positioning, etc.  I have it set up and working in my old UI Overhaul PR.  But to keep this small I left it out for now.~~  The [Oddbird polyfill](https://anchor-positioning.oddbird.net) has been added, and is conditionally applied only if the users browser doesn't support AP.

The menu items are basically the same structure as exists in live, but they could be improved.  I would later create a basic 'button' component that would be used in menubars (navbar, snippetBar) and within menus.  

- Closes #3840 

Is able to cross the divider bar by existing in the `top-layer`:

image

If the window is made small enough that a menu would go off screen, it instead has fallback positions.  This shows a nested snippet menu automatically re-positioning to the left of the parent menu, rather than default right, so it stays within the window if possible:

image

### To Test:

In the Snippet Bar of the editor:
- [x] Clicking a snippet menu opens it
  - [x] it opens below, aligned to left of, the menu trigger
- [x] Menus remain open and only close when a click happens anywhere else
  - [x] the menus stay open if clicking a submenu trigger
  - [x] clicking a menuitem button performs the snippet action and closes the menu
- [x] Menus open in `top-layer`, not obscured by divider
  - [x] Menus use fallback options for positions if the window width is decreased to not allow space.
- [x] Menus are tab-navigable
  - [x] `tab` for next button
  - [x] `shift-tab` for previous button
  - [x] `space` or `enter` to click a button, including opening a submit
  - [x] `esc` to close current menu (closes a submenu but keeps parent menu open)
  - [x] `tab` into opened submenu
- [x] browser support
  - [x] works in chrome
  - [x] works in safari
  - [x] works in firefox 
  - [x] works in iOS
  - [ ] works on Android
  - [x] works on older browsers (polyfill should do this, and i haven't tested *lately* but did test after the bulk of the work was done, on older Firefox (pre-Anchor Positioning/Popover API's).
- [x] no animals were hurt in the production of this PR.

<details><summary><h3>The same list, but as markdown</h3></summary>

```
In the Snippet Bar of the editor:

- [ ] Clicking a snippet menu opens it
  - [ ] it opens below, aligned to left of, the menu trigger
- [ ] Menus remain open and only close when a click happens anywhere else
  - [ ] the menus stay open if clicking a submenu trigger
  - [ ] clicking a menuitem button performs the snippet action and closes the menu
- [ ] Menus open in `top-layer`, not obscured by divider
  - [ ] Menus use fallback options for positions if the window width is decreased to not allow space.
- [ ] Menus are tab-navigable
  - [ ] `tab` for next button
  - [ ] `shift-tab` for previous button
  - [ ] `space` or `enter` to click a button, including opening a submit
  - [ ] `esc` to close current menu (closes a submenu but keeps parent menu open)
  - [ ] `tab` into opened submenu
- [ ] browser support
  - [ ] works in chrome
  - [ ] works in safari
  - [ ] works in firefox 
  - [ ] works in iOS
  - [ ] works on Android
  - [ ] works on older browsers (polyfill should do this, and i haven't tested *lately* but did test after the bulk of the work was done, on older Firefox (pre-Anchor Positioning/Popover API's).
- [ ] no animals were hurt in the production of this PR.
```

</details>


5e-Cleric:
Did this feature have a nice degradation? what will be shown in older browsers?

Gazook89:
What happens in older browsers is the items are absolutely positioned at 0, 0 of the next relative positioned ancestor.  So in the case of the snippet bar, the appear at the top left of the editor panel.  In the case it's a submenu, they stack on top of each other.  In the future, the nav bar menus would stack at the top left of the window. 

That said, I *do* have a whole fallback/polyfill for older browsers that I can add if we want.  It's not really *that much* extra, but I really wanted to limit the scope of the PR, or make clear how simple it *could be*.  

Here is a screenshot with Anchor Positioning turned off, with a menu and submenu opened:
image

Note, that could look a lot better with other improvements like a thin border and a little drop-shadow, which would make it much more clear that two menus are on top of each other.

calculuschild:
Is there a convenient polyfill like we do for `core-js`, so we just import something and don't have to worry about maintaining any custom stuff?
Like we do with this?
https://github.com/naturalcrit/homebrewery/blob/c12bdcd0429f457bce1d7cc9870119b7231c303d/client/homebrew/homebrew.jsx#L1

Gazook89:
Yes, https://anchor-positioning.oddbird.net/

I believe I have mentioned that before and tried it way back when, but couldn't get it work with require.js or whatever. But it could be tried again.

calculuschild:
Ok. At this point, I think we've proven to ourselves that any code change that is not backwards compatible with old browser versions will always cause a panic, and we end up having to rush to make a fix. If we know this one will need a polyfill let's just handle it here.

5e-Cleric:
Yeah here the polyfill is very needed and this feature is great so lets add in the polyfill and testing as soon as we can.

Gazook89:
Trying to get this to work.  I first tried just the basic CDN script based on browser support but hit an issue.  To get around that, I went the "install as NPM package" route, but setup to lazy-load in the case that anchor positioning isn't supported by the browser. 

The issue is that React removes unsupported inline style properties at runtime.  So the way I have this setup is elements with `style={{ positionAnchor: ... }}`, but since that property isn't supported by the browser React cuts it out.  

The polyfill looks through the computed styles of the dom for `positionAnchor` in order to target its work, but if React is removing all those identifiers then it obviously won't work.

So instead, I could use `style={{ '--position-anchor': ... }}` to set a custom property, which is always allowed by browsers and thus by React.  That seems to have some effect and the polyfill *is* finding the correct elements.  But the result is stretching the anchor elements across the whole viewport.

Now I have to figure out why the polyfill just isn't working as expected.

Gazook89:
It might be because I think I have the positioned element as a child of the anchor when they likely should be siblings. 

I believe I nested them in order to work with my own mini polyfill, but if using oddbird then I can not do that. 


So I'll check that when I'm home.

Gazook89:
I think I have the polyfill working on firefox 145 (pre- anchor positioning).  Going to futz with it a bit and go to bed and will push it up to github tomorrow probably.

Gazook89:
I think this is all set if people want to try it.  If you need to install an older version of Firefox to test it, you can find the steps here: https://support.mozilla.org/en-US/kb/install-older-version-firefox#w_i-still-want-to-downgrade-where-can-i-get-the-previous-version

More directly, [here is the directory for version 145](https://ftp.mozilla.org/pub/firefox/releases/145.0/), choose for your OS.   Note that you need to turn off auto updates in the firefox settings, otherwise when you restart FF it'll automatically bump to the latest.

5e-Cleric:
I am not seeing correct alignment of menus without an icon:

image


In some menues this does get a bit weirder:

image

i assume there is a reason for this.

Also, license>shadowdark>logos> seem to have vanished.

Do we not want the original icons in the snippet bar?

The core work is very nice, and i'm happy to be able to open the menus with the divider in any position. Great work.

5e-Cleric:
I also think the menu should close when clicked on a snippet to get the input.

Gazook89:
> the menu should close when clicked on a snippet to get the input.

Agreed.  Will fix.

> I am not seeing correct alignment of menus without an icon...[in some] menus this does get a bit weirder

Yes, agreed.  In previous PR i had this worked out by creating a separate component Button.jsx that was adaptive to "buttons in a menu" and "buttons not in a menu".  It had a lot of smarts, working in many situations. 

In this PR, when I got to this point, I stopped short until I got at least some signal that the base changes to the menus was even desired.  I wanted to present the changes up to this point as plainly as possible.  

So some directions I go from here are:
- Continue work on this PR to improve the buttons.  If going this route, I'd basically just do the basic bit to get it more presentable (akin to what we have now).
- Make a branch *from this branch* as a separate PR, to improve the buttons with some added functionality.
- Merge this PR to the upstream/master as-is, with the promise that an improved button will come along.  

My choice would be the second option, my last choice would be the third.  But really, I just don't want to do more until there was consensus that this is going in the right direction.

5e-Cleric:
> In this PR, when I got to this point, I stopped short until I got at least some signal that the base changes to the menus was even desired. I wanted to present the changes up to this point as plainly as possible.
> 
> So some directions I go from here are:
> 
> * Continue work on this PR to improve the buttons.  If going this route, I'd basically just do the basic bit to get it more presentable (akin to what we have now).
> * Make a branch _from this branch_ as a separate PR, to improve the buttons with some added functionality.
> * Merge this PR to the upstream/master as-is, with the promise that an improved button will come along.
> 
> My choice would be the second option, my last choice would be the third. But really, I just don't want to do more until there was consensus that this is going in the right direction.


I am fine with all three, i agree it is a meaninful change, and unless calc opposes it with a very good argument, i see no problem on merging this now or approving what is done and waiting for the rest of improvements. Whichever way you prefer. Option 2 or 3 would make it easier to review the code.

Gazook89:
> the menu should close when clicked on a snippet to get the input.

Actually, I initially agreed but a second thought:  Is there a benefit to leaving the menu open after clicking a snippet?  Or, is there a benefit to allowing it to remain open if holding some hotkey when you click?  

Something to mull over for a second.  I'm going to go ahead and change it to your recommendation right now and I don't anticipate getting an answer on the above now (or stopping progress this in any way), but just writing down that consideration for posterity.

5e-Cleric:
> > the menu should close when clicked on a snippet to get the input.
> 
> Actually, I initially agreed but a second thought: Is there a benefit to leaving the menu open after clicking a snippet? Or, is there a benefit to allowing it to remain open if holding some hotkey when you click?
> 
> Something to mull over for a second. I'm going to go ahead and change it to your recommendation right now and I don't anticipate getting an answer on the above now (or stopping progress this in any way), but just writing down that consideration for posterity.

I just checked and we don't even close it right now, so i guess... we can do whatever we want...
Since the new one doesn't close on moving the mouse out, i say we close it when we input the snippet, but we retain on shift-click? not sure if we'll ever really use it or have users know about it.

another small thing, lets remove the beta tags on most snippets, they have been stable for like... forever.

Gazook89:
> the menu should close when clicked on a snippet to get the input.

This is now the case, with the latest commit.  I also added a method to check for if a menu item has a `no-dismiss="true"` attribute, in which case clicking that particular menu item does not dismiss the menu.  I thought there might a particular snippet where you might want to keep it open so you could press it repeatedly, but then I went through the list and didn't see anything except maybe the "new page" snippet, but I didn't bother with it.   But having that attribute is still handy because this component should also eventually be used in the Navbar dropdowns and other places.

Still have to do the work on buttons.

Gazook89:
> I am not seeing correct alignment of menus without an icon

Fixed this.  I also noticed that we used some icons that don't actually exist ("fa-mit" and "fa-paizo") so i removed those.

> Also, license>shadowdark>logos> seem to have vanished.

Fixed this.  It was ID collisions, with multiple menus have the same names. It is fixed with React `useId()` safely stripped of special characters.

> Do we not want the original icons in the snippet bar?

They are back in now.

-----
I could probably try adding some tests to this.  Otherwise, please let me know if there are more visible issues.

Gazook89:
> I could probably try adding some tests to this.

I came up with some targets for tests and then told an LLM to "make it so".  I'm not super familiar with tests/Jest and I really didn't want to get bogged down in it.  It is the last commit so it can be popped off if needed.  

What it does:  it tests for html structure in one file and behavior in another.  Both files use a shared helper script for setup and takedown.  

The test is part of a new script in package.json for `test:components` which we can build from later...I don't believe we currently have any tests for other components?

-----
### Small Things
*the images below have some effects that ARE NOT in this PR, just added via browser dev tools*

I think there are some small styling changes that could be done.  Such as a very subtle border and some box-shadow, which would add some contrast between menus and whatever is below it.

image

If wanting to go crazy, we could start to soften the UI a bit with border-radius in places, though I think the boxy look is "classic HB":

image

And visible in above image, as well as on the live site, it is bothersome that some menus with no icons at all still leave a blank "column" or space for icons, which looks dubious.  I can fix that with CSS Grid.  But I'm not sure it belongs in this PR.

------
### Ready to Merge?
Overall, I think this is ready to be reviewed again and hopefully merged.

**The menus all closed** and empty menus (no children) are auto-disabled.
image

**Menus open with clicks, and are not controlled by :hover**
image

**Handles menus at least 3 levels deep** (and actually handles *infinite* levels??)
image

**Menus are not restricted to some container, they appear in the `top layer` and thus cross the splitPane divider**
image

**Menus are tab-navigable**.  Use tab and shift-tab to move through the menus, and `space` to click to open a menu or choose an option, and `esc` to close a menu.
image

5e-Cleric (commented):
If you don't mind, could you move the tests into the tests folder?

client/components/dropdown/dropdown.jsx
```diff
@@ -0,0 +1,124 @@
+/**
+ * A dropdown menu component that uses the Anchor Positioning API to position the elements.  It supports nested submenus as well.
+ * Anchor Positioning is now supported in all major browsers.  If support is needed for older browsers (namely, older firefox)
+ * it is possible to use absolute positioning and calculations/resize observers to position things well enough, but adds another layer of complexity to the code.
```
5e-Cleric: This can be reduced to specify there is a polyfill.js file in use.

client/components/dropdown/dropdown.less
```diff
@@ -0,0 +1,25 @@
+.menu-wrapper {
+	position: relative;
+	place-content: center;
```
5e-Cleric: This property does nothing without display grid or flex, i assume its a leftover.

client/components/dropdown/dropdown.less
```diff
@@ -0,0 +1,25 @@
...
+	}
+}
+
+.menu-list {
+	position         : fixed;
+	z-index          : 1000;
+	inset-block-start: anchor(bottom);
+	inset-inline-start: anchor(left);
+	position-try: flip-inline;
+	color: unset;   // [popover] gets a `canvastext` color value from useragent.
```
5e-Cleric: nothing wrong with this but `inherit` or `revert` might make it more obvious what color this will get. Not sure its worth the change.

5e-Cleric (approved):
Overall great changes, i assume its not ready for merge, since i see an empty file, but looks good, left a few small comments

client/components/dropdown/dropdown.jsx
Gazook89: Yup, thanks.  adjusted.

client/components/dropdown/dropdown.less
Gazook89: yup, thanks, removed.

client/components/dropdown/dropdown.less
Gazook89: inherit works for me, and i agree, it is probably more appropriate.  changed.

Gazook89:
I moved the tests into the `tests/components/dropdown/` folder.  And cleaned up the little things you commented on.  

Regarding the "empty file", i assume you mean the styleMock.js?  I tried removing it, but then tests didn't run.  This is one of those things that an LLM put together, so I queried what was up with that file.  The answer is basically "Jest needs this file to substitute in for the `.less` import, since Jest runs in Node and not a browser. 

<details><summary>The full answer</summary>

> The file is there purely to satisfy Jest when a component imports a stylesheet.
> 
> Jest runs in Node, not a browser bundler, so an import like dropdown.less is not something it can execute on its own. The moduleNameMapper entry in package.json tells Jest: when you see .less or .css, load styleMock.js instead. That mock can be basically empty, because the test only needs the import to resolve, not to actually apply styles.
> 
> So the file’s purpose is:
> 
> Prevent import errors for style files.
> Let component tests focus on DOM structure and behavior.
> Keep the test environment independent from the CSS pipeline.
> If you remove it, Jest will try to resolve dropdown.less directly and fail. The mock can stay minimal, but it does need to exist as a resolvable module.

</details>

I assume that with additional component tests this will continue to be needed, but that this one file will work for all of those tests.

I believe this is **ready**.

5e-Cleric:
> Regarding the "empty file", i assume you mean the styleMock.js? I tried removing it, but then tests didn't run. This is one of those things that an LLM put together, so I queried what was up with that file. The answer is basically "Jest needs this file to substitute in for the `.less` import, since Jest runs in Node and not a browser.

Eh, no? we have plenty of less files everywhere else and no mock file is needed for each of them, so we can do them like they are done elsewhere.

G-Ambatte:
> we have plenty of less files everywhere else and no mock file is needed for each of them

I don't believe that we actually test any rendered elements anywhere else; I half-recall a discussion along the lines of "visual elements don't require automated testing as any issue with them will be immediately noticed".
I'm prepared to learn I've misremembered/am just flat wrong about that, of course. But in keeping with that line of thought for literally every other React component in the project, I'm not sure why testing has become necessary here.

---

> another small thing, lets remove the beta tags on most snippets, they have been stable for like... forever.

I don't disagree that the BETA snippets have been stable for a long time, but removing the tags is unrelated to the purpose of this PR and probably should have been handled in a small focused PR of its own, where the merits of such a change could have been discussed and decided independently of these other menu changes.

That said, it's probably not worth the effort to unwind it from the current state of the PR.

---

On testing, I'm receiving a console error of `Each child in a list should have a unique "key" prop` error for the render method of `Dropdown`; when it is passed a child from SnippetGroup, it is creating multiple sibling components without unique keys.

A little experimentation seems to show that this issue is resolved by changing line 339 of `snippetbar.jsx` to the following:

```
<Dropdown groupName={snippet.name} icon={snippet.icon} key={snippet.name}>
```

---

On an aesthetic note, if the width of the editor panel is less than 1026px, the snippet bar is increased to two lines - the menus are all much wider than they were previously. The Editor tools collapse to the extreme left of the window and the mouseover label is no longer visible on the screen (see image below for example of UNDO being mostly off-screen).

image

---

Attempting to comprehend the `ref` function of AnchoredTrigger/Box is more than I currently have mental capacity to achieve, so I have not yet conducted an actual capital-C-R Code Review.

Gazook89:
> every other React component in the project, I'm not sure why testing has become necessary here

I don't think historical inertia is enough of a reason to *not* test, on it's own.  If we don't want to test because it's hard or something, we don't have to.  But I have found in other projects that testing components *is* helpful.  Being able to run a script to test different combinations of attributes/props rather than making small alterations and reloading the page each time saves time and is more consistent.  There are plenty of things that are not "immediately noticed" that we later go back and patch.  And having tests helps understand what things are important and why.  If *these* tests are not doing that then they can be improved, but I wouldn't say "components don't need tests".

> beta tags

I mostly agree here.  I didn't really care one way or another if we left them in.  But, to support the idea of removal, we don't really have any user-created open issues for any of these, so to some degree they must be "okay".

> unique key

fixed, thank you for pointing that out and offering the fix.

> menus are all much wider

The idea is that the menu trigger should have the same "grid" as it's menu items.  Our current site does not-- the icon space in the trigger is smaller than the icon space in the menu items:

| Live | PR |
|:--:|:--:|
|image | image |

>  tools collapse to the extreme left

This is set by some magic numbers in snippetbar.less on a couple lines, which I can fix quickly for this PR.  In the future, a better layout should be done to avoid these magic numbers.

> mouseover label is no longer visible

The tooltips is not something I'm working on here, but I have another branch not yet pushed to a PR that does tackle this.  The tooltips use static positioning.  They should be upgraded to also using Anchor Positioning and Popover API, just like these menus.  And then it will be a non-issue.   This was another reason that my UI Overhaul PR became so unwieldy-- lots of this UI stuff gets tangled up in a way that is hard to address one thing at a time.  If people would like, I can make a branch from this branch and we can merge that first.  But I'm not going to do that until there is a general mumbling of acceptance.

Thanks for reviewing

Gazook89:
I should also add that I believe that the menus can be made more visually distinct from the menubars they come from, but I didn't do that in this PR.  I didn't want to add a bunch of changes in the `.less` files that weren't directly related to functionality.  But things like a thin border, subtle box shadow, small margins around expanding menus could make menus contrast better with the elements they overlap.

5e-Cleric:
How ready is this? can we merge?

Gazook89:
> How ready is this? can we merge?

I guess I do have a question about how to manage the extra space that the snippet menus take now, related to this:

> tools collapse to the extreme left

I thought I would just adjust the magic numbers we have for the breakpoints in the snippetbar.less for the wrapping, but as I do so I'm thinking:

A. It might be better to just make the snippet bar two rows, permanently.

image

In my older UI Overhaul PR, I had something that was two rows as well, though it was a little more refined.  Moving to two rows with this PR would be a step in that direction:

image


B.  If not "A", then maybe the snippet menu triggers *are* too wide, and that matching the trigger menu spacing to the menu item spacing isn't such a big deal.  

I guess I'm up for whichever, but I would lean towards "A".

I'm out of town for the next week, so I can adjust that shortly after I get back.

5e-Cleric:
> > How ready is this? can we merge?
> 
> I guess I do have a question about how to manage the extra space that the snippet menus take now, related to this:
> 
> > tools collapse to the extreme left
> 
> I thought I would just adjust the magic numbers we have for the breakpoints in the snippetbar.less for the wrapping, but as I do so I'm thinking:
> 
> A. It might be better to just make the snippet bar two rows, permanently.

Not against it, but i do have to say, 70% of the time, i don't have the need for two rows at all. there is plenty of space in screen.

And if we do end up having an editor that we can move out of the screen or to the bottom, the double line would be unnecessary.

That said, it is a change we can undo at any point, i think it is preferred, barring dissagreement from @calculuschild 


> I guess I'm up for whichever, but I would lean towards "A".

So do i

> I'm out of town for the next week, so I can adjust that shortly after I get back.


As always, no hurry.

dbolack-ab (approved):
This is a massive improvement.

client/components/anchorPositioningPolyfill.js
```diff
@@ -0,0 +1,34 @@
...
+and then loads the Oddbird polyfill if support is lacking. 
+*/
+
+let polyfillPromise;
+
+// look for `anchorName` in the computed styles
+const supportsAnchorPositioning = ()=>'anchorName' in document.documentElement.style;
+
+// wait for initial render
+const afterInitialRender = ()=>new Promise((resolve)=>{
```
calculuschild: How necessary is it to wait for the initial render? What does that help us do? (It's also not clear why we have a double requestAnimationFrame happening.)

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -91,7 +91,6 @@ const BrewPage = (props)=>{
 
 //v=====--------------------< Brew Renderer Component >-------------------=====v//
 let renderedPages = [];
-let pageTemplates = [];
```
calculuschild, line 94: This PR is undoing the work from #4593 that sets up a V4 switch and a starting point for persisting style templates across pages. Could these removals be reverted?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -241,10 +226,8 @@ const BrewRenderer = (props)=>{
 		if(props.errors && props.errors.length)
 			return renderedPages;
 
-		if(rawPages.length != renderedPages.length) { // Re-render all pages when page count changes
```
calculuschild, line 244: I think this was a recent change as well from #4740 to fix a variables hoisting issue. Can we revert the deletion here also?

client/homebrew/homebrew.jsx
```diff
@@ -37,15 +37,9 @@ const Homebrew = (props)=>{
 			lang      : ''
 		},
 		userThemes,
-		brews,
```
calculuschild, line 40: More deletions of the already-merged V4 settings

config/default.json
```diff
@@ -7,6 +7,5 @@
 	"local_environments" : ["docker", "local"],
 	"publicUrl" : "https://homebrewery.naturalcrit.com",
 	"hb_images" : null,
-	"hb_fonts" : null,
-	"enablev4" : true
```
calculuschild, line 11: Another one

package.json
```diff
@@ -86,12 +90,12 @@
     ]
   },
   "dependencies": {
-    "@babel/core": "^7.29.7",
-    "@babel/plugin-transform-runtime": "^7.29.7",
+    "@babel/core": "^7.29.0",
```
calculuschild: A number of dependencies getting downgraded here.

themes/V3/Blank/snippets.js
```diff
@@ -330,7 +329,7 @@ export default [
 			},
 			{
 				name        : 'DTRPG Community Content',
-				incon       : 'fab fa-dtrpg',
+				icon       : null,
```
calculuschild, line 332: Are these icon removals intentional?

tests/components/dropdown/dropdown.structure.test.jsx
calculuschild, line 1: I'm not sure this testing is necessary either. From what I can see, it essentially comes down to writing out some HTML, and then reading it back out and checking that that HTML has the same properties it started with. My vote would be to just remove this.

Do you expect to see a lot of HTML getting clobbered in these dropdowns? I would think we can be pretty confident React and the browser aren't going to scramble their structure. There's not any real API or calculation going on here that I think is at risk of throwing an error or giving an unexpected result.

calculuschild (changes requested):
A few comments, mostly just noticing that some unrelated code seems to be getting erased here (maybe just need to rebase on the latest Master).

The functionality is good. My only real complaint is the snippet top-level items are much wider than before, which means I can't display the brewRenderer at 100% size and see all of the snippets. One or the other gets cut off, which didn't happen before.

Before (full page width and full snippet bar visible):
Image

After (full snippet bar gets covered):
Image

Any chance the left and right padding could be reduced maybe? Or remove the down arrows for now? Otherwise we are left with a toolbar that defaults to being partially obscured, which kind of reintroduces the problem this PR was trying to solve in the first place.

Gazook89:
I must have done a rushed merged with upstream/master and botched it, with all that v4 stuff removed.  Or I hadn't realized what that stuff was for, and didn't really see evidence of it being used.  I can't remember the exact though process, but i'll add those things back in.

Regarding the width of the trigger buttons:  I'll drop the arrows on those top level triggers, but keep them in the menus for submenu triggers.

client/components/anchorPositioningPolyfill.js
Gazook89: So with a single RAF, it would wait to wait to initialize the polyfill until just before initial render.  With the nested second RAF, it waits until before the second frame of rendering, which guarantees that the first has happened-- it's been painted to the screen.  This would be done for something like this polyfill which might need to know where elements are before it can measure them, to determine their positions.

However, I removed that initial render step just now on my branch and tested it, and it works fine.  So likely the polyfill already has that baked in/accounted for.

themes/V3/Blank/snippets.js
Gazook89: Yeah.  Correct me if I'm wrong, but I don't think these icons exist with font-awesome, or are otherwise defined in our app anywhere?

fa-dtrpg
fa-mit
fa-mongoosepub
fa-Paizo
fa-shadowdark

I don't see them appear in the live site, either.  I first got turned onto them because they were `truthy` in a conditional statement.

calculuschild: Ok. We do have a place where we add custom icons not in the font awesome package, but if it's already not working then they were probably never defined. We can remove them for now and if someone wants to make icons later we can add them.

calculuschild:
Yeah I believe Abquintic was starting to map out some structure to let us begin introducing V4 features behind a flag. It doesn't do anything yet but only because no V4 features exist yet.

client/homebrew/brewRenderer/brewRenderer.jsx
Gazook89: I believe I've now reverted this all back to #4593.  Unfortunately, I couldn't get git to let me just "revert" that specific commit, it kept giving me an error that I couldn't figure out.  So I manually put things back.  I wish I hadn't needed to do this so that the "blame" could still be on original contributor (dbolack/abquintic) but alas, it was beyond me.

client/homebrew/brewRenderer/brewRenderer.jsx
Gazook89: reverted this as well.

package.json
Gazook89: reverted all of these.

tests/components/dropdown/dropdown.structure.test.jsx
Gazook89: I guess I am looking at it as "things this component should do".  So if down the road we start mucking about with it, changing how it works, we still remember the things it should do.  In our PR's we are supposed to check off a list of things, including "verify things still work".  With these tests, we have that list of things that are supposed to keep working.

calculuschild: I get that. Though when I test those things I'm usually checking "did our function do what it is supposed to / did our API respond correctly", not just asserting the HTML structure. I guess the part I'm struggling with is most of these tests seem like they aren't checking "things it should do", but just static HTML properties or standard DOM behavior. If we write `<button class="xyz">`, React isn’t going to randomly stop outputting that class. So tests that assert “the class is still there” don’t really protect us from regressions in my mind. Similarly, testing that clicking a `<button>` calls its `onClick` handler is just testing that the DOM works.

The tests that do cover actual logic like disabling the trigger when childless, or the dismissal logic? maybe those make sense to keep since they validate conditional behavior. But the ones that just assert static HTML structure and DOM features I think we should drop. This PR has 312 lines of testing for a component that is 99 lines long, which feels excessive.

Gazook89: Okay, i just removed it all.

calculuschild (approved):
Ok, I will go ahead and merge this! Nice work!

PR #4770, opened 2026-04-24
Add image preview

### LINKED ISSUE #2037: Feature Request - Show image thumbnail on hover over URL in editor
This is a 'pie in the sky' request, but it'd be helpful to have a thumbnail appear when mousing over a URL in the editor.   With many images, it can be tough to determine which URL is which (especially if a user hasn't added the alt-text or title attributes).

5e-Cleric:
Adds an image preview on hovering over the image syntax.

### Preview
image

### HTML

```html
<div class="cm-line cm-activeLine">
    <span class="cm-image">
        <span class="cm-link">![homebrew mug](</span>
        <span class="cm-link cm-string cm-url">https://homebrewery.naturalcrit.com/assets/homebrewerymug.png</span>
        <img class="cm-widgetBuffer" aria-hidden="true">
        <img loading="lazy" class="cm-preview" src="https://homebrewery.naturalcrit.com/assets/homebrewerymug.png" contenteditable="false">
        <span class="cm-link">)</span>
    </span>
    <span class="cm-injection">{position:absolute,bottom:20px,left:130px,width:220px}</span>
</div>
```
As far as the double image is concerned, i don't get it, but it seems to be something that codemirror does when you insert widgets, its not like it fetches anything, not sure if there is anything we can do anything about it either.


client/components/codeEditor/codeEditor.jsx
```diff
@@ -73,16 +99,56 @@ const createHighlightPlugin = (renderer, tab)=>{
...
+									attributes : {
+										'style' : `--preview-img:url(${url});`
+									}
+								}).range(node.from, node.to)
+							);
+						}
+					}
+				});
+
 				tokens.forEach((tok)=>{
```
Gazook89: I know this predates this PR, but I would prefer that `tok` just be `token` for readability.

client/components/codeEditor/codeEditor.less
Gazook89, line 1: This could largely be replaced by Popover/Anchor Positioning for auto placement to prevent getting clipped by viewport or the divider.  I don't expect it in this PR, though.  It'd be easier if #4769 got merged first.

client/components/codeEditor/codeEditor.less
```diff
@@ -161,6 +161,48 @@
...
+			content:"";
+			position:absolute;
+			bottom:0;
+			left:0;
+			translate:0 100%;
+			display:block;
+			z-index:1000;
+			pointer-events: none;
+			opacity:0;
+			transition:0.2s opacity;
```
Gazook89: I think the transition from hidden to visible could benefit from a longer delay, something like native browser tooltips/title attributes.  So that when moving your mouse around, you aren't getting flashes of the preview.

client/components/codeEditor/codeEditor.jsx
Gazook89, line 1: Is it too much to write this image preview stuff as an extension, too?  After seeing all the other extensions get renamed/moved I figured this was the direction you were going, and I think it would make sense.

Gazook89 (commented):
You should display something for bad links, either just a "broken image link" icon in the center of the popup, or go further and check to see if it's an Imgur link that is missing a file extension and give a quick tip "Image is missing file extension, check for .jpg/.png".  That is probably one of the most common issues I see people ask about. 

It's also not clear to me why the image is used as a `background-image` in an absolutely positioned/dimensioned div, rather than just an `img`?  The result is an inflexible box with a border, with a smaller image inside it, and that image has effectively another border in the drop-shadow.  Using an actual `img` element would give accurate/tight sizing and the drop-shadow could be skipped, or put on the preview box itself.

The border of the preview box could be determined not by some set value in the stylesheet, but instead by the theme, if the classname matched an existing codemirror theme token class.  Perhaps give it the same color as comments.  Then, the border of this preview would be more harmonious to every theme.


----

For the record, I did check to see what this meant for network requests/lazy loading.  I was worried that it would try to load all the images at once if they were now also displayed in the editor.  But because CM isn't actually requesting the images until it renders the lines, it's effectively lazy loading itself.

client/components/codeEditor/codeEditor.less
5e-Cleric: Yeah, i am not very interested in the whole anchor thing, you can add it, it shoudl work very well with this.

client/components/codeEditor/codeEditor.less
5e-Cleric: good idea

client/components/codeEditor/codeEditor.jsx
5e-Cleric: moving it, yeah

5e-Cleric:
Alright, i was wrong i guess, i managed to do it this time around

5e-Cleric:
> The border of the preview box could be determined not by some set value in the stylesheet, but instead by the theme, if the classname matched an existing codemirror theme token class. Perhaps give it the same color as comments. Then, the border of this preview would be more harmonious to every theme.

because the color of text is set on by the color property, i'd have to set the border to be currentcolor, which is not impossible, but then i couldn't set the background to a color that wouldn't unmatch. The solution here would be to rework the 60 themes into using css variables for the colors, which would take some time. Not doing it for this small feature.

G-Ambatte (commented):
Currently, this spams literally thousands of network requests when a broken image link exists; this does not appear to stop when the image URL is fixed.

Gazook89:
O, probably because it happens on every keystroke?

Might have to do the fetch on hover, rather than fetch it first and only display on hover.

G-Ambatte:
> probably because it happens on every keystroke?

In my (admittedly very limited) testing, it started as soon as a broken URL was entered - copy and paste a working URL and it was fine; delete one character from the URL and instantly generate thousands of fetch requests.

5e-Cleric:
> Currently, this spams literally thousands of network requests when a broken image link exists; this does not appear to stop when the image URL is fixed.
> 
> 
> 

i think i know where this comes from, no problem, should be easy to fix, i was inputing an url to the actual icon for broken image in an an older iteration.

5e-Cleric:
this is fixed now!

G-Ambatte (approved):
Tested and confirmed URL fetch spam no longer occurs.

Mikle103:
when will this go live?

5e-Cleric:
> when will this go live?

With the next patch, probably before the next month ends, its not exactly a crucial update.

PR #4772, opened 2026-04-25
fix save on load

5e-Cleric:
Fixes the save on load time, by adding a first load guard.

Essentially, the save is triggered by the first load of the `saveGoogle` state.

### Why we should fix this

If i open 10 brews and don't save any of them on purpose, the 10 brews go immediately to the top of the updated list even though that is not true, the brew remains the same.

It also places an unnecessary load and takes time of the server.


calculuschild:
I had this solved in a different way on my local branch for merging edit/home/new. I was working toward removing some extraneous useEffect calls and this was one that just doesn't exist anymore. So with that in mind I just copied over the relevant changes here. Hope that's ok.

PR #4774, opened 2026-04-27
Better preview tool collapsing

Gazook89:
## Description

This PR revisits the Preview Toolbar at an aesthetic level. I've never been satisfied with the opening/closing mechanism, or the "glasses" icon.

Now:

- glasses are replaced by left/right caret depending on open state.  Seems more clear to me what the button will do than "glasses". 
- HeaderNav toggle is decoupled from toolbar toggle.  I think that if someone wants to "hide the toolbar", HB should expect that they really want to remove as much UI as possible.  If I could think of a satisfactory way to remove the caret while making it possible to bring the toolbar back, too, I would.   Note, the Header Nav itself doesn't hide when the toolbar is collapsed.
- better `transition`s now, using the new `interpolate-size` css property to be able to transition between `auto` and some fixed size (`100%`). 

Largely stems from the UI Overhaul work, but otherwise this hasn't been discussed, agreed on, etc.  

Click the Show/Hide caret toggle, check tooltips, squash viewport/preview pane, try header nav in both opened/closed toolbar state, check that state is preserved between open/close wherever it needs to be.

Note that in non-chromium browsers there are a couple features that aren't supported (`height: stretch` for one) but that doesn't really have any negative consequences.  Everything is totally fine in those browsers.

PR:
image

image

Current:
image

### Reviewer Checklist

  - [ ] Looks okay?
  - [ ] Toolbar opens and closes
  - [ ] state is preserved between opening/closing
  - [ ] Works in different browsers
  - [ ] works on mobile?

## Next steps

I'd like to re-think the `flex` layout of the toolbar so that it more gracefully collapses as the viewport shrinks.  I'm not a big fan of the vertical space it takes, especially beyond two rows.  But that's going to take some *work*, so leaving that for now.

I will also take a swing at reworking the tooltips component to work with the Anchor Positioning/Popover API's (and Oddbird polyfill) to get better algorithmic positioning, which is something that the Preview toolbar and other menubars would really benefit from.  As an independent PR.

### DIFF
--- a/client/homebrew/brewRenderer/toolBar/toolBar.jsx
+++ b/client/homebrew/brewRenderer/toolBar/toolBar.jsx
@@ -99,172 +99,177 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
 	if(!toolsVisible){
 		return (
 			<div id='preview-toolbar' className='toolBar hidden' role='toolbar'>
-				<div className='group toggleButton'>
-					<button data-tooltip-bottom={`${toolsVisible ? 'Hide' : 'Show'} Preview Toolbar`} onClick={()=>{
+				<div className='group'>
+					<button data-tooltip-right={`${toolsVisible ? 'Hide' : 'Show'} Preview Toolbar`} className='toolBarToggle' onClick={()=>{
 						setToolsVisible(!toolsVisible);
 						localStorage.setItem(TOOLBAR_VISIBILITY, !toolsVisible);
-					}}><i className='fas fa-glasses' /></button>
+					}}><i className='fas fa-caret-right' /></button>
 				</div>
 			</div>
 		)
 	} else {
 		return (
 			<div id='preview-toolbar' className='toolBar visible' role='toolbar'>
-				
-				{/*v=====----------------------< Zoom Controls >---------------------=====v*/}
-				<div className='group' role='group' aria-label='Zoom' aria-hidden={!toolsVisible}>
-					<button
-						id='fill-width'
-						className='tool'
-						data-tooltip-bottom='Set zoom to fill preview with one page'
-						onClick={()=>handleZoomButton(displayOptions.zoomLevel + calculateChange('fill'))}
-					>
-						<i className='fac fit-width' />
-					</button>
-					<button
-						id='zoom-to-fit'
-						className='tool'
-						data-tooltip-bottom='Set zoom to fit entire page in preview'
-						onClick={()=>handleZoomButton(displayOptions.zoomLevel + calculateChange('fit'))}
-					>
-						<i className='fac zoom-to-fit' />
-					</button>
-					<button
-						id='zoom-out'
-						className='tool'
-						onClick={()=>handleZoomButton(displayOptions.zoomLevel - 20)}
-						disabled={displayOptions.zoomLevel <= MIN_ZOOM}
-						data-tooltip-bottom='Zoom Out'
-					>
-						<i className='fas fa-magnifying-glass-minus' />
-					</button>
-					<input
-						id='zoom-slider'
-						className='range-input tool hover-tooltip'
-						type='range'
-						name='zoom'
-						list='zoomLevels'
-						min={MIN_ZOOM}
-						max={MAX_ZOOM}
-						step='1'
-						value={displayOptions.zoomLevel}
-						onChange={(e)=>handleZoomButton(parseInt(e.target.value))}
-					/>
-					<datalist id='zoomLevels'>
-						<option value='100' />
-					</datalist>
-
-					<button
-						id='zoom-in'
-						className='tool'
-						onClick={()=>handleZoomButton(displayOptions.zoomLevel + 20)}
-						disabled={displayOptions.zoomLevel >= MAX_ZOOM}
-						data-tooltip-bottom='Zoom In'
-					>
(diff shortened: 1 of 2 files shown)


5e-Cleric:
Pretty sure we want to keep the header nav button visible either way, and to the left where the nav is going to be, so maybe we should remove it from inside the bar.

Btw if you wouldn't mind, add a small little `pointer-events: none` when the bar is collapsed, because i think we missed that when we made it.

Gazook89:
>  add a small little pointer-events: none when the bar is collapsed,

You are right about the live site.  However, with this PR, the navbar doesn't just hide but it is fully removed.  So that won't be an issue.  

> Pretty sure we want to keep the header nav button visible either way, and to the left where the nav is going to be, so maybe we should remove it from inside the bar.

Can you explain why it should be visible either way?

5e-Cleric:
> > add a small little pointer-events: none when the bar is collapsed,
> 
> You are right about the live site. However, with this PR, the navbar doesn't just hide but it is fully removed. So that won't be an issue.

Great

> > Pretty sure we want to keep the header nav button visible either way, and to the left where the nav is going to be, so maybe we should remove it from inside the bar.
> 
> Can you explain why it should be visible either way?

If only cause its the current behavior. I don't recall if we discussed this, @G-Ambatte might have the answer.

Gazook89:
What if instead of having the toggles for the toolbar and the header nav in the Preview pane, they just live inside a 'View' menu of the top Navbar?

- We aren't having to think about how to collapse it so it's "hidden but also kind of visible", 
- The Toolbar and the HeaderNav can be toggled completely independently, with less overlap
- A 'View' menu is very common in similar programs, and ultimately it would likely get set as a submenu of another menu (Brew --> View --> Toolbar/HeaderNav for example).

5e-Cleric:
> What if instead of having the toggles for the toolbar and the header nav in the Preview pane, they just live inside a 'View' menu of the top Navbar?

top navar is outside of the scope of what happens in the editor, if we're moving them "out of the way" i'd sooner put them in the editor.

truth be told i wouldn't have them be collapsible at all, but i know that's just me.

Gazook89:
>  if we're moving them "out of the way" i'd sooner put them in the editor.

That's no good because the editor isn't available on the Share page.  

>  i wouldn't have them be collapsible at all, but i know that's just me

It's not difficult to offer the option to hide them, and for some users  especially on small screens it is important to be able to use as much screen space for reading as possible.  

> top navbar is outside of the scope of what happens in the editor

It's not entirely out of scope of what the Navbar already does, in terms of manipulating the editor or the preview.  It currently controls whether everything in the editor is swapped for something else (loading other brews or creating new ones), whether the editor saves its contents, or if the preview gets printed.  It's not strictly a navigation bar at all.

5e-Cleric:
> It's not entirely out of scope of what the Navbar already does, in terms of manipulating the editor or the preview. It currently controls whether everything in the editor is swapped for something else (loading other brews or creating new ones), whether the editor saves its contents, or if the preview gets printed. It's not strictly a navigation bar at all.

And i say those things shouldn't be there at all either.

Gazook89:
> And i say those things shouldn't be there at all either.
 
 In the interest of collaboration over dismissal, what do you want the top navbar to have?  Only navigation links that change the route or domain?
 
 - Naturalcrit.com
 - `/`
 - `/changelog`
 - Patreon.com
 - `/new`
 - Reddit.com
 - `/faq`
 - `/migrate`
 - `/vault`
 - [recent items]
 - naturalcrit.com/login
 - `/share`
 - Post to Reddit.com
 
 And not in the navbar:
 
 - Print (doesn't change route)
 - Save (sometimes changes the route, like New->Edit)
 - Drive/db toggle
 - autosave toggle
 - copy Share URL
 
 These items I guess would go into the editors or preview pane?  Maybe Print and Copy Share URL goes into the preview toolbar, and the rest go into the Editor toolbar?

5e-Cleric:
> > And i say those things shouldn't be there at all either.
> 
> In the interest of collaboration over dismissal, what do you want the top navbar to have? Only navigation links that change the route or domain?
> 
> * Naturalcrit.com
> * `/`
> * `/changelog`
> * Patreon.com
> * `/new`
> * Reddit.com
> * `/faq`
> * `/migrate`
> * `/vault`
> * [recent items]
> * naturalcrit.com/login
> * `/share`
> * Post to Reddit.com
> 
> And not in the navbar:
> 
> * Print (doesn't change route)
> * Save (sometimes changes the route, like New->Edit)
> * Drive/db toggle
> * autosave toggle
> * copy Share URL
> 
> These items I guess would go into the editors or preview pane? Maybe Print and Copy Share URL goes into the preview toolbar, and the rest go into the Editor toolbar?

right, navigation vs other functions sounds like a good separation point, we would still retain the navbar as a cluttered place, but maybe we can fix that by sorting the buttons properly in the future (maybe external vs internal links?)

the main issue is where those other buttons should go, to which, frustratingly, i do not have a decent proposal at this time, they could go in the preview toolbar, but the fact that it is inside the preview and that it can be hidden make me think it is not a good choice.

Even if the save option does sometimes change route, i don't think it is perceived as such to users, and it might not do so in the future, since we are combining the elements, who knows, maybe in a near future the new page might be `/edit/new`.

In any case all of this is probably out of the scope of this PR, and should be recorded in an issue, i'll stop blocking you with my stubbornness.

PR #4776, opened 2026-04-29
Lazy Load Images

Gazook89:
## Description

This very simply adds `loading="lazy"` to all `img` elements output by Marked.  Notably, it doesn't change html `` tags that are written in the editor, only the stuff transformed by Marked.

It defers the fetching of images until they are close to the current viewport (I think firefox is within 800px and chrome is more than that).  

It doesn't apply to CSS `background-image`s because those are loaded with the stylesheet.  So this is just inline images.  

- Use the attached brew download to create a new brew. This brew is made by another user.  It is 69 pages and has images using the markdown syntax (not ``).  
[HB-YonderPasttheDragonofIcepirePeak.txt](https://github.com/user-attachments/files/27577272/HB-YonderPasttheDragonofIcepirePeak.txt)

- Open Network tab of browser, filter requests to images
  - optionally, artificially throttle your connection speed to something like "slow 4G" in the network tab (on chrome)
- Reload the page with the brew, and watch the network requests.  As you scroll, you see in the timeline and record of requests the images only loaded as you scroll towards them.
  - You can filter the requests to just `img`.  The little blue lines are individual images popping in as I scroll down.  Looks like this:
image

If you articificially throttle the network speed from the Network tab (very top of the tab) to something like Slow 4G (or even Fast 4G), then clear cache and refresh, you can wait on the first page for everything to load up, and then start to scroll down to page 15. Likely, you will get to the page before the images are finished fetching.  This shows that lazy-loading is working though you might think "well this stinks".  But at least someone on the slow network isn't downloading every image immediately after clicking into an HB link.

## Possible Next Step
I could try to figure out something for transforming html `` tags (not just ones created by Marked).  Hopefully people aren't using Legacy renderer anymore, but still many of those brews exist and get loaded.  This reduces load on the server and "saves the planet" and makes things load faster.  So it may be worth the effort.


Gazook89:
From Matrix:
> @calculuschild: for lazy loading, I vaguely remember we might have tried to implement it before and ran into some issue. I can't seem to find the detail right now, but there is this older conversation about a user doing this and it messed with the PDF output. Can you confirm printing isn't weird if you do lazy loading?
> @dbolack-ab : Yes, lazy loading breaks onLoad lookers

I'll check on this next.  From the little reading I've done just now, it seems that:
- maybe this shouldn't be true in Chrome anymore, but maybe there is an extra wrinkle with the use of the iframe.  I could just be wrong. [Related: [chrome bug](https://issues.chromium.org/issues/40589907#comment37)]
- there is an event `onbeforeprint` that could be used to strip the `loading` attribute (or change it to `eager`). [Related: [MDN - before print](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeprint_event), [SO - older thread, some recommendations towards before print](https://stackoverflow.com/questions/43707762/how-to-lazy-load-images-and-make-them-available-for-print)]

Need to actually try these things and keep looking, will do that later.

Gazook89:
> there is this older conversation about a user doing this and it messed with the PDF output. Can you confirm printing isn't weird if you do lazy loading?

Last commit fixes this, at least for me.  All MarkedJS img tags are still lazy loading, but when you initiate Print via the Get PDF button or hotkey, it first switches all of those img elements to "eager" loading (they just load asap).  It then goes through each of those imgs and waits for them to load before continuing to open the print dialog.

An unwanted consequence is that if you have a slower connection there is a longer delay with no feedback after you click Get PDF before you get the Print dialog.  So it feels like nothing happened, and that stretch of time increases for the slower internet connections.  If this PR is going in the right direction, then I think the next step is to add some popover or indicator that it's "preparing" to print or some such.  

I'm not performance/network whiz, but here are some quick numbers I see in my network panel when emptying cache and doing a hard reload:

Numbers are filtered to only images, no other resources, and these are both on local (not "Live" vs "local")

|   |Master | PR |
|:---| :-----:| :----:|
|  requests | 87 | 54 |
| kB transferred | 29,295 | 5,174 |
| Finish | 5.86-6.07s | 5.92-5.99s |

*4 trials each branch*

That "Finish" time is sort of meaningless for this, right now-- the last items to load are images inside the Themes stylesheets and they come in about 5s after everything else.  Cutting down that gap would be neat, but I haven't put any effort into that.  

Overall, though, a lot less data is getting transferred on initial load.  Obviously stuff has to load *sometime*, but if someone is coming to a 300 page brew with tons of images and using slower speeds, at least they can start scrolling faster or decide to bounce out of the brew before everything is downloaded.

Gazook89:
> An unwanted consequence is that if you have a slower connection there is a longer delay with no feedback after you click Get PDF before you get the Print dialog. So it feels like nothing happened, and that stretch of time increases for the slower internet connections. If this PR is going in the right direction, then I think the next step is to add some popover or indicator that it's "preparing" to print or some such.

I added custom events that fire at the beginning and end of the print cycle (from pressing "Get PDF" or using Ctrl P hotkey up until the print dialog opens), so a little "loading" message is shown instead of "Get PDF" while the images are loaded.

Also updated the failing tests.

I think this is ready to look at when someone has a moment.

5e-Cleric:
> > An unwanted consequence is that if you have a slower connection there is a longer delay with no feedback after you click Get PDF before you get the Print dialog. So it feels like nothing happened, and that stretch of time increases for the slower internet connections. If this PR is going in the right direction, then I think the next step is to add some popover or indicator that it's "preparing" to print or some such.
> 
> I added custom events that fire at the beginning and end of the print cycle (from pressing "Get PDF" or using Ctrl P hotkey up until the print dialog opens), so a little "loading" message is shown instead of "Get PDF" while the images are loaded.
> 
> Also updated the failing tests.
> 
> I think this is ready to look at when someone has a moment.

I am not sure what this is trying to fix, what's the relevance? fixing printing mask load? if so, that should be fixed by chromium's team, this would be a patch over it that should be removed as soon as possible when the real issue is fixed.

Is this trying to fix something else?

Gazook89:
> I am not sure what this is trying to fix, 

When you trigger the print function, it takes the page as it is at that time.  So if we load the page up, and hit "print" before scrolling through the brew, it will only send to the Print Preview images that have been fetched and loaded.  Which will not include anything beyond the viewport.  

So this defers loading the Print Preview/Print Dialog until all images are fetched and loaded.

dbolack-ab (approved):

5e-Cleric:
Testing this... images load fine Live, its only masks that i see fail, does this affect mask-image images too? (i assume not)

As for the deployment, the loading placeholder appeared and was gone in 3 seconds, printing still took the same time to appear (almost 1 minute), and masks are still failing.

The result print is equal for me live and in this deployment, sorry.

This for an image heavy brew of 75 pages that i can't really share.

Maybe i should be using a testing brew.

Gazook89:
This changes nothing about printing (to pdf). It optimizes page load speed by only fetching images in the Preview pane just before they come into view. And, if someone opens a brew and only looks at the first page, it doesn't fetch everything else. 

The work that touches on Print to PDF is added because we need to quickly fetch all those images at that point. 

Regarding masks: this doesn't impact masks because those are loaded with the style sheet-- they can't be lazy loaded (or at least I'm not familiar with that).

5e-Cleric:
> This changes nothing about printing (to pdf). It optimizes page load speed by only fetching images in the Preview pane just before they come into view. And, if someone opens a brew and only looks at the first page, it doesn't fetch everything else.
> 
> The work that touches on Print to PDF is added because we need to quickly fetch all those images at that point.
> 
> Regarding masks: this doesn't impact masks because those are loaded with the style sheet-- they can't be lazy loaded (or at least I'm not familiar with that).

aaaaah, i finally get it, i'm slow these days i guess, alright, testing performance on site then.

Okay, now, I am testing this clearing cache and looking at the network tab, and from the entire brew with dozens of images, just 2 loaded after the initial load. I do see assets loading later, but only masks and watercolors. I might not be understanding how lazy loading works.

Also, according to https://stackoverflow.com/questions/57753240/native-lazy-loading-loading-lazy-not-working-even-with-flags-enabled ,  it is possible some other factors are contributing to lazy loading not really working.

On another note, i cannot see any performance increase, but i also don't see any performance issues, should i be testing with more than 75 pages?

for reference, this is live:

image

This in the PR:

image



Plenty of big images

Actually, if i am reading the waterfall correctly, they are.. loading faster in the PR?

Gazook89:
I've update the original PR comment, the section about QA/testing, with some additional context including a test brew.

I hadn't had that Waterfall column displayed in my own browser, so I'm just reading up on it now. I could be wrong, but my understanding is that Waterfall column is not a good indicator for the effectiveness of this PR.  It tells you the timing of the first assets to load on page load, and specifically the time elapsed between request sent, response received, and download time.  This PR won't change those things, it changes when the fetch request is sent.

5e-Cleric (approved):
Its a small but potentially powerful addition.

dbolack-ab (approved):
Worst case, this will not improve our print problems in the wild. It seems a good thing to put in and evaluate for six months. I didn't see potential for failure

5e-Cleric:
@Gazook89 can i merge this?

Gazook89:
> can i merge this?

Yup.  🚀

PR #4780, opened 2026-05-01
Open Sans: Replace Google Fonts with Installed Fonts

Gazook89:
## Description

Open Sans is the one font that we pull from a CDN rather than have installed font files.  This PR installs the woff2 font files and removes the CDN link.  It installs the same weights (400, 700) as was being pulled from the CDN, and it *does not* install weights 300 and 600 which we had through the CDN but *we don't actually use anywhere on the site*.  

Note: something I noticed is that the live site pulls 300, 400, 600, 700 weights.  But the site only uses 400 & *800* in declared CSS rules.  Since we never loaded a *800* weight, anything that was set to 800 was actually pulling from the 700 weight file.   I checked what actual 800 weight looks like, and it's too thick.  I'm not sure we want to change everythying to actually say "700", or just let it keep working as-is.

Another small change:  the Open Sans font snippet was wrong, it was outputting 'OpenSans' when it should be 'Open Sans'.  Fixed now.

Check that Open Sans loads wherever it is used.  Mostly the UI, but also the Open Sans font snippet.


dbolack-ab:
Include all weights so they are available.

Gazook89:
> Include all weights so they are available.

I just didn't want to include a bunch of font files that aren't getting used.  However, it's possible with more effort to use the variable font version of Open Sans.  Until now, I thought it'd be hard (impossible) to get variable font as a WOFF2, but it seems like [this might have the answer (use command line to subset the unicode ranges we want, and output as WOFF2)](https://clearleft.com/thinking/how-to-use-variable-fonts-in-the-real-world).  

Is it worth the trouble?  I'd like a second and third opinion on that

Also, I honestly think it's weird to have Open Sans in the snippets menu-- just because we load it for the UI doesn't mean we should encourage it in brews.

5e-Cleric:
> Also, I honestly think it's weird to have Open Sans in the snippets menu-- just because we load it for the UI doesn't mean we should encourage it in brews.

It is available to the users, and the license allows it, it would be weird not to have it, in my opinion.

Gazook89:
> It is available to the users, and the license allows it, it would be weird not to have it, in my opinion.

I guess I just think that a lot of the value of Homebrewery is that is opinionated/curated, meaning that we automatically dump users from the very start into a particular theme-- not only a theme, but a very specific existing property.  We don't open with a blank document with no styling like other programs.  Opening with most of the styling preset, narrowly focused, helps people just start writing.

That narrow focus is somewhat diluted by having something like Open Sans (and Code Bold/Light, Courier, TNR, etc) listed with all the other fonts.  

All of Google Fonts, Fontsource.org, Bunny.net (fonts cdn) is available to users, fairly easily.  They all have open licenses.  But we don't have snippets for all of them. Open Sans, being in our codebase (and even currently as only a CDN), doesn't mean it belongs in the 5e PHB theme.  The snippet pattern is clearly established in the snippets that *are* provided: `{{font-family:"..." Dummy Text}}`.  From that pattern it's not hard for a user to understand the convention.  (side note: I think the font snippets should *all* use the double quotes for the font name, not just the ones with spaces, to avoid confusion). 

Another way to think about it:  If we add more fonts to the repo for additional official themes, say a handful for a sci-fi theme, do we make those additional fonts available on every brew?  Right now, on Blank theme, all the same DND fonts are available.  Should these fonts not be scoped to specific themes? And if they are, do we need some subset of fonts available in every theme, or only have themes-specific fonts?

-----

I know I brought up the questionable inclusion of these fonts as snippets, but I didn't mean it to be a blocker to this PR and this PR doesn't actually change that.  In fact, it makes one of those snippets actually *work*. 

It also removes two weights, 300 and 600, which don't even correspond to `normal` (400) and `bold` (800) (for a font snippet that wasn't correct anyway).  For me, those weights fell on the side of "extra stuff not worth including".  

If the loss of those two oddball weights is really a blocker, I can add them back in, I just didn't want it to be something left up to inertia without actually considering it.

calculuschild:
> Another way to think about it: If we add more fonts to the repo for additional official themes, say a handful for a sci-fi theme, do we make those additional fonts available on every brew? Right now, on Blank theme, all the same DND fonts are available. Should these fonts not be scoped to specific themes? And if they are, do we need some subset of fonts available in every theme, or only have themes-specific fonts?

I think there is room for some discussion on this; I do think it makes sense for the snippets at least to be scoped to their relevant themes. But I can see it would be handy to still at least have access to other fonts if you happen to know the font name, no matter what theme you are in.

> Removing font sizes

I'm OK with removing font sizes that aren't used. Bold and Normal are probably sufficient for most users that will just be doing `**bold**` with markdown. I don't think we keep multiple weights of any other fonts.

> If the license allows it

We should include the font license in the repository if possible

___

One minor nitpick. I would make the snippet and font-family name stay as `OpenSans`, one word, for three reasons:
1) For unity and naming predictability with the other font class names. I think the only other case that has spaces is "Times New Roman", which probably would make sense to change or create an alias to `TimesNewRoman` also.
2) While quotes *do* work, it would be nice to not actually need them for any built-in snippets. The `{{ }}` syntax rarely has whitespaces that requires quotes; it's only when users start going off the rails with custom stuff, and it would be nice to keep more of the casual-user-facing stuff consistent and less "cluttery".
2) The Changelog document already uses `OpenSans` for certain things already (with a fix in the brew CSS to make it work). I assume some people have copied that and are already using it.

Gazook89:
> re Theme-specified fonts vs general fonts: I think there is room for some discussion on this;

This PR doesn't need to stop for this discussion, though, right?  

> include font licenses

Okay, I think that can happen in the same discussion as above.  For reference, the SIL license: [https://openfontlicense.org](https://openfontlicense.org/).

> font-family:OpenSans

Removing the spaces doesn't seem right to me. To me, it's an additional layer of confusion if someone knows the font they are looking for, and knows it has spaces.  More confusion than the relatively straightforward use of quotes.  But I don't think there is any way to know for sure which is more or less intuitive.  But by virtue of being in a pre-made snippet menu, where the complexity is sort of irrelevant, I would think that leaning towards being *right*, meaning following the actual name of the font, would make more sense.  

Outside the box:  what if we added utility classes to the codebase for our "generic", non-theme fonts:

```css
.openSans {
  font-family: "Open Sans";
}
```

And then change all those snippets to `{{openSans dummy text}}`.  That way, our font remains `Open Sans`, the name can still be set with `{{font-family:"Open Sans" dummy text}} but `openSans` works as a short class.

Theme authors could follow a similar pattern, but with a little twist:

```css
.gameMechanic {
  font-family: "Open Sans";
}
```

Then that gets included in their theme template brew custom snippets and their users are getting a snippet called "Game Mechanic" but with that Open Sans font (or whatever other font).  In practical, immediate terms, that would mean we could redo the 5ePHB font snippets to output `{{nodestoCaps dummy text}}`, `{{mrsEavesSmallCaps dummy text}}` etc.  Existing snippets would continue to work, they are still getting the same font names, really.

calculuschild:
> This PR doesn't need to stop for this discussion, though, right?

No this is not a blocking issue

calculuschild:
> Removing the spaces doesn't seem right to me. To me, it's an additional layer of confusion if someone knows the font they are looking for, and knows it has spaces.

I would agree, except it goes against what we have done for all the other fonts, which makes this the one case going against the rest of the group (except times new roman). **Nodesto Caps Condensed** and **Scaly Sans Remake** and **Mr Eaves Remake** all have spaces in the font name, but we have been consistent thus far in stripping them out when defining the font face.

For me, breaking the pattern would be more confusing as a user than sticking with the established format.

> ...making special font classes

I'm not opposed to this. I believe we originally put "font-family" into the snippet to provide more of a framework for users to load their own fonts. But that may be better hidden and left for power users to figure out if it removes some clutter for casual users. Not sure.

Gazook89:
Okay, reverted that commit to the snippet.  Good point on the other fonts having their spaces stripped away, I hadn't noticed and I should've before adjusting Open Sans.  

I'll leave utility classes to some future PR. 

Any other issues here I missed or good to go?

Gazook89:
So suggestion from @ericscheid is to just double up the font declarations for both "Open Sans" and "OpenSans", which I'm totally fine with, and I saw no complaints about that path besides just general grumblings that it's "low priority".

I have no problem taking the few minutes to do that for each of the fonts we have already, though I'd skip it for this PR.  It'd happen as either it's own small and easy thing, or part of the earlier discussion about moving the PHB fonts into theme-specific snippets (and more generic fonts into maybe the Blank theme).  

So no changes here, this is still good for review.

5e-Cleric (approved):

5e-Cleric:
If this is ready for merge, we can merge it now.

Gazook89:
> If this is ready for merge, we can merge it now.

It only has one approving review, but merge at your discretion.

PR #4782, opened 2026-05-03
Fixes the next batch of codemirror bugs

### LINKED ISSUE #4771: Bugs with codemirror
Logging this myself as i use the tool in its last version:

- scroll positions between editors not held.
- Sometimes text dissapears on tab swap, scroll brings it back.
- folding is not maintained on tab swap

These are absolutely annoying, and probably hard to fix, so... time consuming for me, yikes

5e-Cleric:
Fixes the following bugs:

- page computing was ocurring when editing the document's css or snippets, but not when switching back to text tab, so scroll to page was failing to find the position. This PR stops the recomputing if not in the text tab

- Folding state was not maintained between tabs
- Tab was somehow broken again
- Add strikethrough highlight

Not sure what's left.


dbolack-ab (approved):

calculuschild (approved):
The bulleted items seem to work for me, and the code looks good. Go ahead and merge if this is all for now.

PR #4797, opened 2026-05-08
make wrapselection work with multiple selections

5e-Cleric:
Wrapselection (and my previous iterations) didn't really take into account bolding more than one selection at a time, and just took the last selection made, this fixes that, by passing the full selection ranges object and looping over it to collect the necessary changes.

Codemirror seems to intuitively get the selection ranges at the end without us telling it speficially, which is very nice.


Gazook89:
So there is a case here where I don't think this works as expected.  If I have a table:

```md
##### Character Advancement
| Experience Points | Level | Proficiency Bonus |
|:------------------|:-----:|:-----------------:|
| 0                 | 1     | +2                |
| 300               | 2     | +2                |
| 900               | 3     | +2                |
| 2,700             | 4     | +2                |
| 6,500             | 5     | +3                |
| 14,000            | 6     | +3                |
```
And I use `Alt/Option` + mouse to select `2,700`, `6,500`, and `14,000` in the first column, and then `Ctrl/Cmd` + `B` to bold all three ranges at the same time, I get this:
```md
| **2,700 **        | 4     | +2                |
| **6,500 **        | 5     | +3                |
| **14,000**        | 6     | +3                |
```
Which renders as:
image

And I get that is because the multi-range selection contains three ranges that are the same length, and that the first two include the spaces inside their selection.  But it is still not what I would *expect*.  

**Can we just add `.trim()` to the assignment of `selection`?**  

----

Second question:  Just earlier today I was looking at this function and posted [this question to the original PR that added this](https://github.com/naturalcrit/homebrewery/pull/4763#issuecomment-4407887083), about how the selection works.  Did you see that, and if it's clear enough, could you try to get that to work in this PR?  Though I'm not entirely sure how it would play well with multi-ranges.  If it doesn't make it in this PR, that's totally fine, but since you are playing around in here I thought i'd ask.

5e-Cleric:
> Can we just add .trim() to the assignment of selection?

pretty sure that will fail, and even if not, that is not what the user would expect, neither is the previous behavior of the editor, the text you select is the one that gets the wrap, not a subset of that.

I am against it.

5e-Cleric:
> Second question: Just earlier today I was looking at this function and posted https://github.com/naturalcrit/homebrewery/pull/4763#issuecomment-4407887083, about how the selection works. Did you see that, and if it's clear enough, could you try to get that to work in this PR? Though I'm not entirely sure how it would play well with multi-ranges. If it doesn't make it in this PR, that's totally fine, but since you are playing around in here I thought i'd ask.

That is interesting, although i don't see how that would work at all, i'd have to get the text from outside the selection, which... i guess we can do...

Gazook89:
I have this done actually. Just haven't had time to share it. 

Both Trim and the other thing

Headed to wedding now so will come back to this tomorrow

5e-Cleric:
> I have this done actually. Just haven't had time to share it.
> 
> Both Trim and the other thing
> 
> Headed to wedding now so will come back to this tomorrow

Are you fine with me merging this and then you opening a PR for that that i can review?

Gazook89:
Ah yes, sorry.

Gazook89 (approved):

PR #4799, opened 2026-05-09
Prevent pattern matching in HTML replace function

G-Ambatte:
## Description

The use of a string in the `replace` function causes the function to attempt pattern matching on the output, resulting in a number of unexpected behaviours.
Changing this to instead pass a function as the parameter disables this behaviour.

TBC

### Reproduction:
1. On /new, create a brew containing `$$`.
2. Save.
3. In /edit, observe that the brew now contains a single `$` and immediately reports an "Out of Sync" error.

### Reviewer Checklist

1. [x] Performing the reproduction steps does not result in a single `$` nor the "Out of Sync" error.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -593,7 +593,7 @@ export default async function createApp(vite) {
 
 		html = html.replace(
 			'<head>',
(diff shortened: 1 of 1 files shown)


5e-Cleric (approved):
Does solve the $ error, not sure if there aren't any crazy errors coming but feel sure enough that is not the case to merge.

calculuschild (approved):

calculuschild:
Merging!

PR #4802, opened 2026-05-10
Log additional information during brew mismatches

G-Ambatte:
## Description

In order to help diagnose the cause of hash mismatches, this PR logs additional information about the error.

None

When the `debugTextMismatch` function fires, this change causes additional data to be added to the log.

### Reviewer Checklist

1. Initiate a hash mismatch
2. Observe that the log window now contains additional data.

### Example of Log Output

image


5e-Cleric (approved):
Seems detailed and a sensible approach

PR #4814, opened 2026-05-13
fix: correct invalid HTML pattern attribute for language input (#4813)

### LINKED ISSUE #4813: pattern attribute for language propery is invalid
Currently, the input for the Language property is:

    <input title="" pattern="/.+/" placeholder="en" type="text" value="">

That pattern is bad. The slashes should not be there (it's html, not javascript). It looks like that's the default pattern value from our <Combobox> function.

It should be more like this:

    pattern="^[a-z]{2,3}(-[A-Z][a-z]{3})?(-[A-Z]{2})?$"

That, btw, also allows for the rare 3 letter regions, 4 letter scripts (e.g. `Hant` for traditional Chinese), and then the final 2 letter language. (We have both in the example drop-list for the input).

----

The default `pattern` attribute value is defined here, as a javascript literal, but should be as a string, and looks to be inserted as a string of the literal js regex:
https://github.com/naturalcrit/homebrewery/blob/989ac071be50901986f4bea1714cd90fb28cd60c/client/components/combobox.jsx#L20

The custom combobox for language is defined here, it looks to be using the default combobox pattern (which is itself bugged), should have `pattern="^[a-z]{2,3}(-[A-Z][a-z]{3})?(-[A-Z]{2})?$"` inserted.
https://github.com/naturalcrit/homebrewery/blob/989ac071be50901986f4bea1714cd90fb28cd60c/client/homebrew/editor/metadataEditor/metadataEditor.jsx#L257-L278

The above 2 suggestions modulo a better understanding of React. I don't even know the assumed syntax for `valuePatterns`.

shuibui:
## Summary

Fixes the invalid `pattern` attribute for the language input field in the Homebrewery editor.

### Problem

The `pattern` attribute for the language input was using a JavaScript regex literal (`/.+/`) instead of a valid HTML pattern string. When passed to the HTML `pattern` attribute, this resulted in an invalid/inoperative validation pattern.

### Changes

1. **combobox.jsx**: Change `valuePatterns` default from `/+.+/` to `'.+'` (proper string)
   - HTML `pattern` attribute expects a string, not a JS regex literal

2. **metadataEditor.jsx**: Add BCP 47 language tag pattern: `^[a-z]{2,3}(-[A-Z][a-z]{3})?(-[A-Z]{2})?$`
   - Supports 2-3 letter language codes (en, eng, hac)
   - Supports 4-letter script codes (e.g., Hant for Traditional Chinese)
   - Supports optional 2-letter region codes (e.g., US, CN)

### Test

- [ ] Enter a language code like 'en', 'eng', 'Hant', 'en-US' in the language field — all should validate correctly
- [ ] Enter invalid input like 'xyz' — should be rejected

Fixes #4813

### DIFF
--- a/client/components/combobox.jsx
+++ b/client/components/combobox.jsx
@@ -17,8 +17,7 @@ const Combobox = createReactClass({
 				suggestMethod           : 'includes',
 				filterOn                : []  // should allow as array to filter on multiple attributes, or even custom filter
 			},
-			valuePatterns : /.+/
-		};
+			valuePatterns : '.+';
 	},
 	getInitialState : function() {
 		this.dropdownRef = React.createRef();
--- a/client/homebrew/editor/metadataEditor/metadataEditor.jsx
+++ b/client/homebrew/editor/metadataEditor/metadataEditor.jsx
@@ -261,6 +261,7 @@ const MetadataEditor = createReactClass({
 					className='language-dropdown'
 					default={this.props.metadata.lang || ''}
 					placeholder='en'
+					valuePatterns='^[a-z]{2,3}(-[A-Z][a-z]{3})?(-[A-Z]{2})?$'
 					onSelect={(value)=>this.handleLanguage(value)}
 					onEntry={(e)=>{
 						e.target.setCustomValidity('');	//Clear the validation popup while typing


5e-Cleric:
hi shuibui, can you confirm to me that you are human and not an ai so i can merge this?

5e-Cleric:
Closing as they do not respond and tests are failing, this is clearly an ai PR.

PR #4837, opened 2026-05-27
[ADMIN]: Proper cleanup function

5e-Cleric:
This PR updates the cleanup function to actually show what it will delete, and to get proper trash brews that can be deleted safely.

Specifically brews that haven't been updated or viewed for over a year, and have no registered author.


calculuschild:
Can we keep the search for tiny brews too? That can be handy sometimes to weed out junk.

client/admin/brewUtils/brewCleanup/brewCleanup.jsx
```diff
@@ -1,8 +1,10 @@
 import React, { useState } from 'react';
 import request from 'superagent';
+import Moment from 'moment';
```
Gazook89, line 3: Since this is admin-facing only, I recommend swapping Moment for the newer [Temporal API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal) to manage dates.  It is supported by Chrome and Firefox, and largely in Safari Technical Preview.  

Will it make a big difference here, where we already use Moment in a few places on the user-facing side?  No.   But the [MomentJS team themselves indicate that there are better alternatives](https://momentjs.com/docs/#/-project-status/), and recommends using Temporal:

>  We recognize that many existing projects may continue to use Moment, but we would like to discourage Moment from being used in new projects going forward. ... We would also like to promote the [Temporal](https://momentjs.com/docs/#/-project-status/future/) addition to the JavaScript language, which is looking for feedback and contributors.

I think once Temporal comes into regular Safari (out of TP), then the user-facing side can be updated as well.

client/admin/brewUtils/brewCleanup/brewCleanup.jsx
```diff
@@ -36,6 +39,43 @@ const BrewCleanup = ({})=>{
 			setPrimed(false);
 		}
 	};
+	const renderBrewList = ()=>{
+		if(!brewCollection) {
+			console.log(brewCollection)
+			return null;
```
Gazook89: This maybe return a line that says "0 brews matched the query" or whatever, rather than just doing nothing besides adding something to the console.

Or, don't `return` out of the method here, but have it create the table below but with:

```
<tbody>
<tr><td colspan=5>No results</td></tr>
</tbody>
```

client/admin/brewUtils/brewCleanup/brewCleanup.jsx
```diff
@@ -1,70 +1,156 @@
...
+			</button>
+			{renderFound('junk')}
+
+			{error && <div className='error noBrews'>{error.toString()}</div>}
+		</div>;
+	};
+	const renderLostBrewCleanup = ()=>{
+		return <div className='lost'>
+			<h3> Lost brews</h3>
+			<p>Removes very short brews to tidy up the database</p>
```
Gazook89: This has the same description as the Junk Brews

> Removes very short brews to tidy up the database

It should indicate what the criteria is.  So maybe update that description such that it summarizes the difference from Junk brews, and then have a disclosure (`details/summary`) element that actually spells out exactly the criteria, for anyone who might look at this page but not already familiar with the code/query behind it.  Then do the same for the Junk brews section as well.

Gazook89 (changes requested):
I don't ever mess around in the admin page or db so I'm wary of giving my stamp to anything that actually *touches* the DB, particularly around deletions.  But I did add some comments about the front end of it.  Hopefully others can pick up the slack on the DB side.

client/admin/brewUtils/brewCleanup/brewCleanup.jsx
5e-Cleric: I think considering this is working and it is admin side, we don't really need any better?
It would move something from working back into the "might be some bugs" stage.

5e-Cleric:
I'm not even sure this should be merged, it is here so everyone can go through the code, but i can run this locally to clean things up.

Gazook89:
> I'm not even sure this should be merged, it is here so everyone can go through the code, but i can run this locally to clean things up.

If it works, I think you should merge it (pending approvals).  It's not out of date with `master`, it provides helpful functions, so it should be merged.

client/admin/brewUtils/brewCleanup/brewCleanup.jsx
calculuschild: I agree with @Gazook89 . Don't necessarily need the specialized summary div, but the description the admin sees should accurately describe what the pipeline is searching.

calculuschild (approved):
Looks good to me as long as the displayed descriptions are updated to match what is being searched under the hood.

5e-Cleric:
image


there, as requested. Awaiting last review.

calculuschild (approved):

PR #4850, opened 2026-06-02
Add back selection tabbing

5e-Cleric:
Earlier this year, i removed tabbing to move the entire line to the right as it was not our usual behavior.

Today i have been asked why didn't that work when the code was selected. Turns out, we did do it, only when multiple lines were selected.

This PR brings back indenting when multiple lines are selected using only one selection (for my own peace of mind)


client/components/codeEditor/extensions/customKeyMaps.js
```diff
@@ -1,14 +1,40 @@
...
+			from   : range.from,
+			to     : range.to,
+			insert : '  '
+		});
+
+		ranges.push({
+			anchor : range.from
+		});
+	}
+	const tr = view.state.update({ changes });
```
Gazook89: What is `tr`?  Can you make that variable name more descriptive?

client/components/codeEditor/extensions/customKeyMaps.js
```diff
@@ -1,14 +1,40 @@
 /* eslint max-lines: ["error", { "max": 300 }] */
 import { keymap } from '@codemirror/view';
 import { undo, redo, indentMore, deleteLine } from '@codemirror/commands';
+import { EditorSelection } from '@codemirror/state';
 import { Prec } from '@codemirror/state';
 
 const insertTab = (view)=>{
-	const { from, to } = view.state.selection.main;
+	const shouldIndent = view.state.selection.ranges.some((range)=>view.state.doc.lineAt(range.from).number !==
```
Gazook89, line 10: A part of me is skeptical that this has to be as *involved* as it is laid out here, to manage the indents.  In this block there is a CodeMirror-provided `indentMore` command, but I don't see that it uses the opposite `indentLess` command that is also provided by CodeMirror.  Those two commands automatically add indentation to all ranges in a selection.

Further, [there is also a `indentSelection` command](https://codemirror.net/docs/ref/#commands.indentSelection) that can correctly set the indentation of selected lines. From CodeMirror's repo:

```
/**
Auto-indent the selected lines. This uses the [indentation service
facet](https://codemirror.net/6/docs/ref/#language.indentService) as source for auto-indent
information.
*/
declare const indentSelection: StateCommand;
/**
Add a [unit](https://codemirror.net/6/docs/ref/#language.indentUnit) of indentation to all selected
lines.
*/
declare const indentMore: StateCommand;
/**
Remove a [unit](https://codemirror.net/6/docs/ref/#language.indentUnit) of indentation from all
selected lines.
*/
declare const indentLess: StateCommand;
```

Is there something I have overlooked or would be missing if we weren't ourselves adding `'  '` to the start of lines, line by line?

Gazook89: [Here is the section all about indentation](https://codemirror.net/docs/ref/#h_indentation), though this stuff likely gets more technical than we need here.

Gazook89 (changes requested):
I'm glad this is getting added, the ability to manage indentation across multiple lines.  The fact that I *can't* do this when making lists in *Github* is annoying, so we'll have a leg-up there.  Thanks for doing it.

I made a little list of expectations below and all are met, except it seems weird that the details of the implementation have to require as much code as is written here.  Codemirror being used as much as it is for writing many types of documents, including code editors, I would expect that this kind of indentation function would be built-in, maybe even the default.  There is another comment to the same effect on those particular lines.

------
- [x] Selecting multiple lines and
  - [x] `tab`ing indents lines
  - [x] `shift-tab`ing dedents lines
- [x] changing indent retains same text selection, adjusted for additional/removed spaces.
- [x] Selecting multiple rows of a single wrapped line and `tab`ing *doesn't* indent the line (replaces selection with a tab).
- [x] Works in
  - [x] Brew Editor
  - [x] Style Editor
  - [x] Snippets Editor
- [ ] Clean code

client/components/codeEditor/extensions/customKeyMaps.js
5e-Cleric: this constant is like a record of the changes, so we can map the old positions to the new ones, codemirror ususally refers to this as a transaction: https://codemirror.net/docs/ref/#state.Transaction

but i can come up with a better name for it,  mappedChanges or something like that

client/components/codeEditor/extensions/customKeyMaps.js
5e-Cleric: We are not adding '  ' line by line, this function either calls indentmore to add indentation to the lines, or swaps the selection for a tab, in case you want to select characters and insert a tab in their place.

About indentLess, i think it was left over from v5, i seem to remember removing it, but with so many PRs, its possible i left it from a conflict. removing it now.

Issue #4812, opened 2026-05-13
Support RTL layouts


AriaMoradi:
### Your idea:

I tried the website with some Persian text and it doesn't show up right.

The layout stay left to right.

ericscheid:
Good catch.

1. we do have a Brew Property setting for specifying the `lang` of the brew .. but
  a. it appears to only accept limited values (which looks like a bug to be investigated.)
  b. `lang` doesn't determine text direction (`dir` does)
  c. we don't have a Brew Property setting for direction. Yet.

2. You can still get RTL text though — just enclose every page content with `{{dir=RTL,lang=fa .... }}` (the opening code on one line by itself, followed by your page content, and the closing braces on a final trailing line).

3. You can override an enclosing {{curly-block}} setting with a {curly-inject} on a line following the text you want to override. This is simpler than wrapping e.g. a single line of brew text with another {{curly-block}}.

4. I'm not even sure the Brew Property for language gets applied to the final rendered content — it only seems to appear in the open-graph meta-data, which is unhelpful. Investigating.


<img width="1417" height="435" alt="code for RTL text" src="https://github.com/user-attachments/assets/1ef53979-b75b-4dbf-b344-1ecc125114bc" />

ericscheid:
## Proposal: 

1. Update Brew Properties panel for a direction setting (RTL and LTR, but full english not acronyms).

2. Apply `lang` and `dir` to `<html>` of the preview iframe and the /share preview iframe. Not to the top-level html document (try it for fun!)

3. Do as two separate PRs — one for applying `lang`, one for adding `dir` feature. Small PRs are good PRs.

firstmatekidd:
I've got the beginning of this working locally....

dbolack-ab:
This should be added the same as the V4 document properties and should be pushed into the CSS stack via the same mechanism as those new properties once that starts to happen.

PR #4874, opened 2026-06-18
Accessibility: Add title to the brew render iframe.

firstmatekidd:
## Description

This adds lang and title tags for accessibility to the brew renderer iframe. The `title` is added as an attribute to the `iframe` but also as a `title` element in the iframe's head since some screen readers like voice over use that instead of the iframe's title. This change means that the iframe shows up in the rotor menu as something more than just "frame"

Bonus, this includes some changes made by the linter to this file.

- Closes # 
  - Currently no issue logged that I'm aware of. I haven't seen a lot of accessibility issues in the list.

No visual change. On mac, you can turn on voice over, open the rotor (shift+cmd+u) and use the left/right arrow keys until you get a list of frames on the page, at which point you will see the change.

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -28,7 +28,8 @@ const PAGE_HEIGHT = 1056;
 const TOOLBAR_STATE_KEY = 'HB_renderer_toolbarState';
 
 const INITIAL_CONTENT = dedent`
-	<!DOCTYPE html><html><head>
+	<!DOCTYPE html><html lang="en"><head>
+	<title>Rendered Brew Content</title>
 	<link href='/homebrew/bundle.css' type="text/css" rel='stylesheet' />
 	<link href="${brewRendererStylesUrl}" rel="stylesheet" />
 	<link href="${headerNavStylesUrl}" rel="stylesheet" />
@@ -91,7 +92,7 @@ const BrewPage = (props)=>{
 
 //v=====--------------------< Brew Renderer Component >-------------------=====v//
 let renderedPages = [];
-let pageTemplates = [];
+const pageTemplates = [];
 let rawPages      = [];
 
 const BrewRenderer = (props)=>{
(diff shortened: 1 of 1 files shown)


Gazook89:
Check out these two related issues: 
#3205 & #4812

Note that one of them (3205) has a little sub issue. 

I'm not sure what, if anything, has already been resolved on those issues since they were last updated. But I would start there, and see if you can find the minimal code needed to close up those two if there is overlap. 

I only took a second look at those because I noticed that your PR didn't set a dynamic lang that matched the brew lang prop. It only set it to "en". Right now I think the lang is set on a child element of the renderer, `.pages`. I think that's where it is applied, rather than the iframe, because it is the closest element to the brew that isn't part of the brew itself.

firstmatekidd:
Those are related to languages, but not the specific change here. I'd be happy to maybe look at them after this one.
- [#3205](https://github.com/naturalcrit/homebrewery/issues/3205) is an open graph thing, the subissue is for the user brew cards and looks like they should definitely be fixed.
- [#4812](https://github.com/naturalcrit/homebrewery/issues/4812) has to do with LTR vs. RTL

I didn't realize there was a way to set the language per-brew. This PR should definitely be modified to use the dynamic language. I'll post an update.

firstmatekidd:
I removed the lang tag from the iframe's `html`. Accessibility-wise, as a general rule, an `html` tag should have a lang tag, and in this case it would help specify that the language of the `title` is english,  but it shouldn't hurt anything not to have it for now and could be sorted in a future PR.  The addition of the `title` is the main point of this PR anyways.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -91,7 +92,7 @@ const BrewPage = (props)=>{
 
 //v=====--------------------< Brew Renderer Component >-------------------=====v//
 let renderedPages = [];
-let pageTemplates = [];
+const pageTemplates = [];
```
Gazook89: I would leave this untouched, as I believe it is scaffolding for future changes.  I'm sure this is part of the linting pass you did, so no harm no foul, I just think it's best to let it lie.

Gazook89 (approved):
Simple enough change, looks good to me.  There was one lint change that I would revert, though.  Commented on it.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: agreed

firstmatekidd: I can revert this change.

Although as a new contributor, it seems like some extra friction that the project has a linting command, but that reviewers are suggesting not to use the linting results.

5e-Cleric: You are very right on that, lots to be said about it, we should also change formatter soon, as eslint has deprecated its formatting rules...

5e-Cleric (approved):
Simple change, sorry this took so long

PR #4884, opened 2026-06-26
V4 metadata UI Additions

dbolack-ab:
## Description

Prospective UI for the new Brew Metadata fields in the parent PR.


Gazook89:
I know it's still draft and early thoughts, but hoping you've pushed to Github in hopes of feedback:

- I think you should keep "Publish" and "Delete Brew" together.  Probably below the new section you've added.
- "Document Specifications" is less descriptive of the new utility than "Print Specifications".  
- It's a lot of noisy and technical inputs, which I suspect a majority of users don't have much use for (I assume most users do not ever intend to physically print).  I think two things might help:
  - Add a select menu for the Paper Size section that has standard options (Letter, A4, A5, etc) that onChange sets the Height and Width.  If Height/Width is changed directly, revert the standard option select to "Custom".  
  - Put Safety and Bleed inside a `details` element, with a `summary` element "Advanced..." or "More..." or similar.

PR #4890, opened 2026-07-02
Add Style property to Config

Gazook89:
Currently, all deployments and local instances of the app have some sort of altered background in the Preview pane, with the `darkred` color and a repeating label.  It does this by adding a `deployment` class to the top-level `homebrew` class, and then styling based on that in `homebrew.less`.

 This drives me a little crazy because I'm interested in having the correct colors and appearance when working on UI stuff.  I could comment out the CSS, but then I have to worry about that pushing that up into the repo.

If I've missed how to disable this, please let me know. 

But in this PR, I've changed the config to accept a `style` property in the config, which sets to an object.  From my `config/local.json`:

```json
{
	"secret" : "secret",
	"web_port" : 8000,
	"enable_v3" : true,
	"enable_themes" : true,
	"hb_images" : null,
	"hb_fonts" : null,
	"mongodb_uri" : "mongodb://127.0.0.1/heroku_cjpfxl1z",
	"style" : {
		"backgroundColor" : null,
		"textLabel"       : false
	}
}
```
I can now set the background color of the Preview here, or leave it null to just take whatever the live site has.  And I can similarly toggle the automated text label ("local" or the PR deployment number).  And this is set in the untracked `local.json`.

Again, if I've missed this already being possible, let me know.  Sometimes it's not entirely clear what all is possible in the config files.


client/homebrew/homebrew.jsx
```diff
@@ -41,16 +41,20 @@ const Homebrew = (props)=>{
 		enablev4
 	} = props;
 
+	console.log(config.color);
```
dbolack-ab: I assume this needs dropped before merge.

dbolack-ab (approved):

client/homebrew/homebrew.jsx
Gazook89: ah yes, thanks.

5e-Cleric (approved):
This should work, will need a deployment to be sure of course, once calc opens it, i can make the appropriate config keys in heroku and we can test.

@calculuschild would you mind deploying this?

dbolack-ab:
Can we set some values in the config.json for the deployment?  Right now it looks like Prod

5e-Cleric:
image
config set

5e-Cleric:
doesnt' seem to change a thing
can we log the config object, or at least the style property if found?.

Mind you this is the env vars i am changing so we may be looking in the wrong place for the style object

Gazook89:
> can we log the config object

I added a `console.table` to log it.  For myself, I change my `/config/local.json` like this:

```
{
	"secret" : "secret",
	"web_port" : 8000,
	"enable_v3" : true,
	"enable_themes" : true,
	"hb_images" : null,
	"hb_fonts" : null,
	"enablev4" : true,
	"mongodb_uri" : "mongodb://127.0.0.1/heroku_cjpfxl1z",
	"style" : {
		"backgroundColor" : "red",
		"textLabel"       : "yolo"
	}
}
```

image

I changed a couple of other things, as well, so try it again now.

client/homebrew/homebrew.jsx
```diff
@@ -46,12 +46,16 @@ const Homebrew = (props)=>{
...
-			const bgText = config?.deployment || 'Local';
-			return {
-				backgroundImage : `url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='100px' width='200px'><text x='0' y='15' fill='%23fff7' font-size='20'>${bgText}</text></svg>")`
+		if(config?.deployment || config?.local) {
+			const bgText = (config?.deployment || config?.style.textLabel) || 'Local';
+			const styleObject = {
+				backgroundImage : `url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' version='1.1' height='100px' width='200px'><text x='0' y='15' fill='%23fff7' font-size='20'>${bgText}</text></svg>")`,
+				backgroundColor : config?.style.backgroundColor ? `${config?.style.backgroundColor}` : null
 			};
+			console.table(styleObject);
```
5e-Cleric:
```suggestion
			console.table(config);
```

calculuschild:
I know this has been out for a while.

If the goal is just a way to change the colors back to default when in a non production environment, I would think the easiest way is just add a boolean config property called "useProductionStyle" or something, and check it at the step where we already check the environment and set the style.

I may be misunderstanding how this works, but it seems like it's built to allow inserting custom titles and backgrounds. Putting style into the config file in my opinion is an easy way for a lot of confusion down the road if we are splintering off secret bits of CSS properties into random places.

Would this work if you just made it a single boolean instead?

5e-Cleric:
> Putting style into the config file in my opinion is an easy way for a lot of confusion down the road if we are splintering off secret bits of CSS properties into random places.
> 
> Would this work if you just made it a boolean instead of requesting background and text colors?

Considering how long we developed without the color, i think its fine if it is just us having a style property in our computers?

Don't care much either way as long as i can still have it.

5e-Cleric:
how is that not working?

image

Gazook89:
a boolean that sets it to production values is fine with me-- I just very specifically don't want my local environment to look different than production.  If I do, I use a browser extension that adds a colored frame around any window with `local` in the address bar and is dismissible.

I went this approach because I figured people might want to choose their own color for some reason, without touching css files.

calculuschild:
Ok. Choosing your own color via config feels like overkill to me, so unless there is an objection I would prefer just a boolean. Avoid splintering styling definitions into into unexpected places.

If we want to actually change the color for aesthetic reasons, that feels like something that should be either a GUI option like the editor themes for users, or you just edit the CSS itself if you are a Dev and want to play around with things.

dbolack-ab:
> Ok. Choosing your own color via config feels like overkill to me, so unless there is an objection I would prefer just a boolean. Avoid splintering where the styling is defined.

I agree with this in principle. Would this boolean disable the use of the deployment class or ?

calculuschild:
Yeah, just turn off the special dev/local styling. When we make the check to turn on the dev styling, also check for this boolean from the config.

5e-Cleric:
Ehm, @calculuschild and @dbolack-ab we already have a boolean for this, it is called "development" in config, the code requires that to exist in order to add the red background.


Might be malfunctioning, can't recall.

calculuschild:
> ...we already have a boolean for this, it is called "development" in config, the code requires that to exist in order to add the red background.

If that's the case, then we can just close this. Can anyone verify that flag works as expected? I can test later but I am in the process of rebuilding my local machine so I don't have an active dev environment.

Gazook89:
I can't verify right now either. But does this property only change that background (the text and the color), or does it have additional effects? To me, "development" alone doesn't make me think it's just changing the style of the preview background.

Gazook89:
I'll raise my question in #4996, and close this.

PR #4891, opened 2026-07-04
Normalize config files for containers to Node 26.4

dbolack-ab:
In looking at a few things I noticed that Circleci was set to use 20.17 in one place, 20.18 in another and our Dockerfile was 22.

This normalizes them all to the same value. This should be compared, IMO, to our Dynos and we should adjust as needed to match.


dbolack-ab:
@calculuschild Can we get a tester here?

calculuschild:
@dbolack-ab  got it

calculuschild:
Looks like deployment works, CircleCI passes. I think this is probably OK to merge.

PR #4897, opened 2026-07-09
More Accessible Dropdowns

Gazook89:
This is a small extension of work to the dropdown menus that shoots to improve structure and styling a bit.

It...
- sets dropdown menus as unordered lists of items
- appropriate roles,
- adds aria-labels,
- adds styles for active menu triggers,
- removes some redundant styling

-----
This isn't the "end run" for accessibility, but I think it moves closer.  Ultimately, to be a true Menubar, it will need a Menubar component that holds these menus and tools and manages focus much more rigorously.  It is supposed to use `tab` to get *into* the Menubar, and then arrow keys to move around, and `tab` again to get out. And, if a menu is open, then other menus can be opened with just hover.  But that will have to come as part of a broader Menubar component.


calculuschild (approved):
I am OK with this. I don't have as good of an eye for UI stuff so if anyone else wants to comment.

5e-Cleric:
> I am OK with this. I don't have as good of an eye for UI stuff so if anyone else wants to comment.

I'd be happy to comment on the UI if i could see it. could we get a depoyment?

5e-Cleric:
If we can get proper colors, i'd appreciate it:
image

client/homebrew/editor/snippetbar/snippetbar.less
```diff
@@ -234,13 +207,6 @@
...
 	}
-	.disabledSnippets {
-		color: grey;
-		cursor: not-allowed;
-
-		&:hover { background-color: #DDDDDD;}
-	}
-
 }
 @container editor (width < 750px) {
```
5e-Cleric: This should fix wrapping for now

```suggestion
@container editor (width < 818px) {
```

5e-Cleric (approved):
Made two comments, otherwise approved

Gazook89:
should be all set then with latest commit.

client/components/dropdown/dropdown.less
```diff
@@ -1,24 +1,32 @@
...
+	syntax: '<color>';
+	inherits: true;
+	initial-value: #DDD;
 }
 
+@property --activeTriggerColor {
+	syntax: '<color>';
+	inherits: true;
+}
+
```
5e-Cleric, line 12: An initial value is required otherwise this rule is invalid.

> The [initial-value](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@property/initial-value) descriptor is optional if the value of the syntax descriptor is the universal syntax definition (syntax: "*"). If the initial-value descriptor is required but omitted, the entire @property rule is invalid and ignored.

from mdn

Gazook89: addressed

PR #4899, opened 2026-07-11
Add `@components` aliases to vite config

Gazook89:
This adds `@components` as an alias for `./client/components` in the vite config.  It also updates all references to that directory to use that alias.  So 1 real change, and then 17 additional files with a single line change on each.

Re-ran build, it works fine.


calculuschild:
Looks good to me. Merging!

PR #4902, opened 2026-07-14
Refactor nav.jsx into Functional Component

Gazook89:
This is a fairly straightforward refactor from class component to functional component for the `nav.jsx` component.  At the same time, it does remove a little cruft, like no more `_.omit()` (and no more Lodash in this file).  

Otherwise everything else stays the same.


5e-Cleric (approved):
Looks alright to me, if it is ready for merge i can do that.

calculuschild:
@Gazook89  Can you fix the conflicts? Looks like a recent merge made some DIVS into `<button>` elements in the same area.

Gazook89:
> @Gazook89 Can you fix the conflicts? Looks like a recent merge made some DIVS into `<button>` elements in the same area.

This is from #4932, which I think we want to preserve.  So I've resolved the conflicts but kept them as buttons.

PR #4905, opened 2026-07-15
Fix safeHTML attribute check

### LINKED ISSUE #4904: Some `on` attributes are not removed
### Renderer

v3

### Browser

Chrome

### Operating System

Windows

### What happened?

Some attributes are not being properly sanitized - see the example element.

Investigation has shown that this is due to `for (const attribute of element.attributes) { ... }` (`safeHTML.js:L35`) skipping some attributes (e.g. `onclick`); although the exact reason for this has not yet been determined.

### Code

```gfm
<div
  id="myDiv"
  style="width:100%;height:50px;border:2px solid magenta;"
  onmouseenter="1"
  onclick="2"
  oninput="3"
  >
  TEST
</div>
```

G-Ambatte:
## Description

This PR changes how the element attributes are parsed as an iterable list, forcing a conversion from a `NamedNodeMap` to an `Array`.

- Closes #4904

Check the element produced by this piece of code.
- In the current master, the `onclick` attribute persists.
- In this PR, all three `on` attributes are removed.

```HTML
<div
  id="myDiv"
  style="width:100%;height:50px;border:2px solid magenta;"
  onmouseenter="1"
  onclick="2"
  oninput="3"
  >
  TEST
</div>
```

### DIFF
--- a/client/homebrew/brewRenderer/safeHTML.js
+++ b/client/homebrew/brewRenderer/safeHTML.js
@@ -32,12 +32,12 @@ function safeHTML(htmlString) {
 			return;
 		}
 		// Check remaining elements for blacklisted attributes
-		for (const attribute of element.attributes){
+		[...element.attributes].forEach((attribute)=>{
 			if(blacklistAttrs.some((test)=>{return test(attribute);})) {
-				element.removeAttribute(attribute.localName);
-				break;
+				element.removeAttribute(attribute.name);
+				return;
 			};
-		};
+		});
 	});
 
 	return div.innerHTML;
--- a/tests/html/safeHTML.test.js
+++ b/tests/html/safeHTML.test.js
@@ -43,6 +43,12 @@ test('Javascript via inline event handler - onMouseOver', function() {
 	expect(rendered).toBe('<div>Hover over me</div>');
 });
 
+test('Javascript via multiple inline event handlers - onClick + onMouseOver', function() {
+	const source = `<div onclick="alert('This is a JavaScript injection via inline event handler')" onmouseover="alert('This is a JavaScript injection via inline event handler')">Hover over or Click me</div>`;
+	const rendered = safeHTML(source);
+	expect(rendered).toBe('<div>Hover over or Click me</div>');
+});
+
 test('Javascript via data attribute', function() {
 	const source = `<div data-code="javascript:alert('This is a JavaScript injection via data attribute')">Test</div>`;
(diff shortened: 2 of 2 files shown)


Gazook89:
When I paste your example code into the live site, I can see the attributes persist.  But nothing can be done with them?  If i change the onclick to `onclick="(e)=>{console.log(e)}"` the element actually disappears.  I haven't been able to get it to perform any function at all.  Am I just testing it wrong?

Overall I don't see anything wrong with this PR, I just can't repeat the issue.

dbolack-ab:
> When I paste your example code into the live site, I can see the attributes persist. But nothing can be done with them? If i change the onclick to `onclick="(e)=>{console.log(e)}"` the element actually disappears. I haven't been able to get it to perform any function at all. Am I just testing it wrong?
> 
> Overall I don't see anything wrong with this PR, I just can't repeat the issue.

I assume you are making that change in the rendered page? I don't think safeHTML is called on update, but rather as an inline of the markdown conversion. @G-Ambatte  ?

G-Ambatte:
> I assume you are making that change in the rendered page? I don't think safeHTML is called on update, but rather as an inline of the markdown conversion. @G-Ambatte ?

@dbolack-ab @Gazook89 

My apologies if I wasn't clear; `safeHTML.js` modifies the rendered output so that the attributes are removed from the element in the BrewRenderer HTML output, so this can be checked and confirmed using Developer Tools of the rendered output. This cleaning function is called from `BrewRenderer` and occurs _every_ time a BrewPage is rendered (including all Style and Theme Style data) before the Markdown rendered output is ever applied to the client browser.

### LIVE
Attributes are not sanitized properly, and arbitrary code execution (ACE) may be achieved.
image
image

### PR
The same code is correctly cleaned; ACE is prevented.
image
image

image
image

G-Ambatte:
> onclick="(e)=>{console.log(e)}"

@Gazook89 Also, this wouldn't work because I don't believe you can pass a function in the attribute property in this manner. You can access some properties with `this`, e.g. `onclick="const test = (e)=>{console.log(e);}; test(this);"`.

Gazook89 (approved):
Looks good to me.

5e-Cleric:
I am not approving this without testing it on a deployment, @calculuschild please

5e-Cleric (approved):
Can confirm this works in removing events in raw html elements. Approved.

PR #4908, opened 2026-07-18
Add a sandbox property to the renderer iframe

ReinaSweet:
Prevents any script execution from within the the brew renderer.

Still allow for print (via allow-modals), same-origin downloads, and top navigation (so that links work).

### DIFF
--- a/client/homebrew/brewRenderer/brewRenderer.jsx
+++ b/client/homebrew/brewRenderer/brewRenderer.jsx
@@ -352,6 +352,7 @@ const BrewRenderer = (props)=>{
 				style={{ width: '100%', height: '100%', visibility: state.visibility }}
 				contentDidMount={frameDidMount}
 				onClick={()=>{emitClick();}}
+				sandbox="allow-same-origin allow-modals allow-top-navigation"
 			>
 				<div className='brewRenderer'
 					onKeyDown={handleControlKeys}


G-Ambatte (approved):
Local testing of this PR shows that functions in the event listeners of elements do not appear to run:
- `alert` does not cause a modal to pop
- `console.log` does not cause a message in the console
- `fetch` does not cause network activity

No other functionality (images, printing, etc) appears to be affected.

Gazook89 (approved):
Why doesn't `alert()` work?  It seems this PR has `allow-modals`, which includes `alert`, so I don't understand that.

But otherwise yes, it all seems to work.  I'm no l33t hacker, so I'm ready to be wrong, but looks good to me.  I think this can live alongside @G-Ambatte 's recent PR #4905 even though this renders it moot?

ReinaSweet:
alert() won't work because the script itself can't run (which would required "allow-scripts").
The closest to an alert that the user can get without allow-scripts would be to use invoker commands.

G-Ambatte:
> alert() won't work because the script itself can't run (which would required "allow-scripts"). The closest to an alert that the user can get without allow-scripts would be to use invoker commands.

I wonder if `safeHTML` should also strip any `command` attributes... Given the current functionality of the feature, it seems unnecessary - but it may be easier to relax restrictions, rather than have to attempt to apply them later.

5e-Cleric:
Has been over a week but build and tests seem to be being run for me, wth?

@calculuschild could use a deployment

G-Ambatte:
> I think this can live alongside @G-Ambatte 's recent PR #4905 even though this renders it moot?

#4905 will sanitize the HTML, even if we render the content outside of an iframe (not that I'm recommending such a thing), and should also provide an additional hurdle to any future iframe sandbox escapes that may be discovered.

Belts and braces, basically.

G-Ambatte:
> Has been over a week but build and tests seem to be being run for me, wth?

@5e-Cleric : I updated to the latest master, which caused the tests to run and complete correctly (see [here](https://app.circleci.com/pipelines/github/ReinaSweet/homebrewery/13/workflows/5fc0b704-4448-4d91-aa8b-a9fca260ff87)). Given that there are two approving reviews and all tests passed, I have merged the PR to master.

calculuschild:
> @calculuschild could use a deployment

@5e-Cleric sorry I'm out of town without my laptop for another week. I'll see if I can do this on my phone later when I have a moment

5e-Cleric:
> > @calculuschild could use a deployment
> 
> @5e-Cleric sorry I'm out of town without my laptop for another week. I'll see if I can do this on my phone later when I have a moment

no need anymore, g-ambatte has gone ahead and merged

PR #4919, opened 2026-07-26
Fix file transfer by non-owners

### LINKED ISSUE #4862: Non-owners able to transfer to Google Drive
It is possible for the document owner to lose editor access to their own documents if a co-author transfers the file into their Google Drive.

1) Owner A creates a file and saves it.
2) Owner A invites Coauthor B to the brew.
3) B makes an edit, saves, and becomes a full Author on the brew.
4) B transfers the file into their own Google Drive:
  - If it was in A's Google Drive, B can still transfer it into Homebrewery storage
  - From Homebrewery storage, B then transfers into their own Google Drive
6) All co-authors and guests can still see the file, but Owner A can no longer open the Edit or Share page
  - If Owner A does not have Google Drive, they get a 401 "Not logged in", as we attempt to use Oauth credentials that don't exist
  - If Owner A does have Google Drive, they get a 404 "File not found", as they do have OAuth credentials but not to B's drive
  - Guests and other coauthors can open the file, because the Service account is used if you are not the owner

5e-Cleric:
Fixes #4862


client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -71,6 +71,7 @@ const EditPage = (props)=>{
 	const [alertTrashedGoogleBrew, setAlertTrashedGoogleBrew] = useState(props.brew.trashed);
 	const [alertLoginToTransfer, setAlertLoginToTransfer] = useState(false);
 	const [confirmGoogleTransfer, setConfirmGoogleTransfer] = useState(false);
+	const [notOwnerToTransfer, setNotOwnerToTransfer] = useState(false);
```
Gazook89: Since this is an alert like the other alerts, it makes sense to name it with the same convention: `alertPermissionToTransfer`, or `alertOwnershipToTransfer` or anything `alert...`.  And then move this line up to line 73 so it's with the other alerts.

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -271,9 +276,19 @@ const EditPage = (props)=>{
 				</div>
 			)}
 
+			{notOwnerToTransfer && (
+				<div className='errorContainer' onClick={closeAlerts}>
+					You must be the Owner to transfer between the Homebrewery and Google Drive!
+					<a target='_blank' rel='noopener noreferrer' href={`https://www.naturalcrit.com/login?redirect=${window.location.href}`}>
+						<div className='confirm'> Sign In </div>
```
Gazook89: I'm not sure it makes sense to have the "Sign In"/"Not Now" set of choices here.  Presumably they are already signed in, or they would hit the `alertLoginToTransfer` error.  Prompting to sign in isn't going to resolve the issue.  

If there was a method of contacting the owner of the brew, we could offer that.  But lacking that, we could include the Owner's username in the notice and then just a simple "Okay" button.

Gazook89 (changes requested):

client/homebrew/pages/editPage/editPage.jsx
5e-Cleric: Good point.

client/homebrew/pages/editPage/editPage.jsx
5e-Cleric: Good idea

dbolack-ab (approved):

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -162,6 +163,9 @@ const EditPage = (props)=>{
 	};
 
 	const handleGoogleClick = ()=>{
+		if(global.account !== brew.authors[0]) {
+			setalertOwnershipToTransfer(true);
```
G-Ambatte, line 167: I suspect we want a `return` here, or else the rest of the function will run?

client/homebrew/pages/editPage/editPage.jsx
```diff
@@ -271,9 +276,19 @@ const EditPage = (props)=>{
 				</div>
 			)}
 
+			{alertOwnershipToTransfer && (
+				<div className='errorContainer' onClick={closeAlerts}>
+					You must be the Owner to transfer between the Homebrewery and Google Drive!
+					The owner of this file is ${brew.authors[0]}
+					<a target='_blank' rel='noopener noreferrer' href={`https://www.naturalcrit.com/login?redirect=${window.location.href}`}>
```
G-Ambatte: Do we want a link to the login system here? From the content, it doesn't seem obvious that it would direct the user to login.

G-Ambatte (changes requested):
Good changes, should prevent the file from being transferred at the UI level.
We could try to prevent it in the back end as well with a similar check.

client/homebrew/pages/editPage/editPage.jsx
calculuschild: Yes. Otherwise co-authors without google accounts will also get the popup telling them to log in. We only want to show one message I think.

client/homebrew/pages/editPage/editPage.jsx
calculuschild: I agree. Echoing @Gazook89 's earlier comment which seems to have been hidden with a recent comment, I don't think we want to prompt a log in here.

client/homebrew/pages/editPage/editPage.jsx
calculuschild: Sign-in prompt is still here, just relabeled "Okay" now. I agree with @Gazook89, a sign-in prompt isn't going to resolve the issue. Either you are a co-author so you are not going to be able to log in as the owner anyway, so you want to just close the popup, or you are the author and are the owner and are already signed in, and will hit the "not logged in to google" popup.

The Okay button should just close the prompt.

calculuschild (approved):
Approve once the comments from @G-Ambatte and @Gazook89  are resolved.

client/homebrew/pages/editPage/editPage.jsx
5e-Cleric: Good call

client/homebrew/pages/editPage/editPage.jsx
5e-Cleric: Another good call, not sure what i was thinking back then.

5e-Cleric:
Okay, flow is now:

- If not owner, say "you're not owner, X is owner" with Okay to close popup
- If owner, but not logged into google, say so, and maybe prompt them to go to the login page to link the google account?
- If Owner and brew is in google drive's trash, confirm to move the brew to storage
- If owner, present confirmation, and transfer on yes

sound good?

5e-Cleric:
Alright, merging!

PR #4920, opened 2026-07-26
Fixes editor theme saved independent from tab

### LINKED ISSUE #4858: Editor Theme is saving independently for both Brew and Style editors
### Renderer

v3

### Browser

Chrome

### Operating System

MacOS

### What happened?

Since the editor update, setting an editor theme is setting the localStorage property independently for each the Brew and Style editors but on the same key.  So if I set my Brew editor to "neat", and the Style editor to "twilight", i can have different themes for each.  But when i reload the page, is pulling "twilight" for both editors.

I believe that we should not allow different themes per editor (and I don't believe we have intended to allow this).  But if for some reason we do want to allow it, we need to save the themes separately in localStorage.

### Code

```gfm

```

5e-Cleric:
Fixes #4858

### DIFF
--- a/client/components/codeEditor/codeEditor.jsx
+++ b/client/components/codeEditor/codeEditor.jsx
@@ -308,7 +308,7 @@ const CodeEditor = forwardRef(
 			view.dispatch({
 				effects : themeCompartment.reconfigure(themeExtension),
 			});
-		}, [editorTheme]);
+		}, [editorTheme, tab]);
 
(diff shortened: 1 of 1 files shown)


Gazook89 (approved):
Looks good to me.  Tested it on local, confirmed it loads the latest theme on each tab whenever it is changed, and it loads the latest theme in all three tabs after a page reload.

Issue #3032, opened 2023-09-05
Make the Homebrewery meet Web Content Accessibility Guidelines(WCAG)


5e-Cleric:
As the title says, i think it would be epic if we made the homebrewery more accessible to people with disabilities, the WCAG international standard provides the necessary information on how to do so.

As per 2023, the most recent stable version is WCAG 2, so i suggest that standard is the one used for this issue.

G-Ambatte:
For future reference - WCAG 2 guidelines: https://www.w3.org/TR/WCAG20/

5e-Cleric:
Adding here some issues that should be solved, related to this:

Editor:
- [ ] Add pointer cursor icon to the entire foldgutter element
- [ ] Add highlight to the foldgutter element
- [ ] Add `not-allowed` cursor icon to snippetbar functions if they are not available (brew history stuff)
- [ ] Remove draggability from thumbnail image preview
- [ ] Normalize text input look (border-radius, padding) in properties editor

Preview:
- [ ] Add `ew-resize` cursor to  toolbar zoom function
- [ ] Increase width of page number to at least 4 digits (4ch + padding)
- [ ] Add `not-allowed` cursor icon to disabled buttons

Navigation:
- [ ] Add `wait` cursor to saving element while saving
- [ ] Add hover effect to storage and new save button

Some of these can be done globally:

`*:disabled { cursor:not-allowed; }`

dbolack-ab:
Is this being broken out of the uI update pR?

PR #4936, opened 2026-08-01
add aria labels to toolbar elements

firstmatekidd:
## Description

This improves screen reader accessibility of the toolbar elements by providing them with aria labels and hiding the icons from the screen reader. A side effect of labeling the buttons is it will also improve voice control.

Specific Changes:
1. The tooltip is (mostly) copied directly to the aria-label. Some were changed (like the page spread) to make the individual button functions more understandable.
2. The icons for each button are hidden from the screen reader. Nesting these within the button was leading to an additional "group" word being read out for each button which could be kind of confusing since there's no actual content inside the button to navigate to.

[#3032 Make the Homebrewery meet Web Content Accessibility Guidelines(WCAG)
](https://github.com/naturalcrit/homebrewery/issues/3032)

Closes # None. This PR gets closer to WCAG accessibility getting more buttons in homebrewery to conform to [WCAG 4.1.2](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html)

Enable screen reade, navigate to the toolbar and listen to what is read by VoiceOver. Previously, many of the elements just read "button" but now they read their function.

### Reviewer Checklist

  - [ ] Feature A does X
  - [ ] Feature B does Y


5e-Cleric (approved):
Looks good so far, if this is ready let me know.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -99,45 +99,54 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
...
 			<div className='group' role='group' aria-label='Zoom' aria-hidden={!toolsVisible}>
 				<button
 					id='fill-width'
 					className='tool'
 					data-tooltip-bottom='Set zoom to fill preview with one page'
+					aria-label='Set zoom to fill preview with one page'
 					onClick={()=>handleZoomButton(displayOptions.zoomLevel + calculateChange('fill'))}
 				>
-					<i className='fac fit-width' />
+					<i aria-hidden='true' className='fac fit-width' />
```
G-Ambatte, line 122: This PR creates a lot of repeated elements with the same general format: `<i aria-hidden='true' className='XYZ' />`.
It might be better to create a React functional component, something like:
```
const Icon = ({ ariaHidden=true, className, children, ...props })=>{
	return <i aria-hidden={ariaHidden} className={className} {...props}>
		{children}
	</i>;
};
```
then these lines could be simply `<Icon className='fac fit-width' />`, the `ariaHidden` property can be set to `false` to disable the new default functionality if/when required, and any further updates to these elements (i.e. for future WCAG-compliance updates) will be centralized to a single location.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -99,45 +99,54 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
...
 				</button>
 				<button
 					id='zoom-to-fit'
 					className='tool'
 					data-tooltip-bottom='Set zoom to fit entire page in preview'
+					aria-label='Set zoom to fit entire page in preview'
```
G-Ambatte, line 128: A better approach may be to create a `Button` component that - when the `aria-label` property is unset - automatically sets it to match the content of `data-tooltip-*`.

client/homebrew/brewRenderer/toolBar/toolBar.jsx
```diff
@@ -166,27 +176,32 @@ const ToolBar = ({ displayOptions, onDisplayOptionsChange, visiblePages, totalPa
...
 						aria-checked={displayOptions.spread === 'flow'}
-					><i className='fac flow-spread' /></button>
+					><i aria-hidden='true' className='fac flow-spread' /></button>
 
 				</div>
 				<Anchored>
-					<AnchoredTrigger id='spread-settings' className='tool' data-tooltip-bottom='Spread options'><i className='fas fa-gear' /></AnchoredTrigger>
+					<AnchoredTrigger id='spread-settings' className='tool' aria-label='Spread options' data-tooltip-bottom='Spread options'>
+						<i aria-hidden='true' className='fas fa-gear' />
+					</AnchoredTrigger>
```
G-Ambatte, line 204: A better approach may be to update the AnchoredTrigger component such that - when the `aria-label` property is unset - automatically sets it to match the content of `data-tooltip-*`; thus ensuring that all future instances are also WCAG compliant.

G-Ambatte (commented):
Overall, these are good changes, and the comments that I have made are intended only as suggestions for discussion, rather than required changes for approval.

firstmatekidd:
Cool suggestions, but I think for now, @5e-Cleric, I'll leave as is for merging and take a note to come back to creating new classes in a future PR.

5e-Cleric:
Okay, nothing wrong with that, btw, i sent you an email to `durstann@yahoo.com`a few days ago, not sure if you got it.

PR #4945, opened 2026-08-08
Make metadata editor more accessible for screen readers and keyboard

firstmatekidd:
## Description

This changes makes the metadata editor more accessible. Mainly accomplished by fixing focus indicators and associating labels with input elements.

Here's a summary of the changes and reasons

- Setting icons to `aria-hidden` prevents the screen reader (at least VoiceOver on Mac) from reading out "group" as part of the element
- input elements are given `id`s and their labels are explicitly associated with them so the labels are read out by screen readers as part of the input element. This gives screen reader users more confidence in knowing which elements are for which purpose.
- `aria-label` added for the show/hide thumbnail button
- The underline is put back for the author links to make it clear that they are links that can be activated. An slight offset was added for the underline to keep legibility.
- `aria-label` is added for the delete tag button on each tag. 
  - There's still something funky going on with keyboard interaction with the tags that's making it hard to edit the tag in place (pressing enter while the text is highlighted deletes the tag while clicking allows in-place editing. Should be a problem for another PR).
- `outline` is reapplied to the `colorButton` class so that focus indicators show up for people navigating with keyboard.

[3032](https://github.com/naturalcrit/homebrewery/issues/3032)

Closes # None. This PR gets closer to WCAG accessibility getting more buttons in homebrewery to conform to [Keyboard accessibility - WCAG 4.1.2](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html) and [Labeling - WCAG 3.3.2](https://www.w3.org/WAI/WCAG22/Understanding/labels-or-instructions.html)

Use the TAB key to navigate by keyboard and use a screen reader (VoiceOver, JAWS, NVDA) to navigate the page.

### Reviewer Checklist

  - [ ] Feature A does X
  - [ ] Feature B does Y


5e-Cleric:
@calculuschild can we get a deployment here¿?

shared/naturalcrit/styles/core.less
```diff
@@ -23,7 +23,6 @@ html,body, #reactRoot {
 	text-transform   : uppercase;
 	text-decoration  : none;
 	cursor           : pointer;
-	outline          : none;
```
5e-Cleric, line 26: I am not sure how will this look, but we dropped the outline for visual reasons, and those have not changed, not sure we want it back.

5e-Cleric:
Have tested, it is a good addition. Merging, thanks @firstmatekidd !

And i ask for the third and final time, do you want to come to the dev chat?

firstmatekidd:
@5e-Cleric I'd be happy to join the dev chat.  Sorry if I missed the earlier offers. Is it the gitter chat linked in the readme?

5e-Cleric:
> @5e-Cleric I'd be happy to join the dev chat.  Sorry if I missed the earlier offers. Is it the gitter chat linked in the readme?

indeed

PR #4961, opened 2026-08-17
Fix broken-image getting into a failure loop on /edit/ pages

ReinaSweet:
This worked only on /new before, and not on /edit/ due to the lack of leading slash.

And because this request gets spammed due to it constantly failing, it causes a rather annoying CPU usage spike, especially if you're watching for network traffic in the console.


5e-Cleric:
Hey, if you can fix this so it actually works, we should be able to get it to the finish line pretty quick.

PR #4970, opened 2026-08-24
Editor settings tab

### LINKED ISSUE #1908: User settings

5e-Cleric:
Adds a new tab, with editor settings.

We can expand on this, or move it elsewhere, the fact that we can now add some individual configs makes me happy, now i can keep adding stuff to the editor and allow people to ignore it completely.

Relatively simple, a few checks and making sure the settings are reflected in all tabs.

In the future, i can split all settings between tabs, if necessary.

Current settings:

- Editor theme (moved from snippetbar)
- Auto Close Brackets
- Show Image previews on hover
- Background Shading of active line
- Show Line Numbers
- Font Size


ericscheid:
Is this attached to an Issue# ?

Are these settings per brew, or per author? 
Or per browser via `localStorage()` ?

Keen to see what this actually looks like =)

Gazook89:
Everything that changes the UI should have a screenshot:

image

<br /><br />

image

client/homebrew/editor/settingsEditor/settingsEditor.less
Gazook89, line 1: This all feels unnecessary.  The Properties Editor is a key/value pair list, and so is the Settings tab, so just use the same CSS file (rename it to "tab-panel.less" or something if you want) and make tweaks if needed.  

If the other PR with the new styling for the Properties editor goes through, this should also match.

Gazook89 (changes requested):
- Move the editor theme to this.  It's an editor setting, and the palette icon it has now is easily confused as a Style tab.
- add an option to disable image link previews
- the same tabs should be available from any editor, as you noted might be the case.  It's jarring to switch to the properties editor and have a tab disappear, and the icons all jump around as a result.

client/homebrew/editor/settingsEditor/settingsEditor.less
5e-Cleric: > the same tabs should be available from any editor, as you noted might be the case. It's jarring to switch to the properties editor and have a tab disappear, and the icons all jump around as a result.

Not sure what you mean by this, all tabs are available at all times...
If you mean that the snippetbar buttons should not disappear, i agree, but it was decided like it is and it is out of scope of this PR.

5e-Cleric:
image

with the latest changes, and the css file removed.

client/homebrew/editor/snippetbar/snippetbar.jsx
Gazook89, line 232: This can be removed, if the editor theme selector is moved to the new settings tab.

client/homebrew/editor/settingsEditor/settingsEditor.jsx
```diff
@@ -0,0 +1,148 @@
...
+			<div className='field'>
+				<label htmlFor='fontSize'>
+					Editor Font Size
+				</label>
+
+				<div className='value'>
+					<small>from 9px to 30px</small>
+					<input
+						id='fontSize'
+						type='number'
```
Gazook89, line 135: Not a hill I would die on, but I believe this should be a range slider, and just options from smallest, small, default, large, largest.  

I don't think there is much reason to dictate specific pixel sizes.  The sizing should be based on `1em` as the default, to match whatever the user has as a browser default.  And then each step is .3em or something, so it's all relative to the browser default.  Otherwise, when we are trying to improve accessibility with this setting, we could actually be hampering it by restricting it to only a certain set of fixed absolute sizes. (between 9px and 30px).

client/homebrew/editor/settingsEditor/settingsEditor.jsx
```diff
@@ -0,0 +1,148 @@
...
+			...settings,
+			[setting] : e.target.type === 'number'
+				? Number(value)
+				: value,
+		};
+
+		updateSettings(updatedSettings);
+	};
+
+	return (
```
Gazook89, line 50: Could we include a small, ~10 line codemirror editor into the settings panel with some placeholder markup so that when selecting you a theme, you can see the theme in action without having to switch back to the brew editor?

This is the downside I didn't think of when i suggested moving it out of the snippet bar and into the panel.  However, I do still believe it needs to be in the settings tab, but just that a sample editor would be helpful. 

The other editor settings would benefit from this as well.

Gazook89 (commented):
More comments on it.  Not sure that'll be all, but it's what I can review right now.

client/homebrew/editor/settingsEditor/settingsEditor.jsx
Gazook89: The same result could be accomplished if the settings tab wasn't a tab at all, but a menu. A menu, opened from the snippet bar by a button would still allow someone to see the editor below it as they change these settings. Similar to the preview bar settings "anchored" component.

calculuschild (approved):
Code all looks good, and functionality seems ok without breaking anything.

I agree the UI tabs shifting around is a bit jarring, and it is slightly obnoxious to not see the editor theme changing without swapping between tabs. But that can be a separate PR to tweak to keep this one from growing too large, 

With that, I vote to merge as-is. It is stable enough that we can lock in these changes and if it were released live now, it is still fully functional. Just needs merge conflicts fixed.

5e-Cleric:
> Code all looks good, and functionality seems ok without breaking anything.
> 
> I agree the UI tabs shifting around is a bit jarring, and it is slightly obnoxious to not see the editor theme changing without swapping between tabs. But that can be a separate PR to tweak to keep this one from growing too large,
> 
> With that, I vote to merge as-is. It is stable enough that we can lock in these changes and if it were released live now, it is still fully functional. Just needs merge conflicts fixed.

Once it is merged, do you think this should be a codemirror panel, just like the search/replace is now?

client/homebrew/editor/settingsEditor/settingsEditor.jsx
5e-Cleric: good call, going from .5em to 2em, roughly 6.5px to 26px

client/homebrew/editor/settingsEditor/settingsEditor.jsx
5e-Cleric: Not doing the small codemirror, but i will go ahead with the panel as soon as i can.

5e-Cleric:
Changes made, nothing too big

calculuschild (approved):
The changes look good to me.

PR #4971, opened 2026-08-25
Migrate shared/markdown.js to an npm module

dbolack-ab:
## Description

Removes the shared/markdown.js file and replaces it with an npm module.

Passes tests. Didn't spot any issues.

## Notes

I initially named this hbfm ( Homebrewery Flavored Markdown ) and, as such, replaced the "Markdown" name for the import with hbfm. However, npm felt this, and the next couple of alternate names I chose were too close to existing module names.  I left the import and export names ***hbfm**.

If anyone has strong feelings about changing this for clarity or the like, I'm not hugely opposed.


dbolack-ab:
It appears that removing the Markdown file from the project blows up the coverage percentage as a whole.

5e-Cleric:
> It appears that removing the Markdown file from the project blows up the coverage percentage as a whole.

With a change as big and crucial as this, i think we're fine setting those to 40%, no, @calculuschild ?

calculuschild:
That's fine. What was the percentage before?

5e-Cleric:
> That's fine. What was the percentage before?

image

5e-Cleric (approved):
Looks alright to me, if we fix the test coverage i think we should be able to test and merge straight away.

5e-Cleric:
@dbolack-ab if you lower the required coverage in package.json, we can merge this and work on test coverage another day.

PR #4973, opened 2026-08-27
Change metadata ui theme

5e-Cleric:
Will have to tweak a lot if editor settings is merged but i don't care.

Offers a programatic theme that changes from dark to light depending on editor theme.

Also extended to the snippet bar.

## Images
Light and dark snippetbar

image

image

---

Light and dark metadata editor

image

image

Disclaimer: The background asset is mine, we can use it.


dbolack-ab (approved):

Gazook89:
Dark mode: I think you should try making the input fields dark as well, though distinguishable from the background. And then the text color something closer white/lighter. 

And the link text needs a different color besides be stock blue.

5e-Cleric:
image

client/homebrew/editor/metadataEditor/metadataEditor.less
Gazook89: This isn't a change you made, but somehow stands out more now that the background of the whole panel is white or black rather than the old neutral gray:  

On the theme dropdown, there is a white background box and then a thin black border, and then white background again.  It makes it look like the field has *focus* when it doesn't.

Live site:

Image

Your PR:

Image

I think that needs some reworking.  Additionally, on the same field, the placeholder text goes behind the caret arrow and looks untidy:

Image

Finally, that caret should be `currentColor` or similar; it needs to match the current text color.

Image

Gazook89 (commented):
Have you considered no borders on the fields, particularly on dark mode?  I feel like it looks a lot cleaner than the high contrast white borders.  Those white borders counteract the appeal of dark modes since the high contrast is very "loud". 

Your PR, dark mode:

Your PR, dark mode, no borders:

Image

Your PR, light mode, no borders:

Image
 


Or, alternatively, don't mess with the borders at all when toggling dark/light mode.    On both dark and light modes, they are lower contrast and less distracting from the actual text.

Your PR, dark mode, original gray borders:

Image

Your PR, light mode, original gray borders:

Image

client/homebrew/editor/metadataEditor/metadataEditor.less
5e-Cleric: image

I have no idea why it had that particular look, but i have never liked it and this is a good opportunity to make it look good

5e-Cleric:
All suggested fixes applied, transitioned the file to use css variables to be able to fix this without manually traversing all the lines.

Gazook89 (commented):
Something weird is happening here with the theme preview.  The preview isn't above all of the backgrounds in the theme menu:

Image

----

also, turning the snippet bar to the same gray as the top navbar in dark mode isn't ideal.  There is no separation at all, and it becomes very unclear where button boundaries are.

Image

----

I have to get to bed so haven't fully reviewed yet.

client/homebrew/editor/metadataEditor/metadataEditor.jsx
```diff
@@ -170,9 +168,9 @@ const MetadataEditor = createReactClass({
 		if(!this.props.metadata.editId) return;
 
 		return <div className='field delete'>
-			<label>delete</label>
+			<label htmlFor='delete-button'>delete</label>
```
G-Ambatte, line 171: I believe that there are no other instances of `htmlFor` in our existing codebase; instead, we have previously used implicit label declaration, whereas the use of `htmlFor` is an explicit declaration pattern. So this is a different construction of the HTML elements, compared to previous/elsewhere in the project.
That said, I do like the explicit declaration of the linked element.

---

After some research, it seems that the implicit style is not recommended for active components (like buttons) as it can interfere with accessibility tools.

G-Ambatte (approved):
Overall, looks to be a perfectly functional set of changes. Tests continue to pass when run locally.
As noted in a comment, there is a change in structure, but this both makes the link of label and input element more explicit AND conforms to accessibility best (or at least, better) practices.

5e-Cleric:
> Something weird is happening here with the theme preview. The preview isn't above all of the backgrounds in the theme menu:
> 
> Image

Not new, also not easily fixable, and out of the scope of this PR (i spent 1h trying to fix it again yesterday, with no results)

The problem arises from stacking contexts, because the preview is contained within the .item div, it cannot appear over subsequent elements with the same z-index as he .item element.

Anchor positioning could solve this glitch, but i do not like it, @Gazook89 might want to take a shot at it.


> also, turning the snippet bar to the same gray as the top navbar in dark mode isn't ideal. There is no separation at all, and it becomes very unclear where button boundaries are.
> 
> Image
> I have to get to bed so haven't fully reviewed yet.

Will change colors a bit, thanks.

calculuschild (approved):
This looks fine to me. Sounds like this will need rework after the editor settings tab is merged in #4970 anyway, but I approve of this change.

My only real comment is to try to compress the image assets for the menu background if they are still used after the rework.

PR #4981, opened 2026-09-02
Prefer position at the start of a tab insertion for the new cursor position

firstmatekidd:
## Description

Currently on `master` when you insert a single tab two `space` characters are inserted and then the cursor is moved two more spaces.

This seems to be because the changeset we get back from the state update was passing `1` to mapPos so it was preferring to find the cursor position at the end of the insertion. This worked for some types of tab presses but not others (see the test cases below). By passing a negative number it prefers to keep the position before the insertion.

[Code Mirror mapPos reference](https://codemirror.net/docs/ref/#state.ChangeDesc.mapPos)

I believe the error was first introduced in this change: https://github.com/naturalcrit/homebrewery/commit/d440c7c6514efd09c68e715ae548a3ba7f553184

- Closes #

Multiple insertion modes should be tested.

* Press `tab` while only having a single cursor and nothing highlight. Two spaces should be inserted and the cursor should be at the end of the two spaces.
* Highlight one character and press `tab`.  The character should be replaced by two spaces and the cursor ends up at the end.
* Highlight multiple characters or words and press `tab`. The selection should be replaced by two spaces and the cursor is at the end of the two new spaces.
* Do the above in any combination using CTRL/CMD + Click to place multiple cursors or create multiple selections.

### Reviewer Checklist

  - [ ] Feature A does X
  - [ ] Feature B does Y

### DIFF
--- a/client/components/codeEditor/extensions/customKeyMaps.js
+++ b/client/components/codeEditor/extensions/customKeyMaps.js
@@ -83,7 +83,7 @@ const insertTab = (view)=>{
 		changes,
 		selection : EditorSelection.create(
 			view.state.selection.ranges.map((range)=>EditorSelection.cursor(
-				mappedChanges.changes.mapPos(range.from, 1) + 2
+				mappedChanges.changes.mapPos(range.from, -1) + 2
 			)
 			)
 		)


5e-Cleric:
I'm sorry, what is the bug?

I don't see any inconsistency, but i may not be looking at it from the right perspective.

firstmatekidd:
Here's the behavior I see locally.  Every time I press tab, two spaces are inserted and then the cursor jumps two more positions past that.

Chrome (152.0.7977.65 (Official Build) (arm64)) on Mac (26.5.2).

wrong tab

Gazook89:
To be 100% clear, this is on *master* only, not on the live production site.  I can confirm on my local machine too.  @firstmatekidd didn't omit this fact, I'm just restating in case it was missed.

5e-Cleric:
> To be 100% clear, this is on _master_ only, not on the live production site. I can confirm on my local machine too. @firstmatekidd didn't omit this fact, I'm just restating in case it was missed.

It was absolutely missed, i must be going blind.

5e-Cleric:
Very weird, not sure how i broke that (and i am 90% sure it was me), thanks for realizing that and figuring out the fix. Merging!

PR #4982, opened 2026-09-02
Refactor of listpage.jsx into functional component

5e-Cleric:
As per the title, this pr refactors the listpage component into a functional version virtually identical to the previous class version.

I had found a big  (455ms) delay to rendering a good amount of brews, but after more testing, it appears this lag is caused by the development tools or other parts of the local environment, and not the actual code.


5e-Cleric:
Can confirm the lag only shows up locally.

PR #4983, opened 2026-09-02
Centralize API.getBrew descended calls to splitTextStyleAndMetadata.

### LINKED ISSUE #4975: splitTextStyleAndMetadata race condition
In PR #4708, a race condition was identified, wherein React could render the page before the `splitTextStyleAndMetadata` function had completed, resulting in the frontmatter (in codefenced blocks) being rendered in the text of the document.
Unfortunately, the PR's suggestion to make this function asynchronous does not appear to prevent the issue from occurring.

Potential fixes:
1. Ensure the split finishes first: run `splitTextStyleAndMetadata` as part of fetching the brew data in `getBrew`, rather than in middleware after the brew data has been returned in `req.brew`. Similarly, the static pages (welcome, legacy, migrate) could use static objects, rather than processing the text into in object on every request.
2. Ensure split finishes first: create a `brewLoad` event and cause the React render to wait until that event fires before beginning the rendering process.
3. Ensure the page is rendered first: run `splitTextStyleAndMetadata` as part of rendering the brew. This would have the intended effect of allowing metadata to be defined in the brew text, so is likely less desirable.

I suspect that the first option is the best solution, but please let me know if I've missed anything - alternative solutions, reasons why a different solution is superior, etc.

dbolack-ab:
## Description

This moves the calls to splitTextStyleAndMetadata inside of getBrew()  and removes those few calls that could potentially have failed a race condition.

This does not address "static" pages nor change tests that called splitTextStyleAndMetadata in their own fashion.

- Closes #4975

Test with large brews on undersized machines.

### Reviewer Checklist

  - [ ] Frontmatter is split as expected
  - [ ] Poorly sized and/or slow devices can correctly load large brews.
  - [ ] Tests still pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12085,21 +12085,6 @@
         "url": "https://opencollective.com/mongoose"
       }
     },
-    "node_modules/mongoose/node_modules/gcp-metadata": {
-      "version": "7.0.1",
-      "resolved": "https://registry.npmjs.org/gcp-metadata/-/gcp-metadata-7.0.1.tgz",
-      "integrity": "sha512-UcO3kefx6dCcZkgcTGgVOTFb7b1LlQ02hY1omMjjrrBzkajRMCFgYOjs7J71WqnuG1k2b+9ppGL7FsOfhZMQKQ==",
-      "extraneous": true,
-      "license": "Apache-2.0",
-      "dependencies": {
-        "gaxios": "^7.0.0",
-        "google-logging-utils": "^1.0.0",
-        "json-bigint": "^1.0.0"
-      },
-      "engines": {
-        "node": ">=18"
-      }
-    },
     "node_modules/mongoose/node_modules/mongodb": {
       "version": "7.5.0",
       "resolved": "https://registry.npmjs.org/mongodb/-/mongodb-7.5.0.tgz",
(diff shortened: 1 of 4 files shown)


calculuschild:
Noting the test brew used in the original PR: [homebrewery.naturalcrit.com/share/MVz83vplntkj](https://homebrewery.naturalcrit.com/share/MVz83vplntkj)

calculuschild (approved):
Code looks ok... I can see that all the moved function calls had `getBrew()` before so they should still be splitting things. This seems like a good move for tidying things up regardless of fixing the race condition.

Haven't been able to test yet though.

calculuschild:
@G-Ambatte  @dbolack-ab  did we ever find *which* `splitTextStyleandMetadata()` call was resulting in the race condition? I still haven't understood how the race condition can occur in the first place since `splitTextStyleandMetadata()` is already synchronous and has to complete before moving on. I would like to understand where the issue came from for testing purposes.

5e-Cleric (approved):
Looks good to me, it makes sense.

calculuschild:
Sorry @dbolack-ab , it looks like this PR needs to be rebased on the upstream master. It's overwriting some changes from #5016 where most of the app.js routes were moved to page-routes.js

PR #4987, opened 2026-09-03
Simple Pinned favorites.

dbolack-ab:
## Description

Implements a version of Favorites/Bookmarks ( using the term Pin ) as discussed today. 

Changes:
 * adds a pinned array to the Mongo document, containing a list of usernames who have pinned it.
 * adds a UI element to share pages on the left of the share title ( a pushpin ) to toggle pinning the document. White is unpinned, red is pinned.
 * Will not allow an author of a Brew to pin it.
 * Adds pinned themes to the theme selection pulldown.

I am unsure that this is absolutely safe for the document, should a user be editing the document when someone chooses to pin it. Perhaps some stronger filtering?

### Reviewer Checklist

  - [ ] Correctly pins a brew via /share
  - [ ] Reload shows brew still pinned
  - [ ] Correctly unpins a brew via /share
  - [ ] Reload shows brew no longer pinned
  - [ ] Pinned theme shows on the theme pulldown
  - [ ] No race conditions with simultaneous editing
  - [ ] Ensure Author updates do not break pins.
- [ ] Ensure Tests are adequate


ericscheid:
Please create an issue for this PR so we've got traceability. Saying "as discussed today" is insufficient

Also:  
The `pinned` state is stored in the brew itself? The brew that is owned by someone else? That you don't have co-author rights to?

Does updating the brew with pinned also affect the modifiedAt and version metadata?

Does cloning the brew clear the pinning user ids?

server/homebrew.model.js
```diff
@@ -65,6 +67,25 @@ HomebrewSchema.statics.getByUser = async function(username, allowAccess=false, f
 	return brews;
 };
 
+HomebrewSchema.statics.setPin  = async function(username, shareId, pin) {
+	// Test for User Ownership of brew
+	const brew = await Homebrew.findOne({ shareId: shareId }, null).exec()
+		.catch((error)=>{});
+	if(brew) {
+		if(!brew.authors.includes(username)) {
```
calculuschild, line 102: Your current implementation does not allow an author to pin their own brews, which is fine for this first implementation. However, the check for authorship happens at two places. Would it be OK to remove the check here, and leave it in the client GUI only? Then this function can focus purely on setting the property as requested, and we don't have a potential scenario where the GUI says one thing and the API says another if we later allow authors to pin their own brews.

server/homebrew.api.js
```diff
@@ -188,8 +190,7 @@ const api = {
 			stub.renderer = stub.renderer || undefined; // Clear empty strings
 			stub = _.defaults(stub, DEFAULT_BREW_LOAD); // Fill in blank fields
 
-
-
+			stub.pinned = (stub?.pinned.length > 0);
```
calculuschild: This seems like it would replace the pinned array with a boolean true/false at the time of loading. Is this intended?

client/homebrew/navbar/metadata.navitem.jsx
```diff
@@ -27,6 +29,18 @@ const MetadataNav = createReactClass({
...
+		if(global.account) {
+			// Don't toggle if Brew Author
+			if(brew.authors.indexOf(global.account.username) > -1) return;
+			await this.setState((prevProps)=>({
+				pinnedBrew : !prevProps.pinnedBrew
+			}));
+			await request.put('/pin')
+				.send({ username: global.account.username,  shareId: this.props.brew.shareId, pin: this.state.pinnedBrew });
+		}
+	},
```
calculuschild, line 43: One trick here, is `await` does not do anything with `this.setState`, since it doesn't return a promise. I believe you are trying to force the state to update before doing the server request, but React doesn't apply state changes until the next render pass anyway.

The more stable way to update the state and use that updated value inside the same function is to just calculate the new value and use that in all the needed places within this function, something like:

```
	togglePinState : async function(){
		if(global.account) {
			// Don't toggle if Brew Author
			if(brew.authors.indexOf(global.account.username) > -1) return;
			const newPinnedBrew = !this.state.pinnedBrew;
			      
			this.setState({pinnedBrew : newPinnedBrew});
			await request.put('/pin')
				.send({ username: global.account.username,  shareId: this.props.brew.shareId, pin: newPinnedBrew });
		}
	},
```

client/homebrew/navbar/metadata.navitem.jsx
```diff
@@ -68,11 +82,17 @@ const MetadataNav = createReactClass({
 	},
 
 	render : function(){
-		return <Nav.item icon='fas fa-info-circle' color='grey' className='metadata'
-			onClick={()=>this.toggleMetaWindow()}>
-			{this.props.children}
-			{this.renderMetaWindow()}
-		</Nav.item>;
+		return <>
+			<Nav.item icon='fas fa-thumbtack' color='grey' className={`brewPin ${this.state.pinnedBrew ? 'active' : 'inactive'}`}
```
calculuschild: This is OK for now, but it does seem like this pin button should be its own navitem.jsx component instead of nested inside of the metadata dropdown navitem.

server/homebrew.api.js
```diff
@@ -240,7 +241,7 @@ const api = {
 	excludePropsFromUpdate : (brew)=>{
 		// Remove undesired properties
 		const modified = _.clone(brew);
-		const propsToExclude = ['_id', 'views', 'lastViewed'];
+		const propsToExclude = ['_id', 'views', 'lastViewed', 'pinned'];
```
calculuschild: This is I assume to avoid overwriting someone's pins if they happen to pin while you are editing the document?

server/app.js
```diff
@@ -460,6 +470,12 @@ export default async function createApp(vite) {
 			}
 		};
 
+		if( global.account) {
+			if(brew.pinned.indexOf(global.account.username) !== -1) brew.pinned = true;
```
calculuschild: I think this check will not work because in `homebrew.api.js getBrew()` we always force brew.pinned to be a boolean. We have already lost the list of users at this point.

server/homebrew.api.js
```diff
@@ -104,7 +104,9 @@ const api = {
 
 		const userThemes = {};
 
-		const brews = await HomebrewModel.getByUser(username, true, fields, { tags: { $in: ['meta:theme', 'meta:Theme'] } });
+		const userBrewThemes = await HomebrewModel.getByUser(username, true, fields, { tags: { $in: ['meta:theme', 'meta:Theme'] }});
+		const pinnedBrewThemes = await HomebrewModel.getByUser(username, true, fields, { tags: { $in: ['meta:theme', 'meta:Theme'] }, pinned: { $in: [username] } });
```
calculuschild: `getByUser` only finds brews where the current user is an author. It won't find pinned brews from other authors.

May need to make a new mongoose call `getByPinned` or something that is nearly identical but uses the pinned field instead of authors field.

calculuschild (changes requested):
I approve of the implementation here. I am not able to test yet but I just noticed a few things that might cause bugs and had some clarifying questions.

I tried to keep the comments to *just* potential bugs. Someone can come along later and nitpick the design after this is merged.

server/homebrew.model.js
dbolack-ab: My thinking was that since this is an endpoint someone could hit it directly. 

If this is considered a minimal or non-concern, sure.

server/homebrew.api.js
dbolack-ab: Yes. 

I didn't think we wanted/needed to pass the array contents back to the sharepage and we don't need to keep evaluating it over and over.

If it makes things cleaner/clearer we could set the boolean in a different object member and null/drop the array from the loaded brew.

client/homebrew/navbar/metadata.navitem.jsx
dbolack-ab: I wondered about that. Makes sense.

client/homebrew/navbar/metadata.navitem.jsx
dbolack-ab: I tried it both ways, but due to the <metadata /> functionally being the title bar, I placed it there.

server/homebrew.api.js
dbolack-ab: I *believe* this will prevent the Author(s) from affecting the pinned list. That's the intent anyway.

server/homebrew.api.js
dbolack-ab: Oh, I forgot a small bit there to permit that with a null username.

A new Mongoose call works. i'll do that for clarity.

server/homebrew.api.js
dbolack-ab: Moved this to a new member and 'delete' the array when the share is loaded.

The array has been renamed to ***pinnedByUsers** and the new member is ***isPinned***

server/app.js
dbolack-ab: That shouldn't happen in getBrew(). Cleaned up.

server/homebrew.api.js
dbolack-ab: Done

server/homebrew.api.js
calculuschild: Works for me

client/homebrew/navbar/navbar.less
```diff
@@ -118,7 +118,14 @@
 			white-space      : nowrap;
 			background-color : transparent;		
 		}
-
+		&.brewPin {
+				&.active i {
+					color: red;
```
5e-Cleric, line 123: Haven't seen the UI, but i recommend our @blue color from our colors.less in the shared folder for active icons.

server/app.js
```diff
@@ -460,6 +481,15 @@ export default async function createApp(vite) {
...
 
+		if( global.account) {
+			if(brew.pinnedByUsers.indexOf(global.account.username) !== -1) brew.isPinned = true;
+			else { brew.isPinned = false; }
+		} else {
+			brew.isPinned = false;
+		}
+
+		delete brew.pinnedByUsers;
+
```
5e-Cleric, line 492: I might be wrong but won't this delete someone else's pinned brew, since all users will be stored within the one field?

server/brewDefaults.js
```diff
@@ -2,32 +2,33 @@ import _ from 'lodash';
...
+	gDrive        : false,
+	trashed       : false,
+	bleedSize     : { top: '.125in', bottom: '.125in', inner: '.125in', outer: '.125in' },
+	safetySpace   : { top: '.25in', bottom: '.25in', outer: '.25in', inner: '.5in' },
+	trimSize      : { width: '8.5in', height: '11in' },
+	columns       : '2',
+	columnGutter  : '.125in',
+	license       : 'None',
+	legalAuthors  : '',
+	pinnedByUsers : []
```
5e-Cleric, line 31: I think we could call this `pinnedBy`, no? not sure if we need it this long.

server/homebrew.model.js
5e-Cleric: I'd normally say it is a minimal concern, but i have been asked to add similar checks on every one of my PRs using an endpoint, so i have mixed feelings here. I say we keep it.

5e-Cleric (commented):
Overall looks good, left a few non-vital comments, will approve once i have properly tested.

PR #4989, opened 2026-09-03
minimal folders backend infrastructure

### LINKED ISSUE #4988: MVP of folders backend infrastructure
### Your idea:

Get the ball rolling on the MVP for https://github.com/naturalcrit/homebrewery/wiki/Folders-etc

- [ ] install the basic plumbing of model + router + api
- [ ] front-end UI for CRUD + addBrewToFolder + removeBrewFromFolder
- [ ] embrace and extend

ericscheid:
## Description

This is the basic plumbing for the folders functionality, based off the MVP on https://github.com/naturalcrit/homebrewery/wiki/Folders-etc.

No front-end UI yet, just the backend plumbing of model+functions+api routing.

#4988 

Once staged/live, the only exposure would be a "folders" array sent in window.__INITIAL_PROPS__, visible via view source.

### DIFF
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -271,7 +271,7 @@ const errorIndex = (props)=>{
 		'100': dedent`Folder operations require a logged in user.`,
 		'101': dedent`A folder with this identifier already exists`,
 		'102': dedent`A bookmarks folder already exists`,
-		'103': dedent`A favourites folder already exists`,
+		'103': dedent`A favorites folder already exists`,
 
 		'104': dedent`Folder could not be found.`,
 		'105': dedent`Folder could not be created.`,
--- a/client/homebrew/pages/errorPage/errors/errorIndex.js
+++ b/client/homebrew/pages/errorPage/errors/errorIndex.js
@@ -15,14 +15,14 @@ const errorIndex = (props)=>{
 		// Default catch all
 		'00' : dedent`
 			## An unknown error occurred!
-			
+
 			We aren't sure what happened, but our server wasn't able to find what you
 			were looking for.`,
 
 		// General Google load error
 		'01' : dedent`
 			## An error occurred while retrieving this brew from Google Drive!
-			
+
 			Google is able to see the brew at this link, but reported an error while attempting to retrieve it.
 
 			### Refreshing your Google Credentials
@@ -39,7 +39,7 @@ const errorIndex = (props)=>{
 		// Google Drive - 404 : brew deleted or access denied
 		'02' : dedent`
 			## We can't find this brew in Google Drive!
-			
+
 			This file was saved on Google Drive, but this link doesn't work anymore.
 			${props.brew.authors?.length > 0
 		? `Note that this brew belongs to the Homebrewery account **${props.brew.authors[0]}**,
@@ -66,7 +66,7 @@ const errorIndex = (props)=>{
 			- **The Google Account may be closed.** Google may have removed the account
 			due to inactivity or violating a Google policy. Make sure the owner can
 			still access Google Drive normally and upload/download files to it.
-			
+
 			If the file isn't found, Google Drive usually puts your file in your Trash folder for
 			30 days. Assuming the trash hasn't been emptied yet, it might be worth checking.
 			You can also find the Activity tab on the right side of the Google Drive page, which
@@ -74,7 +74,7 @@ const errorIndex = (props)=>{
 			the brew was deleted or moved, and by whom.
 			:
 			If the brew still isn't found, some people have had success asking Google to recover
-			accidentally deleted files at this link: 
+			accidentally deleted files at this link:
 			https://support.google.com/drive/answer/1716222?hl=en&ref_topic=7000946.
 			At the bottom of the page there is a button that says *Send yourself an Email*
 			and you will receive instructions on how to request the files be restored.
@@ -91,22 +91,22 @@ const errorIndex = (props)=>{
 		as an author by opening the Edit page for the brew, viewing the {{fa,fa-info-circle}}
 		**Properties** tab, and adding your username to the "invited authors" list. You can
 		then try to access this document again.
-		
+
 		:
 
 		**Brew Title:** ${escape(props.brew.brewTitle) || 'Unable to show title'}
 
 		**Current Authors:** ${props.brew.authors?.map((author)=>{return `[${author}](/user/${encodeURIComponent(author)})`;}).join(', ') || 'Unable to list authors'}
-		
+
 		[Click here to be redirected to the brew's share page.](/share/${props.brew.shareId})`,
 
 		// User is not signed in; must be a user on the Authors List
 		'04' : dedent`
 		## Sign-in required to edit this brew.
-		
+
 		You must be logged in to one of the accounts listed as an author of this brew.
 		User is not logged in. Please log in [here](${loginUrl}).
-		
+
 		:
 
 		**Brew Title:** ${escape(props.brew.brewTitle) || 'Unable to show title'}
@@ -119,10 +119,10 @@ const errorIndex = (props)=>{
 		// Brew load error
 		'05' : dedent`
 		## No Homebrewery document could be found.
-		
+
 		The server could not locate the Homebrewery document. It was likely deleted by
 		its owner.
-		
+
 		:
 
 		**Requested access:** ${props.brew.accessType}
@@ -132,15 +132,15 @@ const errorIndex = (props)=>{
 		// Brew save error
 		'06' : dedent`
 		## Unable to save Homebrewery document.
-		
+
 		An error occurred wil attempting to save the Homebrewery document.`,
 
 		// Brew delete error
 		'07' : dedent`
 		## Unable to delete Homebrewery document.
-		
+
 		An error occurred while attempting to remove the Homebrewery document.
-		
+
 		:
 
 		**Brew ID:**  ${props.brew.brewId}`,
@@ -148,9 +148,9 @@ const errorIndex = (props)=>{
 		// Author delete error
 		'08' : dedent`
 		## Unable to remove user from Homebrewery document.
-		
+
 		An error occurred while attempting to remove the user from the Homebrewery document author list!
-		
+
 		:
 
 		**Brew ID:**  ${props.brew.brewId}`,
@@ -158,10 +158,10 @@ const errorIndex = (props)=>{
 		// Theme load error
 		'09' : dedent`
 		## No Homebrewery theme document could be found.
-		
+
 		The server could not locate the Homebrewery document. It was likely deleted by
 		its owner.
-		
+
 		:
 
 		**Requested access:** ${props.brew.accessType}
@@ -171,17 +171,17 @@ const errorIndex = (props)=>{
 		// Theme Not Valid
 		'10' : dedent`
 		## The selected theme is not tagged as a theme.
-		
+
 		The brew selected as a theme exists, but has not been marked for use as a theme with the \`theme:meta\` tag.
-		
+
 		If the selected brew is your document, you may designate it as a theme by adding the \`theme:meta\` tag.`,
 
 		// ID validation error
 		'11' : dedent`
 		## No Homebrewery document could be found.
-		
+
 		The server could not locate the Homebrewery document. The Brew ID failed the validation check.
-		
+
 		:
 
 		**Brew ID:**  ${props.brew.brewId}`,
@@ -189,9 +189,9 @@ const errorIndex = (props)=>{
 		// Google ID validation error
 		'12' : dedent`
 		## No Google document could be found.
-		
+
 		The server could not locate the Google document. The Google ID failed the validation check.
-		
+
 		:
 
 		**Brew ID:**  ${props.brew.brewId}`,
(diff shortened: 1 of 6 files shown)


server/folder.api.js
```diff
@@ -0,0 +1,119 @@
...
+      message: 'Folder not found',
+      status: 404,
+    };
+
+  res.status(200).send(folder);
+};
+
+
+
+router.post('/folder/',
```
G-Ambatte: If the `requireAccount`(s) here throws an error, (from memory) routes that don't start with `/api/` bounce to ErrorPage. If this is intended, the thrown error should include a HBErrorCode, and have an associated entry in `errors.js`.

ericscheid: OK, adding `HBErrorCode: 101` and updating errorIndex.js.

The other throw errors I'll also update.

client/homebrew/pages/errorPage/errors/errorIndex.js
```diff
@@ -261,10 +261,28 @@ const errorIndex = (props)=>{
...
             Try again in a few minutes.`,
 
 		'91' : dedent` An unexpected error occurred while trying to get the total of brews.`,
+
+		// ####### Folder API errors
+
+		'100': dedent`Folder operations require a logged in user.`,
+		'101': dedent`A folder with this identifier already exists`,
+		'102': dedent`A bookmarks folder already exists`,
+		'103': dedent`A favourites folder already exists`,
```
5e-Cleric:
```suggestion
		'103': dedent`A favorites folder already exists`,
```
US spelling of favorites preferred.

client/homebrew/pages/errorPage/errors/errorIndex.js
```diff
@@ -261,10 +261,28 @@ const errorIndex = (props)=>{
...
+
+		'104': dedent`Folder could not be found.`,
+		'105': dedent`Folder could not be created.`,
+		'106': dedent`Folder to update could not be found.`,
+		'107': dedent`Folder to delete could not be created.`,
```
5e-Cleric: I believe folder could not be found should be just one issue, with the context not being all that important. @calculuschild, @G-Ambatte @dbolack-ab opinions?

5e-Cleric: And pretty sure that 107 error should be "Folder to delete could not be found"

server/folder.api.js
```diff
@@ -0,0 +1,168 @@
...
+    isPublished: req.body.isPublished ?? false,
+  });
+
+  if(!folder)
+    throw {
+      HBErrorCode: 105,
+      name: 'FolderCreate Error',
+      message: 'Folder could not be created',
+      status: 404,
+    };
```
5e-Cleric: if there is an hberror code it should supply the message, instead of manually adding it here.

Same for the other errors in here.

server/folder.model.js
```diff
@@ -0,0 +1,164 @@
+// /server/folder.model.js
+
+import mongoose from 'mongoose';
+import { nanoid } from 'nanoid';
+import { model as BrewModel } from './homebrew.model.js';
+
+
+const FolderSchema = mongoose.Schema({
+  owner:        { type: String, required: true, index: true },
```
5e-Cleric:
```suggestion
  authors:        { type: String, required: true, index: true },
```

As in brews, even if we do not make later a feature to share brews as co-authors, we should use our own terminology, which in this case is Authors.

server/folder.model.js
```diff
@@ -0,0 +1,164 @@
...
+  folderId:     { type: String, required: true, index: true, unique: true, default: () => nanoid(12) },
+  slug:         { type: String, required: true, trim: true, },
+  displayName:  { type: String, required: true, trim: true, default: 'untitled folder', },
```
5e-Cleric:
```suggestion
  title:  { type: String, required: true, trim: true, default: 'untitled folder', },
```

Same as before, we already have names for these things.

server/folder.model.js
```diff
@@ -0,0 +1,164 @@
...
+// Application code validates slug for syntax, and also sibling uniqueness
+
+// Folders can contain sub-folders, as a DAG, not a tree
+// subfolderIds contains outgoing edges in the folder DAG
+// Application code prevents self-references and cycles
+
+// No semantics implied by array order of brewIds or subfolderIds.
+
+// isPublished determines appearance in the User's Published Brews section of their user page
+// isPrivate means non-owners cannot view the folder even if they have the url
```
5e-Cleric: Can a folder be published and private?

server/folder.model.js
```diff
@@ -0,0 +1,164 @@
...
+    { new: true },
+  );
+
+  return folder;
+};
+
+FolderSchema.statics.deleteFolder = async function(owner, folderId) {
+  // TODO: remove dangling references to this folderId. not essential, just tidy.
+  return this.deleteOne({ owner, folderId });
+};
```
5e-Cleric: Question, is this also deleting the brews inside it? and subfolders?

server/folder.model.js
```diff
@@ -0,0 +1,164 @@
...
+
+
+FolderSchema.statics.addBrewToFolder = async function( owner, folderId, brewId) {
+
+  const brewExists = await BrewModel.exists({ brewId });
+  if(!brewExists)
+    return { error: 'BREW_NOT_FOUND' };
```
5e-Cleric, line 114: This should return an error object with a code, not just a name, i think.

5e-Cleric (changes requested):
Looks good, left some comments, i promise to try and add them all together in a review instead of all separate in the future. i always forget.

server/folder.model.js
ericscheid: No. Folders in folders form a DAG, not a tree. Deleting a folder simply deletes those folder → item *relationships*.

Items in a folder continue to exist, and appear wherever else they have memberships.

If an item is not contained by any other folder of this author, and this author would otherwise have edit access, it reappears on the root userpage. That is, _non-filed_ brews and folders appear at the root userpage.

server/folder.model.js
ericscheid: Yes. It would only be visible to that author, but it would appear in their Published Brews section of their userpage.

A little confusing, but less impactful than changing isPublished into a ternary from a binary.

server/folder.model.js
ericscheid: I have the `Api` functions returning error objects, and the model functions returning whatever results they generate (e.g. see deleteFolderApi → deleteFolder → deleteOne).

Problem is, addBrewToFolder can fail due to brew not found, or folder not found, and so I needed a way to distinguish them. Can certainly return { HBErrorCode: 111 or 112 } instead.

Otherwise I'd be checking for the error situation in both someFunctionApi *and* someFunction, which doesn't seem DRY. 

That is, check if result of `findOneAndUpdate` is `null`, and if so return an error object (instead of simply the `null` generated), and then in the functionApi checking for a HBErrorCode to build a new error object to throw. 

Note also that a given model function could be called by different Api functions, and so can't presume which HBErrorCode would ultimately be needed. Can get around that with extra HBErrorCodes for primitive/generic errors (e.g. Folder not found, Brew not found), which the caller then detects and handles as they need (e.g throw "Folder to add brew to could not be found.", or "Folder to remove brew from could not be found.", or "Folder to delete could not be created.", etc).

client/homebrew/pages/errorPage/errors/errorIndex.js
ericscheid: I'm using these for api responses, and a generic "folder could not be found" could be generated by different api calls. I used distinct error codes because that's what show up in the jest tests. 

Could be I don't know how to read the jest reports well enough.

Happy to go with more generic error codes.

server/folder.api.js
ericscheid: what's the code for grabbing the error message from errorIndex.js ?

ericscheid:
@5e-Cleric 
> Looks good, left some comments, i promise to try and add them all together in a review instead of all separate in the future. i always forget.

Most are easy to resolve, no problem.

The one of what model functions returns is a little unclear on which path to pursue. Is it only the two functions that return a custom error messages that need updating/conforming, or all the model functions (instead of simply returning the mongoose function result, which could be `null`)?

server/folder.model.js
```diff
@@ -0,0 +1,164 @@
...
+
+FolderSchema.statics.deleteFolder = async function(owner, folderId) {
+  // TODO: remove dangling references to this folderId. not essential, just tidy.
+  return this.deleteOne({ owner, folderId });
+};
+
+
+FolderSchema.statics.addBrewToFolder = async function( owner, folderId, brewId) {
+
+  const brewExists = await BrewModel.exists({ brewId });
```
ericscheid: brewId is not a field on `BrewModel` .. it should be searching `shareId: brewId`.

ericscheid: new commit inbound.

server/folder.api.js
```diff
@@ -0,0 +1,172 @@
...
+    isPublished: req.body.isPublished ?? false,
+  });
+
+  if(!folder) {
+   const error = {
+      HBErrorCode: 105,
+      name: 'FolderCreate Error',
+      status: 404,
+    };
+    error.message = ErrorIndex({folderId: 'xyz', folderName: 'abc'})[error.HBErrorCode.toString()];
```
calculuschild, line 41: Following our gitter chat, these API calls should not be pulling from ErrorIndex. Just put the message here as you would like it to appear in the log.

Then when the client side is developed later, you can choose to use the same message there, or something more detailed.

server/folder.model.js
calculuschild: Sometimes there are cases like this where there is a common root error (brew not found) that can be triggered from multiple places, and you want to distinguish them, while also calling out the root error.

In that case you can `throw` (not just `return`) the original error, then catch it in the calling function and append any detail needed, and throw again. Examples:

https://github.com/naturalcrit/homebrewery/blob/aeb5a74452894d442fa8499a0ba79e68eb94cdb1/server/homebrew.api.js#L170-L175

https://github.com/naturalcrit/homebrewery/blob/aeb5a74452894d442fa8499a0ba79e68eb94cdb1/server/homebrew.api.js#L312-L316

You don't need a unique error code here as long as every function that calls it applies its own error code. In other words, every error should *eventually* have an error code applied somewhere in the stack.

___


Keep in mind that the only errors that actually get logged are the ones that are never caught (eventually they land in the app.js error handler), or if you explicitly do a console.error/warn/log somewhere. So if you want "BREW_NOT_FOUND" to actually appear anywhere, it needs to be appended to the calling function's error, or you need to add an explicit console.error('BREW_NOT_FOUND") here.

server/folder.api.spec.js
```diff
@@ -0,0 +1,422 @@
...
+
+      expect(FolderModel.createFolder).toHaveBeenCalledWith('testuser', {
+        title       : 'Test',
+        slug        : undefined,
+        isPublished : false,
+      });
+    });
+
+    it('should throw when the folder could not be created', async ()=>{
+      FolderModel.createFolder.mockResolvedValue(undefined);
```
calculuschild: My instinct is this is not going to work. You are replacing the entire function of `createFolder` with "return undefined", which means the part of the code that actually generates the 105 error is not going to be executed. We are testing an empty function here.

server/folder.api.spec.js
```diff
@@ -0,0 +1,422 @@
...
+      await api.createFolderApi(req, res);
+
+      expect(FolderModel.createFolder).toHaveBeenCalledWith('testuser', {
+        title       : 'Test',
+        slug        : undefined,
+        isPublished : false,
+      });
+    });
+
+    it('should throw when the folder could not be created', async ()=>{
```
calculuschild, line 107: I don't think this test is testing what you want. If the goal is simply "when the model returns undefined, throw an error", then yes, this works.

But I'm assuming the scenario you are testing is "empty body should result in an error", there is nothing in this test that actually reads the `body` field.

If you aren't checking for valid `body` fields in your `api.createFolderApi()` but just relying on the model `save()` function to fail, that becomes trickier to test since we don't currently have any Mongo DB simulation set up. There are ways to do that (and maybe we should).

Just wanted to be sure that was clear. Not necessarily blocking this PR since there's not a better way to do it at the moment.

server/folder.model.spec.js
```diff
@@ -0,0 +1,599 @@
...
+
+      const result = await Folder.getByUser('testuser', false);
+
+      expect(Folder.find).toHaveBeenCalledWith({
+        author    : 'testuser',
+        isPrivate : false,
+      });
+
+      expect(result).toBe(folders);
+    });
```
calculuschild, line 169: A lot of these model tests I don't think are testing what you think. This test essentially breaks down to:

1) Create a fake function that always returns `[]`;
2) Call that fake function with `{'testuser', false}` as inputs
3) Confirm the fake function had `{'testuser', false}` as inputs
4) Confirm the fake function indeed returned `[]`;

This isn't actually testing any of the behavior of Folder.getByUser, other than confirming that at some point the fake version of `find` was called.

server/folder.api.js
ericscheid: Yup, reworking the folder errors into a local thing (keeping numbers in errorIndex.js as registry). 

Also reworking model vs api calls → returns/throws to maintain separation of concerns.

server/folder.model.spec.js
ericscheid: getting a crash course in jest =/

good points .. lemme fix the model & api to a saner architecture for handling errors between FolderModel and Api functions (and which doesn't use ErrorIndex) ... and then start tacking all the tests again.

server/folder.api.js
ericscheid: after much discussion, won't be accessing `ErrorIndex()` in its current state.

client/homebrew/pages/errorPage/errors/errorIndex.js
ericscheid: Architecture for error messaging is being reworked.

1. FolderModel will throw  various `{ HBErrorCode: nnn, operation: "CreateFolder", message: "...", source: err }` depending on which part borked
2. functionFolderApi will catch those and insert appropriate http status values (404, 409, 500, 218, etc), and then throw that (or ignore as maybe)
3. folderErrorMessages is an array in folder.model.js
4. a copy of the folderErrror codes will be maintained in errorIndex.js, acting as a registry
5. a _whole_ bunch of tests will get rewritten

server/folder.api.spec.js
ericscheid: Appreciate the eyes-on.

i'm keeping tests as atomic as possible — IIRC, this one simply tests api.createFolderApi responds properly to FolderModel.createFolder returning null instead of a folder.

That said, am revamping FolderModel.createFolder to throw instead of return if operation failed some how, and will throw a more informative object than `null`.

I imagine there should be tests in folder.model.spec.js that will have to test what happens if Mongo fails (and hence test the various error paths there).

server/folder.model.spec.js
ericscheid: and yes, the testing of omitting isPrivate 3rd-party folders should be done along the lines of set up a folder with isPrivate and author: "other" and go from there.

PR #4993, opened 2026-09-05
Userpage UI Rework

5e-Cleric:
This PR brings a rework of the userpage UI, moving away from the basic page dnd themed approach, and into more options.

All the styles are drafts, open to suggestions so everybody kinda likes it.

I have yet to store the mode in localstorage, but i'll do so.

I'll upload screenshots.

## The three modes: Grid

image

## List

image

## Cards

image

### DIFF
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -19,6 +19,7 @@ const ListPage = ({ brewCollection = [{ title: '', class: '', brews: [] }], navI
 	const [sortType, setSortType] = useState(query?.sort || null);
 	const [sortDir, setSortDir] = useState(query?.dir || null);
 	const [groupVisibility, setGroupVisibility] = useState({});
+	const [layoutMode, setLayoutMode] = useState('grid');
 
 	const groupVisibilityRef = useRef(groupVisibility);
 	const sortTypeRef = useRef(sortType);
@@ -259,31 +260,61 @@ const ListPage = ({ brewCollection = [{ title: '', class: '', brews: [] }], navI
 			const sortedBrewGroup = sortedBrewCollection[idx];
 			const visible = groupVisibility[brewGroup.class];
 
-			return (
-				<div key={idx} className={`brewCollection ${brewGroup.class ?? ''}`}>
-					<h1
-						className={visible ? 'active' : 'inactive'}
-						onClick={()=>{
-							toggleBrewCollectionState(brewGroup.class);
-						}}>
-						{brewGroup.title || 'No Title'}
-					</h1>
+			return (<>
+				<h1
+					className={visible ? 'active' : 'inactive'}
+					onClick={()=>{
+						toggleBrewCollectionState(brewGroup.class);
+					}}>
+					{brewGroup.title || 'No Title'}
+				</h1>
+				<div key={idx} className={`brewGroup ${brewGroup.class ?? ''} ${layoutMode}`}>
+
 					{visible ? renderBrews(sortedBrewGroup.brews) : <></>}
 				</div>
-			);
+			</>);
 		});
+	};
+
+	const handleLayoutChange = (e, mode)=>{
+		setLayoutMode(e.target.checked ? mode : 'grid');
+		return;
+	};
+
+	const renderLayoutModeOptions = ()=>{
+		return (
+			<div className='layout-container'>
+				<div className='layout-option' title='grid'>
+					<label>
+						<i className='fas fa-grip'></i>
+						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'grid')} checked={layoutMode === 'grid'} />
+					</label>
+				</div>
+				<div className='layout-option' title='list'>
+					<label>
+						<i className='fas fa-grip'></i>
+						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'list')} checked={layoutMode === 'list'} />
+					</label>
+				</div>
+				<div className='layout-option' title='card'>
+					<label>
+						<i className='fas fa-table'></i>
+						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'card')} checked={layoutMode === 'card'} />
+					</label>
+				</div>
+			</div>
+		);
 	};
 
 	return (
 		<div className='listPage sitePage'>
-			<link href='/themes/V3/Blank/style.css' type='text/css' rel='stylesheet' />
-			<link href='/themes/V3/5ePHB/style.css' type='text/css' rel='stylesheet' />
 			{navItems}
 			{renderSortOptions()}
 			{renderTagsOptions()}
+			{renderLayoutModeOptions()}
 
 			<div className='content V3'>
-				<div className='page'>{renderBrewCollection(brewCollection)}</div>
+				<div className='brewCollection'>{renderBrewCollection(brewCollection)}</div>
 			</div>
 		</div>
 	);
--- a/client/homebrew/pages/basePages/listPage/listPage.jsx
+++ b/client/homebrew/pages/basePages/listPage/listPage.jsx
@@ -286,20 +286,20 @@ const ListPage = ({ brewCollection = [{ title: '', class: '', brews: [] }], navI
 			<div className='layout-container'>
 				<div className='layout-option' title='grid'>
 					<label>
-						<i className='fas fa-grip'></i>
 						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'grid')} checked={layoutMode === 'grid'} />
+						<i className='fas fa-grip'></i>
 					</label>
 				</div>
 				<div className='layout-option' title='list'>
 					<label>
-						<i className='fas fa-grip'></i>
 						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'list')} checked={layoutMode === 'list'} />
+						<i className='fas fa-grip'></i>
 					</label>
 				</div>
 				<div className='layout-option' title='card'>
 					<label>
-						<i className='fas fa-table'></i>
 						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'card')} checked={layoutMode === 'card'} />
+						<i className='fas fa-table'></i>
 					</label>
 				</div>
 			</div>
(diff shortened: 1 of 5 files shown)


dbolack-ab:
I really like what I see here. 

Once you get the mode stored and we can do some live tests I *think* this is good to go as is and we can "perfect" the styling once it is on the dev instance.

Don't let perfect be the enemy.

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -259,31 +260,61 @@ const ListPage = ({ brewCollection = [{ title: '', class: '', brews: [] }], navI
...
+		setLayoutMode(e.target.checked ? mode : 'grid');
+		return;
+	};
+
+	const renderLayoutModeOptions = ()=>{
+		return (
+			<div className='layout-container'>
+				<div className='layout-option' title='grid'>
+					<label>
+						<i className='fas fa-grip'></i>
```
ericscheid, line 289: if you swap the order of the `<i />` and `<input />` elements not only will this look less weird, but also make styling the two elements easier.

For example:
```css
label {
  cursor: pointer;
}

label input[type="radio"] {
  display: none;
}

label i {
  color: #999;
  font-size: 24px;
}

label input[type="radio"]:checked + i {
  color: #007bff;
}
```

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -259,31 +260,61 @@ const ListPage = ({ brewCollection = [{ title: '', class: '', brews: [] }], navI
...
+						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'grid')} checked={layoutMode === 'grid'} />
+					</label>
+				</div>
+				<div className='layout-option' title='list'>
+					<label>
+						<i className='fas fa-grip'></i>
```
ericscheid, line 295: This list view option uses same icon as grid view.

Maybe use fa-list, or fa-bars ?

client/homebrew/pages/basePages/listPage/listPage.jsx
```diff
@@ -259,31 +260,61 @@ const ListPage = ({ brewCollection = [{ title: '', class: '', brews: [] }], navI
...
+						<input name='layout-mode' type='radio' onChange={(e)=>handleLayoutChange(e, 'list')} checked={layoutMode === 'list'} />
+					</label>
+				</div>
+				<div className='layout-option' title='card'>
+					<label>
+						<i className='fas fa-table'></i>
```
ericscheid, line 301: Given this is essentially a variant on the grid format, being vertical instead of horizontal shaped cards .. maybe use fa-grip-vertical. 

Then, put this control next to the fa-grip control, exploiting the gestalt principle of proximity to aid the viewer's understanding.

ericscheid (changes requested):
Looks great.

A couple of minor changes for UX/UI.

Gazook89:
Apologies in advance.  I tried for brevity, but had a long list.  These are my notes, I can go into more depth on any of them, and I can also submit my own PR either to your fork/branch/PR or wait for you to merge whatever and then submit my changes on it's own PR (if people generally agree and you just want to submit what you have).  I understand you likely didn't come into these files looking for a fight.

- the list and grid view have the same icon.  
- I don't think a second row in the toolbar is necessary for the different views.
- dramatically reduce the inline padding on small screens
- published and unpublished shouldn't be separated into two `details` elements, it should just be a filterable/sortable option with a visual indicator on the brews to show if they are published or not.

### grid view
- I don't personally think it's been demonstrated that both cards and grid are needed
- don't center cards that are on their own row.
- the action links (edit, view, ...) should always be visible.  perhaps vertically on the right edge as it is now, but remove the need to hover.
- the grid/card views can be done with actual CSS Grid so that fill the whole width of the grid container and responsively grow/shrink within certain bounds before kicking items to the next row.  See the example video after the card view section
- Or, perhaps better in the near future, drop the "grid" idea and use [CSS Grid Lanes](https://gridlanes.webkit.org/).  This will work better for items with varying lengths of descriptions.  Descriptions wouldn't have to be cut-off just to fit into a fixed size box, without awkward gaps between neighboring items.  Available in Safari TP, and Chrome/Edge behind a flag.  Likely in all browsers by end of year.  [webkit link for more info](https://webkit.org/blog/17660/introducing-css-grid-lanes/)
  - A very quick example using devtools on live site below.  No other changes to the brewItems are made for this example, but further work would be done to soften it up.  This would have best visual impact with a mix of brewItems with cover photos and some without, where cover photos could be clicked to reveal the description.
  
image

### card view
- set a much smaller gap.  preferably by removing the `margin` on the brewItems entirely, and using `gap: 20px`.
- not sure the grid layout of the `info` hits the mark yet.
  - too much spacing
  - emphasize the brew title by de-emphasizing the info stats.
  
### Video of css grid responsively sizing brewItems and filling available space:

https://github.com/user-attachments/assets/05a9f1d3-7253-4745-afc3-4b33bfabde6b

### list view
- set h2 brew title to same size as the rest of the text
- remove icons from each brewItem (user, views, pages, etc) and instead use a header row with those icons/text
- remove coloring of action buttons.  As-is, I think it's too low of contrast, and I think it'll complicate any future dark mode.
- the tags list:
  - my brew with 4 tags is only showing one of them, even at very wide width?
  - hovering that tag is extremely finicky, flashes in and out
  - when hovering is working, i can see there is a `:after` element with more tags, but strangely only _one_ more tag?
  - when hovering is working, the :after element is obscured by a `title` tooltip attribute showing the full list of tag names.
  - I think you should try setting the tags as an inline element so they come immediately after the title of the brew, rather than in a "column"
  - reduce gap between brew items to 1px.  thicker horizontal lines trick your eyes into moving side to side when in fact I think you want the eye to be moving vertically down the list on any given column (scanning for a brew title).  The 1px gap is enough for helping move across a row without losing your place.  If further visual help is needed for that, a hovered/focused (or *interest invoker api*) effect that tints the row would do.
- Here is some of those changes, quickly:

image

5e-Cleric:
> Apologies in advance. I tried for brevity, but had a long list. These are my notes, I can go into more depth on any of them, and I can also submit my own PR either to your fork/branch/PR or wait for you to merge whatever and then submit my changes on it's own PR (if people generally agree and you just want to submit what you have). I understand you likely didn't come into these files looking for a fight.

Looking for a fight indeed


> * the list and grid view have the same icon.

Changed it to:

image

I plan on doing at least an icon for the card layout, but i don't see why this PR has to be held back because of that.

> * I don't think a second row in the toolbar is necessary for the different views.

There is no second bar, and no third, all three bars are now condensed into one and wrapped  appropiately.


> * dramatically reduce the inline padding on small screens

The inline padding of what? the page, the wrapper, the brew items???

> * published and unpublished shouldn't be separated into two `details` elements, it should just be a filterable/sortable option with a visual indicator on the brews to show if they are published or not.

This is a big change I don't actually agree with, if you can get votes for it, i'll do it.


> ### grid view
> * I don't personally think it's been demonstrated that both cards and grid are needed

Nothing has been demonstrated, I think they are useful in their extremes. Cards don't actually look that good without thumbnails, i prefer to maintain all 3 modes.

> * don't center cards that are on their own row.
Fixed

> * the action links (edit, view, ...) should always be visible.  perhaps vertically on the right edge as it is now, but remove the need to hover.

this is just preserving the current UX, you know that, but i agree, they will be visible.

> * the grid/card views can be done with actual CSS Grid so that fill the whole width of the grid container and responsively grow/shrink within certain bounds before kicking items to the next row.  See the example video after the card view section

There is no need for grid in all cases, i fail to see why we'd want a grid in the grid view, but it is used in cards now.

As for the video, it does not play for me at all.

> * Or, perhaps better in the near future, drop the "grid" idea and use [CSS Grid Lanes](https://gridlanes.webkit.org/).  This will work better for items with varying lengths of descriptions.  Descriptions wouldn't have to be cut-off just to fit into a fixed size box, without awkward gaps between neighboring items.  Available in Safari TP, and Chrome/Edge behind a flag.  Likely in all browsers by end of year.  [webkit link for more info](https://webkit.org/blog/17660/introducing-css-grid-lanes/)
>   
>   * A very quick example using devtools on live site below.  No other changes to the brewItems are made for this example, but further work would be done to soften it up.  This would have best visual impact with a mix of brewItems with cover photos and some without, where cover photos could be clicked to reveal the description.
> 
> image

Oh no, a masonry grid is great for displaying images in pinterest, but here there is a simple order to the items that must be easy to follow, which a masonry layout breaks completely. Very much opposed.


> ### card view
> * set a much smaller gap.  preferably by removing the `margin` on the brewItems entirely, and using `gap: 20px`.

Further decreasing the gap does not remove extra space, because the content is centered within itself, for achieving a smaller space, flex would have to be used, which would break the last item or items and center them too.
image



> * not sure the grid layout of the `info` hits the mark yet.

Yeah, i want it to stay in place no matter the length of data, (say 30k views vs 1 view) so i use grid, but then spaces are left behind when the data isn't that long. It is either that or the placement won't match between cards.

>   * too much spacing
>   * emphasize the brew title by de-emphasizing the info stats.

Probably a good idea, still have to decide on fonts too


> ### list view
> * set h2 brew title to same size as the rest of the text

No? that way it won't be a title? I guess i can reduce it a bit.

> * remove icons from each brewItem (user, views, pages, etc) and instead use a header row with those icons/text

There is no need to turn this into a table, there is plenty of space as is.

> * remove coloring of action buttons.  As-is, I think it's too low of contrast, and I think it'll complicate any future dark mode.

I can test that out.


> * the tags list:
>   
>   * my brew with 4 tags is only showing one of them, even at very wide width?

That is not my code, all tags are visible:

image


>   * hovering that tag is extremely finicky, flashes in and out

I haven't added nor modified the hover, that is the title attribute.

>   * when hovering is working, i can see there is a `:after` element with more tags, but strangely only _one_ more tag?

not my code, i have not touched the title attribute

>   * when hovering is working, the :after element is obscured by a `title` tooltip attribute showing the full list of tag names.

What after element? i have not added any after element, there is only the title tooltip

>   * I think you should try setting the tags as an inline element so they come immediately after the title of the brew, rather than in a "column"

They are an inline element, that wraps when there is no more space, just like the title overflows into ellpisis when there is no more space.

>   * reduce gap between brew items to 1px.  thicker horizontal lines trick your eyes into moving side to side when in fact I think you want the eye to be moving vertically down the list on any given column (scanning for a brew title).  The 1px gap is enough for helping move across a row without losing your place.  If further visual help is needed for that, a hovered/focused (or _interest invoker api_) effect that tints the row would do.

There is no gap between elements, there is only a gap if the element doesn't fill the assigned space.
image
The way to fill those "gaps" would be to use flex rather than grid, and then the information would not be aligned between items.



> * Here is some of those changes, quickly:
> 
> image

I see, but that makes the tags unaligned with each other, i do not like that at all.

I am not looking for a fight either, but damn we disagree on stuff.

PR #4994, opened 2026-09-05
Add tests for static-assets.mv.js

dbolack-ab:
## Description

Brings file coverage to 100%

I added brotli files for the files in shared/naturalcrit/styles/fonts for testing against.


5e-Cleric:
Lets see if i get this, so, to test the fonts we need to add in copies of them in a compressed format?

5e-Cleric (approved):

shared/naturalcrit/styles/fonts/CODE Bold.otf.br
calculuschild, line 1: We shouldn't need to make .br copies of assets for the test. We fetch assets from the /build folder, which should already contain the .br files as part of the build process.

tests/routes/static-content.test.js
```diff
@@ -0,0 +1,37 @@
...
+	request = supertest.agent(app).set('X-Forwarded-Proto', 'https');
+});
+
+describe('Tests for static assets and cache settings', ()=>{
+	it('Test Brotli compressed CSS file', async ()=>{
+		const results = await request.get('/fonts/fonts/fonts.css.br');
+		expect(results.header['content-type']).toBe('application/octet-stream');
+		expect(results.header['cache-control']).toBe('public, max-age: 43200, must-revalidate');
+	});
+
```
calculuschild, line 18: We don't explicitly request .br files in normal operation. That's what `static-assets.mv.js` is for. It takes a request for font.woff2, searches for a .br copy of the same file, and returns it if it exists.

This test bypasses all that logic by requesting the .br directly, We should be able to test a request for a normal file extension, and then confirm that our asset handler correctly located the brotli copy (something like `expect(res.headers['content-encoding']).toBe('br')`).

calculuschild: So, it appears this has revealed a side-effect of switching to Vite. Our build process no longer performs brotli compression by default, so there are no .br files to serve.

Font files weren't compressed in the first place (.woff2 is already compressed), but we did compress our .js and .css bundles back then.

That said, I still vote we don't add .br copies of things to our source just for testing. We can test the part we actually use, and it's ok if the brotli server isn't 100% covered for now. When/if we figure out if it's worth having Vite compress stuff for us [maybe with a plugin](https://medium.com/@tuhinkairi/improve-your-react-performance-by-2x-using-vite-compress-fba062ccad0e) , we can test against those.

shared/naturalcrit/styles/fonts/CODE Bold.otf.br
dbolack-ab: If they were being generated pre-vite, they are not now.  So I generated entities that I could trust to exist for testing purposes.

dbolack-ab:
Closing per vote.

calculuschild:
The test spec file is ok, I just voted to remove the brotli tests specifically.

PR #4996, opened 2026-09-07
Fix background color in local installations

G-Ambatte:
This PR changes the default value of the `development` config property to FALSE, and adds it to the checks in `homebrew.jsx` to set the class for the background color in exactly the same way that it was already set for the background object (containing the SVG for deployment name, etc).

### Screenshots

`development`: FALSE

image

---

`development`: TRUE

image

### DIFF
--- a/client/homebrew/homebrew.jsx
+++ b/client/homebrew/homebrew.jsx
@@ -61,7 +61,7 @@ const Homebrew = (props)=>{
 	if(brew.pureError) {
 		return (
 			<Router>
-				<div className={`homebrew${(config?.deployment || config?.local) ? ' deployment' : ''}`} style={backgroundObject()}>
+				<div className={`homebrew${(config?.deployment || (config?.local && config?.development)) ? ' deployment' : ''}`} style={backgroundObject()}>
 					<Routes>
 						<Route path={brew.originalUrl} element={<WithRoute el={ErrorPage} brew={brew} />} />
 					</Routes>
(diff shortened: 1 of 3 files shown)


5e-Cleric (approved):

Gazook89:
As noted in #4890:

> But does this property only change that background (the text and the color), or does it have additional effects? To me, "development" alone doesn't make me think it's just changing the style of the preview background.

As far as i can tell, the `development` config property *only* changes this preview background, nothing else.   I had assumed that this property was doing more than just that, which is my mistake, but I think it could be more clear.  Possibly the name of the property could better indicate what it is setting, `developmentStyling` or something (totally open to names).

5e-Cleric:
> As noted in #4890:
> 
> > But does this property only change that background (the text and the color), or does it have additional effects? To me, "development" alone doesn't make me think it's just changing the style of the preview background.
> 
> As far as i can tell, the `development` config property _only_ changes this preview background, nothing else. I had assumed that this property was doing more than just that, which is my mistake, but I think it could be more clear. Possibly the name of the property could better indicate what it is setting, `developmentStyling` or something (totally open to names).

Open to it, sounds good to me.

G-Ambatte:
Pro: a single config flag lets us set multiple items
Con: a single flag does not allow fine-grained control over your own instance

It might be scoop creep, as this is a bigger underlying issue, rather than the issue that this PR seeks to solve.

calculuschild (changes requested):
Question: is `development` ever going to be true when local is not true? It seems we could simplify this

```suggestion
- (config?.deployment || (config?.local && config?.development))
+ (config?.deployment || config?.development)
```

I also agree it makes sense to just name this `developmentStyle`. Much more clear what the meaning is, and less overlap with `local/deployment` which start to sound like similar things.

5e-Cleric:
> Question: is `development` ever going to be true when local is not true? It seems we could simplify this
> 
> ```
> - (config?.deployment || (config?.local && config?.development))
> + (config?.deployment || config?.development)
> ```
> 
> I also agree it makes sense to just name this `developmentStyle`. Much more clear what the meaning is, and less overlap with `local/deployment` which start to sound like similar things.

My original thinking was we could use this flag for other things that local users don't want but we do, like maybe the login system, or the admin page

but... i don't care either way, got more important things on my plate.

calculuschild:
Made the small changes. Merging!

PR #4997, opened 2026-09-07
Fix page title

G-Ambatte:
This PR changes the logic for determining the page title. It should no longer be possible to have page titles with a leading hyphen.

The logic breaks down as follows:

1. Set the result to "The Homebrewery:
2. If a brew title exists, prepend it with a hyphen separator
3. If a staging environment is detected (stage/PR deployment/local install) then prepend the appropriate descriptor, again with a hyphen separator
4. Set the page title to the result

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,6 +21,15 @@
 
   			let prefix = '';
 
+			// Default page title
+			let result ='The Homebrewery';
+
+			// Add brew title, if it exists
+			if (title) {
+  				result = `${title} - ${result}`;
+  			}
+
+			// Add prefix for Stage/PR Deployment/Local install
   			if (url && url?.includes('://homebrewery-stage.')) {
   				prefix = `Stage `;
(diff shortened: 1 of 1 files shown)


5e-Cleric (approved):

dbolack-ab:
I believe we had the order on this flipped before this was broken a bit back.

dbolack-ab (approved):

5e-Cleric:
I have not checked the code but the order should be:

`local | PR XXXX` - `title` - The Homebrewery 

That way the most specific information comes first, in order of importance.
A few examples:

`The Homebrewery - NaturalCrit` (Homepage)

`Changelog - The Homebrewery` (Changelog, live)

`Local - How to brew wine - The Homebrewery` (edit, local)

`PR #4997 - 500 page test - The Homebrewery`

Does that make sense?

PR #5004, opened 2026-09-09
Fix SharePage crash on FireFox

G-Ambatte:
After the recent update, SharePage is crashing on Firefox. Investigation has shown that an error is being thrown in `fetchThemeBundle` due to the function being called with the `setError` parameter set to `undefined`, rather than an actual function.

This PR creates a default value for the `setError` and `setThemeBundle` functions, which is an empty function (`() => {}`).

### DIFF
--- a/shared/helpers.js
+++ b/shared/helpers.js
@@ -146,7 +146,7 @@ const printCurrentBrew = async ()=>{
 	}
 };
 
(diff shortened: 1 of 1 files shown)


calculuschild:
Looks good to me. I will make an issue that sharePage should probably have the same `error-navitem` component as editPage so it can display errors like this.

PR #5006, opened 2026-09-09
Extract common "handlePageChange" to a reusable hook

calculuschild:
## Description
Make a custom hook to contain the common components of Edit/Home/New pages. in the transition to merging these components. All pages will be able to pull from "useCommonEditPageFunctions" which should interact with state and everything OK.

There should be no functional change. If this works, will merge, and continue extracting common components piece by piece to make sure nothing breaks at each step.

Eventually this hook will be transformed into a basePage component as all the commonality merges over.


calculuschild:
Page editing of editor/style/metadata remains functional on all three pages. Merging. No one shall ever know...

PR #4972, opened 2026-08-27
Add additional coverage tests for shared/helpers.js

dbolack-ab:
## Description

Adds tests for snippet helpers in shared/helpers.js as a new test file in the tests/html path.

Adds test:helpers script to npm
Adds new test to circleci

PR #5007, opened 2026-09-09
Add "Style" snippets

### LINKED ISSUE #4998: Snippets for Style Editor
### Your idea:

Extending custom brew snippets to work in the Style Editor would be nice.  Either

- displaying the whole list of available custom snippets in *both* the Brew Editor and the Style Editor, or
- using a syntax for denoting something is a "style" snippet (ex: `\snippet(style) Snippet Name`), or
- a new syntax to "split" the Snippet Editor to create a block for 'Brew snippets' and another for 'Style snippets', or
- Literally have two editors in that pane, stacked, one for Brew and one for Style

While I am raising the feature requesting, I'm prepared to be told I'm wrong.  I haven't personally needed Style Snippets.  I figure that if I'm making a template I would create a stylesheet that addresses all of the Brew snippets I have, and changes to the styles of those snippets would happen with CSS custom properties and pre-made alternates (ex, "Blue Stat Block", "Red Stat Block", etc).  

But just because I haven't run into the need doesn't mean that other people won't, or that I wouldn't find a purpose for it.  But it's definitely a question of "add the feature, or educate people on alternative methods" sort of thing.

Raising it for this reddit thread, [How to make snippets?](https://old.reddit.com/r/homebrewery/comments/1w9ekmx/how_to_make_snippets/)

dbolack-ab:
## Description

Adds a new snippet opening tag `\style`. This tells the snippets parser to place the snippet in a separate menu object that has a view of `style`.

As a clarifier, `\text` is an in-place replacement for `\snippet` to indicate  a text-inserting snippet.

Some of this is a little ugly trying to dance around refactoring too much.

To-do:
 - [ ] Update Tests
 - [ ] Update style editor syntax highlighting

- Closes #4998
- Blocked by #4972

### Reviewer Checklist

DO NOT TEST WITH BREWS YOU WANT TO RECOVER. CLONE FIRST.

  - [ ] `\style` is picked up and added to Style editor menu
  - [ ] `\style` is inherited from Brew templates.
  - [ ] User snippets still work
  - [ ] Theme-inherited snippets still work.

### DIFF
--- a/client/homebrew/editor/snippetbar/snippetbar.jsx
+++ b/client/homebrew/editor/snippetbar/snippetbar.jsx
@@ -149,7 +149,8 @@ const Snippetbar = createReactClass({
 		}
 
 		const userSnippetsasJSON = brewSnippetsToJSON(this.props.brew.title || 'New Document', this.props.brew.snippets, this.props.themeBundle.snippets);
-		compiledSnippets.push(userSnippetsasJSON);
+		compiledSnippets.push(userSnippetsasJSON.snippets);
+		compiledSnippets.push(userSnippetsasJSON.styles);
 
 		return compiledSnippets;
 	},
--- a/server/homebrew.api.js
+++ b/server/homebrew.api.js
@@ -216,8 +216,8 @@ const api = {
 				`${text}`;
 		}
 		const metadata = _.pick(brew, ['title', 'description', 'tags', 'renderer', 'theme']);
-		const snippetsArray = brewSnippetsToJSON('brew_snippets', brew.snippets, null, false).snippets;
-		metadata.snippets = snippetsArray.length > 0 ? snippetsArray : undefined;
+		const snippetsStyles = brewSnippetsToJSON(null, brew.snippets, null, false);
+		metadata.snippets = snippetsStyles ? snippetsStyles : undefined;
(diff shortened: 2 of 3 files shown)


5e-Cleric:
By this naming logic, the brew snippets should be called `text` or `brew`? what about `styleSnippet`?

Gazook89:
`\snippet(style)` or similar. Or (css). 

Then that can work for (brew) too, which would also be the default so it is backwards compatible as all. My two cents, but I'm not hung up on it.

dbolack-ab:
> By this naming logic, the brew snippets should be called `text` or `brew`? what about `styleSnippet`?

I'm not sure I understand the question. Did you mean replacing the `\snippet` with alternate text such as `\brew` or `\text` ?

If ya'll feel it would be confusing to leave it as `\snippet`, I don't know. `brewtext` ? `

Gazook89:
I think it's fine to leave normal brew snippets as `\snippet`.

I think it's also fine to have `\snippet(brew)` work exactly the same.

And for style snippets, `\snippet(style)` would create a style snippet.

@5e-Cleric is just pointing out that consistency in names is desirable.  Having one be `\snippet` and another be `\style` doesn't make clear that both are custom snippets.  It sort of it makes it look like one is for custom snippets, and the other is something to do with styles and maybe it's additional CSS or...?

dbolack-ab:
> And for style snippets, `\snippet(style)` would create a style snippet.

Oof. That's awkward. And long.

If we need to explain that a tag is a snippet concept in the snippet window, I do not think that user would need it.

PR #5009, opened 2026-09-10
New/Edit/Home Move setup effects to common functions

calculuschild:
Moves the page setup effects (register key listeners, load settings from localstorage, etc.) into the useCommonEditFunctions hook.

Should be no function change.

### DIFF
--- a/client/homebrew/pages/editPage/editPage.jsx
+++ b/client/homebrew/pages/editPage/editPage.jsx
@@ -8,7 +8,6 @@ import { hbfm } from 'hbmarkedwrapper';
 import _                                      from 'lodash';
 
 import { DEFAULT_BREW_LOAD }                  from '../../../../server/brewDefaults.js';
-import { printCurrentBrew, fetchThemeBundle } from '@shared/helpers.js';
 
 import useCommonEditPageFunctions from '../../utils/commonEditPageFunctions.js'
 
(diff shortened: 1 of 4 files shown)


5e-Cleric:
I assume the functions still have to be called, no? and exported if necessary

calculuschild:
@5e-Cleric 

Actual functions that are being called somewhere in the page do need to be exported; I did this with `handleBrewChange()` for example.

image

But plain `useEffect` blocks just run automatically when you call `useCommonEditorFunctions()`, as if it were in the main component.

Or more accurately, everything in the `useCommonEditorFunctions` hook runs, and part of that is creating the `handleBrewChange()` function. It's like we are literally injecting a segment of React code into the middle of the Home/New/Edit component. The only trick is you have to pass in any dependent values it needs to be aware of so it can compile. 

So calling this hook will both run all the "useEffect" stuff inside, as well as give us the name of any functions that we need access to:

image

The plan is to to continue to gradually shift everything into this `useCommonEditorFunctions()`, testing and merging as we go, until only the unique pieces are left in each page. This means common states and things that are no longer needed for unique actions will also start moving in there, and will no longer need to be passed in as dependencies either. This way I can refactor bit by bit without breaking something and falling too far behind master.

The last step is to turn it into a basePage with the same contents, since that's a little easier to follow the component structure. But that shouldn't be too hard because 95% of things will now be common in one place.

calculuschild:
All moved features are working on all three pages, except CTRL+S to save on HomePage which never did anything.

PR #5016, opened 2026-09-13
Split app.js

5e-Cleric:
This PR splits the page routes out of app.js into their own file, to keep app.js' size down, from 700 lines to 400, rounding up.

Separates concerns and should help in future development, as it will make it easier to navigate these files.


server/app.js
```diff
@@ -132,134 +129,25 @@ export default async function createApp(vite) {
...
+        welcomeTextLegacy,
+        migrateText,
+        changelogText,
+        faqText,
+        getBrew,
+        getUsersBrewThemes,
+        HomebrewModel,
+        GoogleActions,
+        sanitizeBrew,
+    }));
```
calculuschild, line 130: My only comment is we should try to not pass in all of these if we can just import them in the new file. For example:

`welcomeText` should probably be created in the new page-routes file, because it no longer used in here.

`GoogleActions` can just be `import`-ed in the other page.

i.e., unless page-routes needs the *same instance* of these items, it should handle its own imports and dependencies as much as possible, so the two files aren't needlessly coupled together. If a function is now needed in both files, move it to the `helpers.js`.

5e-Cleric: Great comment, thanks

PR #5026, opened 2026-09-14
fix deletion of orphan stubs

5e-Cleric:
Changes the check for google brew to properly returned the expected error code, fixing our deletion code. It now does delete orphaned stubs when the google drive document is not found.


calculuschild (approved):
If you have confirmed this works via testing, then the code change looks good to me.

dbolack-ab (approved):

PR #5040, opened 2026-09-21
Live update sharepage from edit

### LINKED ISSUE #5041: Update Share Page when brew object is updated in edit
Like a livereload from edit to share page, this is relatively simple to make since #4034 was merged. Thanks to G-Ambatte for that progress.

#5040 should close this issue.

5e-Cleric:
Pretty short for what it is.

Creates an api endpoint to serve the brew object, and updates the necessary fields only, to prevent huge rerenders. Quite simple.


5e-Cleric:
I should probably do a test for this little feature.

G-Ambatte:
It works well when the change is in the rendered view or it triggers a change in the number of pages, but doesn't seem to trigger an update when the change is only to the internal content and page is out of the current view.

5e-Cleric:
> It works well when the change is in the rendered view or it triggers a change in the number of pages, but doesn't seem to trigger an update when the change is only to the internal content and page is out of the current view.

As far as i know how this works, brewRenderer won't rerender if the page is outisde the faux-ppr, and it would just render it when it comes to it, no? did i get this wrong?

5e-Cleric:
> > It works well when the change is in the rendered view or it triggers a change in the number of pages, but doesn't seem to trigger an update when the change is only to the internal content and page is out of the current view.
> 
> As far as i know how this works, brewRenderer won't rerender if the page is outisde the faux-ppr, and it would just render it when it comes to it, no? did i get this wrong?

Huh, can confirm your findings, its very interesting, starting to debug

5e-Cleric:
Right, this is brewRenderer not reloading the page since the current viewed pages have not changed, and then forgetting to cache that result as the new subset.

5e-Cleric:
Alright that should do it, requesting reviews.

Gazook89:
So to be clear, if I'm editing a document, those changes are reflecting live in the Share page without having to reload share page?

Does this happen only when it is saved (auto saves)?

client/homebrew/pages/sharePage/sharePage.jsx
```diff
@@ -36,19 +38,43 @@ const SharePage = (props)=>{
...
+		const response = await request
+			.get(`/api/fetch/${currentBrew.shareId}`)
+			.catch((error)=>{
+				console.log('error at fetching updated brew: ', error);
+			});
+		if(response.ok && !!response.body.brew) {
+			const updatedBrew = response.body.brew;
+
+			setCurrentBrew((prev)=>{
+				const changed = Object.keys(updatedBrew).filter((key)=>{
```
calculuschild: This `changed` doesn't seem to be used for anything. Is it needed? We already know the document is changed because version number changed right?

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -230,21 +232,28 @@ const BrewRenderer = (props)=>{
...
 			renderedPages.length = 0;
+			renderedPageTexts.length = 0;
 			pageTemplates.length = 0;
 		}
 
 		// Render currently-edited page first so cross-page effects (variables, links) can propagate out first
-		if(rawPages.length > props.currentEditorCursorPageNum -1)
+		if(rawPages.length > props.currentEditorCursorPageNum -1) {
 			renderedPages[props.currentEditorCursorPageNum - 1] = renderPage(rawPages[props.currentEditorCursorPageNum - 1], props.currentEditorCursorPageNum - 1);
+			renderedPageTexts[props.currentEditorCursorPageNum - 1] = rawPages[props.currentEditorCursorPageNum - 1];
```
calculuschild: What is the purpose of the `renderedPageTexts` array? We already have

- `renderedPages[]` which contains the parsed HTML
- `rawPages[]` which contains the Markdown

It seems like `renderedPageTexts[]` just becomes a copy of `rawPages[]?`

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -230,21 +232,28 @@ const BrewRenderer = (props)=>{
...
 		_.forEach(rawPages, (page, index)=>{
 			const varsOnPageRegex = /([!$]?)\[((?!\s*\])(?:\\.|[^\[\]\\])+)\]/g; // Find out if there are any vars on the page.
 			const forceRender = checkHoists &&
 				!props.hoisted &&
 				(page.match(varsOnPageRegex));  // forceRender forces pages outside of the PPR range to render if true.
 			                                    // This is necessary on the first load to fully populate the variable table.
-			if((isInView(index) || !renderedPages[index] || forceRender) && typeof window !== 'undefined'){
-				renderedPages[index] = renderPage(page, index); // Render any page not yet rendered, but only re-render those in PPR range
+
+			const pageTextChanged = renderedPageTexts[index] !== page;
```
calculuschild: This logic to detect page content changes was already handled in the `useMemo()` below, with `props.text` as a dependency. Any reason that logic needs to move up here? This will add a lot of overhead comparing every page every time a rerender happens.

client/homebrew/brewRenderer/brewRenderer.jsx
```diff
@@ -313,7 +322,10 @@ const BrewRenderer = (props)=>{
 	};
 
 	const renderedStyle = useMemo(()=>renderStyle(), [props.style, props.themeBundle]);
-	renderedPages = useMemo(()=>renderPages(), [props.text, displayOptions]);
+	renderedPages = useMemo(
+		()=>renderPages(),
+		[props.version, props.currentBrewRendererPageNum, props.currentEditorCursorPageNum, displayOptions]
```
calculuschild: We now only re-parse Markdown when brew version changes, brewRendererPage changes, or EditorCursorPage changes.

But actually typing into a brew will no longer be reflected in the brew preview, because props.text has been removed here.

EditorCursor position should also not trigger a re-parse. Simply scrolling to a different editor page and clicking does not change the brew contents.

I would also replace `currentBrewRendererPageNum` with `centerPage`. It's the same value, but using the local version instead of the value passed up and then back down props will prevent a duplicate rerender.

client/homebrew/pages/sharePage/sharePage.jsx
5e-Cleric: No, this makes sure only the changed fields get changed, so if title is changed, the text or styles are not and the brew doesn't rerender in a big flash.

As far as i know it is needed.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: rawpages contains the new and updated brew
renderedPages contains the html
renderedPageText contains the raw text from the previous brew iteration

The key here being because we have PPR involved, a page that is not in view or close does not get rendered when scrolling because there is nothing telling it to.

This compares the pages and checks if the page has changed, so we can re-render that one page.

Am not sure why doesn't the whole thing rerender at the moment, but this seemed like a clever way to fix it at the time. Maybe there is some more clever fix?

calculuschild: If you just need it to rerender when scrolling, I would just add the `centerPage` state to the `useMemo()` entry below. It's the same value as `currentBrewRendererPage` but won't trigger as many duplicate rerenders because it doesn't have to pass up and down through props.

client/homebrew/pages/sharePage/sharePage.jsx
calculuschild: What I mean is, `const changed` gets calculated, but the value is not used anywhere. Whatever problem you think this is solving, the way it's written, it isn't doing anything.

calculuschild: Either way, I don't think it would make a difference. Values that aren't changed will just overwrite themselves with the same value, and nothing will appear any different. We don't rerender the whole brew like that.

client/homebrew/brewRenderer/brewRenderer.jsx
5e-Cleric: lets chat about this in real time in gitter because i am getting lost. please, you can open  gitter where you are i assume

5e-Cleric:
> So to be clear, if I'm editing a document, those changes are reflecting live in the Share page without having to reload share page?

That is correct


> Does this happen only when it is saved (auto saves)?

happens on manual saves and on auto saves, the code that sends the signal to the sharepage is in the saving api.

Gazook89:
I think this particular update should be well noted in changelog/announcements, as some people may be surprised to see changes happening as they view brews, and some may be dismayed to know that opening a brew in a tab isn't enough to "lock it in place", say if an author deletes large portions.

calculuschild (approved):
Looks good to me. I vote to merge this.

5e-Cleric:
Yay!

